Issue #6, opened 2020-10-07
Can't compile due to "error: reached the type-length limit"


xiaochuanyu:
```
❯ cargo build
   Compiling lychee v0.3.0 (/Users/xiaochuanyu/stuff/lychee)
error: reached the type-length limit while instantiating `<std::vec::IntoIter<std::future:...]>>, ()}]>, ()}]>}]>>::Future}]>`
    --> /Users/xiaochuanyu/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/src/libcore/iter/traits/iterator.rs:2015:5
     |
2015 | /     fn fold<B, F>(mut self, init: B, mut f: F) -> B
2016 | |     where
2017 | |         Self: Sized,
2018 | |         F: FnMut(B, Self::Item) -> B,
...    |
2024 | |         accum
2025 | |     }
     | |_____^
     |
     = note: consider adding a `#![type_length_limit="2332107"]` attribute to your crate
```
I tried doing whats suggested but got similar error again with higher suggested limit.
Eventually adding `#![type_length_limit="7912782"]` to top of `main.rs` worked.

mre:
I can reproduce it on my machine. There seems to be a bug in the async/futures stack somewhere that blows up those types. 🤔 It seems to be a recursion problem during code generation.

mre:
Looks like it's a recent issue with the mail checking crate: https://github.com/amaurymartiny/check-if-email-exists/issues/568

xiaochuanyu:
@mre fyi I tried downgrading `check-if-email-exists` but ran into compile errors due its dependencies.
I found a workaround instead, see #9.

mre:
Nice! Thanks for that @xiaochuanyu.
Really weird. Given that this unblocks the other PR and that it's a good idea to use the latest toolchain anyway I'm very happy to merge this.

mre:
@pawroman found this ticket: https://github.com/rust-lang/rust/issues/54540. I'll mention it in the `check-if-email-exists` thread from above.
Guess we can close this. 👏

untitaker:
Was this issue really fixed or just worked around for in CI? Lychee still can't be compiled locally on osx.

mre:
Are you on Rust 1.47?

untitaker:
I was not, sorry for the noise.

mre:
No worries. Thanks for trying lychee. We'd be interested to hear about your experience. You could comment here or open issues if you find any. :smiley:

PR #7, opened 2020-10-09
Add exclude private URLs feature

### LINKED ISSUE #5: Exclude private URLs
We should have flags for excluding private, link-local, and loopback hosts.

Basically this means creating the new commandline flags and mapping them to the functions in std::net (linked below):

* `--exclude-private`: [Exclude private addresses](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_private)
* `--exclude-link-local`: [Exclude link-local addresses](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_link_local)
* `--exclude-loopback`: [Exclude loopback](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.is_loopback)

Newcomers welcome!

pawroman:
Currently, the excludes are based purely on IP address ranges and only work for websites (no support for mail addresses). The IPv6 support is implemented, to a degree possible with Rust standard library as of 1.47.0. See commit messages and code comments for more details.

Note: while IP address exclude works, it's possible that some URLs will resolve to link-local, private or loopback addresses. While this is not very likely, it's a possibility. To address this, we'd have to DNS-resolve all URIs and then run them through IP exclude rules. I have skipped this part because of extra effort involved.

Closes #5.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -1,4 +1,5 @@
 use crate::extract::{self, Uri};
+use crate::options::LycheeOptions;
 use anyhow::anyhow;
 use anyhow::{Context, Result};
 use check_if_email_exists::{check_email, CheckEmailInput};
@@ -6,6 +7,7 @@ use hubcaps::{Credentials, Github};
 use indicatif::ProgressBar;
 use regex::{Regex, RegexSet};
 use reqwest::header::{self, HeaderMap, HeaderValue};
+use std::net::IpAddr;
 use std::{collections::HashSet, convert::TryFrom, time::Duration};
 use url::Url;
 
@@ -68,12 +70,42 @@ impl From<reqwest::Error> for Status {
     }
 }
 
+/// Exclude configuration for the link checker.
+pub(crate) struct Excludes {
+    regex: Option<RegexSet>,
+    private_ips: bool,
+    link_local_ips: bool,
+    loopback_ips: bool,
+}
+
+impl Excludes {
+    pub fn from_options(options: &LycheeOptions) -> Self {
+        Self {
+            regex: RegexSet::new(&options.exclude).ok(),
+            private_ips: options.exclude_private,
+            link_local_ips: options.exclude_link_local,
+            loopback_ips: options.exclude_loopback,
+        }
+    }
+}
+
+impl Default for Excludes {
+    fn default() -> Self {
+        Self {
+            regex: None,
+            private_ips: false,
+            link_local_ips: false,
+            loopback_ips: false,
+        }
+    }
+}
+
 /// A link checker using an API token for Github links
 /// otherwise a normal HTTP client.
 pub(crate) struct Checker<'a> {
     reqwest_client: reqwest::Client,
     github: Github,
-    excludes: Option<RegexSet>,
+    excludes: Excludes,
     scheme: Option<String>,
     method: RequestMethod,
     accepted: Option<HashSet<reqwest::StatusCode>>,
@@ -85,7 +117,7 @@ impl<'a> Checker<'a> {
     /// Creates a new link checker
     pub fn try_new(
         token: String,
-        excludes: Option<RegexSet>,
+        excludes: Excludes,
         max_redirects: usize,
         user_agent: String,
         allow_insecure: bool,
@@ -195,8 +227,8 @@ impl<'a> Checker<'a> {
         }
     }
 
-    fn in_excludes(&self, input: &str) -> bool {
-        if let Some(excludes) = &self.excludes {
+    fn in_regex_excludes(&self, input: &str) -> bool {
+        if let Some(excludes) = &self.excludes.regex {
             if excludes.is_match(input) {
                 return true;
             }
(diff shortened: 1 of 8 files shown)


pawroman:
I quite honestly don't know why the build is failing. The tests run on my local machine just fine (Rust 1.47.0, Linux, x86_64).

pawroman:
The CLI options parser generated weird flag names for this:

```
Usage: target/debug/lychee [OPTIONS]

Positional arguments:
  inputs                 Input files

Optional arguments:
  -h, --help             show help
  -v, --verbose          Verbose program output
  -m, --max-redirects MAX-REDIRECTS
                         Maximum number of allowed redirects (default: 10)
  -t, --threads THREADS  Number of threads to utilize (defaults to  number of cores available to the system
  -u, --user-agent USER-AGENT
                         User agent (default: curl/7.71.1)
  -i, --insecure         Proceed for server connections considered insecure (invalid TLS) (default: false)
  -s, --scheme SCHEME    Only test links with given scheme (e.g. https)
  -e, --exclude EXCLUDE  Exclude URLs from checking (supports regex)
  -E, --exclude-private  Exclude private IP address ranges from checking
  --exclude-link-local   Exclude link-local IP address range from checking
  --exclude-loopback     Exclude loopback IP address range from checking
  -H, --headers HEADERS  Custom request headers
  -a, --accept ACCEPT    Comma-separated list of accepted status codes for valid links
  -T, --timeout TIMEOUT  Website timeout from connect to response finished (default: 20)
  -M, --method METHOD    Request method (default: get)
```
Any preferences as what to should the flags be?

I have a proposal:
- Drop the short flags, leaving only long ones AND
- Have a `--exclude-all-private` flag, mapped to `-E` that would enable all of them (`--exclude-private --exclude-link-local --exclude-loopback`).

What do you think?

pawroman:
> I quite honestly don't know why the build is failing. The tests run on my local machine just fine (Rust 1.47.0, Linux, x86_64).

I was able to work out that it uses Rust 1.46.0, as included in `ubuntu-latest` image: https://github.com/actions/virtual-environments -- see: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md 

I have installed Rust 1.46.0 locally using rustup, and re-ran the tests, and they did fail indeed, with the same message.
Then I checked out the `master` branch (without my changes) and re-ran the tests, they failed.

Then I installed Rust 1.45.2 and re-ran the tests for `master` branch. It did not fail. Then I checked my branch, it did not fail either.

So, it's likely a problem with Rust 1.46.0... Some compiler bug perhaps? I'm not sure what the next steps should be. Let me know what you think!

PS: For reference, here's how to install toolchains:

```
$ rustup toolchain install 1.45.2
$ rustup default 1.45.2
```

pawroman:
For the record: adding

```rust
#![type_length_limit = "7912782"]
```

At the very top of `main.rs` makes it work with Rust 1.46.0.

mre:
@xiaochuanyu found a fix in #6. As you guessed, it's likely an issue with the 1.46.0 toolchain. Updating to the latest version seems to fix the issue.

src/checker.rs
```diff
@@ -191,17 +223,44 @@ impl Checker {
...
         false
     }
 
+    fn in_ip_excludes(&self, uri: &Uri) -> bool {
+        if let Some(ipaddr) = uri.host_ip() {
+            if self.excludes.loopback_ips && ipaddr.is_loopback() {
+                return true;
+            }
+
+            // Note: in a pathological case, an IPv6 address can be IPv4-mapped
```
mre, line 241: Oh wow. That's both shocking and hilarious. 😆

src/checker.rs
```diff
@@ -285,12 +363,14 @@ mod test {
         checker
     }
 
+    fn website_url(s: &str) -> Uri {
```
mre, line 366: nice!

src/extract.rs
```diff
@@ -23,6 +24,17 @@ impl Uri {
             Uri::Mail(_address) => None,
         }
     }
+
+    pub fn host_ip(&self) -> Option<IpAddr> {
```
mre, line 28: nicely done

mre:
@pawroman can you reintegrate `master`? Merged the changes by @xiaochuanyu. 
Apart from that it looks great!

pawroman:
> @pawroman can you reintegrate `master`? Merged the changes by @xiaochuanyu.
> Apart from that it looks great!

Thanks for the review. I have rebased against master, good to see that the CI is green now :)

Any thoughts on the CLI options? https://github.com/hello-rust/lychee/pull/7#issuecomment-706229204

mre:
> I have a proposal:
> * Drop the short flags, leaving only long ones AND
> * Have a --exclude-all-private flag, mapped to -E, that would enable all of them (--exclude-private --exclude-link-local --exclude-loopback).

Sounds very reasonable to me. 👍

pawroman:
> Sounds very reasonable to me. +1

:+1: 

Implemented, along with a test that runs the program against a test file with the flag.

This should now be ready to merge.

mre:
Just gave it a spin locally and it does what it says on the can. Great work and thanks for your contribution @pawroman. 😃

PR #9, opened 2020-10-13
Use action-rs Actions instead

xiaochuanyu:
I don't know why but having most up-to-date Rust toolchain addressed #6.
This PR makes it so that we use `actions-rs/toolchain@v1` which always installs the latest stable versions.
The code is mostly copied from https://github.com/marketplace/actions/rust-cargo.


mre:
Thanks for looking into this and providing a workaround @xiaochuanyu.

PR #10, opened 2020-10-16
Actions improvements: add rustfmt and clippy checks, run tests

pawroman:
Following up on changes in #9, I have noticed that the new action definition only performs a `build`, without running the tests.

This PR introduces the following changes:
- The tests are ran as part of the `test` actions workflow, no longer merely building the code.
- The new `lints` actions workflow adds checks for:
  - `cargo fmt` (checks if all code is formatted using `rustfmt`)
  - `cargo clippy` (checks if there's any warnings produced by [clippy](https://github.com/rust-lang/rust-clippy)). In the process of adding this, I have addressed all the clippy warnings. There's one warning which I silenced, see: https://github.com/hello-rust/lychee/pull/10/files#diff-80d406c60c48b139f4acbeb1b43605d26970cf1de2e39e423ebcf701051e63faR86 -- That I think should be tackled in a separate PR.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,7 +15,8 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-      - uses: actions-rs/cargo@v1
+      - name: Run cargo test
+        uses: actions-rs/cargo@v1
         with:
           command: test
   lints:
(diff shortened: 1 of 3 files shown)


mre:
Great improvement; merged! As for the silenced warning I created #12 which proposes a builder pattern for the Checker. So yes, let's tackle that separately. 👍 
Thanks for the PR @pawroman.

Issue #18, opened 2020-10-21
Add support for basic auth


mre:
Sometimes websites require basic auth to be retrieved.
We should add a flag for basic auth support to lychee.

Maybe something like `--basic-auth "username:password"`.

Hyper (our http client) recently [moved typed headers into a separate crate](https://github.com/hyperium/hyper/issues/1950).

Here is the header we could use ([docs](https://docs.rs/headers/0.3.2/headers/struct.Authorization.html)):

```rust
use headers::Authorization;

let basic = Authorization::basic("username", "password");
```

We could pass that to hyper when we create the client instance.
Pull requests greatly appreciated. Add a comment here if you'd like to work on this.

FabianBG:
Hi @mre, looking to practice some Rust so this looks good, can you assign it to me ?? :D

mre:
Aye it's yours. :clap:

FabianBG:
Pr donne can you review it pls.

FabianBG:
Some problems with fmt y cllippy :P but already fixed

mre:
Resolved by @FabianBG. Yay! 😀

PR #20, opened 2020-10-23
Add support for basic auth #18

FabianBG:
[Issue #18](https://github.com/hello-rust/lychee/issues/18)
* Add headers crate to type headers and create auth header
* Add cmd param basic-auth to set property to the checker
* Add simple test to test if with auth headres is no broken

Signed-off-by: FabianBG <f4b4g3@gmail.com>

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -5,10 +5,11 @@ use crate::{
 use anyhow::anyhow;
 use anyhow::{Context, Result};
 use check_if_email_exists::{check_email, CheckEmailInput};
+use headers::{HeaderMap, HeaderValue};
 use hubcaps::{Credentials, Github};
 use indicatif::ProgressBar;
 use regex::{Regex, RegexSet};
-use reqwest::header::{self, HeaderMap, HeaderValue};
+use reqwest::header;
 use std::net::IpAddr;
 use std::{collections::HashSet, convert::TryFrom, time::Duration};
 use tokio::time::delay_for;
@@ -137,12 +138,11 @@ impl<'a> Checker<'a> {
         verbose: bool,
         progress_bar: Option<&'a ProgressBar>,
     ) -> Result<Self> {
-        let mut headers = header::HeaderMap::new();
+        let mut headers = HeaderMap::new();
         // Faking the user agent is necessary for some websites, unfortunately.
         // Otherwise we get a 403 from the firewall (e.g. Sucuri/Cloudproxy on ldra.com).
         headers.insert(header::USER_AGENT, HeaderValue::from_str(&user_agent)?);
         headers.insert(header::TRANSFER_ENCODING, HeaderValue::from_str("chunked")?);
-
         headers.extend(custom_headers);
 
         let builder = reqwest::ClientBuilder::new()
--- a/src/main.rs
+++ b/src/main.rs
@@ -4,8 +4,9 @@ extern crate log;
 use anyhow::anyhow;
 use anyhow::Result;
 use futures::future::join_all;
+use headers::authorization::Basic;
+use headers::{Authorization, HeaderMap, HeaderMapExt, HeaderName};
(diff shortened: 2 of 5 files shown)


src/checker.rs
```diff
@@ -5,10 +5,11 @@ use crate::{
 use anyhow::anyhow;
 use anyhow::{Context, Result};
 use check_if_email_exists::{check_email, CheckEmailInput};
+use headers::{Authorization, HeaderMap, HeaderMapExt, HeaderValue};
 use hubcaps::{Credentials, Github};
 use indicatif::ProgressBar;
 use regex::{Regex, RegexSet};
-use reqwest::header::{self, HeaderMap, HeaderValue};
+use reqwest::header::{self};
```
mre, line 12: Guess we don't need `self` here?

src/checker.rs
```diff
@@ -136,13 +137,21 @@ impl<'a> Checker<'a> {
...
+        let mut headers = HeaderMap::new();
         // Faking the user agent is necessary for some websites, unfortunately.
         // Otherwise we get a 403 from the firewall (e.g. Sucuri/Cloudproxy on ldra.com).
         headers.insert(header::USER_AGENT, HeaderValue::from_str(&user_agent)?);
         headers.insert(header::TRANSFER_ENCODING, HeaderValue::from_str("chunked")?);
 
+        // Add basic header if exist
+        if let Some(b) = basic_auth {
+            let params: Vec<&str> = b.split(':').collect();
+            let auth = Authorization::basic(params[0], params[1]);
```
mre, line 151: This could fail in case there is just one element in `params`, e.g. if `b` is `hello` without a colon. Consider the snippet [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cc9ad42d4a093a9c4139302431a61af3).

I'd prefer to do the parsing outside of the checker and pass in the `Authorization` header as part of `custom_headers`. 

The parsing function could be in `main.rs` for now and have a signature like `fn parse_auth(auth: &str) -> Authorization`. This way we could add some isolated unit tests just for the parsing.

What do you think?

src/options.rs
```diff
@@ -133,6 +133,14 @@ pub(crate) struct Config {
     #[structopt(short, long, help = "Base URL to check relative URls")]
     #[serde(default)]
     pub base_url: Option<String>,
+
+    #[structopt(
+        short = "B",
```
mre, line 138: Since it's such a specific feature, I'd rather skip the short argument. In my book it's better to spell that one out on the command-line to be explicit that we're passing credentials here.

mre (changes requested):

mre:
Thanks for the PR @FabianBG. Appreciate it. Added some comments; let me know what you think.

src/checker.rs
FabianBG: Sure !!

src/checker.rs
FabianBG: Yehh sound great to me and can be useful for another types of auths in the future

FabianBG:
@mre thanks for the comments and I on that

FabianBG:
Fixes applied, pls review it and tell me if there is something more I can do.

src/main.rs
```diff
@@ -155,6 +162,17 @@ fn parse_statuscodes(accept: String) -> Result<Option<HashSet<http::StatusCode>>
     Ok(Some(statuscodes))
 }
 
+fn parse_basic_auth(auth: &str) -> Result<Authorization<Basic>> {
+    let params: Vec<_> = auth.split(':').collect();
+    if params.len() != 2 {
+        return Err(anyhow!(
+            "Basic auth value should be of the form username:password, got {}",
```
mre, line 169: pefect

mre (approved):

mre:
I wouldn't change a single thing. Very nice! Thanks for the PR @FabianBG. Merging. 🌈

PR #22, opened 2020-10-25
Make GITHUB_TOKEN optional

pawroman:
This also makes the token possible to pass in from CLI args.

I got bitten by this a few times, and decided that issuing a warning instead of failing would be a better choice in terms of user experience.

Happy to tweak the behavior to if you think the new default is not an improvement over the old one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,9 @@
 ## What?
 
 This thing was created from [Hello Rust Episode
-10](https://hello-rust.show/10/). It's a link checker that treats Github links
-specially by using a `GITHUB_TOKEN` to avoid getting blocked by the rate
+10](https://hello-rust.show/10/). It's a link checker.
+
+For GitHub links, it can optionally use a `GITHUB_TOKEN` to avoid getting blocked by the rate
 limiter.
 
 ![Lychee demo](./assets/lychee.gif)
@@ -78,13 +79,20 @@ This comparison is made on a best-effort basis. Please create a PR to fix outdat
 cargo install lychee
 ```
 
-Set an environment variable with your token like so `GITHUB_TOKEN=xxxx`.
+Optional (to avoid being rate limited for GitHub links): set an environment variable with your token
+like so `GITHUB_TOKEN=xxxx`, or use the `--github-token` CLI option. This can also be set in the
+config file.
 
 Run it inside a repository with a `README.md` or specify a file with
 
 ```
 lychee <yourfile>
 ```
+
+### CLI return codes
+
+- `0` for success (all links checked successfully or excluded/skipped as configured)
+- `1` for any failures
 
 ## Comparison
 
--- a/README.md
+++ b/README.md
@@ -89,10 +89,11 @@ Run it inside a repository with a `README.md` or specify a file with
 lychee <yourfile>
 ```
 
-### CLI return codes
+### CLI exit codes
 
 - `0` for success (all links checked successfully or excluded/skipped as configured)
-- `1` for any failures
+- `1` for any unexpected runtime failures or config errors
+- `2` for link check failures (if any non-excluded link failed the check)
 
 ## Comparison
 
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -60,6 +60,10 @@ impl Status {
     pub fn is_success(&self) -> bool {
         matches!(self, Status::Ok(_))
     }
+
+    pub fn is_excluded(&self) -> bool {
+        matches!(self, Status::Excluded)
+    }
 }
 
 impl From<reqwest::Error> for Status {
@@ -110,7 +114,7 @@ impl Default for Excludes {
 /// otherwise a normal HTTP client.
 pub(crate) struct Checker<'a> {
     reqwest_client: reqwest::Client,
-    github: Github,
+    github: Option<Github>,
     includes: Option<RegexSet>,
     excludes: Excludes,
     scheme: Option<String>,
@@ -126,7 +130,7 @@ impl<'a> Checker<'a> {
     // of arguments is short
     #[allow(clippy::too_many_arguments)]
     pub fn try_new(
-        token: String,
+        github_token: Option<String>,
         includes: Option<RegexSet>,
         excludes: Excludes,
         max_redirects: usize,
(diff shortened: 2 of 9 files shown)


src/checker.rs
```diff
@@ -419,7 +437,7 @@ mod test {
     }
 
     #[tokio::test]
-    async fn test_exponetial_backoff() {
+    async fn test_exponential_backoff() {
```
pawroman, line 440: Just a typo fix.

README.md
```diff
@@ -78,14 +79,21 @@ This comparison is made on a best-effort basis. Please create a PR to fix outdat
...
 Run it inside a repository with a `README.md` or specify a file with
 
 ```
 lychee <yourfile>
 ```
 
+### CLI return codes
+
+- `0` for success (all links checked successfully or excluded/skipped as configured)
+- `1` for any failures
```
pawroman, line 95: This seems to be true for other errors bubbled up to `main()` (such as invalid value for `--method`)

mre: Good point. We could use a different exit codes for pure link errors and config errors.
Proposing `1` for link errors and `2` for config errors.

src/checker.rs
```diff
@@ -389,7 +407,7 @@ mod test {
 
     fn get_checker(allow_insecure: bool, custom_headers: HeaderMap) -> Checker<'static> {
         let checker = Checker::try_new(
-            "DUMMY_GITHUB_TOKEN".to_string(),
+            None,
```
mre, line 410: Much better. 😃

src/main.rs
```diff
@@ -86,8 +89,16 @@ async fn run(cfg: Config, inputs: Vec<String>) -> Result<i32> {
     } else {
         None
     };
+
+    if cfg.github_token.is_none() {
```
mre, line 93: Guess the default use-case would be non-Github links so I'd remove that warning as it might be confusing. How about we extend the status message from above?
Right now it is
```rust
Status::Error("GitHub token not specified".to_string()),
```
How about
```rust
Status::Error("GitHub token not specified. Use `--github-token` flag / `GITHUB_TOKEN` env var.".to_string()),
```

src/options.rs
```diff
@@ -133,6 +133,15 @@ pub(crate) struct Config {
     #[structopt(short, long, help = "Base URL to check relative URls")]
     #[serde(default)]
     pub base_url: Option<String>,
+
+    #[structopt(
+        short,
```
mre, line 138: Personally I'd remove the `short` option in case we'd like to support an option where `-g` is fitting. Can't think of anything right now but I'd rather keep the set of options small for now and make it quite explicit that we're passing a github token here.
I'm fine either way, though. 😉

mre:
Sorry, I merged in #23 so you might have to reintegrate `master`. 😅 
Added some comments. Let me know what you think. Otherwise PR looks good.

src/options.rs
pawroman: Makes sense to me :+1:

README.md
pawroman: Because of [how it currently works in Rust (return 1 for any Result::Err bubbled up to main)](https://github.com/rust-lang/rust/blob/a85e94927622665a9e9022de0d33a890a2e32d43/library/std/src/process.rs#L1844) and how the lychee code is structured, I propose we reverse that.

E.g. `1` for config errors and unexpected runtime errors, and `2` for link errors. Rationale: it's way simpler to bubble all "unexpected" errors to main using the `Result` enum and rely on Rust to exit the process with status code `1`. Where we can control the status code easily without much effort, e.g. link errors, we can return `2`.

This could be expanded to be `3` for specific config errors in the future, if need be, but it would require extra effort to classify errors and to normalize the `?` handling in `main()/run()`.

I will make the change to show what I mean; we can then discuss if it's OK with you.

pawroman:
> Sorry, I merged in #23 so you might have to reintegrate `master`. sweat_smile
> Added some comments. Let me know what you think. Otherwise PR looks good.

Addressed the comments now and rebased against master.

src/checker.rs
```diff
@@ -445,7 +465,10 @@ mod test {
         let end = start.elapsed();
 
         assert!(matches!(res, Status::Failed(_)));
-        assert!(matches!(end.as_secs(), 7));
+
+        // on slow connections, this might take a bit longer than nominal backed-off timeout (7 secs)
+        assert!(end.as_secs() >= 7);
+        assert!(end.as_secs() <= 8);
```
mre, line 471: Think there's a way to specify an epsilon. 🤔

mre: Nope. There are [some workarounds](https://stackoverflow.com/questions/30856285/assert-eq-with-floating-point-numbers-and-delta), but nothing built-in. 😬

src/main.rs
```diff
@@ -20,6 +19,17 @@ use checker::{Checker, Excludes, Status};
 use extract::Uri;
 use options::{Config, LycheeOptions};
 
+/// A C-like enum that can be cast to `i32` and used as process exit code.
+enum ExitCode {
+    Success = 0,
+    // NOTE: exit code 1 is used for any `Result::Err` bubbled up to `main()` using the `?` operator.
```
mre, line 25: Thanks for documenting this behavior.

src/main.rs
```diff
@@ -123,7 +133,12 @@ async fn run(cfg: Config, inputs: Vec<String>) -> Result<i32> {
         print_summary(&links, &results);
     }
 
-    Ok(results.iter().all(|r| r.is_success()) as i32)
+    let success = results.iter().all(|r| r.is_success() || r.is_excluded());
+
+    match success {
+        true => Ok(ExitCode::Success as i32),
+        false => Ok(ExitCode::LinkCheckFailure as i32),
```
mre, line 140: Hey that's kinda cool!

tests/cli.rs
```diff
@@ -28,4 +27,73 @@ mod cli {
             .stdout(contains("Successful: 0"))
             .stdout(contains("Errors: 0"));
     }
+
```
mre, line 30: These tests are all great!

mre (approved):

mre:
Good work my sir. Chapeau. 🥂🧐🎩

PR #35, opened 2020-11-24
Add skip_missing flag, add Input enum (closes #26)

### LINKED ISSUE #26: Better handle input files arguments
I have two suggestions for improvements in terms of handling the input files arguments:

1. Allow to pass in links from stdin, e.g. follow the standard-ish Unix convention of specifying `-` as input file, instructing the program to read in from STDIN instead. E.g. this should work fine:

```bash
# - means read from stdin instead from file
$ echo "https://google.com" | lychee -
```

2. Issue some kind of warning or error for non-existing files. Perhaps erroring out instead of issuing a warning can be an optional flag. This should help the user experience in the face of typos and other mishaps.

pawroman:
Handle input arguments in a more structured way, with the ability to skip missing files.

Done:
- [x] Implementation of #26 
- [x] Ensure skip_missing flag is handled properly (the default behavior has been changed to error if one of the input files is missing (doesn't apply for globs or URLs))
- [x] Improve docs for input types
- [x] Refactor inputs to be more generic and extensible
- [x] Parallelize the `collect_links` method for faster operation with multiple inputs/files
  - The next step (some time in the future) could be to dispatch each link for checking as soon as it's extracted, making the program much faster for large amounts of inputs and links
- [x] Tests for all input types handling, including integration tests


mre:
Draft looks good to me. 👍

src/main.rs
```diff
@@ -262,3 +264,220 @@ mod test {
         assert_eq!(expected, actual);
     }
 }
+
+#[cfg(test)]
+mod cli_test {
```
pawroman, line 269: I moved the cli tests here to be able to use the `get_mock_server` helper

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
+    use std::str::FromStr;
+
+    const TEST_STRING: &str = "http://test-string.com";
+    const TEST_URL: &str = "https://test-url.org";
+    const TEST_FILE: &str = "https://test-file.io";
+    const TEST_GLOB_1: &str = "https://test-glob-1.io";
+    const TEST_GLOB_2_MAIL: &str = "test@glob-2.io";
+
+    #[tokio::test]
+    async fn test_collect_links() -> Result<()> {
```
pawroman, line 243: I think this test should be enough here, but I'm happy to add more (e.g. tests for all `Input` methods) if you think it makes sense.

pawroman:
@mre It took way longer than I thought, but it should be OK to go now

src/main.rs
pawroman: Ok, that breaks `cargo test` when running these tests from clean slate.

I will revert and explore ways of sharing the helper.

pawroman: After struggling with this a bit, I think it will be hard to do without introducing a library.

I ended up duplicating these functions :disappointed: -- I think we should re-visit this after #46 is merged.

README.md
```diff
@@ -93,7 +115,7 @@ config file.
 ### CLI exit codes
 
 - `0` for success (all links checked successfully or excluded/skipped as configured)
-- `1` for any unexpected runtime failures or config errors
+- `1` for missing inputs and any unexpected runtime failures or config errors
```
mre, line 118: Eventually we might want to have different exit codes for each case but it's out of the scope of this PR.

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
+
+impl Input {
+    pub(crate) fn new(value: &str, glob_ignore_case: bool) -> Self {
+        if value == STDIN {
+            Self::Stdin
+        } else {
+            match Url::parse(&value) {
+                Ok(url) => Self::RemoteUrl(url),
+                Err(_) => {
+                    // this seems to be the only way to determine if this is a glob pattern
```
mre, line 51: Looks like it: https://github.com/rust-lang-nursery/glob/issues/72

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
+
+    pub async fn get_contents(
+        &self,
+        file_type_hint: Option<FileType>,
+        skip_missing: bool,
+    ) -> Result<Vec<InputContent>> {
+        use Input::*;
+
+        match self {
+            // TODO: should skip_missing also affect URLs?
```
mre, line 75: Good question and no idea. Let's leave that edge-case open for now. Pretty sure we'll eventually know the right answer. 😉

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
+            input: Input::FsPath(path.into()),
+        };
+
+        Ok(input_content)
+    }
+
+    async fn stdin_content(file_type_hint: Option<FileType>) -> Result<InputContent> {
+        let mut content = String::new();
+        let mut stdin = stdin();
+        stdin.read_to_string(&mut content).await?;
```
mre, line 144: Guess eventually we could read line-by-line for `stdin` in case people use lychee in a long-running process to avoid blocking and buffering. The question is if we ever gonna hit this case.

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
+                    contents.push(content);
+                }
+                Err(e) => println!("{:?}", e),
+            }
+        }
+
+        Ok(contents)
+    }
+
+    async fn path_content<P: Into<PathBuf> + AsRef<Path>>(path: P) -> Result<InputContent> {
```
mre, line 131: As a remark, I really like the implementations for each data type.

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
+            Self::String(s) => s.clone(),
+        }
+    }
+}
+
+/// Fetch all unique links from a slice of inputs
 /// All relative URLs get prefixed with `base_url` if given.
 pub(crate) async fn collect_links(
-    inputs: Vec<String>,
+    inputs: &[Input],
```
mre, line 178: Nice!

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
     base_url: Option<String>,
+    skip_missing_inputs: bool,
 ) -> Result<HashSet<Uri>> {
     let base_url = match base_url {
         Some(url) => Some(Url::parse(&url)?),
         _ => None,
     };
 
-    let mut links = HashSet::new();
+    let (contents_tx, mut contents_rx) = tokio::sync::mpsc::channel(32);
```
mre, line 187: Guess we could use `MAX_CONCURRENCY` instead of 32 here to make that configurable? It's not a big deal, though.

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
-            }
-        };
+        tokio::spawn(async move {
+            let contents = input.get_contents(None, skip_missing_inputs).await;
+            sender.send(contents).await
+        });
+    }
+
+    // receiver will get None once all tasks are done
+    drop(contents_tx);
```
mre, line 200: Interesting. I forgot that for the client_pool channel. Is that always needed? If so I will add it.

src/collector.rs
```diff
@@ -1,63 +1,282 @@
...
+    while let Some(result) = contents_rx.recv().await {
+        for input_content in result? {
+            let base_url = base_url.clone();
+            let handle =
+                tokio::task::spawn_blocking(move || extract_links(&input_content, base_url));
+            extract_links_handles.push(handle);
+        }
+    }
+
+    // Note: we could dispatch links to be checked as soon as we get them,
```
mre, line 214: Great idea. I guess returning a `Stream` instead of a `HashSet` could be an option?

src/collector.rs
mre: Looks perfectly fine to me. Thanks!

src/extract.rs
```diff
@@ -13,6 +14,27 @@ pub(crate) enum FileType {
     Plaintext,
 }
 
+impl Default for FileType {
+    fn default() -> Self {
+        Self::Plaintext
+    }
+}
+
+impl<P: AsRef<Path>> From<P> for FileType {
```
mre, line 23: Very nice use of `AsRef` and `Default` above.

src/extract.rs
```diff
@@ -13,6 +14,27 @@ pub(crate) enum FileType {
...
+    /// Detect if the given path points to a Markdown, HTML, or plaintext file.
+    fn from(p: P) -> FileType {
+        let path = p.as_ref();
+        match path.extension() {
+            Some(ext) => match ext.to_str().expect("Couldn't convert OsStr to str") {
```
mre, line 28: I wonder if we want to use `TryFrom` instead of `From` for this function to be able to return that error here instead of panicking.

src/options.rs
```diff
@@ -140,21 +160,30 @@ pub struct Config {
     #[serde(default = "method")]
     pub method: String,
 
-    #[structopt(short, long, help = "Base URL to check relative URls")]
+    /// Base URL to check relative URLs
```
mre, line 163: So nice of you to move those help strings out and make them actual comments.

tests/cli.rs
```diff
@@ -1,20 +1,58 @@
...
+    fn main_command() -> Command {
+        // this gets the "main" binary name (e.g. `lychee`)
+        Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("Couldn't get cargo package name")
+    }
+
+    fn fixtures_path() -> PathBuf {
+        Path::new(module_path!()).parent().unwrap().join("fixtures")
+    }
+
+    // TODO: duplicate of test_utils
```
mre, line 22: Can we remove those mock server functions and import them from `test_utils` instead or is there a blocker somewhere? In that case I'd rephrase the todo a bit and describe the problem to make it easier to fix in the future. 😉

tests/cli.rs
```diff
@@ -83,4 +112,136 @@ mod cli {
...
+    }
+
+    #[tokio::test]
+    async fn test_stdin_input_multiple() {
+        let mut cmd = main_command();
+        let mock_server_a = get_mock_server(http::StatusCode::OK).await;
+        let mock_server_b = get_mock_server(http::StatusCode::OK).await;
+
+        // this behavior (treating multiple `-` as separate inputs) is the same as most CLI tools
+        // that accept `-` as stdin, e.g. `cat`, `bat`, `grep` etc.
```
mre, line 146: Ha! Today I learned... 😆

mre (approved):

mre:
Very good PR! Thanks for tackling this @pawroman. I really like how you improved code quality across the board and paid attention to the smallest details like fixing comments and making the code more idiomatic. Thanks for that.
I just added a few veeery minor notes but we don't have to tackle any of them right now. Just would like to get your feedback on this at some point, but feel free to merge right away if you like. 😃

README.md
pawroman: Yeah, agreed

src/collector.rs
pawroman: Nice idea. This would work perfectly with #55

src/collector.rs
pawroman: Sure, I was going to add it, but managed to overlook it :sweat:

src/collector.rs
pawroman: According to the [tokio docs](https://docs.rs/tokio/0.2.23/tokio/sync/index.html#mpsc-channel):

> 
    // The `rx` half of the channel returns `None` once **all** `tx` clones
    // drop. To ensure `None` is returned, drop the handle owned by the
    // current task. If this `tx` handle is not dropped, there will always
    // be a single outstanding `tx` handle.

src/collector.rs
pawroman: Yeah, this is more or less what I mean here.

How about we tackle this in #55 ?

src/extract.rs
pawroman: I can't really imagine a realistic use case where `OsStr` would not be convertible to `str` (UTF-8). It would require some malice from the users (such as including zero bytes or whatever nastiness). Because of that, I'm inclined to say we should not pollute the API and just stick with `expect`.

Alternatively, we could convert each of the literals below to `OsString` and match against that... :thinking: Using `lazy_static` should actually make it cheaper too...

Let me see what I can do :wink:

tests/cli.rs
pawroman: Yeah I tried to fix this, but:

https://github.com/lycheeverse/lychee/pull/35#discussion_r533652892

It might require `lib.rs` to fix properly.

src/extract.rs
pawroman: Actually, ended up using `PartialEq` for `OsString` https://doc.rust-lang.org/std/ffi/struct.OsString.html#impl-PartialEq%3C%26%27_%20str%3E

src/extract.rs
```diff
@@ -13,6 +14,27 @@ pub(crate) enum FileType {
...
+
+impl<P: AsRef<Path>> From<P> for FileType {
+    /// Detect if the given path points to a Markdown, HTML, or plaintext file.
+    fn from(p: P) -> FileType {
+        let path = p.as_ref();
+        match path.extension() {
+            Some(ext) => match ext {
+                _ if ext == "md" => FileType::Markdown,
+                _ if (ext == "htm" || ext == "html") => FileType::HTML,
+                _ => FileType::Plaintext,
```
pawroman, line 31: We discussed whether we should `unwrap`, use `TryFrom` etc. Turns out, `OsStr` implements `PartialEq` for `&str`, so... I think we have a winner here :wink:

src/options.rs
```diff
@@ -1,12 +1,24 @@
...
+const MAX_CONCURRENCY: usize = 128;
+const MAX_REDIRECTS: usize = 10;
+
+// this exists because structopt requires `&str` type values for defaults
+// (we can't use e.g. `TIMEOUT` or `timeout()` which gets created for serde)
+lazy_static! {
+    static ref TIMEOUT_STR: String = TIMEOUT.to_string();
+    static ref MAX_CONCURRENCY_STR: String = MAX_CONCURRENCY.to_string();
+    static ref MAX_REDIRECTS_STR: String = MAX_REDIRECTS.to_string();
+}
```
pawroman, line 21: I got dismayed by these values being `String`.

I think at some point we should overhaul the opts+config definition and parsing.

src/main.rs
mre: Yeah no worries.

src/collector.rs
mre: Ouch then I'll add the explicit drop. 👍

src/collector.rs
mre: totes

src/extract.rs
mre: Rust is awesome.

mre: and kudos for finding that

src/options.rs
mre: oh yeah, it's getting a bit out of hand

mre:
So happy to merge this. A major step forward and built with great attention to detail. Looking forward to whatever you'll tackle next.

PR #38, opened 2020-11-24
Properly announce lychee user agent

mre:
Just a couple of cleanups I made while browsing the source-code. 
@pawroman I'm not sure if it's worth your time to do a code-review here but I'll keep that open for a while just in case. 😉 
Perhaps the biggest change is that we're now using `lychee/0.3.1` as the user agent name. It can still be overwritten of course.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,13 +1,13 @@
 [package]
 authors = ["Matthias Endler <matthias-endler.de>"]
-description = "A boring link checker for my projects (and maybe yours)"
-documentation = "https://github.com/hello-rust/lychee/blob/master/README.md"
+description = "A glorious link checker"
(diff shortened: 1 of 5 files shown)


pawroman (approved):

Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [package]
 authors = ["Matthias Endler <matthias-endler.de>"]
-description = "A boring link checker for my projects (and maybe yours)"
-documentation = "https://github.com/hello-rust/lychee/blob/master/README.md"
+description = "A glorious link checker"
```
pawroman, line 3: :rocket:

mre: Haha, the world shall know about the glory that is this link checker. 🤣

Issue #13, opened 2020-10-18
Make lychee usable as a library


mre:
Realized that some other projects might profit from a good link checker.
One such project that comes to mind is [zola](https://github.com/getzola/zola), which uses a [custom but basic link checker](https://github.com/getzola/zola/blob/2d1c95432290478547316522d1f9219ac38e5b13/components/site/src/link_checking.rs) right now.

TODO:
* [x] Split code into `lib.rs` and `main.rs`
* [x] Use `lib.rs` from `main.rs`
* [x] Bonus: split lib into separate crate that can be published on crates.io

If anyone is interested in giving this a shot, feel free to add a comment below so I can assign it to you. Thanks!

mre:
Done in #46

PR #46, opened 2020-11-30
Make lychee usable as a library #13

mre:
This splits up the code into a `lib` and a `bin` to make it reusable in other crates.
Along the way I made some design decisions that I'd like to discuss:

* The exclude struct was previously built from the config with `Excludes::from_options(&cfg)`. However the config file is independent from the link checking and I didn't want to couple the two. Therefore I introduced a few more parameters for the builder:    

    ```rust
    pub struct ClientBuilderInternal {
        /// Exclude links matching this set of regular expressions
        excludes: Option<RegexSet>,
        /// Exclude all private network addresses
        exclude_all_private: bool,
        /// Exclude private IP addresses
        exclude_private_ips: bool,
        /// Exclude link-local IPs
        exclude_link_local_ips: bool,
        /// Exclude loopback IP addresses (e.g. 127.0.0.1)
        exclude_loopback_ips: bool,
        /// other, previously existing fields
    }
    ```

  With this I hope we can make the builder as flexible as possible while having a clear separation between the `bin` and the `lib`.
* In line with the previous point I renamed `include()` to `includes()` in the builder to use the plural for both excludes and includes.
* I kept the code in a single crate (instead of using a workspace) in the hope of publishing the lychee library under the same name on crates.io: lychee. That should avoid confusion in the future as to which crate to use.
* Added  `#[deny(missing_docs)]` to enforce documentation for public code in the lib. This now serves as a public interface and I want to treat it like a contract with clear documentation.
* Kept the `client_pool` inside the lib for now as it might help with checking a lot of links. One might argue that it doesn't belong there as its implementation is rather opinionated. I didn't want to put it into the `bin` folder, though as I'd only like to have the runtime configuration code in there.

Feedback welcome. 

TODO

* [ ] Add skeptic for testing the `README.md` examples.
* [ ] Add convenience method for adding custom headers.


Cargo.toml
```diff
@@ -4,6 +4,13 @@ description = "A glorious link checker"
 documentation = "https://github.com/lycheeverse/lychee/blob/master/README.md"
 edition = "2018"
 homepage = "https://github.com/lycheeverse/lychee"
+keywords = [
+    "link",
+    "checker",
+    "cli",
+    "link-checker",
+    "validator",
+]
```
pawroman, line 13: :tada:

src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+*
+* #[tokio::main]
+* async fn main() -> Result<(), Box<dyn Error>> {
+*   let client = ClientBuilder::default().build()?;
+*   let url = Url::parse("https://github.com/lycheeverse/lychee")?;
+*   let response = client.check(Website(url)).await;
+*   assert!(matches!(response.status, Status::Ok(_)));
+*   Ok(())
+* }
+* ```
```
pawroman, line 24: Just sanity checking: is this actually being run as a doctest?

src/client.rs
```diff
@@ -34,21 +35,57 @@ pub struct Client {
...
+    /// Exclude loopback IP addresses (e.g. 127.0.0.1)
+    exclude_loopback_ips: bool,
+    /// Maximum number of redirects before returning error
     max_redirects: usize,
+    /// User agent used for checking links
     user_agent: String,
+    /// Ignore SSL errors
     allow_insecure: bool,
+    /// Valid URI schemes (e.g. https, http)
     scheme: Option<String>,
```
pawroman, line 61: I'd change the doc to say `Allowed` rather than `Valid` and add more context of what it affects (link discovery).

src/client.rs
```diff
@@ -34,21 +35,57 @@ pub struct Client {
...
+    /// Exclude links matching this set of regular expressions
+    excludes: Option<RegexSet>,
+    /// Exclude all private network addresses
+    exclude_all_private: bool,
+    /// Exclude private IP addresses
+    exclude_private_ips: bool,
+    /// Exclude link-local IPs
+    exclude_link_local_ips: bool,
+    /// Exclude loopback IP addresses (e.g. 127.0.0.1)
+    exclude_loopback_ips: bool,
```
pawroman, line 53: I'm in two minds about it.

1:

We ask the users to provide complex objects from other crates like `custom_headers: HeaderMap` or `accepted: Option<HashSet<http::StatusCode>>` -- yet we "explode" the `Excludes` which should be part of the public API.

I get the "removing coupling" argument, but maybe the functionality of producing `Excludes` from config could be moved to config, and then we could have a nice public constructor/builder for `Excludes` what would go into the library's public API?

2:

Accepting just primitive types makes the builder much simpler and straightforward to use.

Honestly, don't know what's the right choice here, but I'm happy to give the current approach a try.

pawroman (approved):
> Feedback welcome.

Looks all good to me.

Left a code comment about `excludes`, as I'm in two minds about it.

src/lib.rs
mre: Yup, and it passes. 😄

mre:
```
   Doc-tests lychee

running 1 test
test src/lib.rs - client (line 8) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

from https://github.com/lycheeverse/lychee/pull/46/checks?check_run_id=1471385577#step:4:674

mre: As a side-note, I'd really like to support 

```rust
 let response = client.check("https://github.com/lycheeverse/lychee").await;
```

instead of

```rust
*   let url = Url::parse("https://github.com/lycheeverse/lychee")?;
*   let response = client.check(Website(url)).await;
```

to simplify library usage. 
Of course it's not that easy because otherwise `check()` would have to return a `Result` and I don't like that. (The URL parsing could fail). Alternatively we return a `Response::Error` which is a special enum variant that indicates that parsing failed - but then we treat `Response` as some sort of `Result` which isn't very clean either.
The last option I can see is to go with the slightly less ergonomic

```rust
 let response = client.check("https://github.com/lycheeverse/lychee".try_into()?).await;
```

or 

```rust
 let response = client.check(Request::try_from("https://github.com/lycheeverse/lychee")?).await;
```

Not sure if it's worth the trouble, though. It would at least allow us to accept different input types in general, which could make it extensible after 1.0.
Did I miss an option? Feedback welcome.

src/client.rs
mre: I can feel you. The one good thing about the builder is that not all options have to be configured at once. I assume that many users would simply pass `exclude_all_private(true)` and perhaps a regex pattern if needed. Probably many wouldn't even touch the options and go with the defaults, which is great. 😄 

The other thing I like about "unrolling" the config options like this is that the IDE can provide useful completions. E.g. if a user isn't sure what exclude they might need, they might simply type `exclude` and see what the IDE suggests, whereas they'd have to import the ExcludeBuilder and learn how to configure it before passing it to the ClientBuilder, which could break their train of thought.

On the other hand your argument about having other complex types is true. It was a direct translation from the previous constructor, but one could argue that it's time to break that up and make life a little easier for the user by providing an `add_header()` and an `accept()` method and simply creating the more complex types during the build.
Would that make sense?

src/lib.rs
pawroman: Very interesting ideas! Maybe we could provide 2 interfaces: one simplified e.g. "just pass as &str and I will tell you if it worked", and the "usual" one with explicit type handling.

But let's move this discussion to a dedicated place.

src/client.rs
pawroman: > make life a little easier for the user by providing an add_header() and an accept() method and simply creating the more complex types during the build.
Would that make sense?

Definitely! I'm very much a fan of this approach. This would be the simplest API from the human-friendliness perspective. And if we had both `add_header(&str, &str)` and `add_headers(HeaderMap)` that would cover all the use cases.

src/lib.rs
mre: Agreed. #61 is where it's at.

src/client.rs
mre: Done. 👍

mre:
Will rebase on top of `master`.
One question that came up was if we want to make `extract.rs` part of the library. 
Some people would probably be using the client in combination with a set of existing links but some might want to have a convenient way to retrieve those links from their documents somehow.
I wonder what's the more common scenario.
If in doubt I'd opt for not making the extractor public initially as it adds quite a big surface area to the public API.

mre:
Also, did I get it right that `test_utils` will be part of the lib and imported from the binary?
If so wouldn't that leak some testing internals of lychee and technically mean a breaking change if we modify it?
Moving it out into the binary also feels a bit awkward because it's not exactly CLI code we're testing here.
Probably I'm missing something.
To rephrase, what was the reason why we needed to duplicate the `test_utils` code?

pawroman:
> Also, did I get it right that `test_utils` will be part of the lib and imported from the binary?
> If so wouldn't that leak some testing internals of lychee and technically mean a breaking change if we modify it?
> Moving it out into the binary also feels a bit awkward because it's not exactly CLI code we're testing here.
> Probably I'm missing something.
> To rephrase, what was the reason why we needed to duplicate the `test_utils` code?

Yeah, I managed to get rid of the duplication with 2 caveats:

1) The test utils needed to be exposed as public API from `lib.rs`, without `test` attribute
2) wiremock needed to be moved to normal requirements, not just dev requirements

I was going to open an issue for this to discuss further, since I don't like how it's implemented now, and I don't like the way I managed to "fix" it.

Perhaps the problem could be solved by adding a sub-crate to the project?

pawroman:
>  One question that came up was if we want to make `extract.rs` part of the library.

I think it makes sense to do that. We have all this cool functionality, why keep it to ourselves? 😉

mre:
> I think it makes sense to do that. We have all this cool functionality, why keep it to ourselves? 😉

You're right. I moved it to the lib.
Exposed it as a separate module, though and not on the root level to be able to change that more easily in the future.

mre:
Was able to avoid the code duplications for the testing code as per your instructions. 🥳 

> Perhaps the problem could be solved by adding a sub-crate to the project?

Yes. I was actually wondering about that. We'd probably need similar testing code for benchmarking so a separate crate might be the cleanest way to go about it.
We can collect ideas in an issue but I don't think it's very pressing to change it right now. 😄

src/test_utils.rs
```diff
@@ -1,22 +1,17 @@
-#![cfg(test)]
-
 use http::StatusCode;
 use wiremock::matchers::path;
 use wiremock::{Mock, MockServer, ResponseTemplate};
 
 // TODO: used in cli tests (as duplicate)
```
pawroman, line 5: This can now be removed 😄

pawroman:
> Yes. I was actually wondering about that. We'd probably need similar testing code for benchmarking so a separate crate might be the cleanest way to go about it.
> We can collect ideas in an issue but I don't think it's very pressing to change it right now. 😄

Added a new issue for this: #62 

As far as the PR goes, it looks great now! :shipit:

pawroman:
Oh, just one more thing I forgot to mention: now that we ship a decent library interface, it would be great to publish the releases to [crates.io](https://crates.io/)!

src/test_utils.rs
mre: done

mre:
Thanks again for the collaboration @pawroman. We're on a good track!

Issue #40, opened 2020-11-29
Publish images to Docker Hub


pawroman:
To make the project more user friendly, and easier to integrate into existing workflows, we should add Docker image builds for releases/tags and publish them to Docker Hub.

mre:
I registered the lycheeverse namespace on Dockerhub now so we can proceed with building the image.
Propose to base in off of alpine or debian slim. Contributions welcome.

mre:
This one looks great: https://github.com/marketplace/actions/publish-docker

mre:
Proposing to use Debian for the base image as it provides better perf in comparison to alpine builds, and SSL support is easier to configure according to https://blog.logrocket.com/packaging-a-rust-web-service-using-docker/.

pawroman:
> Proposing to use Debian for the base image

Absolutely agreed. Debian slim is a great base from my experience. Can't say the same about alpine (it's a bit non-standard and causes all sorts of random issues)

mre:
Yay, we have an official image now and the latest `master` build is already published. 🐳 
It's based on Debian slim and automatically generated whenever we push to `master` or release a new tag.

PR #52, opened 2020-11-30
Add Docker image #40

mre:
Adds an initial Docker image to the repo.

*TODO*

* [x] Set up automatic deployment of the image after every push to `master`.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,21 @@
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Publish Docker Image
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Publish to Registry
+        uses: elgohr/Publish-Docker-Github-Action@master
+        with:
+          name: lycheeverse/lychee
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          snapshot: true
+          tag_semver: true
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - name: Publish to Registry
-        uses: elgohr/Publish-Docker-Github-Action@master
+        uses: elgohr/Publish-Docker-Github-Action@3.01
         with:
           name: lycheeverse/lychee
           username: ${{ secrets.DOCKER_USERNAME }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM rust:1.48 as builder
 
 RUN USER=root cargo new --bin lychee
 WORKDIR /lychee
-COPY ./Cargo.toml ./Cargo.toml
+COPY Cargo.toml Cargo.toml
 RUN cargo build --release
 RUN rm src/*.rs
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,20 @@ FROM rust:latest as builder
 
 RUN USER=root cargo new --bin lychee
 WORKDIR /lychee
+
+# Just copy the Cargo.toml and trigger a build so 
+# that we compile our dependencies only.
+# This way we avoid layer cache invalidation
+# if our dependencies haven't changed,
+# resulting in faster builds.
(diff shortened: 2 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,21 @@
...
+      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
+
+name: Publish Docker Image
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Publish to Registry
+        uses: elgohr/Publish-Docker-Github-Action@master
```
pawroman, line 15: Perhaps we should pin this to a tagged version to avoid breakage? I don't know if actions support semantic versioning.

Latest tag seems to be `3.01`: https://github.com/elgohr/Publish-Docker-Github-Action/tree/3.01

Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM rust:1.48 as builder
```
pawroman, line 1: Our pipelines/tests use latest stable Rust, maybe we should use latest stable here too? E.g. `FROM rust:latest` or `FROM rust:1`. Not sure how github actions docker cache will behave if we supply `rust:latest`, e.g. if the cache won't prevent us keeping up to date with stable Rust.

Alternatively, we should also run the pipelines/tests for the pinned version, as well as latest stable.

The rationale behind these two approaches is to make sure the packaged version is also always tested before it's released.

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN cargo build --release
+RUN rm src/*.rs
+
+ADD . ./
+
+RUN rm ./target/release/deps/lychee*
+RUN cargo build --release
+
+
+FROM debian:buster-slim
```
pawroman, line 15: Worth separating the stages with a comment

Makefile
```diff
@@ -0,0 +1,27 @@
...
+.PHONY: docker-build
+docker-build: ## Build Docker image
+	docker build -t $(IMAGE_NAME) .
+
+.PHONY: docker-run
+docker-run: ## Run Docker image
+	docker run $(IMAGE_NAME)
+
+.PHONY: push
+push: ## Push image to Docker Hub
```
pawroman, line 18: I'd change this to `docker-push`

Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM rust:1.48 as builder
+
+RUN USER=root cargo new --bin lychee
```
pawroman, line 3: I don't see the code being copied into the container... Are we just building hello world for now?

.github/workflows/docker.yml
mre: True. Done 👍

Dockerfile
mre: Oh good point. Big oversight from me. 
Guess `rust:latest` should be fine. Haven't found much information on how long the images from Docker Hub get cached, just some vague general information from [here](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows). It does not say anything about the Docker cache time but I doubt it would be very long given that so many pipelines depend on the `latest` tag. In the worst case we can change it back to `rust:1`.

Dockerfile
mre: In this step yes. We just copy the `Cargo.toml` and trigger a build so that we compile only our dependencies.
A bit further down we `ADD . ./` to copy the source code and run the build again. This should only compile lychee itself. A little trick to reduce build times. 
Of course it only helps for local builds because the Github action starts with an empty environment but we could add [caching](https://github.com/marketplace/actions/build-docker-images-using-cache) for that step in the future.

Dockerfile
mre: True. Added.

Makefile
mre: nice catch

mre:
@pawroman thanks for the review. Addressed your comments. Just two more open questions. Let me know what you think and if we're good to merge. 😃

Dockerfile
pawroman: Makes sense, let's try that.

Dockerfile
pawroman: Wow, OK. I see. Hacky!

Let's add a comment explaining these shenanigans.

mre: True. 😆 It's kind of a standard approach for other package managers, too. Most notably node.js. Then again it's a good idea to comment that nonetheless, which I now did. 👍

mre:
I... guess we're clear here. Merging this in.

PR #60, opened 2020-12-01
Update troubleshooting guide, add .editorconfig

pawroman:
Having noticed a typo I have decided to fix inconsistent whitespace in the troubleshooting document as well.

While at it, I added `.editorconfig` file (see [EditorConfig](https://editorconfig.org/)) which should make it easier to maintain consistent indentation and white space handling throughout the project.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.rs]
+indent_size = 4
+
+[Makefile]
+indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,5 +9,11 @@ trim_trailing_whitespace = true
 [*.rs]
 indent_size = 4
 
+[*.html]
+indent_size = 2
+
+[*.yml]
+indent_size = 2
+
 [Makefile]
 indent_style = tab
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ![Rust](https://github.com/hello-rust/lychee/workflows/Rust/badge.svg)
 
-A fast, async, resource-friendly link checker written in Rust.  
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 ![Rust](https://github.com/hello-rust/lychee/workflows/Rust/badge.svg)
 
-A fast, async, resource-friendly link checker written in Rust.
```
mre, line 5: Actually added those two spaces to enforce an extra line break. 😉 
My intention was to make that sentence stand out a bit.

TROUBLESHOOTING.md
```diff
@@ -1,22 +1,26 @@
-# Toubleshooting Guide
+# Troubleshooting Guide
```
mre, line 1: whoopsie

.editorconfig
```diff
@@ -0,0 +1,19 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.rs]
+indent_size = 4
```
mre, line 10: Wouldn't that be handled by `rustfmt`? Just wondering. 😅

pawroman: I guessed it wouldn't hurt :shrug:

README.md
pawroman: Ok. A backslash (`\`) will provide the same effect. I shall update.

pawroman: And done. The line breaks should be preserved now, without trailing whitespaces!

mre: Ah, didn't know that `\` also works. That's nice because it makes the intent more explicit. Thanks.

.editorconfig
mre: True. It's unlikely that rustfmt will change its default indentation I guess, so nevermind. 😆

mre:
Great work with this PR @pawroman. Rock on! 🤘

PR #65, opened 2020-12-04
Publish releases on crates.io

mre:
See #59 
Pipeline code taken from https://github.com/EmbarkStudios/tryhard/blob/main/.github/workflows/rust-ci.yml

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,7 +19,8 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-  lints:
+  
+  lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


pawroman:
Have you thought about syncing git tag with Cargo crate version in `Cargo.toml`?

My worry is that the tags and version in `Cargo.toml` can get out of sync. At the very least, we should check if they are in sync before publishing.

mre:
> Have you thought about syncing git tag with Cargo crate version in Cargo.toml?

Nope, but it's a great idea. Not sure how to do this, though. Are there any other projects that do that already that we can learn from?

pawroman:
> Not sure how to do this, though. Are there any other projects that do that already that we can learn from?

I can do some research into it. Not a blocker for this PR though.

mre:
Thanks, then I'll merge it.
I think I'll add support for binary releases to Github, too and then cut a new 0.4.0 release.

PR #66, opened 2020-12-04
Update GitHub token info, add my repo as user

pawroman:
I was able to use lychee successfully with my links repo: https://github.com/pawroman/links

In the process, I decided to add instructions for GitHub token, and add my repo as first user :wink: 

Here's how to make it work with GitHub actions: https://github.com/pawroman/links/blob/master/.github/workflows/main.yml#L12-L25 -- It's not ideal, but it's a start. It will be useful for creating a dedicated GitHub action.

Straight off the bat, I see an issue with `--progress` flag in actions. Will investigate in spare time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,6 @@ You can run lychee directly from the commandline.
 
 #### Using cargo
 
-### Using cargo
-
 ```
 cargo install lychee
 ```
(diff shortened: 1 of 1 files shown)


mre:
Congrats for being the first user! So cool to see. 
Yeah progress will be a tough one. It sends terminal control commands to change a line. In GitHub it just looks like gibberish I assume. Maybe `verbose` would be the way to go for CI and the GitHub action. That said I'd love to make the final GitHub action flexible enough to accept lychee parameters.

PR #69, opened 2020-12-06
Update demo .gif in README.md

pawroman:
The .gif is looping and on black background. It's also much smaller in terms of file size.

Check out how it looks like in the README.md: https://github.com/pawroman/lychee/blob/update_gif/README.md

I'm happy to tweak it if needed. :smile:


mre:
That's an instant merge for me. Looks 💯 times better than the current one. The emojis look much nicer, too!
And it's finally looping. 🥳

mre:
Another advantage is that it covers the full width of the `README.md`. 😆

pawroman:
Awesome.

In case you're curious, I have used the [peek](https://github.com/phw/peek) screen recorder -- maybe it will come in handy :)

mre:
On second thought I have some ideas on how to make it even better:
- The background color could be white to better match with the rest of the `README.md` and not stick out too much.
- Typing out the initial command could be 30-50% faster. 
- Showing the lychee and Rust version in the prompt is fancy but might look outdated pretty quickly with newer Rust and lychee versions. 

These are all very minor and we can totally keep the recording as it is right now. Just wanted to give you some more feedback. 😜

pawroman:
> On second thought I have some ideas on how to make it even better:
> 
>     * The background color could be white to better match with the rest of the `README.md` and not stick out too much.
> 
>     * Typing out the initial command could be 30-50% faster.
> 
>     * Showing the lychee and Rust version in the prompt is fancy but might look outdated pretty quickly with newer Rust and lychee versions.
> 
> 
> These are all very minor and we can totally keep the recording as it is right now. Just wanted to give you some more feedback. 😜

Good feedback, especially about the versions. For faster typing we could use [doitlive](https://github.com/sloria/doitlive).

I can re-record the gif, but please don't force me to use a white background, I can't stand it 😜

mre:
> I can re-record the gif, but please don't force me to use a white background, I can't stand it 😜

No worries. 😆

Issue #64, opened 2020-12-04
Expose the link check as library


drahnr:
Hey, awesome tool and as a matter of fact [cargo-spellcheck](http://github.com/drahnr/cargo-spellcheck) has a bit of an overlap (I just wasn't aware of `lychee`'s existence!).

It would be awesome if `lychee` could be used as a library as well so that arbitrary cmark content can be fed into it, this would be particularly interesting for not re-implementing the logic as part of https://github.com/drahnr/cargo-spellcheck/pull/113

Thanks!

mre:
Hey @drahnr,
you're in luck. #46 will make that happen and it lands... \*checking notes\*... it has just landed. 🎉 
Can you give it a shot and report back? We'd be happy to add any missing pieces.

mre:
BTW added cargo-spellcheck to https://github.com/analysis-tools-dev/static-analysis.
It's live on the website now in case you want to give it an upvote: https://analysis-tools.dev/tag/rust

mre:
Closing this as library support is available now. We can create new issues for library feature requests.  
We'd love to get your feedback after trying lychee. You can create a new issue or comment here. 😄

drahnr:
There are two issues that I can see right now:

* cargo manifest does not expose a library
* a release is pending to include those changes (`0.3.1` was cut right before #64 landed)

Or maybe I missed smth. Thanks in advance!

mre:
Will cut a new release 0.4.0 after #74 is merged. Then we can test the CD pipeline as well. 😄

mre:
@drahnr can you give it another shot? 0.4.0 is out on now on crates.io.

drahnr:
Cool thanks!

There are few things I think that could be improved upon:

* Reduce the exposed module structure - imho it's better to have fewer modules when each of them only contains a single or two entities
* Provide true errors (one that impl's `Send + Sync + 'static + std::error:Error`) rather than using `anyhow`. I'd recommend `thiserror` :)
* Currently one example is provided for the simple case, it would be nice if that was shown as the entry one for docs.rs and expanded in order to showcase the useage of `InputContent` with `collect_links`.

Other than that, it's nice and simple to use!

drahnr:
Another issue that arises, is regarding the use of `tokio 0.2`, lychee's futures will not run nicely with `tokio 0.3`: 
```
[...]

PR #74, opened 2020-12-09
Build binaries for Linux, Windows, and macOS for every release

mre:
This creates lychee binaries for every release. 
See #59 for more info.

Created a fake release [here](https://github.com/mre-early-access/lychee/releases/tag/v0.6.2) to show how the output looks like.

Some notes:
* mips builds are not supported by `ring` at the moment, which is a sub-dependency. reqwests has support for rustls through a [feature flag](https://github.com/seanmonstar/reqwest/blob/master/Cargo.toml#L37), but I doubt that it supports mips either.
* With around 26 minutes, the windows builds take the longest time. (2-3x as long as macOS and Linux.) I don't know why this is; the compile step is just slower. ¯\_(ツ)_/¯
* Introducing a feature flag for using [vendored openssl](https://docs.rs/openssl/0.10.31/openssl/#vendored) as it prevents headaches with the OpenSSL shared library on Linux.
* Adding `openssl-sys` as direct dependency. This allows us to pass the `vendored-openssl` feature flag to this dependency and compile openssl from source.
* Using `bruceadams/get-release@v1.2.2` to pull in the tag name from a release. This allows us to create a new release from the Github UI while the pipeline can still detected the respective tag for a release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,24 +39,24 @@ jobs:
           - aarch64-unknown-linux-gnu
           # See https://github.com/briansmith/ring/issues/562
           # - mips-unknown-linux-musl
-          # - #mipsel-unknown-linux-musl
+          # - mipsel-unknown-linux-musl
       fail-fast: false
     steps:
       - name: Install musl tools
         if: contains(matrix.target, 'musl')
-        run: sudo apt-get install musl-tools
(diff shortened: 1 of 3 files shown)


mre:
Going ahead and merging this to unblock #64. We can always revisit this pipeline and fix any potential issues.

PR #76, opened 2020-12-14
Add output file support (fixes #53)

### LINKED ISSUE #53: Report output in JSON format
## Introduction

We only print a summary to the console at the end of the process at the moment.
Having machine-readable output would allow other tools and CI pipelines to consume
lychee's output.

## How to contribute

Link checking statistics end up in a `stats` object that is defined here:
https://github.com/lycheeverse/lychee/blob/master/src/stats.rs

You'd have to implement `Serialize` and `Deserialize` from [`serde`](https://serde.rs/) for this struct:
https://github.com/lycheeverse/lychee/blob/d2e349cb29606d3a4a999dcf5dce1b5108a8c0f9/src/stats.rs#L10-L18
After that the output can be automatically converted to JSON format (as well as HTML, SQL, CSV, XML, YAML, TOML and others if you want).

Here is a good example of the full process:
https://github.com/serde-rs/json#creating-json-by-serializing-data-structures

## Mentorship

If you need more guidance don't be afraid to ask here. We will do our best to help you get started.

mre:
This PR adds two new flags:

`--output` for defining an output file for the status report.
`--format` for setting the output format (only `string` and `json` supported for now)

I honestly don't know if it makes sense to include formats other than string or JSON.
For example, MD and HTML are not really machine-readable. YAML is not
a great standard format for this use-case. Open for discussions, though.

These features come in handy for the new [lychee Github action](https://github.com/lycheeverse/lychee-action/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1836,6 +1836,7 @@ dependencies = [
  "regex",
  "reqwest",
  "serde",
+ "serde_json",
  "shellexpand",
  "structopt",
  "tempfile",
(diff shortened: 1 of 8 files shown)


mre:
I'm gonna merge that right away because the implementation is pretty straightforward and I need the feature to make the lychee action more robust.
The only thing I'm not too happy about is the naming of the `fmt` function. Considered having something like `ResponseStatus::format`, but I didn't want to couple the formatting enum with the stats so I kept the two separate.
We could revisit that in the future, though.

PR #92, opened 2021-01-02
Fix HTML parsing for non-closed elements like <link>

pawroman:
Addresses #83 (does not solve all the uncovered issues, but fixes early parsing stop).

The XML parser we use requires all tags to be closed by default,
and if they aren't (like HTML5 `<link>` elements), it simply gives up
on further parsing.  This change makes it ignore such issues.

Also uncover a bug with the current parser (it simply won't parse
elements like `<script defer src="..."></script>`) -- e.g. elements
with no attribute values.

The XML parser is an XML parser and will have to be replaced with
HTML aware parser in the future.

### DIFF
--- a/fixtures/TEST_HTML5.html
+++ b/fixtures/TEST_HTML5.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


mre:
This is great! Thanks a lot for tackling that. 🐠

PR #93, opened 2021-01-02
Run clippy for all targets, including tests

pawroman:
The test code should also be linted.

Also fixes the code in tests that was failing the lints.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-  
+
   lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -37,7 +37,8 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings
+          # --all-targets makes it lint tests too
+          args: --all-targets -- --deny warnings
```
mre, line 41: Thanks for expanding `-D` to `--deny` here. Makes it easier to remember what each argument is doing. 👍

mre:
Thanks @pawroman. Nice catch.

PR #98, opened 2021-01-06
Use html5ever for HTML link extraction

pawroman:
html5ever is a HTML specific parser, compared to previous XML parser:

- Way more compatible with HTML constructs
- Finds a lot more links
- Doesn't require hacks or tweaks

TODO:

- [x] Add tests for HTML5 quirks, e.g. issues uncovered in #83 
- [x] Use wildcards for attribute matching


pawroman:
I have checked this against examples provided by @Michalng in #83 -- it seems to be on par with plain text extractor now, and as an added benefit should work with relative links which the plaintext mode doesn't do.

src/extract.rs
```diff
@@ -311,10 +332,7 @@ mod test {
             Uri::Website(Url::parse("https://example.com/images/icon.png").unwrap()),
             Uri::Website(Url::parse("https://example.com/css/style_relative_url.css").unwrap()),
             Uri::Website(Url::parse("https://example.com/css/style_full_url.css").unwrap()),
-            // TODO BUG: the JS link is missing because the parser can't properly deal
-            //           with `<script defer src="..."></script>` (tags that have attributes with no value)
-            // Uri::Website(Url::parse("https://example.com/js/script.js").unwrap()),
-
+            Uri::Website(Url::parse("https://example.com/js/script.js").unwrap()),
```
mre, line 335: So nice.

src/extract.rs
```diff
@@ -57,75 +59,94 @@ fn extract_links_from_markdown(input: &str) -> Vec<String> {
...
     }
-    urls
 }
 
-// Extracting unparsed URL strings from a plaintext
+/// Determine if element's attribute contains a link / URL.
+fn elem_attr_is_link(attr_name: &str, elem_name: &str) -> bool {
+    matches!(
+        (attr_name, elem_name),
+        ("href", "a")
```
mre, line 125: Just a thought: would wildcards work here? E.g. instead of `("href", "a")` we use `("href", _)`. So essentially we assume any `href` is a valid link. Same for `src` and friends. Can't think of any element where this is not the case (and I might be wrong). Or was the idea to be explicit in what we accept? 🤔 
[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=44b796a6d28f4402aa8aa2b59e91adff).

mre:
Very nice work so far!

src/extract.rs
pawroman: Nice idea!

You're right, `href` and `src` seem only to be used for links: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes -- I guess it wouldn't hurt to wildcard these attributes.

pawroman:
Ready for review now!

mre (approved):
lgtm :shipit:

mre:
Hope it's fine that I just went ahead and merged this. I'll tag a new release for it, 0.5.0. 🤖

pawroman:
> Hope it's fine that I just went ahead and merged this. I'll tag a new release for it, 0.5.0. 🤖

Nice, thanks!

PR #99, opened 2021-01-07
Add no check (dump links only) flag

pawroman:
I'm not very fixed on the flag's name, it's the best I could come up with after a few minutes of thinking about it. Suggestions for different names welcome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,8 @@ FLAGS:
         --glob-ignore-case       Ignore case when expanding filesystem path glob inputs
         --help                   Prints help information
     -i, --insecure               Proceed for server connections considered insecure (invalid TLS)
+        --no-check               Don't perform any link checking. Dump all the links extracted from inputs that would be
+                                 checked
     -n, --no-progress            Do not show progress bar.
                                  This is recommended for non-interactive shells (e.g. for continuous integration)
         --skip-missing           Skip missing input files (default is to error if they don't exist)
--- a/README.md
+++ b/README.md
@@ -174,8 +174,8 @@ FLAGS:
         --glob-ignore-case       Ignore case when expanding filesystem path glob inputs
         --help                   Prints help information
     -i, --insecure               Proceed for server connections considered insecure (invalid TLS)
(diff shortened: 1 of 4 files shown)


pawroman:
@mre any thoughts on this so far?

mre:
@pawroman apart from the `README.md` usage update (great feature BTW) this would be ready to merge from my side.

pawroman:
> @pawroman apart from the README.md usage update (great feature BTW) this would be ready to merge from my side.

Fixed. I have also reworded the help text a little tiny bit for clarity.

pawroman:
Actually, I'm having some second thoughts about the help text.

With current implementation, the `--exclude`s have no effect on the dumped links. 

@mre do you think that we should:
1. Leave it as is
2. Change the implementation to respect excludes etc
3. Adjust the help text to say that it's going to dump all links found in inputs (the exclude flags not affecting the output)
4. Mark each link in the output somehow, similar to `--verbose` (indicating which links would be skipped)

mre:
I'd go with option 2: Change the implementation to respect excludes.
The reasoning is that I think of the command as a "dry-run". If you agree, perhaps we should rename the flag to `--dry-run` to make it a little clearer?
Alternatively if our main goal is to test the link detection with that flag, we could skip the exclude parts and perhaps name the flag `--dump` or `--dump-links`.

mre:
Update: I moved the `exclude` behavior into a separate mod called `filter`. With that it should be a little easier to remove the excluded links before dumping them.

pawroman:
> Thanks. Haven't looked at it yet, but my suspicion is that it's because we do the parsing before the rejection of mail addresses in general.

Thanks for the heads up. Once I have some spare time, I will re-visit this PR with that in mind.

mre:
Renamed the flag to `--dump` now. Also removed the banner as it was printed to `stdout` on my machine and other Unix tools also don't have any header output for similar cases AFAIK.  It looks a bit cleaner to me.

What's missing is the exclude part.

mre:
> What's missing is the exclude part.

Added that now.

PR #134, opened 2021-02-08
Ensure destructors are ran before std::process::exit

pawroman:
See comments in code for more details.

This *might* be related to #131, but I'm not sure. Definitely worth a shot though.


mre:
That's great. Definitely cleaner, thanks!

PR #135, opened 2021-02-08
Mention GitHub Action in README.md

pawroman:
It needs some more publicity. :smile: 

Also mention GitHub token in TROUBLESHOOTING.md


mre:
> It needs some more publicity. 😄

Haha. I thought we mention it in the README.md somewhere, but we don't, so ✌️.

PR #143, opened 2021-02-15
 Show input source in status output

mre:
If an error occurs during link checking,
it is important to know where the error occurred.
Therefore the request and response objects now contain a the input
source as a field. This makes error tracking easier.

### DIFF
--- a/src/bin/lychee/stats.rs
+++ b/src/bin/lychee/stats.rs
@@ -128,9 +128,19 @@ mod test_super {
         expected_map.insert(
             Input::Stdin,
             vec![
-                website("http://example.com/failed"),
-                website("http://example.com/redirect"),
-            ],
+                Response {
+                    uri: website("http://example.com/failed"),
+                    status: Status::Failed(http::StatusCode::BAD_GATEWAY),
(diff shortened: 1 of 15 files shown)


src/bin/lychee/stats.rs
```diff
@@ -1,65 +1,137 @@
...
+            Error(_) => self.errors += 1,
+            _ => self.successful += 1,
         }
+
+        if matches!(
+            response.status,
+            Failed(_) | Timeout(_) | Redirected(_) | Error(_)
+        ) {
+            let fail = self.fail_map.entry(response.source).or_default();
+            fail.push(response.uri);
```
mre, line 57: This is a bit sad. It means that we don't show the response status in the output yet. That's because `Status` is an enum which wraps around `http::StatusCode`, which is not `Serialize`. Maybe we find a way to do that in the future. In the worst case we can push a String with the pre-formatted status code + the uri. I'd like to avoid that, though to keep the machine-readable output clean and parseable without grep/cut/...

mre: Fixed. We now use the `Response` type with full status/error output.

PR #145, opened 2021-02-18
Improve lychee ergonomics

mre:
Adds a new function `lychee::check()`, which removes
a lot of boilerplate for simple cases. Adjusted the code,
tests, and documentation.
The downside is that `check` now returns a Result, so
we have to use `?` to get to the response. That's because
we have to account for the case where the given string is
not a valid URI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,25 +205,35 @@ ARGS:
 ## Library usage
 
 You can use lychee as a library for your own projects.
-Simply add it as a dependency and build your client:
+Here is a "hello world" example:
 
 ```rust
-use lychee::{Request, Input, ClientBuilder, Status};
-use lychee::Uri::Website;
-use url::Url;
+use std::error::Error;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+  let response = lychee::check("https://github.com/lycheeverse/lychee").await?;
+  println!("{}", response);
+  Ok(())
+}
+```
+
+This is equivalent to the following snippet, in which we build our own client:
+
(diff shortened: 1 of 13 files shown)


mre:
So this PR includes a few more changes:

1. Add doc-comment tests
2. Change exclude behavior to make includes more intuitive
3. Improve progress visualization
4. Add support for colored output

About 2.:

Previously an include alone would not mean that only the included
patterns were checked. Only if an exclude was given as well,
the includes would make a difference. Now, the includes on their
own will work as expected.
Moved the exclude methods into the Exclude mod.
Also changed the order of exclude tests to do the fast lookup
ones before the regex ones.
Added tests to guarantee behavior in the future.

I could move those changes into a separate PR. Then again the benefit would be very low at this point.
So I'll go ahead and merge it as part of this existing PR.
What the changes all have in common is that they improve ergonomics in a wider sense, so I'll go ahead and update the title before merging.

README.md
```diff
@@ -149,7 +149,7 @@ token with no extra permissions is enough to be able to check public repos links
 There is an extensive list of commandline parameters to customize the behavior,
 see below for a full list.
 
-```
+```sh
```
pawroman, line 152: I think this is causing weird syntax highlight when it's rendered: https://github.com/lycheeverse/lychee#commandline-parameters

mre: true, let me fix it

mre: done in `master`.

pawroman: Awesome 😎

Issue #146, opened 2021-02-18
URL reference to other Markdown treated as E-Mail check


joesan:
I'm having a Markdown in which I refer to other Markdown files using the @ notation. For example., to link to another Markdown, I use 

`[please have a look here](@/posts/some/path/mk.md)`

When I run it through the action, the above gets checked as E-Mail. Is there a way to ignore such links or to interpret it as a link instead of an E-Mail?

mre:
If you want to exclude those links while still checking mail addresses, you can try 
```
--exclude '@/posts'
```

(Have not tested that, but it's probably close to what you want.)

If you want to skip all mail addresses from getting checked, try 

```
--exclude-mail          
```

mre:
That said, I'm not sure if these links should be detected at all. The `@` in front of a path is a common pattern in static site generators, so I wonder if we should treat them as differently, e.g. as relative URLs by stripping away the `@/` and applying `base-url` if set.

joesan:
I was thinking about that as excluding them might not be a good idea as these are indeed valid HTTP links pointing to other Markdown files. So I like your second approach where you check them by prepending a base-url if set, if not check for the file in the same location.

mre:
Cool, agreed. Will see what I can do.

joesan:
Could you post a reference to the source file that I should look at? I might try to get this implemented!

mre:
Oh that would be cool. Guess you could start by adding your snippet from above as a test right below this one:
https://github.com/lycheeverse/lychee/blob/ae2d02b8a0bb5e186478c5cf6a2d1ecad0c5ac64/src/extract.rs#L263-L268
Then you'd have to follow the trail and see where it fails.

mre:
And this here needs to be adjusted to check for '@/':
https://github.com/lycheeverse/lychee/blob/ae2d02b8a0bb5e186478c5cf6a2d1ecad0c5ac64/src/uri.rs#L53-L55

Maybe this will also be relevant:
https://github.com/robinst/linkify/pull/8#issuecomment-778605835
linkify only allows URLs with a schema at the moment. They are looking for help to push schema-less URLs over the finishing line.

joesan:
Fixed in PR https://github.com/lycheeverse/lychee/pull/150

PR #150, opened 2021-02-20
added implementation for #146

joesan:
# Check internal Markdown links and add it as a Website struct with a base_url
# Check for E-Mails and ignore them (added a unit test)


joesan:
@mre I'm not sure why it fails with the publish check. Could you please take a look and merge it in?

.gitignore
```diff
@@ -1 +1,14 @@
-/target
+# Generated by Cargo
```
mre, line 1: Nice. Thanks for adding those missing entries

src/uri.rs
```diff
@@ -44,15 +44,21 @@ impl TryFrom<&str> for Uri {
     type Error = anyhow::Error;
 
     fn try_from(s: &str) -> Result<Self> {
+        // We want to check for internal links
+        // For example., links to other internal Markdown
+        // Example input: '@/internal.md'
+        let is_internal_link = s.ends_with(".md");
```
mre, line 50: We have to cover a few more cases here:

* [`@/pages/about.markdown`](https://www.getzola.org/documentation/content/linking/)
* [`@/pages/about.md#example`](https://www.getzola.org/documentation/content/linking/)

These are just two from the top of my head. Probably more cases will come up in the future. I suggest to create a helper function 

```
fn is_internal_link(link: &str) -> bool
```

and add some unit tests for these cases. Then we can just use that function here.

src/extract.rs
```diff
@@ -267,6 +267,43 @@ mod test {
         assert_eq!(links, HashSet::new())
     }
 
+    #[test]
+    fn test_markdown_internal_url() {
+        let input = "This is [an internal url](@/internal.md).";
+        let links: HashSet<Uri> = match Url::parse("https://localhost.com/") {
```
mre, line 273: Why do a `match` statement here?
It might be needed but I don't get it. 😄 At least in the test below (`test_skip_markdown_email`) you didn't have a `match` because you knew the output already. Was expecting something like this here, too.
If you are worried about the `Result` type, you can just `unwrap()` it here. Since we are in a unit test that's fine as `unwrap()` will merely panic, which is what will be recognized as a failed test.

mre:
> @mre I'm not sure why it fails with the publish check.

That's because the current `master` branch has dependencies to crates not published on crates.io. As a result the `master` branch version can not be published on crates.io either. In short: it is to be expected. Will be resolved once https://github.com/async-email/async-smtp/pull/36 is merged and published.

src/uri.rs
joesan: Agreed. Where is the good place to put this function? Have got an idea?

mre: Since it's just a helper function you can put it into `src/uri.rs` as well. Just make sure that it's outside of the Uri struct as it's independent of that (to avoid close coupling).

joesan: I'll look into this tonight or early tomorrow!

src/extract.rs
joesan: Will fix this!

src/uri.rs
joesan: Added additional tests and few more cases

mre (approved):

mre:
LGTM! Thanks a lot @joesan. Merging. 😊

Issue #28, opened 2020-11-05
Add youtube and vimeo link check support


pawroman:
For youtube links, it's necessary to scrape the content of the website to work out if the video still exists. For vimeo, they don't really like fake user agents.

The https://www.youtube.com/oembed API will return 404 for missing videos and can be used to check youtube links.

We special case GitHub already, we could special case YouTube and Vimeo too. The mechanism to specialize these kinds of links should be made extensible though.

mre:
@pawroman and me discussed that we could have a `workarounds` or `overwrites` module for these cases.
Perhaps a structure like this would work:
```rust
struct Overwrite {
  match: regex::Regex,
  user_agent: String,
  headers: HashSet<Header>,
  // ...
}
```
We could think about more generic approaches, but I'd be careful not to generalize too early and only add the overwrites we need in the beginning. Any ideas?

mre:
I'm working on this in #156.

Tested vimeo and it seems to be okay with a fake user agent (?):

```
 curl -sIXGET  -A "totally-legit-useragent" https://vimeo.com/510814675
 ```

This gives me a 200.

```
 curl -sIXGET  -A "totally-legit-useragent" https://vimeo.com/510814675invalid
 ```

This gives me a 404 as expected.

Perhaps they changed something on their end?

mre:
This should be fixed now in `master` because #156 is merged. 🎉

PR #156, opened 2021-02-22
Add support for website quirks

mre:
This is an attempt at #28. 
I've added a structure called `Quirk` and a few tests for Twitter and YouTube.

~~It's very hacky right now and definitely a work-in-progress. Early feedback welcome.
Need to do a lot of cleanup before this can be merged.~~

This is not perfect, but I'd say it's okay to merge this now.
The YouTube oembed check doesn't work somehow; I get a 404 at least.
I don't know why, but we can revisit this in the future. Maybe you have an idea @pawroman? 😉


src/quirks/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+            Quirk {
+                // https://stackoverflow.com/a/19377429/270334
+                pattern: Regex::new(r"^(https?://)?(www\.)?(youtube\.com|youtu\.?be)").unwrap(),
+                rewrite: |request| {
+                    let mut out = request;
+                    let original_url = out.url();
+                    let urlencoded: String =
+                        url::form_urlencoded::byte_serialize(original_url.as_str().as_bytes())
+                            .collect();
+                    *out.method_mut() = Method::HEAD;
```
pawroman, line 41: It seems that making a GET request works here, while HEAD does not (verified with curl).

src/quirks/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub rewrite: fn(Request) -> Request,
+}
+
+#[derive(Debug, Clone)]
+pub struct Quirks {
+    quirks: Vec<Quirk>,
+}
+
+impl Quirks {
+    pub fn init() -> Self {
```
pawroman, line 20: Why not `new`?

If the intention was to indicate that it's going to "init standard quirks" then maybe we should be a bit more verbose with the name, e.g.

`pub fn standard_quirks() -> Self {`

It's just a nit-picking suggestion though.

pawroman (commented):
Nice work, looks solid. Left a few minor comments, but otherwise all good with me!

src/quirks/mod.rs
mre: You're right. Come to think of it, there is a default way to handle that scenario and that is... 🥁 🥁 🥁 the `Default` trait. 🎉 
So I'll implement that instead. `Quirks::init()` becomes `Quirk::default()` then, which is closer to what I'd like to convey.

src/quirks/mod.rs
mre: You're right. Checked on my machine and got the same results. Fixed now and added a unit test for verification. Thanks!

Issue #161, opened 2021-02-25
Medium.com links parsed as mail links


gj:
E.g.:

```
Errors in new-docs/content/any/project/performance/n_plus_one.md
↯ medium.com/@bretdoucette/n-1-queries-and-guides-avoid-them-a12f02345be5 (Invalid mail address: medium.com/@bretdoucette/n-1-queries-and-guides-avoid-them-a12f02345be5)
```

Presumably this is a bit too eager:

https://github.com/lycheeverse/lychee/blob/cefe38ee25d433dd3b7c61a96ccacbec66aa70a8/src/uri.rs#L59-L61

mre:
> Presumably this is a bit too eager
 
Exactly. Any idea on how to fix that check?
One idea would be to try and parse the part before the `@` as a URL. If that succeeds then assume that the entire link is a URL.

Similar to this:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=aff61feeddf28ae7084cda1015f2ef08

Although in this case the link does not begin with a protocol, so it counts as a relative URL without a base, which cannot be parsed by the `url` crate. (You can check by removing the `https://` in the example above.)

@joesan perhaps you have a better idea?

mre:
In the meantime you could use `--exclude-mail` in case you don't want to check mail addresses anyway. 😅

gj:
> Although in this case the link does not begin with a protocol, so it counts as a relative URL without a base, which cannot be parsed by the `url` crate. (You can check by removing the `https://` in the example above.)

The link actually does begin with `https://` — GitHub Actions UI helpfully (🙄) removes the protocol when formatting URLs.

> Exactly. Any idea on how to fix that check?
> One idea would be to try and parse the part before the `@` as a URL. If that succeeds then assume that the entire link is a URL.

If you only care about URLs that include a protocol, then that might work. But if `example.com` should be treated as a URL, there might be issues since `example.com@example.com` is a valid email address.

It looks like that `url` crate parses `mailto:` links just fine, so I would probably try something like:

1. Try parsing the input string with `Url::parse` (without manually stripping out `mailto:`).
2. If (1) fails, try parsing it as an email.
3. If (1) succeeds and the parsed `scheme` is `"mailto"`, try parsing the parsed `path` as an email, same as (2).
4. If (1) succeeds and the parsed `scheme` is `"http"`/`"https"`, great, we have a URL.
[...]

PR #164, opened 2021-02-27
Added partial fix for #161

joesan:
URL's that does not contain a scheme, but contains a @ are being treated as E-Mail. So added a Regex to check for E-Mail


mre:
Thanks for the PR and for helping out with #161 @joesan. Much appreciated!
To be honest I'm not totally sold on the regex implementation yet. E-Mails are notoriously hard to parse and me might miss an edge case. There is a great talk and a nice blog post about the topic [here](https://github.com/kdeldycke/awesome-falsehood#emails).
If we see this as a temporary workaround that will be removed once https://github.com/robinst/linkify/pull/8 is merged, it would be fine from my side. In that case we should add a comment to the pull request to the code so we won't forget. 😉
What's your opinion?

joesan:
> Thanks for the PR and for helping out with #161 @joesan. Much appreciated!
> To be honest I'm not totally sold on the regex implementation yet. E-Mails are notoriously hard to parse and me might miss an edge case. There is a great talk and a nice blog post about the topic [here](https://github.com/kdeldycke/awesome-falsehood#emails).
> If we see this as a temporary workaround that will be removed once [robinst/linkify#8](https://github.com/robinst/linkify/pull/8) is merged, it would be fine from my side. In that case we should add a comment to the pull request to the code so we won't forget. 
> What's your opinion?

Valid point. E-Mail addresses are meant to be created such that they are simple enough to be addressed. I see that the RFC allows for an E-Mail like this:

`phil.h\@\@ck@haacked.com`

But I would not want to send any E-Mail to such an address that one looks like a hacking endpoint. But nevertheless, you are right. In this case, I would not merge this PR as I do not like adding something with comments saying that this will be removed after something else is fixed. So going by the strict E-Mail validation using the RFC, I would say let us skip this PR.

mre:
Okay then, let's close this. The unit tests might be valuable in the future, so we should revisit this in the future.

Issue #78, opened 2020-12-14
Add recursive option


styfle:
It would be nice to pass a URL and have it crawl the entire website recursively looking for dead links.

In order to avoid crawling the entire internet, it should stop recursing once a request no longer matches the original domain.

mre:
Yeah it has been discussed a few times already and it's high on the todo-list. 
Will require quite some restructuring I guess.
Right now the flow is

```
main -> extractor -> channel -> client (link checker) -> main
```

but there is no connection back to the extractor.

frederickjh:
The lack of recursive spidering make this project unusable for my purpose of checking all links internal and external on a website. I am trying to find a replacement for  [michaeltelford/broken_link_finder ](https://github.com/michaeltelford/broken_link_finder). It is written in Ruby and with out super user access a the new place where this will run it is impossible to install.  I am looking for a "portable" replacement. I worked with michaeltelford to get his project into a much more usable state. Check out that projects issue queue for some of the reasoning that went into the development.

In any case regarding this issue. The spidering should stop with links found on the current domain, but the links found to external sources should still be checked.

mre:
#165 is getting very close to completion. It implements the functionality described. If you like to support, please build the version from that branch and test it. Feedback on the pull request is appreciated.

frederickjh:
@mre I am new to rust but it seemed pretty straight forward as to how to build from [Working on an Existing Cargo Package](https://doc.rust-lang.org/cargo/guide/working-on-an-existing-project.html). However I have run into an issue. At first I thought it was a credential issue but it looks like a 404 issue.

![image](https://user-images.githubusercontent.com/864159/112447909-2145f680-8d52-11eb-81eb-cac2454cca6c.png)

> Caused by:
>  Unable to update https://github.com/amaurym/async-smtp?branch=am-fast-socks#eac57391

That GitHub URL returns a 404 so I am not sure how to proceed to build.

Rust information:
```shell
$ rustc --version
rustc 1.50.0 (cb75ad5db 2021-02-10)
$ rustup --version
rustup 1.23.1 (3df2264a9 2020-11-30)
info: This is the version for the rustup toolchain manager, not the rustc compiler.
info: The currently active `rustc` version is `rustc 1.50.0 (cb75ad5db 2021-02-10)`
$ cargo --version
cargo 1.50.0 (f04e7fab7 2021-02-04)
```

The GitHub repository **amaurym/async-smtp** does not seem to exist anymore.
[...]

PR #165, opened 2021-02-28
Recursion Support (closes #55, #78)

### LINKED ISSUE #55: Introduce extractor pool
## Introduction

As of now we send each URI we'd like to check to a client pool in `main.rs`. Code is here:
https://github.com/lycheeverse/lychee/blob/d2e349cb29606d3a4a999dcf5dce1b5108a8c0f9/src/main.rs#L128-L135

This is not ideal for a few reasons:

- All links get extracted on startup. This is a slow process that can take up to a few seconds for long link lists.
  It's not necessary to block the client during this step, though as we could lazy-load the links on demand from the inputs.
- There is no clear separation of concerns between `main` and the link extraction. Ideally the responsibilities could should be split up to make testing and refactoring easier.

We already use a channel for sending the links to check to the client pool. We could use the same abstraction for extracting the links, too in form of an extractor pool.

In the future this would allow implementing some advanced features in an extensible way:

- Recursively check links: Push newly discovered websites into the input channel of the extractor pool
- Skip duplicate URLs: Filter input links with a `HashSet` or even a Bloom filter (for constant memory-usage) that is maintained by the extractor pool before sending it to the client pool.
- Request throttling: Group requests per website and apply some throttling to not overload the server.

## How to contribute

1. Create an extractor pool similar to our [client pool](https://github.com/lycheeverse/lychee/blob/master/src/client_pool.rs)
2. Spawn the pool inside `main` on startup, pass the channel to the pool and start processing the inputs.

(The other end of the channel the channel is already passed to the client pool.)

mre:
⚠️ This is on hold as we started implementing a stream-based approach in #330, which might supersede this branch soon.  Recursion support will be added once this gets merged. We can probably rebase it on top of `master` then and make adjustments in this PR (or create new PR). Closing this for now.

----

This is a super basic recursion implementation of recursion as discussed in #78.
It doesn't use an extractor pool as discussed in #55. I think it's not needed at this point.
The current implementation just checks the responses, filters out websites, and pushes them back into the request queue.

I deliberately avoided to add the recursion support into the client. It would break separation of concerns. Also didn't want to add a separate recursion handler to the library as it I wouldn't expect any gains from that. Recursion handling is surprisingly tricky, because there are quite a few design decisions to make. (Recurse indefinitely or stop at a predefined recursion depth? Recurse on all domains or just the set of input domains?)
So most people using the library would have to make the decisions based on their use-case. The current recursion depth is part of the request/response struct, so it won't have to be wrapped in another struct at least.

TODO:
* [X] Cache requests
* [X] Filter out domains that were not in the set of inputs
* [x] Check recursion level
* [x] Find out why recursion does not terminate
* [x] Cleanup
* [ ] Integration test?


joesan:
Could you let me know what this task is about? The title does give me a hint, but I fail to understand the context?

mre:
Sorry, forgot to link the issue. Here it is: #78.

mre:
From my side this should be good to go.
@pawroman, @joesan any final comments before merging this?

src/bin/lychee/options.rs
```diff
@@ -245,6 +245,15 @@ pub struct Config {
...
     pub format: Format,
+
+    /// Enable recursion (make sub-requests for detected links)
+    #[structopt(short, long)]
+    #[serde(default)]
+    pub recursive: bool,
+
+    /// Stop link checking beyond this maximum recursion depth. (Recommended for large inputs.)
+    #[structopt(long)]
+    pub depth: Option<usize>,
```
pawroman, line 256: I'd consider allowing this to be signed (e.g. `isize`), so that users could explicitly express the need for infinite recursion with numbers less than zero (e.g. `-1`).

src/bin/lychee/stats.rs
```diff
@@ -146,11 +149,13 @@ mod test_super {
                     uri: website("http://example.org/failed"),
                     status: Status::Failed(http::StatusCode::BAD_GATEWAY),
                     source: Input::Stdin,
+                    recursion_level: 0,
                 },
                 Response {
                     uri: website("http://example.org/redirect"),
                     status: Status::Redirected(http::StatusCode::PERMANENT_REDIRECT),
                     source: Input::Stdin,
+                    recursion_level: 0,
```
pawroman, line 158: For convenience and to cover the "most common case", it might be worth considering adding an alternative constructor for `Response` which would set `recursion_level` to 0.

Or even have `::new` constructor which would have this behavior (recursion = 0) and another one named something like `::with_recursion` which would allow setting recursion level.

src/lib.rs
```diff
@@ -1,37 +1,37 @@
...
-* #[tokio::main]
-* async fn main() -> Result<(), Box<dyn Error>> {
-*   let client = ClientBuilder::default().build()?;
-*   let response = client.check("https://github.com/lycheeverse/lychee").await?;
-*   assert!(matches!(response.status, Status::Ok(_)));
-*   Ok(())
-* }
-* ```
-*/
+//!
```
pawroman, line 1: Great to make this a doc-comment!

But isn't a blank line here creating a newline in the rendered docs? Just curious.

src/bin/lychee/main.rs
```diff
@@ -184,6 +231,68 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
...
+        if recursion_level > depth {
+            // Maximum recursion depth reached; stop link checking.
+            return Ok(0);
+        }
+    }
+
+    if !response.status.is_success() {
+        return Ok(0);
+    }
+    if cache.contains(response.uri.as_str()) {
```
pawroman, line 254: Nitpick: I'm a bit skeptical of the name `cache` here. We're not really caching, we're merely "checking if this URI was checked before, and should therefore be skipped".

src/bin/lychee/main.rs
```diff
@@ -184,6 +231,68 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
...
+        return Ok(0);
+    }
+    cache.insert(response.uri.to_string());
```
pawroman, line 257: Since we're allowing *infinite* recursion, this data structure can really take up the entire memory with sufficient number of links. Depending on the expected maximum number of links checked in recursive mode, we might consider a [bloom filter](https://en.wikipedia.org/wiki/Bloom_filter) here:
- HashSet is probably fine if the number of links checked is in the order of millions or fewer
- Bloom filter might be a good idea if the number of links checked is in the order of tens of millions or more

---

Super-rough calculations below. (Note that I'm just calculating the size of this one structure, there's probably going to be more memory overheads from various bits of the program).

Assumptions:
- 64 bit machine, e.g. 8 bytes per pointer
- 64 bytes per link on average
- 1 million links

Calculations:

- Link string storage: 64 MB
- String overhead (24 bytes per string): 24 MB
  source: https://cheats.rs/#owned-strings
- HashSet overhead is 1 byte per entry: 1MB
  source: https://github.com/rust-lang/hashbrown#features

So, HashSet with 1 million unique links will take up ~64+24+1 = ~89 MB (~85 MiB)

For comparison, a Bloom Filter with ~1e-7 false positive probability is expected to take up:
- ~4MiB for 1 million links: https://hur.st/bloomfilter/?n=1M&p=1.0E-7&m=&k=
- ~80MiB for 20 million liks: https://hur.st/bloomfilter/?n=20M&p=1.0E-7&m=&k=
- ~4GiB for 1 billion links: https://hur.st/bloomfilter/?n=1G&p=1.0E-7&m=&k=

So, to be honest, HashSet should be absolutely fine for the most users. But we should make a remark in the README that memory usage when using recursive mode with tens millions of links might cause RAM usage issues.

pawroman (changes requested):
Nice work! Great to see this implemented.

I have left a few code comments and I have one generic comment: I think we should add at least one simple test for the recursive link checking.

src/bin/lychee/main.rs
lenisha: Hi, would it be merged soon? and Docker image updated with the change to make GH action use the latest?

mre:
TBH I'm not super happy with the current impl anymore as I count the links in the queue and then close the channel after all links got checked. It can lead to subtle bugs I think.
There must be a better way. Does anyone have an idea?

src/bin/lychee/main.rs
mre: @pawroman: Good idea with the bloom filter. I'll add that as soon as #208 is merged.
@lenisha: Hopefully yes. I don't like the current implementation anymore (see comment below), but I think it's not gonna be a blocker either way.

src/bin/lychee/main.rs
mre: How about `seen`?

src/lib.rs
mre: Didn't check yet, but I assumed cargo would be smart enough to trim empty lines at the beginning and the end.

lebensterben:
We can have a dedicated `Collector` type that collects links.

- First a channel for `Input` is created. The `Collector` is the receiver end and we initially send some `Input` to the channel.
- While `Collector` is still receiving new `Input`, it sends `Request`. (If max recursive level is not reached, etc)
- While `ClientPool` is still receiving new `Request`, it spawns `Client`s to validate the `Request`.
- Once `Client` finished validation, it sends the `Response`, which contains `Input`.
- The `Input` in the `Response` is fed back to the channel for `Collector` to receive.

src/bin/lychee/main.rs
```diff
@@ -160,9 +183,33 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
...
+    let input_domains: HashSet<String> = input_domains(inputs);
+
+    // We keep track of the total number of requests
+    // and exit the loop once we reach it.
+    // Otherwise the sender would never be dropped and
+    // we'd be stuck indefinitely.
+    let mut curr = 0;
+
+    while curr < total_requests {
+        curr += 1;
```
mre, line 195: @lebensterben, I'm not too happy about the above three lines. Is there a better way to handle the request queue compared to counting the number of total requests like I did?

mre:
Yup, that's correct.
More generally though, is there a better way to handle the request queue. (Added a comment to the code in question.)

src/bin/lychee/main.rs
lebensterben: Check this
https://docs.rs/tokio-stream/0.1.5/tokio_stream/#returning-a-stream-from-a-function
more examples:
https://tokio.rs/tokio/tutorial/streams

lebensterben: tokio-stream crate has a wrapper type for Reciever of the channel.
That should work without counting the number.
It's like Future trait, you can poll it.
It's also like Iterator trait, `next` returns None when it's over.

mre: Ah that's great. Sounds like what we need.

mre:
Will put this on hold once again as we started implementing a stream-based approach in #330, which might supersede this branch soon. Sorry to everyone waiting on recursion support to land, but I'd like to get this right instead of merging a buggy solution prematurely.

PR #177, opened 2021-03-13
Fix URLs with '@' parsing as emails

### LINKED ISSUE #161: Medium.com links parsed as mail links

pawroman:
Only consider a link an email if it fails to parse as URL.

Also use a proper email validation instead of a simple '@' check.

This uses the fast_chemail crate which parses email links according
to the HTML specification (which is much more practical than checking
for RFC 5322 formatted emails).  It's also worth noting that
fast_chemail is used internally (albeit indirectly) by the
check_if_email_exists crate.  This means that email addresses
not considered valid by fast_chemail wouldn't pass link checks
anyway.

Fixes #161.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1381,6 +1381,7 @@ dependencies = [
  "deadpool",
  "derive_builder",
  "doc-comment",
+ "fast_chemail",
  "futures",
  "glob",
  "headers",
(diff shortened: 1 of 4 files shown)


mre:
Brilliant PR. Thanks so much @pawroman! ⭐

PR #178, opened 2021-03-13
Avoid extra newlines in output

pawroman:
Ensure that no extra newlines are printed in the output.

See comparison below:

| Before | This MR |
|---|---|
| empty-file-before | empty-file-after |
| non-verbose-before | non-verbose-after |
|  verbose-before | verbose-after |


mre (approved):

mre:
Love it. Thanks!

PR #197, opened 2021-03-28
Assume HTML in case there is no extension (e.g. for URLs)

mre:


mre:
Current error (https://github.com/lycheeverse/lychee/pull/197/checks?check_run_id=2214388798):

```
thread 'collector::test::test_collect_links' panicked at 'assertion failed: `(left == right)`
  left: `{Website(Url { scheme: "https", username: "", password: None, host: Some(Domain("test-url.org")), port: None, path: "/", query: None, fragment: None }), Website(Url { scheme: "https", username: "", password: None, host: Some(Domain("test-glob-1.io")), port: None, path: "/n", query: None, fragment: None }), Website(Url { scheme: "https", username: "", password: None, host: Some(Domain("test-file.io")), port: None, path: "/n", query: None, fragment: None }), Website(Url { scheme: "http", username: "", password: None, host: Some(Domain("test-string.com")), port: None, path: "/", query: None, fragment: None }), Mail("test@glob-2.io")}`,
 right: `{Website(Url { scheme: "http", username: "", password: None, host: Some(Domain("test-string.com")), port: None, path: "/", query: None, fragment: None }), Website(Url { scheme: "https", username: "", password: None, host: Some(Domain("test-file.io")), port: None, path: "/", query: None, fragment: None }), Website(Url { scheme: "https", username: "", password: None, host: Some(Domain("test-glob-1.io")), port: None, path: "/", query: None, fragment: None }), Website(Url { scheme: "https", username: "", password: None, host: Some(Domain("test-url.org")), port: None, path: "/", query: None, fragment: None }), Mail("test@glob-2.io")}`', src/collector.rs:308:9
```


Super confused why `path` is `"/n"` instead of `"/"`. If anyone has an idea, let me know!

src/extract.rs
```diff
@@ -77,9 +77,7 @@ fn extract_links_from_html(input: &str) -> Vec<String> {
 fn walk_html_links(mut urls: &mut Vec<String>, node: &Handle) {
     match node.data {
         NodeData::Text { ref contents } => {
-            // escape_default turns tab characters into "\t", newlines into "\n", etc.
-            let esc_contents = contents.borrow().escape_default().to_string();
```
mre, line 81: @pawroman, was there a reason why we call `escape_default` here? Removed them because they cause problems with links at the end of a line.
If we need the calls for some other purpose, we can add them back.

Cargo.toml
```diff
@@ -69,6 +69,7 @@ predicates = "1.0.7"
 uuid = { version = "0.8.2", features = ["v4"] }
 tempfile = "3.2.0"
 doc-comment = "0.3.3"
+pretty_assertions = "0.7.1"
```
pawroman, line 72: Nice!

src/extract.rs
```diff
@@ -204,6 +203,9 @@ mod test {
 
     #[test]
     fn test_file_type() {
+        // Assume HTML in case there is no extension (that's including URLs
+        // without a path)
```
pawroman, line 207: What about plain text UNIX files? I know this is being picky, but something like plain-text README would be unnecessarily parsed as HTML.

How about this: default to HTML for URLs and to plaintext for paths/files?

src/extract.rs
pawroman: I think this was to avoid having un-escaped newline, tab, etc. characters in URLs.

To be fair, I recall seeing it in html5ever [example code](https://github.com/servo/html5ever/search?q=escape_default) and just assuming it was the reasonable thing to do.

I guess if no tests are failing then we can remove this, unless a reasonable counter-argument could be found. I assume that the newline and tab characters will be url-escaped anyway.

src/extract.rs
mre: You're right. Fixed that now.

Issue #183, opened 2021-03-16
relative broken links not detected


lenisha:
Relative Links in the github pages index.html are broken but not detected:

https://github.com/lenisha/linkchecker-test/runs/2122896789?check_suite_focus=true
```
+ lychee '--exclude=^(javascript|chrome):.*' --verbose --accept=200,403,429 --base-url https://lenisha.github.io/linkchecker-test/   https://lenisha.github.io/linkchecker-test/ --output /tmp/lychee/out.md
+ exit_code=0
+ '[' 0 -ne 0 ']'
+ cat /tmp/lychee/out.md
+ echo
+ echo ::set-output name=exit_code::0
📝 Summary
---------------------
🔍 Total...........17
✅ Successful......17
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
🚫 Errors...........0
```

while  index.thml has broken link https://github.com/lenisha/linkchecker-test/blob/main/index.html#L53
```
<a class="nav-link pr-3" href="./abouterror.html">About</a>
```
that points to https://lenisha.github.io/linkchecker-test/abouterror.html

Could it be also checked locally? Similar to discussion #15

lenisha:
@mre  is there any option to detect problems with relative links right now?
I tried to go against the live website with base-url
```
lychee '--exclude=^(javascript|chrome):.*' --verbose --base-url https://lenisha.github.io/linkchecker-test/   https://lenisha.github.io/linkchecker-test/ 
```

and tried against files files in PR
```
lychee  --exclude="^(javascript|chrome):.*"   --accept=200,403,429 "**/*.md" "**/*.html"
```

non have found the simple error
```
<a class="nav-link pr-3" href="./abouterror.html">About</a>
```

pawroman:
> @mre is there any option to detect problems with relative links right now?

There should be, once https://github.com/lycheeverse/lychee/pull/99 is merged. Right now, link-parsing debugging is a bit cumbersome (e.g. you have to modify the source).

mre:
No clue why it doesn't find the relative links, though. @pawroman do you know why?

lenisha:
weird thing is when I explicitly point to index.html page it finds it
```
lychee --verbose '--exclude=^(javascript|chrome):.*' --accept=200,403,429 --base-url https://lenisha.github.io/linkchecker-test/ https://lenisha.github.io/linkchecker-test/index.html --output /tmp/lychee/out.md
+ exit_code=2
[...]

Issue #82, opened 2020-12-15
Improve library support


mre:
In #64 @drahnr mentioned a few improvement areas for library support:

* [ ] Reduce the exposed module structure - it's better to have fewer modules when each of them only contains a single or two entities
* [ ] Provide true errors (one that impl's `Send + Sync + 'static + std::error:Error`) rather than using anyhow. So use `thiserror` instead of `anyhow`.
* [ ] Currently one example is provided for the simple case. It would be nice if that was shown as the entry one for docs.rs and expanded in order to showcase the usage of `InputContent` with collect_links. We should create an examples folder which contains this and other use-cases.
* [ ] Provide the byte- and character ranges of the link that was being checked, would be very helpful in order to craft meaningful error messages.
* [ ] At the moment the library depends on tokio. Make the library executor-agnostic and not rely on any particular runtime.

## Contributing

Contributions very welcome.
Add a comment here if you'd like to tackle one of the points above or simple send a pull request right away. 😃

mre:
@drahnr mentioned that this is how you get byte ranges with cmark: https://github.com/drahnr/cargo-spellcheck/blob/master/src/documentation/markdown.rs#L90-L92 and as cmark operates on characters, you essentially need to consume the characters upon the start byte bound and count the chars within the byte range to get a char range.

untitaker:
I have no plans to work on this, but I will assist with PRs somebody else makes

PR #208, opened 2021-04-03
refactor

lebensterben:
Changes:

- The binary component and library component are separated as two packages in the same workspace.
  - `lychee` is the binary component, in `lychee-bin/*`.
  - `lychee-lib` is the library component, in `lychee-lib/*`.
  - Users can now install only the `lychee-lib`, instead of both components, that would require fewer dependencies and faster compilation.
  - Dependencies for each component are adjusted and updated. E.g., no CLI dependencies for `lychee-lib`.
  - CLI tests are only moved to `lychee`, as it has nothing to do with the library component.
- `Status::Error` is refactored to contain dedicated error enum, `ErrorKind`.
  - The motivation is to delay the formatting of errors to strings. Note that `e.to_string()` is not necessarily cheap (though trivial in many cases). The formatting is no delayed until the error is needed to be displayed to users. So in some cases, if the error is never used, it means that it won't be formatted at all.
- Replaced `regex` based matching with one of the following:
  - Simple string equality test in the case of 'false positivie'.
  - URL parsing based test, in the case of extracting repository and user name for GitHub links.
  - Either cases would be much more efficient than `regex` based matching. First, there's no need to construct a state machine for regex. Second, URL is already verified and parsed on its creation, and extracting its components is fairly cheap. Also, this removes the dependency on `lazy-static` in `lychee-lib`.
- `types` module now has a sub-directory, and its components are now separated into their own modules (in that sub-directory).
- `lychee-lib::test_utils` module is only compiled for tests.
- `wiremock` is moved to `dev-dependency` as it's only needed for `test` modules.
- Dependencies are listed in alphabetical order.
- Imports are organized in the following fashion:
  - Imports from `std`
  - Imports from 3rd-party crates, and `lychee-lib`.
  - Imports from `crate::*` or `super::*`.
- No glob import.
- I followed suggestion from `cargo clippy`, with `clippy::all` and `clippy:pedantic`.

---

Benchmark results of running the refactored version and the current origin/HEAD version is listed below.
The sample size is very small (n=6)
Frankly, I don't know which part contribute to most of improvement.


lebensterben:
The data is obtained by `heaptrack`, and I did some t-tests.

My conclusion is there's no significant improvement in speed, but in memory usage. 


The difference in run-time is insignificant.
```
data:  c(126.91, 129.91, 133.03, 130.06, 126.17, 128.88) and c(125.21, 137.37, 111.35, 128.21, 127.47, 118.25)
t = 1.1916, df = 5.7553, p-value = 0.2802
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -4.854592 13.887925
sample estimates:
mean of x mean of y
 129.1600  124.6433
```

My refactored version has 3.22% fewer calls to memory allocation in total.
```
data:  c(5290434, 5240945, 5275566, 5263350, 5262841, 5244248) and c(5434387, 5309354, 5455931, 5449631, 5484685, 5496514)
t = -6.1582, df = 5.7684, p-value = 0.000975
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -245945.6 -105093.8
sample estimates:
mean of x mean of y
  5262897   5438417
```

The memory allocation calls per second is less significant, and it's 7.03% fewer.
```
calls/s:
data:  c(41687, 40341, 39658, 40470, 41713, 40689) and c(43403, 38649, 48996, 42505, 43025, 46480)
t = -2.0728, df = 5.5129, p-value = 0.0877
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -6802.6509   635.9842
sample estimates:
mean of x mean of y
 40759.67  43843.00
```

The difference in temporary memory allocation in total and per second are not significant
```
data:  c(510614, 604372, 515672, 514999, 517575, 501116) and c(522805, 450923, 525175, 513906, 524612, 517626)
t = 0.93237, df = 9.3099, p-value = 0.3747
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -25758.59  62192.25
sample estimates:
mean of x mean of y
 527391.3  509174.5
```

```
data:  c(4023, 3882, 3876, 3959, 4102, 3888) and c(4175, 3282, 4716, 4008, 4115, 4377)
t = -0.79212, df = 5.371, p-value = 0.4618
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -656.7887  342.4554
sample estimates:
mean of x mean of y
 3955.000  4112.167
```

My refactored version has 12.79% fewer peak memory usage, and it's very significant.
```
peak heap:
data:  c(29.88, 29.43, 30.81, 31.18, 30.89, 31.02) and c(33.69, 35.33, 34.23, 34.08, 35.28, 36.04)
t = -9.0237, df = 9.4371, p-value = 6.048e-06
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -5.295475 -3.184525
sample estimates:
mean of x mean of y
   30.535    34.775
```

My refactored version has 4.14% fewer peak resident memory.
```
data:  c(108.36, 111.94, 111.99, 108.42, 118.81, 107.58) and c(114.86, 116.04, 115.83, 116.75, 116.85, 115.59)
t = -2.7615, df = 5.3159, p-value = 0.03724
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -9.195788 -0.410879
sample estimates:
mean of x mean of y
 111.1833  115.9867
```


The only **downside**, I have significantly much more leaked memory (25.90% increase).
My guess is this probably is caused by my usage of `Box<dyn Error>` (as a result of refactoring `Status::Error`) or static reference of `RegexSet`.

This should not be a big deal though. And notice that `heaptrack`'s analysis on memory leak is not as fine as `valgrind`, so it could overestimate.


```
data:  c(279.68, 301.38, 299.75, 322.15, 300.96) and c(242.67, 251.02, 240.71, 232.52, 232.52, 234.05)
t = 8.4081, df = 5.572, p-value = 0.0002269
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 43.52331 80.21469
sample estimates:
mean of x mean of y
  300.784   238.915
```

lebensterben:
In the latest forced push, I replaced regex based matching with url parsing based matching.
Note that we stored URL in URI, where URL is parsed and verified when created.
And extracting host, path, or any other components from URL is cheap.

So for example, to extract repo and user name from a github URL, there's really no need to use regex at all.

lebensterben:
Updated benchmark with a few more optimizations (replacing regex with string matching and etc):

```
$total_runtime

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = -1.0993, df = 17.69, p-value = 0.2864
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -10.401486   3.261486
sample estimates:
mean of x mean of y
  125.394   128.964


$calls_to_allocation_functions

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = -1.2081, df = 10.879, p-value = 0.2526
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -172992.12   50492.72
sample estimates:
mean of x mean of y
  5312026   5373276


$calls_to_allocation_functions_per_second

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = 0.72679, df = 16.662, p-value = 0.4774
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -1386.879  2841.079
sample estimates:
mean of x mean of y
  42483.7   41756.6


$temporary_memory_allocations

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = 2.7132, df = 16.527, p-value = 0.01503
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
  2887.776 23282.224
sample estimates:
mean of x mean of y
 520759.9  507674.9


$temporary_memory_allocations_per_second

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = 1.9783, df = 17.864, p-value = 0.06353
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -13.54774 446.54774
sample estimates:
mean of x mean of y
   4164.4    3947.9


$peak_heap_memory_consumption_MB

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = -0.37202, df = 11.858, p-value = 0.7164
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -1.414081  1.002081
sample estimates:
mean of x mean of y
   34.785    34.991


$`peak_RSS_(including_heaptrack_overhead)_MB`

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = -0.41003, df = 10.717, p-value = 0.6899
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -3.958803  2.718803
sample estimates:
mean of x mean of y
  116.284   116.904


$total_memory_leaked_KB

	Welch Two Sample t-test

data:  SD[ID == "develop", ..var] and SD[ID == "master", ..var]
t = -19.724, df = 9.0098, p-value = 1.012e-08
alternative hypothesis: true difference in means is not equal to 0
95 percent confidence interval:
 -41.89709 -33.27691
sample estimates:
mean of x mean of y
  197.761   235.348
```

mre:
This is great! Love the improvements. 👏
One thing: can we keep lychee for the name of the library instead of lychee-lib? I'd like to keep the old name because it looks a bit nicer when calling it. 🙂

lebensterben:
Need to do more investigation on the performance changes.

Then should `lychee-bin` be the name for the binary?

Also, the CI would fail unless the library component is actually published.

mre:
> Need to do more investigation on the performance changes.

Thanks. 👍 

> Then should lychee-bin be the name for the binary?

Yes, please.

> Also, the CI would fail unless the library component is actually published.

All good. Just let me know when this is ready for review and I'll try it locally and then we can merge this.

lebensterben:
figure1

It looks good, but there are a few outliers.

lebensterben:
To replicate, you need to install [heaptrack](https://github.com/KDE/heaptrack)

```bash
git checkout develop
cargo clean && cargo build --release --all
mkdir develop
cd develop
for _i in {0..9}; do heaptrack ../target/release/lychee --no-progress -E https://wiki.archlinux.org/index.php/List_of_applications &> /dev/null; done
popd

git checkout master
cargo clean && cargo build --release
mkdir master
cd master
for _i in {0..9}; do heaptrack ../target/release/lychee --no-progress -E https://wiki.archlinux.org/index.php/List_of_applications &> /dev/null; done
popd

echo "ID, total_runtime, calls_to_alloc, calls_to_alloc_per_second, temp_mem_alloc, temp_mem_alloc_per_second, peak_heap_mem_MB, peak_RSS_MB, total_mem_leaked_KB" > data.csv
for _i in ./develop/heaptrack*; do echo "ID: developer"; heaptrack_print $_i | tail -n 6; done >> data.csv
for _i in ./master/heaptrack*; do echo "ID: master"; heaptrack_print $_i | tail -n 6; done >> data.csv
```

The next part is data cleaning. You can do that in `vim`
- `:2,$s/[[:alnum:]| ()]*: //`
- `:2,$s/s\.$\|MB$\|KB$//`
- `:2,$s/\ (/\r/`
- `:2,$s/\/s)$//`
- `:2,$j`
- `:2s/ \(developer\|master\)/\r\1/g`
- `:2,$s/ /, /g`

Next is to generate the output, you need `R` and install a few packages

```R
install.packages(c("ggplot2", "ggpubr", "stringi", "data.table", "rlang"))

library(ggplot2)
library(ggpubr)
library(stringi)
library(data.table)
library(rlang)

DT <- data.table::fread(file = "data.csv")

pps <-
  lapply(names(DT)[-1],
         function (var) {
           df <- DT[, .SD, .SDcols = var, by = ID]
           caption <-
             DT %>% wilcox_test(new_formula(sym(var), sym("ID"))) %>% get_test_label(detailed = TRUE)
           names(df) <- c("ID", "Measure")
           ggplot(df, aes(x = ID, y = Measure, color = ID)) +
             geom_boxplot() +
             geom_jitter() +
             labs(x = "", y = "", caption = caption) +
             theme_minimal() +
             theme(legend.position = "none")
         })

labels <-
  sapply(names(DT)[-1],
         function (name) {
           name %>%
             stri_replace_all_fixed("_", " ") %>%
             stri_trans_totitle() %>%
             stri_replace_last_fixed("b", "B")
         },
         USE.NAMES = FALSE)

pp <-
  eval(expr(
    ggarrange(
      !!!pps,
      labels = labels,
      ncol = 2,
      nrow = 4,
      font.label = list(size = 10)
    )
  ))

ggexport(
  pp,
  filename = "figure1.png",
  width = 1000,
  height = 1400,
  res = 100
)
```

lebensterben:
figure1

When running on test data (those under `fixtures`), my refactored version is way much better.

lebensterben:
I'm replacing ad-hoc errors (via anyhow) to structural errors (via this-error).
The biggest disadvantage of anyhow is that errors are basically strings which is readable to human readers, but the computer library is not able to differentiate different types of errors.

mre:
> I'm replacing ad-hoc errors (via anyhow) to structural errors (via this-error).

Yeah that was on the list anyway. Thanks for tackling this as well. 😀

lebensterben:
@mre 

I had handwritten the errors without using `this-error`.
You can take a look.

As for the naming:
If you want the library component to keep the name `lychee`, it will cause breaking error in CIs and other applications that uses `cargo install lychee`. That's why I have the binary component to take the name.

lebensterben:
If this looks good I will do some final clean ups.

lychee-lib/src/client.rs
```diff
@@ -168,92 +153,70 @@ impl ClientBuilder {
...
+    pub async fn check<T, E>(&self, request: T) -> Result<Response>
+    where
+        Request: TryFrom<T, Error = E>,
+        ErrorKind: From<E>,
+    {
+        let Request { uri, source } = Request::try_from(request)?;
+        let status = match uri {
+            _ if self.filter.is_excluded(&uri) => Status::Excluded,
+            Uri::Website(_) => self.check_website(&uri).await,
+            Uri::Mail(_) => self.check_mail(&uri).await,
```
mre, line 171: Love those idiomatic improvements.

lychee-lib/src/client.rs
```diff
@@ -262,35 +225,32 @@ impl Client {
...
-            Some(result) => {
-                // Accept everything that is not invalid
-                !matches!(
-                    result.is_reachable,
-                    check_if_email_exists::Reachable::Invalid
-                )
-            }
+        let result = results.get(0).unwrap();
+
+        // TODO: We should not be using a HTTP status code for mail
```
mre, line 233: Guess this TODO can be removed now, because you fixed it.

lychee-lib/src/client_pool.rs
```diff
@@ -25,7 +26,9 @@ impl ClientPool {
             let client = self.pool.get().await;
             let tx = self.tx.clone();
             tokio::spawn(async move {
-                let resp = client.check(req).await.expect("Invalid URI");
+                // Client::check() may fail only because Request::try_from() may fail
+                // here request is already Request, so it never fail
```
mre, line 30: never fail -> never fails

lychee-lib/src/collector.rs
```diff
@@ -1,32 +1,43 @@
...
+};
+
+use crate::{
+    extract::{extract_links, FileType},
+    Request, Result,
+};
 
 const STDIN: &str = "-";
 
+/// Links which to be validated.
```
mre, line 23: How about "Links which need to be validated" or "Links which should be validated"?

lychee-lib/src/collector.rs
```diff
@@ -78,25 +90,23 @@ impl InputContent {
...
+            // this seems to be the only way to determine if this is a glob pattern
+            let is_glob = glob::Pattern::escape(value) != value;
+
+            if is_glob {
+                Self::FsGlob {
+                    pattern: value.to_owned(),
+                    ignore_case: glob_ignore_case,
                 }
+            } else {
+                Self::FsPath(value.into())
```
mre, line 109: Wow, that refactor is impressive.

lychee-lib/src/extract.rs
```diff
@@ -78,16 +74,15 @@ fn walk_html_links(mut urls: &mut Vec<String>, node: &Handle) {
...
+            urls.append(&mut extract_links_from_plaintext(
+                &contents.borrow().escape_default().to_string(),
+            ));
         }
 
         NodeData::Comment { ref contents } => {
-            for link in extract_links_from_plaintext(&contents.escape_default().to_string()) {
-                urls.push(link);
-            }
+            urls.append(&mut extract_links_from_plaintext(
```
mre, line 83: Nice

lychee-lib/src/extract.rs
```diff
@@ -122,15 +115,11 @@ fn walk_html_links(mut urls: &mut Vec<String>, node: &Handle) {
...
-    matches!(
-        (attr_name, elem_name),
-        ("href", _)
-            | ("src", _)
-            | ("srcset", _)
-            | ("cite", _)
-            | ("data", "object")
-            | ("onhashchange", "body")
-    )
+    matches!(attr_name, "href" | "src" | "srcset" | "cite")
```
mre, line 118: Kind of find the older version more readable here. 
Instead of two matches I prefer the redundant `_`, because it indicates that the second value in the tuple is insignificant for that match (e.g. in the case of `("href", _)`).
It's highly subjective, though.

lychee-lib/src/extract.rs
```diff
@@ -155,20 +144,15 @@ pub(crate) fn extract_links(
...
-                if !Path::new(&link).exists() {
-                    if let Some(base_url) = &base_url {
-                        if let Ok(new_url) = base_url.join(&link) {
-                            requests.insert(Request::new(
-                                Uri::Website(new_url),
-                                input_content.input.clone(),
-                            ));
-                        }
-                    }
+        if let Ok(uri) = Uri::try_from(link.as_str()) {
```
mre, line 147: Maybe we should move that block into its own method. The nested if lets are not the easiest to read. We can do that at a future point in time, though.

lychee-lib/src/filter/excludes.rs
```diff
@@ -37,54 +36,36 @@ impl Default for Excludes {
...
+        self.regex.as_ref().map_or(false, |re| re.is_match(input))
     }
 
-    pub fn false_positive(&self, input: &str) -> bool {
-        lazy_static! {
-            static ref FALSE_POSITIVES: RegexSet = RegexSet::new(FALSE_POSITIVE_REGEX).unwrap();
-        }
-        FALSE_POSITIVES.is_match(input)
+    pub fn is_false_positive(input: &str) -> bool {
+        input == FALSE_POSITIVE_PAT[0]
```
mre, line 45: Even though `FALSE_POSITIVE_PAT` only contains one string so far, I wonder if we still want to iterate through all elements of that slice already so that we don't forget to update this code once we start adding more strings to `FALSE_POSITIVE_PAT`. What do you think?

lychee-lib/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub mod collector;
+pub mod extract;
+pub mod test_utils;
+
+pub use client::check;
+pub use client::ClientBuilder;
+pub use client_pool::ClientPool;
+pub use collector::Input;
+pub use types::*;
+pub use uri::Uri;
```
mre, line 50: Do you think now is a good time to combine those modules into viewer exports?
(https://github.com/lycheeverse/lychee/issues/82)
If not, we can also do it in a separate PR of course.

lychee-lib/src/types/error.rs
```diff
@@ -0,0 +1,173 @@
+use std::{any::Any, convert::Infallible, fmt::Display, hash::Hash, path::PathBuf};
```
mre, line 1: So we decided against using this-error? Shouldn't it avoid a lot of that boilerplate? You probably had a good reason for not using it.

lychee-lib/src/uri.rs
```diff
@@ -15,30 +16,56 @@ pub enum Uri {
...
+
+    pub(crate) fn extract_github(&self) -> Option<(&str, &str)> {
+        debug_assert!(
+            matches!(self, Uri::Website(_)),
+            "Should only be called on a Website type!"
+        );
+        let url = if let Uri::Website(url) = self {
+            url
+        } else {
+            unsafe { std::hint::unreachable_unchecked() }
```
mre, line 56: Just wanted to mention that I still have mixed feelings about the use of unsafe. 😄 
It's fine and we can merge it like that, but I don't see the need for introducing unsafe if the perf improvement is not significant and there are no other significant upsides. It might be beneficial, but we'd have to measure that. 
Yes, I'm aware that this totally safe in this case with the `debug_assert`, but why introduce unsafe when we can have a fully "safe" crate with 99% of the upsides? 
My main concern is about future refactorings and missing an edge-case. 
Sorry, don't want to start another long discussion. It's just a note that we have to revisit that topic and we might revert that decision in the future. 😅

lychee-lib/src/uri.rs
```diff
@@ -15,30 +16,56 @@ pub enum Uri {
...
+        };
+        if matches!(url.domain(), Some("github.com") | Some("www.github.com")) {
```
mre, line 58: Thanks for removing the old regex match. 👍

mre:
That pull request is great! I learned a lot about making the code more ergonomic.
Just added a few more hints and then we're good to go.

Let's keep `lychee-lib` because you have a point about `cargo install lychee`. My guess is that more people will use the cli than the library. 👍

lychee-lib/src/extract.rs
lebensterben: I can combine the two `if let` to one
```rust
            if let Some(new_url) = base_url.as_ref().and_then(|u| u.join(&link).ok()) {
                requests.insert(Request::new(
                    Uri::Website(new_url),
                    input_content.input.clone(),
                ));
            }
```

lychee-lib/src/client.rs
lebensterben: Will remove in next commit

lychee-lib/src/client_pool.rs
lebensterben: Will fix in next commit

lychee-lib/src/collector.rs
lebensterben: Will fix in next commit with "Links which need to be validated"

lychee-lib/src/extract.rs
lebensterben:
```rust
    matches!(
        (attr_name, elem_name),
        ("href", _)
            | ("src", _)
            | ("srcset", _)
            | ("cite", _)
            | ("data", "object")
            | ("onhashchange", "body")
    )
```

yes, looks a bit clearer.

lychee-lib/src/filter/excludes.rs
lebensterben: In the future we may want. So this is not future-proof.

But it's really hard to come up with a second 'false-positive' example.
Also, I haven't thought thoroughly but the naming 'false-positive' is not clear enough.

lychee-lib/src/lib.rs
lebensterben: Let's do that in a separate PR.
This PR is already growing too large....

lychee-lib/src/types/error.rs
lebensterben: No. It's just easier to write by hand in my current workflow so I started without this-error.

It's way easy to implement `this-error` if you want to.
There's one tiny gain for writing them by hand, which is the faster compilation. It might only be faster for a few seconds at most. If that's true, consider that lychee is used in CI, it could translate to a considerable improvement overall.

lychee-lib/src/uri.rs
lebensterben: Well for this particular case, I'm able to avoid writing any explicit `unwrap`, `expect`, `unrechable` or `std::hint::unreachable_unchecked`.

```rust
        if let Uri::Website(url) = self {
            if matches!(url.domain(), Some("github.com") | Some("www.github.com")) {
                if let Some(mut path) = url.path_segments() {
                    if let Some(owner) = path.next() {
                        if let Some(repo) = path.next() {
                            return Some((owner, repo));
                        }
                    }
                }
            }
        }

```

lebensterben:
Well a few things happened during last few days.

I further slimmed downed the new `ErrorKind` and figured out that we can have `HttpError(reqwest::Error)`, instead of `HttpError(String, StatusCode)`. Note that we can extract the error message and status code from `reqwest::Error`. 

The only downside is we are not able to create create a superficial http error with empty message and a status code. The impact is minimal because we only need to do that in tests ( specifically `lychee::stat::test` module), and I figured out a way to fix this by returning a real `reqwest::Error` from a mock server. This would be a bit more resource intensive during test but no impact on any other parts of the program.

I updated dependencies, and ordered in alphabetically. Now `wiremock` would be a `dev-dependency` and `lychee-lib::test_utils` is only compiled for tests. The downside is I have to duplicate two functions in `lychee` (binary component)'s test modules. But cutting down dependency would greatly benefit CI usage of lychee.

lebensterben:
And about the performance gain.

I wrote quite a lot on the benchmarks. To sum up
- I found no significant change in speed when testing on the real world case ( that arch wiki page with hundreds of external links ). But my refactored version cuts off many memory allocations.
- I found HUGE improvement in both speed and memory allocations when testing on `fixutres/*`. It's actually 4 times faster.

So
- The bottom line is my refactor is not making things worse.
- My refactor seems to optimize extremely well with certain cases.

My guess is, `fixtures/*` contains links which are either right away excluded, or known to work or known to fail. There's hardly any timeouts. So timeout is the real culprit why you see no improvement in speed.

To verify my hypothesis, I made a markdown file with top 1,000 websites in Alexa's rank, and removed any of them that timed out.

```text
𝛌> hyperfine --min-runs 5 -i './develop/lychee --no-progress -E -t 5 top-1k-no-timeout.md' './master/lychee --no-progress -E -t 5 top-1k-no-timeout.md'
Benchmark #1: ./develop/lychee --no-progress -E -t 5 top-1k-no-timeout.md
  Time (mean ± σ):     29.645 s ±  1.900 s    [User: 1.379 s, System: 0.471 s]
  Range (min … max):   27.431 s … 31.543 s    5 runs

  Warning: Ignoring non-zero exit code.

Benchmark #2: ./master/lychee --no-progress -E -t 5 top-1k-no-timeout.md
  Time (mean ± σ):     33.484 s ±  6.254 s    [User: 1.392 s, System: 0.474 s]
  Range (min … max):   26.550 s … 39.519 s    5 runs

  Warning: Ignoring non-zero exit code.

Summary
  './develop/lychee --no-progress -E -t 5 top-1k-no-timeout.md' ran
    1.13 ± 0.22 times faster than './master/lychee --no-progress -E -t 5 top-1k-no-timeout.md'
```

As you can see, my refactored version IS faster for about 13%.

lebensterben:
I'm going to commit my local changes.
Please take a look of the changes and I will rebase them.

lychee-lib/src/filter/mod.rs
```diff
@@ -22,63 +21,65 @@ impl Filter {
...
-        // No regex includes/excludes at all?
-        if self.includes.is_empty() && self.excludes.is_empty() {
-            // Not excluded unless it's a known false positive
-            return self.excludes.false_positive(request.uri.as_str());
-        }
-        // Includes take precedence over excludes
-        if self.includes.regex(request.uri.as_str()) {
+
+        let input = uri.as_str();
+        if self.includes.is_empty() {
```
lebensterben, line 42: Here is one thing you haven't mentioned.

From L42 to L66 is some non-obvious optimization.
There were duplicated checks because some previous checks implied some latter ones.

The conditionals here are carefully organized so it doesn't miss any case while it won't check the same thing twice.

lebensterben:
Well well well...

Just find out I can further optimize `lychee-lib::Uri`. It would be a wrapper of `url::Url`.
There's no need to differentiate whether its `Uri::Website(_)` or `Uri::Mail(_)` at creation, because mail address would have scheme `mailto`.

Also, currently we check GitHub url by matching the domain "www.github.com" and "github.com". I will change it to matching host "github.com". So something like `https://raw.githubusercontent.com/lycheeverse/lychee/cefe38ee25d433dd3b7c61a96ccacbec66aa70a8/.gitignore` will also work.

(In the future we definitely will want to support other git repos as well.)

lebensterben:
Well I should stop now.

I end up refactored some of the tests. I introduced two macros and it's much shorter now.

One significant change is, for the integration change, it's now matching the json output instead of matching the CLI output. This should be more robust.

lychee-lib/src/extract.rs
mre: Looks good

lychee-lib/src/filter/excludes.rs
mre: True. Let's keep it like that then. Also can't come up with a better name at the moment. Maybe something like `IGNORE_PATTERNS` or `KNOWN_PHONY_URLS` could work. 🤔 We can come up with something better in the future.

lychee-lib/src/filter/mod.rs
mre: Yeah, I saw that but I forgot to mention it. Great work!

lychee-lib/src/lib.rs
mre: Sounds good. 👍

lychee-lib/src/types/error.rs
mre: Sure, then let's keep your version and if we see that we want to move to `this-error` in the future, we can always do so. 👍

lychee-bin/src/main.rs
```diff
@@ -257,14 +260,13 @@ mod test {
     #[test]
     fn test_parse_statuscodes() {
         let actual = parse_statuscodes("200,204,301").unwrap();
-        let expected: HashSet<StatusCode> = [
+        let expected = array::IntoIter::new([
```
mre, line 263: wohoo! 🤩

lychee-bin/src/stats.rs
```diff
@@ -35,18 +35,9 @@ pub(crate) struct ResponseStats {
...
-            total: 0,
-            successful: 0,
-            failures: 0,
-            timeouts: 0,
-            redirects: 0,
-            excludes: 0,
-            errors: 0,
-            fail_map,
-        }
+        Self::default()
```
mre, line 40: bam! Great one

lychee-bin/src/stats.rs
```diff
@@ -137,58 +159,35 @@ mod test_super {
...
+    #[tokio::test]
+    async fn test_stats() {
+        let stata = [
+            StatusCode::OK,
+            StatusCode::PERMANENT_REDIRECT,
+            StatusCode::BAD_GATEWAY,
+        ];
+
         let mut stats = ResponseStats::new();
-        stats.add(Response(
```
mre: Where did all the boilerplate go? 😆

lychee-bin/tests/cli.rs
```diff
@@ -23,87 +34,127 @@ mod cli {
...
+            Ok(())
+        }};
     }
 
-    /// Test that a GitHub link can be checked without specifying the token.
     #[test]
-    fn test_check_github_no_token() {
-        let mut cmd = main_command();
-        let test_github_path = fixtures_path().join("TEST_GITHUB.md");
+    fn test_exlucde_all_private() -> Result<()> {
```
mre, line 90: Small typo: exlucde -> exclude

lychee-lib/src/client.rs
```diff
@@ -20,17 +23,17 @@ const DEFAULT_USER_AGENT: &str = concat!("lychee/", env!("CARGO_PKG_VERSION"));
...
 #[derive(Debug, Clone)]
 pub struct Client {
-    /// The underlying reqwest client instance that handles the HTTP requests
+    /// Underlying reqwest client instance that handles the HTTP requests.
     reqwest_client: reqwest::Client,
-    /// Github API client
-    github: Option<Github>,
-    /// Filtered domain handling
+    /// Github API token.
+    github_token: Option<Github>,
```
mre, line 29: Wait, shouldn't that actually be `github_client`? After all, we call `Github::new(user_agent, Credentials::Token(token)` further down below and store that as `github_token`. 🤔

lychee-lib/src/client.rs
```diff
@@ -257,56 +260,36 @@ where
...
 
     #[tokio::test]
     async fn test_nonexistent() {
-        let template = ResponseTemplate::new(404);
-        let mock_server = MockServer::start().await;
-        Mock::given(method("GET"))
-            .respond_with(template)
-            .mount(&mock_server)
-            .await;
+        // let mock_server = get_mock_server(StatusCode::NOT_FOUND).await;
```
mre, line 273: Nit: we can remove that commented code.

lychee-lib/src/collector.rs
```diff
@@ -47,26 +46,13 @@ impl Serialize for Input {
...
-                write!(f, "{}", path.to_str().unwrap_or_default())
-            }
-            Input::Stdin => {
-                write!(f, "stdin")
-            }
-            Input::String(_) => {
-                write!(f, "raw input string")
-            }
-        }
+        f.write_str(match self {
```
mre, line 49: Cool! Didn't know you could do that.

lychee-lib/src/filter/mod.rs
```diff
@@ -99,75 +100,74 @@ mod test {
     const V6_MAPPED_V4_PRIVATE_CLASS_A: &str = "http://[::ffff:10.0.0.1]";
     const V6_MAPPED_V4_LINK_LOCAL: &str = "http://[::ffff:169.254.0.1]";
 
+    fn get_host_v4(s: &str) -> Option<Ipv4Addr> {
```
mre, line 103: Given that `get_host_v4` and `get_host_v6` are only used in tests and not expected to fail, do you think it makes sense to just use `unwrap`? 
So instead of `Option<Ipv4Addr>`, we could just return `Ipv4Addr` directly.

Would make the call more simple:

```rust
get_host_v4(V4_PRIVATE_CLASS_A).ok_or(())?
```

becomes 

```rust
get_host_v4(V4_PRIVATE_CLASS_A)
```

lychee-lib/src/test_utils.rs
```diff
@@ -1,45 +1,45 @@
...
+    Uri::from(Url::parse(url).expect("Expected valid Website URI"))
 }
 
-pub fn mail(address: &str) -> Uri {
-    Uri::Mail(address.to_owned())
+pub(crate) fn mail(address: &str) -> Uri {
+    if address.starts_with("mailto:") {
+        Url::parse(address)
+    } else {
+        Url::parse(&(String::from("mailto:") + address))
```
mre, line 41: Format could also work here, right?

```
Url::parse(&format!("mailto:{}", address))
```

lychee-lib/src/types/error.rs
```diff
@@ -1,65 +1,53 @@
...
-use tokio::task::JoinError;
-use url::ParseError;
 
 use crate::Uri;
 
 /// Kinds of status errors.
 #[derive(Debug)]
 #[non_exhaustive]
 pub enum ErrorKind {
-    ReqwestError(reqwest::Error),
```
mre: Thanks for removing that wart. 👍

lychee-lib/src/types/error.rs
```diff
@@ -69,31 +57,33 @@ impl Hash for ErrorKind {
...
-            Self::InvalidUri(uri) => format!("Invalid URI: {}", uri),
-            Self::InvalidEmailAddress(uri) => format!("Invalid mail address: {}", uri),
-            Self::InvalidHeader(e) => e.to_string(),
-            Self::InvalidGlobPattern(e) => e.to_string(),
-            Self::MissingGitHubToken => {
-                "GitHub token not specified. To check GitHub links reliably, \
-                 use `--github-token` flag / `GITHUB_TOKEN` env var."
-                    .to_owned()
+            Self::UrlParseError(s, (url_err, None)) => {
+                write!(f, "Cannot parse {} as webisite url ({})", s, url_err)
```
mre, line 77: webisite -> website

lychee-lib/src/types/error.rs
```diff
@@ -118,56 +108,55 @@ impl From<std::io::Error> for ErrorKind {
...
     }
 }
 
-// FIXME: Remove?
-impl<T> From<ErrorKind> for crate::Result<T> {
-    fn from(e: ErrorKind) -> Self {
-        Err(e)
+impl From<Infallible> for ErrorKind {
+    fn from(_: Infallible) -> Self {
+        // tautological
```
mre, line 159: 😆

lychee-lib/src/uri.rs
```diff
@@ -1,200 +1,222 @@
...
+use fast_chemail::parse_email;
 use serde::{Deserialize, Serialize};
 use url::Url;
 
 use crate::{ErrorKind, Result};
 
-/// Lychee's own representation of a URI, which encapsulates all support formats
+/// Lychee's own representation of a URI, which encapsulates all support formats.
+///
+/// If the scheme is `mailto`, it's a email address.
```
mre, line 11: Nit: a email address -> an email address

mre: Or "a mail address"

lychee-lib/src/uri.rs
```diff
@@ -1,200 +1,222 @@
...
-            Self::Mail(_) => None,
+        match self.url.host()? {
+            url::Host::Domain(_) => None,
+            url::Host::Ipv4(v4_addr) => Some(v4_addr.into()),
+            url::Host::Ipv6(v6_addr) => Some(v6_addr.into()),
         }
     }
 
+    // TODO: Support GitLab etc.
     pub(crate) fn extract_github(&self) -> Option<(&str, &str)> {
```
mre, line 58: So nicely done. Bonus points for getting rid of `unreachable_unchecked`. 💚

lychee-lib/src/uri.rs
```diff
@@ -1,200 +1,222 @@
...
+        }
+    }
+}
+
 impl TryFrom<&str> for Uri {
     type Error = ErrorKind;
 
     fn try_from(s: &str) -> Result<Self> {
-        // Check for internal Markdown links
-        let is_link_internal = is_internal_link(s);
```
mre: Doesn't that mean that relative links in Markdown won't work anymore?
https://github.com/lycheeverse/lychee/issues/146
Or is that covered by your refactoring somehow?

mre: Tests are passing, so it seems to be fine.

mre:
Just added a few more comments, but apart from that it's all nice and dandy. Guess you can rebase after that and we'll be good to go.

mre:
Guess by now you have touched almost all the lines in the codebase. 😆

mre:
Merged in #217 to fix #183. Guess you have to merge that into your branch. Hope that's okay. 😅

lychee-bin/src/main.rs
lebensterben: This is the latest stablized feature.

lychee-bin/tests/cli.rs
lebensterben: Fixed

lychee-bin/src/stats.rs
lebensterben: It's outdated. Which line do you mean?

lychee-lib/src/client.rs
lebensterben: Yep. Fixed

lychee-lib/src/client.rs
lebensterben: Removed.

lychee-lib/src/filter/mod.rs
lebensterben:
```rust
    macro_rules! assert_ip_address {
        (v4: $ip:expr, $predicate:tt) => {
            let res = if let Host::Ipv4(ipv4) = Url::parse($ip).map_err(|_| ())?.host().ok_or(())? {
                ipv4.$predicate()
            } else {
                false
            };
            std::assert!(res);
        };
        (v6: $ip:expr, $predicate:tt) => {
            let res = if let Host::Ipv6(ipv6) = Url::parse($ip).map_err(|_| ())?.host().ok_or(())? {
                ipv6.$predicate()
            } else {
                false
            };
            std::assert!(res);
        };
    }

    #[test]
    fn test_const_sanity() -> Result<(), ()> {
        assert_ip_address!(v4: V4_PRIVATE_CLASS_A, is_private);
        assert_ip_address!(v4: V4_PRIVATE_CLASS_B, is_private);
        assert_ip_address!(v4: V4_PRIVATE_CLASS_C, is_private);

        assert_ip_address!(v4: V4_LOOPBACK, is_loopback);
        assert_ip_address!(v6: V6_LOOPBACK, is_loopback);

        assert_ip_address!(v4: V4_LINK_LOCAL, is_link_local);

        Ok(())
    }
```

lebensterben: Returning `Ipv4Addr` or `Ipv6Addr` without wrapping them in `Option` is error-prone. A small typo would break it.

lychee-lib/src/test_utils.rs
lebensterben: Yes. I just don't like calling `format!`, or anything using format strings like `write!`, unless not doing so considerably reduce readability.

These macros calls `format_args!` and that's not really efficient for simple cases like this.

See the call graph: https://lib.rs/crates/ufmt

lychee-lib/src/types/error.rs
lebensterben: fixed

lychee-lib/src/uri.rs
lebensterben: A mail address.
Fixed.

lychee-lib/src/uri.rs
lebensterben: Yep I should mention this.
I don't get the point of `ignore` that test.
What's the background for this check?

lebensterben:
Will merge #217 later.

lebensterben:
Ready to rebase and merge.

I've merged #217 and fixed a bug. (the ignored test is passed)

mre:
@lebensterben, guess there is one more CI issue to fix before we can finally merge this:

```
 error[E0277]: the trait bound `types::error::ErrorKind: From<reqwest::error::Error>` is not satisfied
   --> src/types/error.rs:126:45
    |
126 |             hubcaps::Error::Reqwest(e) => e.into(),
    |                                             ^^^^ the trait `From<reqwest::error::Error>` is not implemented for `types::error::ErrorKind`
```

Source: https://github.com/lycheeverse/lychee/pull/208/checks?check_run_id=2331614860

lebensterben:
@mre 
This doesn't make any sense...
It passes the `cargo test` CI, and how's it possible to fail?

lebensterben:
BTW...

I've found other interfaces for further refactor. But I will keep them in another PR.
I'll be back to the HTTPS PR later.

lebensterben:
@mre 

You didn't merge it right. Now the problem should be fixed.

lebensterben:
Since it passes both [cargo test](https://github.com/lycheeverse/lychee/pull/208/checks?check_run_id=2338463910) and [cargo lint](https://github.com/lycheeverse/lychee/pull/208/checks?check_run_id=2338463875), I'm going to squash those commits.

lebensterben:
You can merge with `git merge -s theirs`. That's the safest way.

mre:
> You didn't merge it right. Now the problem should be fixed.

Resolved the conflict via Github directly as it only showed a single file to be changed: the `Cargo.toml` in the project root. It looked like a trivial change because the diff only showed that the dependencies had moved into the individual `lib` and `bin` crates, so I wonder how this could have happened. My only guess is that Github squashed the commits right after and deleted the history. 😨 Anyway, sorry for that.

I'm seeing a list of conflicting files now. Given my bad experiences with the last merge, perhaps you could try to resolve that this time to avoid further churn.

Other than that, let's merge it in. 😊

lebensterben:
@mre 
I just merged them and there's no conflict now.

mre:
Hate to say it, but there is one more lint error. 😅 
Also our old friend is back:

```rust
   Compiling lychee-lib v0.6.0 (/home/runner/work/lychee/lychee/target/package/lychee-lib-0.6.0)
error[E0277]: the trait bound `types::error::ErrorKind: From<reqwest::error::Error>` is not satisfied
   --> src/types/error.rs:127:45
    |
127 |             hubcaps::Error::Reqwest(e) => e.into(),
    |                                             ^^^^ the trait `From<reqwest::error::Error>` is not implemented for `types::error::ErrorKind`
    |
    = help: the following implementations were found:
              <types::error::ErrorKind as From<(PathBuf, std::io::Error)>>
              <types::error::ErrorKind as From<(std::string::String, url::ParseError)>>
              <types::error::ErrorKind as From<(std::string::String, url::ParseError, fast_chemail::ParseError)>>
              <types::error::ErrorKind as From<Infallible>>
            and 6 others
    = note: required because of the requirements on the impl of `Into<types::error::ErrorKind>` for `reqwest::error::Error`
```

I swear I didn't touch it this time. 🤞 

Also, shouldn't 

```
cargo install --path lychee-bin
```

just work? I'm getting

```rust
error[E0277]: the trait bound `url::Url: uri::_::_serde::Serialize` is not satisfied
  --> lychee-lib/src/uri.rs:15:5
   |
15 |     /// Website URL or mail address
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `uri::_::_serde::Serialize` is not implemented for `url::Url`
   |
   = note: required by `uri::_::_serde::ser::SerializeStruct::serialize_field`

error[E0277]: the trait bound `url::Url: Deserialize<'_>` is not satisfied
  --> lychee-lib/src/uri.rs:15:5
   |
15 |     /// Website URL or mail address
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `url::Url`
   |
   = note: required by `next_element`

error[E0277]: the trait bound `url::Url: Deserialize<'_>` is not satisfied
  --> lychee-lib/src/uri.rs:15:5
   |
15 |     /// Website URL or mail address
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `url::Url`
   |
   = note: required by `next_value`

error: aborting due to 3 previous errors
```

What's the right command to install it locally?

lebensterben:
@mre 

Mystery solved:
```text
   --> src/types/error.rs:127:62
    |
127 |             hubcaps::Error::Reqwest(e) => Self::ReqwestError(e),
    |                                                              ^ expected struct `reqwest::Error`, found struct `reqwest::error::Error`
    |
    = note: perhaps two different versions of crate `reqwest` are being used?
```

lebensterben:
This is some restriction of rust. These two are the same struct.
I cannot use `reqwest::error::Error` because `reqwest::error` is private.
But rustc cannot figure out they're the same.

I solved it by introduced a new enum variant.

mre:
Nicely done. Should `cargo install --path lychee-bin` work?

lebensterben:
Yes it works.
```
   Compiling reqwest v0.11.3
   Compiling hubcaps v0.6.2 (https://github.com/softprops/hubcaps.git#a9e6616e)
   Compiling check-if-email-exists v0.8.21
   Compiling lychee-lib v0.6.0 (/home/lucius/GitHub/lycheeverse/lychee-lib)
   Compiling lychee v0.6.0 (/home/lucius/GitHub/lycheeverse/lychee-bin)
    Finished release [optimized] target(s) in 1m 40s
  Installing /home/lucius/.cargo/bin/lychee
   Installed package `lychee v0.6.0 (/home/lucius/GitHub/lycheeverse/lychee-bin)` (executable `lychee`)
```
Once `lychee-lib` is published, `lychee` should pass the publish check.

mre:
Yaaaay! You've done it. 🚀 Great success. Thanks a lot for pushing through and for your patience. 👏

mre:
I'll publish lychee-lib now. 
P.S.: we forgot to adjust the `Dockerfile`, but it's not a big deal; will fix.

lychee-lib/src/uri.rs
drahnr: Extracting this in a separate `fn extract_github_info(url: &_) -> Option<(String, String)> { .. }` would allow you to simply use the `?` operator (`Option<_>` impls `Try`) and there would not need to be a single nesting layer :smiley:

mre: Good idea @drahnr. Would you be willing to open a PR for that?

drahnr: Actually, it was already implemented as `extract_github` in a follow up which I missed

mre: All good then. 👍

PR #218, opened 2021-04-12
fix `deny(missing_docs)` lint

soruh:
This makes the `deny(missing_docs)` lint into an inner attribute instead of an outer one, making it actually deny missing docs


lebensterben:
#208 has `// #![deny(missing_docs)]`. We can enable it at a later time.
And also fixed the crate-level documentation.

So this PR is superseded.

mre:
This is fixed with #231 now. 🥳

PR #219, opened 2021-04-12
Clippy pendantic and more refactors

soruh:
- Check the code with `clippy::pedantic`, fix most of it's complaints and ignore the rest (mostly documentation tasks).
- Manually converted some code into using methods on `Option` and `Result` instead of matches.
  
P.S.: my rustfmt seems to have collapsed some `use`s and comment, if you dislike this I can try to undo that manually.

P.P.S.: sorry for creating another relatively large PR :)

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -89,9 +89,9 @@ use crate::{
 /// A C-like enum that can be cast to `i32` and used as process exit code.
 enum ExitCode {
     Success = 0,
-    // NOTE: exit code 1 is used for any `Result::Err` bubbled up to `main()` using the `?` operator.
-    // For now, 1 acts as a catch-all for everything non-link related (including config errors),
-    // until we find a way to structure the error code handling better.
+    // NOTE: exit code 1 is used for any `Result::Err` bubbled up to `main()` using the `?`
+    // operator. For now, 1 acts as a catch-all for everything non-link related (including
+    // config errors), until we find a way to structure the error code handling better.
     #[allow(unused)]
     UnexpectedFailure = 1,
     LinkCheckFailure = 2,
@@ -109,7 +109,8 @@ fn main() -> Result<()> {
 fn run_main() -> Result<i32> {
     let mut opts = LycheeOptions::from_args();
 
-    // Load a potentially existing config file and merge it into the config from the CLI
+    // Load a potentially existing config file and merge it into the config from the
+    // CLI
     if let Some(c) = Config::load_from_file(&opts.config_file)? {
         opts.config.merge(c)
     }
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -62,6 +62,7 @@
 use ring as _;
 
 use std::iter::FromIterator;
+use std::ops::Not;
 use std::{collections::HashSet, fs, str::FromStr, time::Duration};
 
 use anyhow::{anyhow, Context, Result};
@@ -123,10 +124,10 @@ fn run_main() -> Result<i32> {
             tokio::runtime::Builder::new_multi_thread()
                 .worker_threads(threads)
                 .enable_all()
-                .build()?
+                .build()
         }
-        None => tokio::runtime::Runtime::new()?,
-    };
+        None => tokio::runtime::Runtime::new(),
+    }?;
 
     runtime.block_on(run(cfg, opts.inputs()))
 }
@@ -198,16 +199,14 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
     .await
     .map_err(|e| anyhow!(e))?;
 
-    let pb = if cfg.no_progress {
-        None
-    } else {
+    let pb = cfg.no_progress.not().then(|| {
         let bar =
             ProgressBar::new(links.len() as u64).with_style(ProgressStyle::default_bar().template(
                 "{spinner:.red.bright} {pos}/{len:.dim} [{elapsed_precise}] {bar:25} {wide_msg}",
             ));
         bar.enable_steady_tick(100);
-        Some(bar)
-    };
+        bar
+    });
 
     let (send_req, recv_req) = mpsc::channel(max_concurrency);
     let (send_resp, mut recv_resp) = mpsc::channel(max_concurrency);
@@ -254,11 +253,13 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
         println!("{}", stats_formatted);
     }
 
-    if stats.is_success() {
-        Ok(ExitCode::Success as i32)
+    let exit_code = if stats.is_success() {
+        ExitCode::Success
     } else {
-        Ok(ExitCode::LinkCheckFailure as i32)
-    }
+        ExitCode::LinkCheckFailure
+    };
+
+    Ok(exit_code as i32)
 }
 
 fn read_header(input: &str) -> Result<(String, String)> {
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -90,9 +90,9 @@ use crate::{
 /// A C-like enum that can be cast to `i32` and used as process exit code.
 enum ExitCode {
     Success = 0,
-    // NOTE: exit code 1 is used for any `Result::Err` bubbled up to `main()` using the `?`
-    // operator. For now, 1 acts as a catch-all for everything non-link related (including
-    // config errors), until we find a way to structure the error code handling better.
+    // NOTE: exit code 1 is used for any `Result::Err` bubbled up to `main()` using the `?` operator.
+    // For now, 1 acts as a catch-all for everything non-link related (including config errors),
+    // until we find a way to structure the error code handling better.
     #[allow(unused)]
     UnexpectedFailure = 1,
     LinkCheckFailure = 2,
@@ -110,8 +110,7 @@ fn main() -> Result<()> {
 fn run_main() -> Result<i32> {
     let mut opts = LycheeOptions::from_args();
 
-    // Load a potentially existing config file and merge it into the config from the
-    // CLI
+    // Load a potentially existing config file and merge it into the config from the CLI
     if let Some(c) = Config::load_from_file(&opts.config_file)? {
         opts.config.merge(c)
     }
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -82,10 +82,9 @@ macro_rules! fold_in {
 )]
 pub(crate) struct LycheeOptions {
     /// The inputs (where to get links to check from).
-    /// These can be: files (e.g. `README.md`), file globs (e.g.
-    /// `"~/git/*/README.md"`), remote URLs (e.g. `https://example.org/README.md`) or standard input (`-`).
-    /// Prefix with `--` to separate inputs from options that allow multiple
-    /// arguments.
+    /// These can be: files (e.g. `README.md`), file globs (e.g. `"~/git/*/README.md"`),
+    /// remote URLs (e.g. `https://example.org/README.md`) or standard input (`-`).
+    /// Prefix with `--` to separate inputs from options that allow multiple arguments.
     #[structopt(name = "inputs", default_value = "README.md")]
     raw_inputs: Vec<String>,
 
@@ -222,8 +221,7 @@ pub(crate) struct Config {
     #[serde(default)]
     pub(crate) basic_auth: Option<String>,
 
-    /// GitHub API token to use when checking github.com links, to avoid rate
-    /// limiting
+    /// GitHub API token to use when checking github.com links, to avoid rate limiting
     #[structopt(long, env = "GITHUB_TOKEN")]
     #[serde(default)]
     pub(crate) github_token: Option<String>,
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -82,9 +82,10 @@ macro_rules! fold_in {
 )]
 pub(crate) struct LycheeOptions {
     /// The inputs (where to get links to check from).
-    /// These can be: files (e.g. `README.md`), file globs (e.g. `"~/git/*/README.md"`),
-    /// remote URLs (e.g. `https://example.org/README.md`) or standard input (`-`).
-    /// Prefix with `--` to separate inputs from options that allow multiple arguments.
+    /// These can be: files (e.g. `README.md`), file globs (e.g.
+    /// `"~/git/*/README.md"`), remote URLs (e.g. `https://example.org/README.md`) or standard input (`-`).
+    /// Prefix with `--` to separate inputs from options that allow multiple
+    /// arguments.
     #[structopt(name = "inputs", default_value = "README.md")]
     raw_inputs: Vec<String>,
 
@@ -221,7 +222,8 @@ pub(crate) struct Config {
     #[serde(default)]
     pub(crate) basic_auth: Option<String>,
 
-    /// GitHub API token to use when checking github.com links, to avoid rate limiting
+    /// GitHub API token to use when checking github.com links, to avoid rate
+    /// limiting
     #[structopt(long, env = "GITHUB_TOKEN")]
     #[serde(default)]
     pub(crate) github_token: Option<String>,
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -261,7 +261,7 @@ impl Config {
             Err(e) => {
                 return match e.kind() {
                     ErrorKind::NotFound => Ok(None),
-                    _ => Err(Error::from(e)),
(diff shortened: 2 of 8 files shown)


soruh:
Ah it seems I accidentally used some nightly features

src/extract.rs
```diff
@@ -130,7 +135,14 @@ fn elem_attr_is_link(attr_name: &str, elem_name: &str) -> bool {
     // over at: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
     matches!(
         (attr_name, elem_name),
-        ("href" | "src" | "srcset" | "cite", _) | ("data", "object") | ("onhashchange", "body")
+        // TODO: re-enable this once `or-patterns` become stable:
```
mre, line 138: TIL I learned about or patterns. 😍

src/bin/lychee/main.rs
```diff
@@ -184,10 +185,12 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
...
-    match stats.is_success() {
-        true => Ok(ExitCode::Success as i32),
-        false => Ok(ExitCode::LinkCheckFailure as i32),
-    }
+    let exit_code = match stats.is_success() {
+        true => ExitCode::Success,
+        false => ExitCode::LinkCheckFailure,
+    };
+
+    Ok(exit_code as i32)
```
mre, line 193: Nice!

src/client.rs
```diff
@@ -140,17 +140,11 @@ impl ClientBuilder {
...
-                    Some(github)
-                }
-            }
-            None => None,
-        };
+        let github = token
+            .and_then(|token| {
+                (!token.is_empty()).then(|| Github::new(user_agent, Credentials::Token(token)))
+            })
+            .transpose()?;
```
mre, line 147: Aha! `transpose` does make sense here.

mre:
Great improvements! Thanks a lot for this.
I'd like to wait with the merge until #208 is in. Most likely you'd have to rebase your changes on top of `master` once that is done. Hope that's okay for you.

src/bin/lychee/main.rs
```diff
@@ -127,17 +130,15 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
     )
     .await?;
 
-    let pb = match cfg.no_progress {
-        true => None,
-        false => {
-            let bar = ProgressBar::new(links.len() as u64)
-                .with_style(ProgressStyle::default_bar().template(
+    let with_progress = !cfg.no_progress;
+    let pb = with_progress.then(|| {
```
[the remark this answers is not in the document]
soruh: I can see that I'll revert it when I rebase on the new master when #208 is merged.

src/client.rs
soruh: Yeah I was so happy when I found out about transpose

soruh:
Seems like a good Idea although doing this again on the new master might actually be less work than rebasing.

lebensterben:
There's no need to merge this.
I've already applied `clippy::pendantic` in #208.

soruh:
Which is why the plan is to rebase on top of your PR.
Also as far I can see you didn't actually enable clippy::pedantic but only applied it's suggestions which means that it will not be checked against by CI.

lebensterben:
@soruh 
That's true. But you shouldn't rebase on top of my PR because you will see many rebase conflicts. It's much easier just to discard this one and start a new PR which adds the clippy attributes.
P.S. I've enabled both `clippy::pedantic` and `clippy::all`.

src/bin/lychee/main.rs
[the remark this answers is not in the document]
mre: #208 is merged now. 😄

mre:
@soruh, would you like to rebase on top of the current master, create a new PR with the improvements or are all those changes covered by #208 already? 😊

soruh:
I think there are some additional changes, I'll try to rebase it when I find the time and will make a new PR if there are significant changes left / close this one if there aren't.

mre:
Thanks!

soruh:
I just rebased my code onto master, there is not a huge amount left but still enough for keeping this open I think.

I re-added my old solution of converting file extensions to lower-case before checking them. By now I have realized that this is probably not the optimal solution (as is always heap allocates). A version of `to_lowercase` that returns a `Cow` or just checking for the uppercase variants of the extensions directly may be preferable.

I personally like the `.not().then(..)` pattern but am of course open to change it to something more traditional if preferred.

There a a few things I would like an opinion on (annotated with `REVIEW`).

lychee-bin/src/main.rs
```diff
@@ -62,6 +62,7 @@
 use ring as _;
 
 use std::iter::FromIterator;
+use std::ops::Not;
```
lebensterben, line 65: -1 unnecessary

lychee-bin/src/main.rs
```diff
@@ -122,10 +123,10 @@ fn run_main() -> Result<i32> {
             tokio::runtime::Builder::new_multi_thread()
                 .worker_threads(threads)
                 .enable_all()
-                .build()?
+                .build()
         }
-        None => tokio::runtime::Runtime::new()?,
-    };
+        None => tokio::runtime::Runtime::new(),
+    }?;
```
lebensterben, line 129: ok. but easy to miss when reading.

lychee-bin/src/options.rs
```diff
@@ -259,7 +259,7 @@ impl Config {
             Err(e) => {
                 return match e.kind() {
                     ErrorKind::NotFound => Ok(None),
-                    _ => Err(Error::from(e)),
+                    _ => Err(e.into()),
```
lebensterben, line 262: No I deliberately avoid using Into::into().
It's hard to figure out what it is converting to.

lychee-bin/src/main.rs
```diff
@@ -197,16 +198,14 @@ async fn run(cfg: &Config, inputs: Vec<Input>) -> Result<i32> {
     .await
     .map_err(|e| anyhow!(e))?;
 
-    let pb = if cfg.no_progress {
-        None
-    } else {
+    let pb = cfg.no_progress.not().then(|| {
```
lebensterben, line 201: Does this increase any readability or performance?

lychee-lib/src/client.rs
```diff
@@ -141,13 +142,15 @@ impl ClientBuilder {
...
-            Some(ref token) if !token.is_empty() => Some(Github::new(
-                self.user_agent.clone(),
-                Credentials::Token(token.clone()),
-            )?),
-            _ => None,
-        };
+        let github_token = self
+            .github_token
+            .as_ref()
+            .and_then(|token| {
```
lebensterben, line 148: Again.
When I refactored these code, I deliberately avoided this kind of pattern.

lychee-lib/src/client.rs
```diff
@@ -236,6 +239,11 @@ impl Client {
 
     pub async fn check_mail(&self, uri: &Uri) -> Status {
         let input = CheckEmailInput::new(vec![uri.as_str().to_owned()]);
+
+        // REVIEW: This panics with an out-of-bounds error
+        // if there are no results from check_email
+        // (previous versions returned false)
+        // these parenthesis also seem unnecesary
         let result = &(check_email(&input).await)[0];
```
lebensterben, line 247: Unnecessary. This will not happen.
Read the source code here https://docs.rs/check-if-email-exists/0.8.21/src/check_if_email_exists/lib.rs.html#194-214

lychee-lib/src/extract.rs
```diff
@@ -125,6 +132,8 @@ fn elem_attr_is_link(attr_name: &str, elem_name: &str) -> bool {
     // over at: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
     matches!(
         (attr_name, elem_name),
+        // TODO: re-enable once `or-patterns` become stable in Rust 1.53
+        // ("href" | "src" | "srcset" | "cite", _) | ("data", "object") | ("onhashchange", "body")
         ("href", _)
```
lebensterben, line 137: When I refactored this part, it's requested to keep the current form for readability.

lychee-lib/src/extract.rs
```diff
@@ -224,11 +233,17 @@ mod test {
     #[test]
     fn test_file_type() {
         // FIXME: Assume plaintext in case a path has no extension
+        // REVIEW: can we just fix this or is there some compatablility
+        // reason we can't ?
         // assert_eq!(FileType::from(Path::new("/")), FileType::Plaintext);
+        // assert_eq!(FileType::from("test"), FileType::Plaintext);
         assert_eq!(FileType::from("test.md"), FileType::Markdown);
+        assert_eq!(FileType::from("TEST.MD"), FileType::Markdown);
```
lebensterben, line 241: -1
On modern file system it's case sensitive.
According to mimeinfo of `text/markdown`, valid extension is one of `.md` `.mkd` and `.markdown`.

If a `.MD` extension is recognized as markdown, it IS A BUG.

lychee-lib/src/types/response.rs
```diff
@@ -64,7 +64,7 @@ impl Display for ResponseBody {
             }
             Status::Timeout(Some(code)) => format!(" [{}]", code),
             Status::Error(e) => format!(" ({})", e),
-            _ => "".to_owned(),
+            _ => "".into(),
```
lebensterben, line 67: No.
When converting a &str to String, I deliberately use `.to_owned()` to indicate the memory allocation.

lychee-lib/src/collector.rs
```diff
@@ -222,11 +222,9 @@ pub async fn collect_links(
     skip_missing_inputs: bool,
     max_concurrency: usize,
 ) -> Result<HashSet<Request>> {
-    let base_url = if let Some(url) = base_url {
-        Some(Url::parse(&url).map_err(|e| (url, e))?)
-    } else {
-        None
-    };
+    let base_url = base_url
```
lebensterben, line 225: Again, not necessarily more readable.

lebensterben (changes requested):

lychee-bin/src/main.rs
soruh: We are creating a runtime in both cases and are checking if that failed after which I find more readable.

lychee-bin/src/options.rs
soruh: We are calling `.into()` inside of an `Err(...)`, the `Error::` doesn't give you any new information here and just makes the more verbose.

lychee-bin/src/main.rs
soruh: I think the functional style is clearer but I agree that the `.not()` may not be (see my original comment).

lychee-bin/src/main.rs
lebensterben: the `?` operator in `}?;` is quite easy to miss.

lychee-bin/src/main.rs
lebensterben: The issue is not with `not`, the main issue is the `functional style`.
`std::option::Option::then()` is a fairly new addition. If it's the more idiomatic way of handling Boolean conditionals, it'd be added much earlier.

lychee-lib/src/client.rs
soruh: The question is unnecessary?
I don't really see how that part of the code guarantees that the result will never be empty as it just maps over the input and it doesn't really answer why the parenthesis is necessary.
Since you aparetnly have found a sound way of reasoning that this can not panic I would appreciate if you could suggest a comment to go here that explains it.

lychee-bin/src/options.rs
lebensterben: How do you figure out what's the type of `E` in a `Result<T, E>`?
`Error::from(e)` clearly shows `E` is of type `Error`, while `e.into()` has no such information.

lychee-lib/src/extract.rs
soruh: Are you saying that handling an html file incorrectly because it is called "index.HTML" is okay? Or are you only referring to Markdown files? In that case I guess If we do what I proposed in my comment above we can only check if certain extension match insensitively, I don't see how handling files with extensions that are not _technically_ in spec is an improvement

lychee-lib/src/types/response.rs
soruh: This doesn't allocate; I tough about using `String::new()` but decided that this is clearer.

lychee-lib/src/collector.rs
soruh: I disagree but again, I guess it's a matter of opinion if you think the functional or the imperative style is more readable

lychee-lib/src/client.rs
soruh: I agree that the `not()` should be a `!` but think this is way more readable than the `Some(ref token) if !token.is_empty() => Some(` incantation it replaced. I agree that this is a matter of opinion though.

lychee-lib/src/client.rs
lebensterben: `input` is of type `CheckEmailInput`, which contains a vector of URL https://docs.rs/check-if-email-exists/0.8.21/src/check_if_email_exists/util/input_output.rs.html#77

`check_email` returns an `impl Future<output = Vec<CheckEmailOutput>>`, where the vector of output has same dimension as the vector of URL.

It's very obvious. As long as the input is non empty, the output is not. So it won't panic.
---
You can remove the parenthesis though.

lychee-lib/src/client.rs
lebensterben: No.

I deliberately keep `Some(ref token)`. this make it very clear of the type signature.
( Calling `as_ref()` has the same effect. )

The not readable part is the nested `and_then` and `then`. Unless you add explicit type signature, it's less clear. But if you did, it's too verbose.

lychee-lib/src/extract.rs
lebensterben: I'm not talking about the REVIEW comment.

What I'm taking is the additional `assert_eq` tests you added are not sound. They should be removed.

lychee-lib/src/types/response.rs
lebensterben: The whole code base consistently use `str::to_owned()` for this type of conversion.
When you refactor you need to take the overall coding style and convention into account.

soruh:
I will be reverting some of the changes of the original rebase (at least the `.not()`) and will see about the style related ones once there is a consensus on them.

Regarding the lowercase file extension handling I do think it makes sense to do but can remove it if again if deemed unnecessary.

P.S. @lebensterben I would like to inform you that the way you have reviewed my code has honestly been a very bad experience; This had been the only time that I did not look forward to reading a review of my code.
I personally don't mind that much but I hope you don't always do this as I can't imagine it encourages new contributors to continue contributing (it certainly hasn't for me).

lychee-lib/src/types/response.rs
soruh: Sure, I'll revert it

lychee-lib/src/extract.rs
soruh: Ah, that makes sense, are you okay with changing that to a ".HTML" test or are uppercase extension not technically in-spec in general?

lebensterben:
FYI when I refactor, there are a few things taken into account

- Does the refactor make it more performant?
- Does the refactor make it more readable?
- Does the refactor conform to idiomatic Rust coding styles and the overall style of this code base.

Also, I always prefer to make things more explicit. Without writing type signatures everywhere, a way to achieve this to use more specific function/methods.
E.g.
- For &str to String, use `to_owned` instead of `into` or `to_string`.
- For general conversion of T to U, use `U::from(T)` instead of `T.into()`.

lychee-lib/src/client.rs
soruh: Ah, I didn't see the line above which creates the input because it had been cut off by the github diff and assumed that was supplied by the user, sorry for the confusion.

lychee-lib/src/extract.rs
lebensterben: To incorporate them is very complicated.
For modern Linux/Unix/FreeBSD systems, the file system should be case-sensitive. Thus these are unknown schemes and are recognized as planttext. (But newer Linux kernel supports opt-in case-insensitive feature)
For macOS, HFS/HFS+ defaults to case-insensitive if operating system is installed before 10.3. Otherwise it's case-sensitive. For APS, it's case-sensitive.
For Windows, NTFS is actually case sensitive internally, but defaults to case-insensitive in user space.

If you really want to incorporate those tests, you need conditional compilations based on operating system and whehther the filesystem is case sensitive.

My opinion is these cases are too complicated to handle. Just remove those tests.

mre:
@soruh are you still planning to work on this or shall we close the PR and open a new one if time allows?

soruh:
Ah yeah sorry, I think I'll just close this.

PR #222, opened 2021-04-13
More refactor

lebensterben:
- Major changes in `lychee-lib::filter` module:
  - Fields in `Excludes` except the `RegexSet` is now moved to `Filter`.
  - `Filter` contains `Option<Excludes>` and `Option<Includes>`, which are
    wrapper struct of `RegexSet` instead of `Option<RegexSet>`. As a result
    the code now looks cleaner.
  - Factored out some filtering logics to dedicated functions.
    - It's possible to write tests for those functions in addition to tests
      for the `Filter` struct.
  - Added docs to `Filter::is_excluded` and reorgnized the code.
- placed `derive_builder` by `typed_builder`:
  - The internal interface very ugly, as admitted by the author, but we no
    longer have nested `Option`s like before.
  - As a result, the `Client` building is much easier to read.
  - Main benefit of `typed_builder` is, the arguments feeded to builder is
    checked at compile time instead of run-time.
- Fixed a bug in `lychee::tests::usage` and `lychee-lib::stats::test`.
  - Now it will clear environment variable which would otherwise cause an
    issue if `GITHUB_TOKEN` is set.
- Updated dependencies.

### DIFF
--- a/lychee-lib/src/filter/excludes.rs
+++ b/lychee-lib/src/filter/excludes.rs
@@ -1,76 +1,23 @@
 use regex::RegexSet;
-use std::net::IpAddr;
-
-use crate::Uri;
-
-/// Pre-defined exclusions for known false-positives
-static FALSE_POSITIVE_PAT: &[&str] = &[r"http://www.w3.org/1999/xhtml"];
 
 /// Exclude configuration for the link checker.
-/// You can ignore links based on regex patterns or pre-defined IP ranges.
-#[allow(clippy::struct_excessive_bools)]
+/// You can ignore links based on regex patterns.
 #[derive(Clone, Debug)]
 pub struct Excludes {
     /// User-defined set of excluded regex patterns
-    pub regex: Option<RegexSet>,
-    /// Example: 192.168.0.1
-    pub private_ips: bool,
-    /// Example: 169.254.0.0
-    pub link_local_ips: bool,
-    /// For IPv4: 127.0.0.1/8
-    /// For IPv6: ::1/128
-    pub loopback_ips: bool,
-    /// Example: octocat@github.com
-    pub mail: bool,
-}
-
-impl Default for Excludes {
-    fn default() -> Self {
-        Self {
-            regex: None,
-            private_ips: false,
-            link_local_ips: false,
-            loopback_ips: false,
-            mail: false,
-        }
-    }
+    pub(crate) regex: RegexSet,
 }
 
 impl Excludes {
     #[inline]
     #[must_use]
-    pub fn regex(&self, input: &str) -> bool {
-        self.regex.as_ref().map_or(false, |re| re.is_match(input))
-    }
-
-    #[must_use]
-    pub fn is_false_positive(input: &str) -> bool {
-        input == FALSE_POSITIVE_PAT[0]
-    }
-
-    #[must_use]
-    pub fn ip(&self, uri: &Uri) -> bool {
-        match uri.host_ip() {
-            Some(ip_addr) if self.loopback_ips && ip_addr.is_loopback() => true,
-            // Note: in a pathological case, an IPv6 address can be IPv4-mapped
-            //       (IPv4 address embedded in a IPv6).  We purposefully
-            //       don't deal with it here, and assume if an address is IPv6,
-            //       we shouldn't attempt to map it to IPv4.
-            //       See: https://tools.ietf.org/html/rfc4291#section-2.5.5.2
-            Some(IpAddr::V4(v4_addr)) if self.private_ips && v4_addr.is_private() => true,
-            Some(IpAddr::V4(v4_addr)) if self.link_local_ips && v4_addr.is_link_local() => true,
-            _ => false,
-        }
+    pub fn is_match(&self, input: &str) -> bool {
+        self.regex.is_match(input)
     }
 
     #[inline]
     #[must_use]
-    pub const fn is_mail_excluded(&self) -> bool {
-        self.mail
-    }
-
-    #[inline]
     pub fn is_empty(&self) -> bool {
-        self.regex.as_ref().map_or(true, RegexSet::is_empty)
+        self.regex.is_empty()
     }
 }
--- a/lychee-lib/src/filter/mod.rs
+++ b/lychee-lib/src/filter/mod.rs
@@ -1,74 +1,147 @@
 mod excludes;
 mod includes;
 
+use std::net::IpAddr;
+
 pub use excludes::Excludes;
 pub use includes::Includes;
 
 use crate::uri::Uri;
 
+/// Pre-defined exclusions for known false-positives
+static FALSE_POSITIVE_PAT: &[&str] = &[r"http://www.w3.org/1999/xhtml"];
+
+#[inline]
+#[must_use]
+pub fn is_false_positive(input: &str) -> bool {
+    input == FALSE_POSITIVE_PAT[0]
+}
+
 /// A generic URI filter
 /// Used to decide if a given URI should be checked or skipped
+#[allow(clippy::struct_excessive_bools)]
 #[derive(Clone, Debug, Default)]
 pub struct Filter {
-    pub(crate) includes: Includes,
-    pub(crate) excludes: Excludes,
-    pub(crate) scheme: Option<String>,
+    pub includes: Option<Includes>,
+    pub excludes: Option<Excludes>,
+    // TODO: accept multiple scheme
+    // TODO: includes scheme and excludes scheme
+    // TODO: excludes_mail should be merged to excludes scheme
+    // allowed scheme
+    pub scheme: Option<String>,
+    /// Example: 192.168.0.1
+    pub exclude_private_ips: bool,
+    /// Example: 169.254.0.0
+    pub exclude_link_local_ips: bool,
+    /// For IPv4: 127.0.0.1/8
+    /// For IPv6: ::1/128
+    pub exclude_loopback_ips: bool,
+    /// Example: octocat@github.com
+    pub exclude_mail: bool,
 }
 
 impl Filter {
+    #[inline]
     #[must_use]
-    pub fn new(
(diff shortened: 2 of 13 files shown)


lebensterben:
Let's merge this after #208 and I'll combine rebase it.

lychee-lib/src/filter/excludes.rs
```diff
@@ -1,8 +1,7 @@
 use regex::RegexSet;
-use std::net::IpAddr;
-
-use crate::Uri;
 
+// TODO: renaming
+// TODO: should just be added to the Excludes set.
```
mre, line 4: That's a great idea! We could call it `default_excludes` or something

lychee-lib/src/filter/excludes.rs
```diff
@@ -11,61 +10,25 @@ static FALSE_POSITIVE_PAT: &[&str] = &[r"http://www.w3.org/1999/xhtml"];
...
-    pub fn is_mail_excluded(&self) -> bool {
-        self.mail
+    pub fn is_match(&self, input: &str) -> bool {
+        self.regex.is_match(input)
     }
 
     #[inline]
     pub fn is_empty(&self) -> bool {
-        self.regex.as_ref().map_or(true, RegexSet::is_empty)
+        self.regex.is_empty()
```
mre, line 32: 🤦‍♀️ nice catch

mre: Ah okay, I get it now. It wasn't useless before, it's just that the new builder doesn't require these type shenanigans anymore.

lychee-lib/src/filter/mod.rs
```diff
@@ -1,72 +1,140 @@
...
 
+use std::net::IpAddr;
+
 pub use excludes::Excludes;
 pub use includes::Includes;
 
 use crate::uri::Uri;
 
+/// Pre-defined exclusions for known false-positives
+static FALSE_POSITIVE_PAT: &[&str] = &[r"http://www.w3.org/1999/xhtml"];
```
mre, line 12: Note: this line is also still in `excludes.rs`.

lychee-lib/src/filter/mod.rs
```diff
@@ -1,72 +1,140 @@
...
+    fn is_includes_match(&self, input: &str) -> bool {
+        matches!(self.includes, Some(ref includes) if includes.is_match(input))
+    }
+
+    #[inline]
+    fn is_excludes_match(&self, input: &str) -> bool {
+        matches!(self.excludes, Some(ref excludes) if excludes.is_match(input))
+    }
+
+    /// Determine whether a given [`Uri`] should be excluded.
```
mre, line 88: Thanks for documenting the behavior.

lychee-lib/src/filter/mod.rs
```diff
@@ -1,72 +1,140 @@
...
+    ///
+    /// # Details
+    ///
+    /// 1. If any of the following conditions is met, the URI is excluded:
```
mre, line 92: conditions is met -> conditions are met

lychee-lib/src/filter/mod.rs
```diff
@@ -1,72 +1,140 @@
...
+    ///   - If it's a mail address and it's configured to ignore mail addresses.
+    ///   - If the IP address belongs to a type that is configured to exclude.
```
mre, line 94: is configured to exclude -> is configured to be excluded

lychee-lib/src/filter/mod.rs
```diff
@@ -1,72 +1,140 @@
...
-                // In case we have includes and no excludes,
-                // skip everything that was not included
-                return true;
-            }
-        } else if self.excludes.regex(input) {
-            // Excluded explicitly
+
+        if is_false_positive(input)
+        // Exclude well-known false-positives
+        // Performed after checking includes to allow user-overwriddes
```
mre, line 126: overwriddes -> overwrites

lychee-lib/src/filter/mod.rs
```diff
@@ -1,72 +1,140 @@
...
+            || self.is_excludes_empty()
+        // Previous checks implies input is not explicitly included,
```
mre, line 128: implies -> imply

lychee-lib/src/filter/mod.rs
```diff
@@ -1,72 +1,140 @@
...
+        // if excludes rules is empty, then *presumably excluded*
+            || self.is_excludes_match(input)
+        // If excludes rules matches input, then
+        // *explicitly excluded*
+        {
```
mre, line 133: In the above if conditions, I'd personally also move the comment before the condition. At least I find it easier to read this way, but it's subjective of course. Just wonder if `rustfmt` would also change the indentation in that case.

lychee-lib/src/lib.rs
```diff
@@ -20,7 +20,7 @@
 //!
 //! #[tokio::main]
 //! async fn main() -> Result<()> {
-//!   let client = ClientBuilder::default().build()?;
+//!   let client = ClientBuilder::default().client()?;
```
mre, line 23: Do you think we could add a `ClientBuilder::client()` function, which just calls `ClientBuilder::default().client()?` internally? Would be easier to read and reduce some boilerplate for the user.

mre:
I love it. Added some comments. Other than that all good.

lychee-lib/src/filter/excludes.rs
lebensterben: In a second thought, this is doable but will makes the filtering messy.

Because when only `includes` are provided without `excludes`, we require URLs to be explicitly included.

If we add false positives as default excludes, this mechanism need to be changed.

So the updated code removed these TODOs.

lychee-lib/src/filter/mod.rs
lebensterben: I removed that from `exclude.rs`.

lychee-lib/src/filter/mod.rs
lebensterben: Fixed in new commit

lychee-lib/src/filter/mod.rs
lebensterben: No, the grammar is correct.

Consider `there's work to be done` and `there's work to do`.
Active infinitive focuses on the agent of action.
This is the documentation of the behaviour of a `Filter`. So the focus in on `Filter`.

If you use passive infinitive, the focus transferred to the receiver, which is the `Url`.

lychee-lib/src/filter/mod.rs
lebensterben: This is a common misusage.

See the section title: https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

lychee-lib/src/filter/mod.rs
lebensterben: Fixed in next commit

lychee-lib/src/filter/mod.rs
lebensterben:
```rust
        if
        // Exclude well-known false-positives
        // Performed after checking includes to allow user-overwriddes
        is_false_positive(input)
                // Previous checks imply input is not explicitly included,
                // if excludes rules is empty, then *presumably excluded*
                || self.is_excludes_empty()
                // If excludes rules matches input, then
                // *explicitly excluded*
            || self.is_excludes_match(input)
        {
            return true;
        }
```

This is also hard to read

lychee-lib/src/lib.rs
lebensterben: I can do
`ClientBuilder::client() == ClientBuilder::builder().build().client() == ClientBuilder::default().client()`

But if you want to supply additional configuration, it becomes
`ClientBuilder::builder().foo().bar().build().client()`
which is too different.

lebensterben:
The publish error is due to `lychee-lib` on `docs.rs` uses `derive-builder`, so it cannot find the `typed-builder` dependency.
Again we need to first publish `lychee-lib`.

Also, this is a breaking change so it would be `0.7.0`, or `0.7.0.alpha`.

lebensterben:
I will make an attempt to improve the builder interface.
I'll try to put the internal builder in a non-public module. And expose a new builder struct with better API interface.
The downside of typed-builder is that it's less flexible. E.g. you cannot provide custom build() function. But I will try to overcome that limitation.

So hold on and don't merge now.

lychee-lib/src/lib.rs
lebensterben: The conclusion is I cannot...

`ClientBuilder::builder()` returns a `ClientBuilderBuilder<T1, ..., TN>`, where the generic parameters corresponds to each field of `ClientBuilder`.

When you set options, it just store it in the position of one of the generic parameter.
Then you call `ClientBuilderBuilder<T1, ..., TN>::build()`  to get a `ClientBuilder`.

`ClientBuilder::default()` is equivalent to `ClientBuilder::builder()::build()`.
I've already made a `ClientBuilder::client()`, but it takes a finished `ClientBuilder` with all fields set up.

So we have to accept the change in API.

The generated internal builder from `typed-builder` is ugly.
It's a DFA, of initial state `ClientBuilderBuilder<(), .., ()>` to a finished state `ClientBuiulderBuilder<T1, ..., TN>`.
So for example, say we have `ClientBuilder{ foo: u8, bar: u16}`.
It will generate `ClientBuilderBuilder<T1, T2>`.
And something like
```rust
impl ClientBuilderBuilder<T1, T2> {
  fn foo<U: Into<u8>>(self, value: U) -> Self<u8, T2>;
  fn bar<U: Into<u16>>(self, value: U) -> Self<T1, u16>;
  fn build(self) -> ClientBuilder;
}
```

It's possible to hack around to achieve same API as before. But it's painful.

mre: No worries. It's an acceptable tradeoff given the benefits. Thanks for the investigation.

PR #224, opened 2021-04-15
Optional features

lebensterben:
- Some optional features are unnecessary and are removed.
- `console` can pad strings, but it currently needs a patch. And `pad`
  is no longer needed.
- Added optional features to `lychee`
  - `config_file` enables setting options from a TOML file.
  - `json_output` enables outputting in JSON format.
  - `indicatif` enables progress bar.
  - By default they're all enabled.
  - Unsupported CLI options are hidden accordingly.

Next:
- Replace `hubcaps` with `github-rs` to see whether M1 architecture can build `lychee`.
- Updated CI to check different combinations of features.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -123,19 +123,19 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "1.3.1"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9315f8f07556761c3e48fec2e6b276004acf426e6dc068b2c2251854d65ee0fd"
+checksum = "fcb9af4888a70ad78ecb5efcb0ba95d66a3cf54a88b62ae81559954c7588c7a2"
 dependencies = [
  "concurrent-queue",
  "fastrand",
  "futures-lite",
  "libc",
  "log",
- "nb-connect",
  "once_cell",
  "parking",
  "polling",
+ "socket2 0.4.0",
  "vec-arena",
  "waker-fn",
  "winapi",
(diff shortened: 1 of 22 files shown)


lebensterben:
I will rebase once #222 is merged.

mre:
> Replace hubcaps with github-rs to see whether M1 architecture can build lychee.

github-rs doesn't seem to be actively maintained at this point (last commit in May 2019). 
In fact I used github-rs before I [moved to hubcaps](https://github.com/lycheeverse/lychee/commit/3650c673df8c76ee1d5628efbfe5675d1a535599#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) because it was more actively maintained and had support for async reqwest, which github-rs [does not](https://github.com/github-rs/github-rs/issues/72) afaik.

lebensterben:
@mre 
But hubcaps doesn't work well on M1.

mre:
That's true. If we can we should support them with adding that. My feeling is we'll run into other issues with github-rs further down the line. To be honest I don't mind keeping the explicit dependency on ring right now to have M1 support.

lebensterben:
github-rs will not be the big issue because GitHub API is stable, and we only need very minimal API (Token and Repository).

The true problem is the rust-native-tls it depends on. This doesn't rely on ring but it is not actively developed.

So we can use github-rs for aarch.apple.darwin, and in the long run M1 compatibility should get better.

Also hubcaps is not well-maintained. It uses old dependencies which is a serucity risk.

mre:
That's fine with me, even though I'd say depending on two outdated crates is more of a security risk than just one. I'd personally just keep ring as an explicit dependency for now and try to get an updated version of hubcaps out, which would help other downstream crates as well.

lebensterben:
There's something wrong with CI.
It's not reading the `#[cfg()] attributes.

mre:
Humm... no clue what's going on yet.

mre:
With #330 merged, can you rebase your changes on top of `master` @lebensterben?

mre:
@lebensterben can you rebase this to take a look at the failing `#[cfg()]` flags on CI again?

lebensterben:
@mre 
it may take a few days.
I've another big repository to maintain.

lebensterben:
After some review I concluded that it's much easier to restart a PR...
So I'm closing this.

PR #225, opened 2021-04-15
More refactor

lebensterben:
- Major changes in `lychee-lib::filter` module:
  - Fields in `Excludes` except the `RegexSet` is now moved to `Filter`.
  - `Filter` contains `Option<Excludes>` and `Option<Includes>`, which are
    wrapper struct of `RegexSet` instead of `Option<RegexSet>`. As a result
    the code now looks cleaner.
  - Factored out some filtering logics to dedicated functions.
    - It's possible to write tests for those functions in addition to tests
      for the `Filter` struct.
  - Added docs to `Filter::is_excluded` and reorgnized the code.
- placed `derive_builder` by `typed_builder`:
  - The internal interface very ugly, as admitted by the author, but we no
    longer have nested `Option`s like before.
  - As a result, the `Client` building is much easier to read.
  - Main benefit of `typed_builder` is, the arguments feeded to builder is
    checked at compile time instead of run-time.
- Fixed a bug in `lychee::tests::usage` and `lychee-lib::stats::test`.
  - Now it will clear environment variable which would otherwise cause an
    issue if `GITHUB_TOKEN` is set.
- Updated dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -430,7 +430,7 @@ dependencies = [
  "ansi_term 0.11.0",
  "atty",
  "bitflags",
- "strsim 0.8.0",
+ "strsim",
  "textwrap 0.11.0",
  "unicode-width",
  "vec_map",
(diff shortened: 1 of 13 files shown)


lebensterben:
Well #222 was closed since I've renamed the branch to work with git-flow.

mre:
lgtm. There is one final error in https://github.com/lycheeverse/lychee/pull/225/checks?check_run_id=2354928844 that should be an easy fix.

lebensterben:
You can build it locally, since `cargo test` passes.
`cargo publish` instead look for the package from registry, which is the one published yesterday. So it complains about not finding the function.

Again we need to publish the `lychee-lib` first.

P.S. And I haven't changed the version number. It could be v0.7.0.alpha since the API changes.

mre:
Works locally. Let's merge it. 😄

PR #246, opened 2021-05-10
Don't panic on unknown status codes

### LINKED ISSUE #234: Handle unknown status codes
A Lychee worker panics when encountering a website with an invalid and unaccepted status code.
For example, LinkedIn returns the status code 999 by default.

```
echo https://www.linkedin.com/in/endlermatthias | lychee -
```

This results in the following output:

```
thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code', lychee-lib/src/types/status.rs:67:21
```

A workaround is to accept the status code like so:

```
echo https://www.linkedin.com/in/endlermatthias | lychee - --accept 999
```

Also we can handle LinkedIn with a quirk by using a different user agent.

But in general we are missing a case where a request returns an unexpected status code here:

https://github.com/lycheeverse/lychee/blob/d651b1ef7bb6e42cf96c334be3556c71a5574293/lychee-lib/src/types/status.rs#L60-L68

@lebensterben is that something you could tackle?

mre:
Fixes #234

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ Cargo.lock
 
 # IntelliJ generated files
 *.idea
+
+# VSCode generated files
+.vscode
--- a/lychee-bin/src/stats.rs
+++ b/lychee-bin/src/stats.rs
@@ -15,9 +15,8 @@ pub(crate) fn color_response(response: &ResponseBody) -> String {
     let out = match response.status {
         Status::Ok(_) => style(response).green().bright(),
         Status::Excluded | Status::Unsupported(_) => style(response).dim(),
-        Status::UnknownStatusCode(_) => style(response).yellow().bright(),
         Status::Redirected(_) => style(response),
-        Status::Timeout(_) => style(response).yellow().bright(),
+        Status::UnknownStatusCode(_) | Status::Timeout(_) => style(response).yellow().bright(),
         Status::Error(_) => style(response).red().bright(),
     };
     out.to_string()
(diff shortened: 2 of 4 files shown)


pawroman:
Thinking about it, it's the most reasonable thing to do when we simply don't know what the status code might mean.

Code looks good, too. Nice little PR :+1:

mre:
Finally merged. 🎉

fauust:
I have just hit this bug, I can confirm that the latest image from docker is correctly handling this:

```console
❯ docker run -it lycheeverse/lychee:0.7.1 --version
lychee 0.7.1
❯ docker run -it lycheeverse/lychee:0.7.1 https://mariadb.org
⠂ 129/156 [00:00:00] ████████████████████░░░░░ ✔ https://www.continuent.com/resources/blog/testing-time-and-test-suites-managemen
thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code', lychee-lib/src/types/status.rs:73:26
❯ docker run -it lycheeverse/lychee:latest --version
lychee 0.7.2
❯ docker run -it lycheeverse/lychee:latest https://mariadb.org
📝 Summary
---------------------
🔍 Total..........156
✅ Successful.....151
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
⚠ Unknown..........1
🚫 Errors...........4
```

mre:
Thanks for the feedback. The only thing that irritates me is that the ⚠️ emoji is not colored in the output. I tried to fix that, but so far no luck.

PR #274, opened 2021-07-04
docs: lychee is also available on Termux

thunder-coding:


mre:
Thanks so much @thunder-coding. Wonderful to support another platform. 😊

Issue #302, opened 2021-08-31
Load --exclude from a file


dblock:
We like having a link checker in https://github.com/opensearch-project/OpenSearch, but we also often encounter URLs that no longer work in things like license files. Therefore we can't just fix the links. This leads to a fairly long list of URLs to exclude, which we keep adding to a GitHub action job, which reads poorly.

In https://github.com/opensearch-project/OpenSearch/pull/1189 I've hacked our workflow to read the exclusions from a file, but it's not an ideal implementation :) 

Should we add `--exclude-file` to lychee?

lebensterben:
@dblock 
lychee currently support config file. so at least you can put those urls in it.

dblock:
> @dblock
> lychee currently support config file. so at least you can put those urls in it.

Oh where/how? I see README mentions a config file, but can't find a reference/example.

lebensterben:
indeed.

mre:
Good idea!

dblock:
This was implemented in #306, and released in 7.0.1.

PR #306, opened 2021-09-01
Implements --exclude-file

dblock:
Implemented https://github.com/lycheeverse/lychee/issues/302.

This is my first rust code, so talk to me like a 2yo :)

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -63,6 +63,8 @@ use ring as _;
 
 use std::iter::FromIterator;
 use std::{collections::HashSet, fs, str::FromStr, time::Duration};
+use std::io::{self, BufRead};
+use std::fs::File;
 
 use anyhow::{anyhow, Context, Result};
 use headers::{authorization::Basic, Authorization, HeaderMap, HeaderMapExt, HeaderName};
@@ -113,6 +115,18 @@ fn run_main() -> Result<i32> {
     if let Some(c) = Config::load_from_file(&opts.config_file)? {
         opts.config.merge(c)
     }
+
+    // Load excludes from file
+    for path in &opts.config.exclude_file {
+        let file = File::open(path).expect("No such file");
+        opts.config.exclude.append(
+            &mut io::BufReader::new(file)
+                .lines()
+                .map(|l| l.expect("Could not read line"))
+                .collect()
+        );
+    }
+
     let cfg = &opts.config;
 
     let runtime = match cfg.threads {
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -118,13 +118,10 @@ fn run_main() -> Result<i32> {
 
     // Load excludes from file
     for path in &opts.config.exclude_file {
-        let file = File::open(path).expect("No such file");
-        opts.config.exclude.append(
-            &mut io::BufReader::new(file)
-                .lines()
-                .map(|l| l.expect("Could not read line"))
-                .collect(),
-        );
+        let file = File::open(path)?;
+        opts.config
+            .exclude
+            .append(&mut io::BufReader::new(file).lines().collect::<Result<_, _>>()?);
     }
 
     let cfg = &opts.config;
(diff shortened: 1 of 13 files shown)


lychee-bin/src/main.rs
```diff
@@ -113,6 +115,18 @@ fn run_main() -> Result<i32> {
...
         opts.config.merge(c)
     }
+
+    // Load excludes from file
+    for path in &opts.config.exclude_file {
+        let file = File::open(path).expect("No such file");
+        opts.config.exclude.append(
+            &mut io::BufReader::new(file)
+                .lines()
+                .map(|l| l.expect("Could not read line"))
```
lebensterben, line 125: there's no need to manually unwrap a Result in Iterator.
If you've a Itertor<Item = Result<T, E>>, you can collect it directly to a Result<Vec<T>, E>
- If there's any Err, it returns an Err
- If all elements arr Ok, it returns Ok(Vec<T>)

lebensterben:
I think your original proposal of `.lycheeignore` is great.
We should make it default.

And this also allows the possibility for per-directory exclusion list.

dblock:
> I think your original proposal of `.lycheeignore` is great.
> We should make it default.
> 
> And this also allows the possibility for per-directory exclusion list.

It's a new construct for Lychee, so I think we should think about it more. The thing with things like `.gitignore` is that they git walks the parent path to find the closest .gitignore until it reaches the git root. This won't be possible here, so maybe it's too much magic? 

One way to implement that would be to default --exclude-file to `.lycheeexclude` if that exists. Can be built on top of this PR.

lebensterben:
>  The thing with things like .gitignore is that git walks the parent path to find the closest .gitignore until it reaches the git root. 

lychee supports 'glob'.
so you can do something like `lychee ~/notes/**/*.md`.
when expanding the glob pattern it's already walking the directories.

lychee-bin/src/main.rs
```diff
@@ -113,6 +115,18 @@ fn run_main() -> Result<i32> {
     if let Some(c) = Config::load_from_file(&opts.config_file)? {
         opts.config.merge(c)
     }
+
+    // Load excludes from file
+    for path in &opts.config.exclude_file {
+        let file = File::open(path).expect("No such file");
```
mre, line 121: Instead of `.expect (...)` can you use a `?` here? Makes it a little cleaner.

mre (changes requested):

mre:
Thanks for the PR. 
I also like the idea of a `.lycheeignore`. In contrast to the `.gitignore`, a `.dockerignore` will only be parsed from the build context (usually the current directory), so I would not be too concerned about the differences between the implementations as long as we document the behavior and add tests. 

As you both mentioned it could be a nice addition in the future.

lychee-bin/src/main.rs
mre: @dblock I resolved this by pushing to your branch if you want to have a look at the diff.

lychee-bin/src/main.rs
mre: Resolved by pushing my changes to the branch.

lychee-bin/tests/cli.rs
```diff
@@ -364,10 +364,76 @@ mod cli {
...
+        let expected = r#"{"total":10,"successful":10,"failures":0,"timeouts":0,"redirects":0,"excludes":0,"errors":0,"fail_map":{}}"#;
         let output = fs::read_to_string(&outfile)?;
         assert_eq!(output.split_whitespace().collect::<String>(), expected);
         fs::remove_file(outfile)?;
         Ok(())
     }
+
+    /// Test excludes
+    #[test]
+    fn test_exclude_wildcard() -> Result<()> {
```
mre, line 376: Thanks for adding these additional exclude tests and increasing the test coverage. ❤️

mre:
I'll merge this now. Expect a new release soon. Created a ticket for discussing the future addition of `.lycheeignore` at https://github.com/lycheeverse/lychee/issues/308. 
Thanks for first (but hopefully not last) your contribution @dblock. ⭐ And welcome to the Rust community! 🦀

PR #309, opened 2021-09-03
Fix publish workflow

### LINKED ISSUE #305: Publish workflow fails
Since https://github.com/lycheeverse/lychee/pull/225 the publish CI check is failing, which seems to be "by design". This was confusing to me trying to contribute for the first time - I wasn't sure whether I broke something. 

Maybe that workflow should be disabled/removed until it can succeed?

dblock:
Fixes #305 by using [cargo-publish-all](https://crates.io/crates/cargo-publish-all) which creates a workspace first, then publishes lychee-lib into it, then publishes lychee-bin using what it just published. So now we can increment the version of lychee-lib and lychee-bin to the *next* iteration which is the code that we have now on master.

This requires a newer version of rust-type-builder with the fix for https://github.com/idanarye/rust-typed-builder/issues/57.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -37,7 +37,7 @@ jobs:
           args: --all-targets -- --deny warnings
 
   publish-check:
-    name: Publish Check
+    name: publish check
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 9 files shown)


mre:
Hey that's great, thanks! Your persistence paid of.

dblock:
I think the actual publication should also change to use `cargo-publish-all`, I'll leave this to you.

mre:
🤔 you have a point there

PR #318, opened 2021-09-09
Add support for tokio-console

mre:
This enables debugging and improvement of asynchronous Rust code.
It is an optional feature that is still
experimental (but can be quite helpful)

More info: https://github.com/tokio-rs/console

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,7 +202,7 @@ dependencies = [
  "fast_chemail",
  "hostname 0.1.5",
  "log",
- "nom",
+ "nom 5.1.2",
  "pin-project",
  "pin-utils",
  "serde",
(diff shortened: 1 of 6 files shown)


lebensterben:
should this be a dev dependency?

we should always avoid over-bloating the binary to ensure a fast deployment in CI environment.

mre:
You're absolutely right. Thanks for catching that.

mre:
@lebensterben I'm getting "dev-dependencies are not allowed to be optional". Given that `tokio-console` is marked as `optional` anyway, wouldn't it be fine if it was a normal dependency then? If not, what would be the correct approach in your opinion?

lebensterben:
@mre you're definitely right.
I failed to notice that it's already optional.

PR #325, opened 2021-09-13
Add a description on DockerHub registry

fauust:
This adds a simple CI step to update the DockerHub registry with the README.md file.

For the moment, the registry overview is empty:

screenshot_20210913_164222

See #321

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,3 +21,10 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
           snapshot: true
           tag_semver: true
+      - name: Update DockerHub description
+        uses: peter-evans/dockerhub-description@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: lycheeverse/lychee
+          readme-filepath: README.md


mre:
Nice! 💪

fauust:
Hmm, I did not check that you are using token for authentication to the registry, this action still does not support it.
https://github.com/peter-evans/dockerhub-description/issues/10

mre:
Oh okay. Two options: either we switch to username/password or we disable the docs publishing step until the issue gets resolved. I have my doubts that this will be anytime soon, but I also don't want to switch to user/pw. Suggestions? 🤔

fauust:
> I have my doubts that this will be anytime soon,

This is not going to happen and since some month Docker folks are radically changing the offer for free accounts.

I had to disable my 2FA authentication and go back to a user/pwd authentication (which I did not like too) and I understand that this is not a move that you want.

So for the moment I have no solution to propose but I will investigate in the next days.
Maybe you can rollback for the moment.

fauust:
There is a manual solution in the mean time. You can edit the description directly in the registry (in markdown), see bellow:

screenshot_20210914_185737

You may want to fetch the banner from a public server (`assets/banner.svg` will not work on DockerHub).

This is clearly not a nice and definitive option, but if the `README.md` does not change very often, it should be OK for the moment.

This is the reason why I think having a description is important for this project:
https://hub.docker.com/search?q=lychee&type=image

mre:
Great idea. I've updated the docs now and added a tagline.

image

mre:
Would also love to set an icon/logo/avatar, but that's only possible with a pro account it seems. 😕 That little lychee would look amazing IMHO. 😆

fauust:
@mre remember to rollback this merge as the CI is now broken because of me ;)

mre:
Cool. Rolled back the change. Thanks anyway for looking into this. Let's try again at a later point in time. 😃

fauust:
Or when there is a better registry alternative for free accounts ;)

PR #330, opened 2021-09-14
Improve concurrency with streams

mre:
The goal of this effort is to avoid blocking the runtime as much as possible.
This is a multi-step approach:

* [x] Change `input::get_contents` from returning`Vec<InputContent>` to `Stream<Item=Result<InputContent>>`.
* [x] Change `Collector::collect_links` from returning `Result<HashSet<Request>>` to `Stream<Item=Result<InputContent>>`. (1)
* [x] Remove deadpool from the dependencies in lychee-lib and only work with streams throughout the call-stack.
   https://github.com/lycheeverse/lychee/pull/355
* [x] ~Remove `ClientPool` as reqwest already has a built-in pool.~
* [x] Simplify `main.rs` by removing mpsc channels and iterating over streams directly.
* [x] ~Optional: Move the link cache out of the collector. Use a [XOR filter](https://github.com/ayazhafiz/xorf) instead of a HashSet. It uses less memory.~

See #193, https://github.com/lycheeverse/lychee/issues/21#issuecomment-913877831

@TimoFreiberg fyi


TimoFreiberg:
Making `Collector::collect_links` return a Stream is pretty straightforward: https://github.com/TimoFreiberg/lychee/commit/a05400b28bf13baf40cf0b6a3af3f34b29c0967b.
I'm tempted to add [the progressbar changes](https://github.com/lycheeverse/lychee/pull/193/files#diff-75ca6f70c4137d8c64514f688fa83f080ea786c3b6f87fa6b11214be199ed995R132) from my previous PR, do you want them here or not?

I'm not 100% sure how the rework of `main.rs` will look like. I think we'll still need at least one Tokio Task?
Because changing one of the Tasks spawned in `main.rs` to handle Results is something I would like to discuss with you beforehand 😄

mre:
@TimoFreiberg your changes look good. I wonder how to proceed. Shall I just merge in the changes from your branch into this PR or shall we continue working on your fork?

mre:
The progressbar changes can be added here, too yes.

TimoFreiberg:
I think merging my changes into your branch is fine. I don't have access to
a computer this weekend so feel free to continue on your own :)

TimoFreiberg:
Alright, I have a proposal for a minimal change in the tokio tasks here: https://github.com/lycheeverse/lychee/commit/98cdfba9585016205b8e9b718167b688ecfc56e2

mre:
This is great. I have to merge in your changes into my branch when I find a few minutes. 👍

mre:
>  feel free to merge TimoFreiberg@28ad683

Worked nicely on my machine as well. Thanks for fixing that!

> What do you think?

Great idea. In the meantime we added the `ClientPool` back in https://github.com/lycheeverse/lychee/pull/355 because we ran into the same issues. I'll benchmark your approach, to see how it compares to the `ClientPool`. My hope is that it's equally fast so that we can remove the pool to simplify the code and remove dependencies. 👍 

I also merged a couple commits for basic directory support. Want to land that together with streams, because it's way easier and more elegant to implement it with `yield` instead of collecting all files into a Vector before returning (which kind of would defeat the point of walking files, heh).

MichaIng:
Time to change the rustfmt rules?
```diff
-        assert_eq!(FileType::from(Path::new("http://foo.com/index.html")), FileType::Html);
+        assert_eq!(
+            FileType::from(Path::new("http://foo.com/index.html")),
+            FileType::Html
+        );
```
As long as it's not a large number of arguments, or very long ones, I don't agree that multiple lines look cleaner here 😄.

https://rust-lang.github.io/rustfmt/?version=v1.4.36&search=#fn_args_layout
- The default "Tall" actually seems reasonable, only the break point may be too low, for my personal taste at least.
- However, wanted to push that door open that it indeed can be configured very well 🙂.

mre:
Heh, I prefer the `rustfmt` multi-line default here. That just goes to show that formatting is very subjective. 😉 
On a more practical note, I found that splitting complex statements over multiple lines makes for cleaner diffs because sometimes only a single part gets changed, so the diff might be on a single line, which is easier to review for me.

MichaIng:
Definitely a question of personal taste 🙂. Totally fine to keep it like that, I see the benefits: On GitHub, also changes within a line are made visible nicely, but on in a regular diff. Just good to know that and how it is possible to adjust these rules to match own coding standards.

mre:
@MichaIng I agree. Good to know that we can customize it if the need arises.

mre:
Okay so I did quite some more refactoring now. The tests are failing and stuff, but it builds and runs, heh.
I haven't done any sophisticated testing yet, but from what I can see, this branch is _crazy_ fast -- especially on big directories. Startup time is literally instant. Not surprising, given that we do everything in a stream now.
It also handles directories as inputs now, which is very nice for testing.
Can't wait for this to land.

@MichaIng, @TimoFreiberg, @untitaker wanna take it for a spin? My hardware is not the newest, so I'd be interested in some benchmarks. 😄 Don't forget to compile in release mode.

mre:
PS: I decided that caching is out of the scope of this PR. It's already quite big.

MichaIng:
Awesome, many thanks for your work 👍. Yes I agree that it is huge and caching deserves an own PR. Also it allows to better compare the effect of each when implementing in separate steps.

I'll setup a test branch on our two `lychee` checked projects tomorrow to compare with the main branch 🙂.

TimoFreiberg:
Only performed a rudimentary test on lychee's README.md - quite busy right now unfortunately.
If I can do more to help, I could really use a benchmark command I can copy-paste as I don't have much brain left over in the evening😅 

```
$ command time target/release/lychee -- README.md
Issues found in 1 input(s). Find details below.

[README.md]:
✗ https://example.org/README.md: Network error while trying to connect to an endpoint via reqwest

🔍 45 Total ✅ 44 OK 🚫 1 Error (HTTP:1)
Command exited with non-zero status 2
0.05user 0.05system 0:08.02elapsed 1%CPU (0avgtext+0avgdata 22796maxresident)k
0inputs+0outputs (0major+2345minor)pagefaults 0swaps

```

mre:
Thanks @TimoFreiberg.
I used hyperlink for some testing.
Found a way to execute the stream in parallel and it's way faster again. The problem is that I need to fix the error handling.
Running into some issues with async Rust.
If you or anyone has an idea, here's the thread in the Rust support forum: https://users.rust-lang.org/t/async-stream-with-item-result/68066

untitaker:
@mre since you're extending the API, which command line should I replace this with? I assume that "directory is now an input" means I can add it as posarg, and absence of globs may affect performance

```
lychee --offline -b . '**/*.htm*'
```

untitaker:
`time lychee --offline -n .` from this branch on sentry-docs takes

```
real    0m2.888s
user    0m12.182s
sys     0m0.564s
```

while `time lychee --offline -b . '**/*htm*' -n` from master takes

```
real    0m5.379s
user    0m19.369s
sys     0m5.381s
```

this is a great improvement, and generally lychee seems to be in the same ballpark as hyperlink (somewhere between muffet/liche and hyperlink), but as for filesystem-only checking I think there are bigger areas of improvement that lychee can make than anything in relation to parallelism.

See [this flamegraph](https://gist.githubusercontent.com/untitaker/23809824e1cef69b1bd67e214ba7334c/raw/ce59ff91189c6d6a274c75cbf1e38a8fe9b23d4d/lol.svg). This is an interactive SVG with embedded JS generated via `cargo flamegraph`, for some reason you need to download it from github and open it as local file to get the most readable output. Maybe Content-Security-Policy? Who knows.

Two big blocks:

* indicatif continues to be _really slow_. In fact the above test was closer to running at 1 minute before I added `-n`. I think this is not a good default experience

* html parsing is slow. I know that using html5ever's `Tokenizer` more than doubled hyperlink's runtime. I see that you're constructing an entire tree in lychee instead of iterating over tokens. [Shamelessly plugging my own html parser at this point,](https://github.com/untitaker/html5gum) but swapping out html5ever with literally anything else, or replacing the document parser with just the [Tokenizer](https://docs.rs/html5ever/latest/html5ever/tokenizer/struct.Tokenizer.html) will give you some boost. Quickly skimming over what you do with the element tree, you don't really need anything but tokens.

MichaIng:
I tested our two sites and here is the execution time release => stream branch:
- 8s (613 links) => 7s (607 links)
- 18s (3347 links) => 12s (3313 links)

So jep, it's faster 👍.

As can be seen, also the amount of checked links was reduced somehow, while the checked inputs didn't change. I know a cross-file cache wasn't implemented, but it has been changed from a link to a request cache, if I see right? May it be related to that that some links which were seen as distinct/different before are after processing seen as identical requests or so?

mre:
If we can somehow get https://users.rust-lang.org/t/async-stream-with-item-result/68066 resolved then I can push my local changes which give me a cool 2x improvement. 
Using only the tokenizer is a great idea. We can then return a token stream from the extractor as well. Actually we could even par_iter on the tokens (with rayon) for another potential boost. 
I remember that someone posted a link to a discussion about indicatif's performance issues. Something about locking but I forgot. I'll post it here if I can dig it up again.

mre:
@untitaker this should also work now:
```
lychee --offline -b . .
```

(not tested)

Note that this will include markdown files as well at the moment. I still need to add the overwrite option for that. (Something like `--extensions html,htm` could work in the future.)

MichaIng:
About HTML parsing: At least for some feature requests it is required, either on the inputs or even on the target HTML document:
- #259
- #185

But probably it can be done conditionally, only if an option/flag requires it.

untitaker:
I don't think it's required for anchors but yes, if lychee doesn't consider itself to be feature complete, moving to a tokenizer is limiting future features too much

MichaIng:
> I don't think it's required for anchors

Meant is to also check whether the URL `https://foo.bar#fragment` does not only respond well but that the fragment is valid as well, hence that an element with the ID `fragment` exists in the target HTML. For this, the whole link target needs to be downloaded (at least when it is `content-type: text/html`) and parsed for a matching HTML element. Sounds heavy, but indeed it is useful when e.g. checking internal links in a documentation with large pages where you definitely want to link to a specific heading instead of the start of the page.

untitaker:
> For this, the whole link target needs to be downloaded (at least when it is content-type: text/html) and parsed for a matching HTML element.

I understand but that requires parsing more documents, not any particular kind of HTML parsing that a tokenizer wouldn't support. you need to find the start tag that contains the right id. any start tag is fine, you don't even need to find the right one. You don't need to understand hierarchy of elements for example.

mre:
Okay I think we're done here.

```
❯❯❯ hyperfine "lychee --base '/Users/mendler/Downloads/sentry-docs/public' '/Users/mendler/Downloads/sentry-docs/public' --offline --dump --no-progress"
Benchmark 1: lychee --base '/Users/mendler/Downloads/sentry-docs/public' '/Users/mendler/Downloads/sentry-docs/public' --offline --dump --no-progress
  Time (mean ± σ):      4.589 s ±  0.379 s    [User: 23.370 s, System: 7.124 s]
  Range (min … max):    3.995 s …  5.167 s    10 runs
```

```
Summary
  'lychee --base '/Users/mendler/Downloads/sentry-docs/public' '/Users/mendler/Downloads/sentry-docs/public/**/*.html' --offline --dump' ran
    1.35 ± 0.11 times faster than 'lychee-old --base '/Users/mendler/Downloads/sentry-docs/public' '/Users/mendler/Downloads/sentry-docs/public/**/*.html' --offline --dump'
```

Using jwalk for directory traversal:

```
Summary
  'lychee --base '/Users/mendler/Downloads/sentry-docs/public' '/Users/mendler/Downloads/sentry-docs/public' --offline --dump' ran
    1.38 ± 0.16 times faster than 'lychee-old --base '/Users/mendler/Downloads/sentry-docs/public' '/Users/mendler/Downloads/sentry-docs/public/**/*.html' --offline --dump'
```

So in summary we're around 35% to 50% percent faster with the stream-based version. I honestly expected a bit more, but at least the architecture is more promising now and the cores get fully used from the beginning many files. This way the startup time is much better and it shouldn't get stuck when it runs out of memory.

I'm planning to merge this tomorrow so this is the "final comment period" for the changeset. Everything else that isn't covered yet (optimizing the extractor, caching) will be handled by a separate PR. Thanks all for your support!

mre:
🥳 🥳 🥳

PR #332, opened 2021-09-16
Don't try to publish Docker image on fork

mre:
Discussion: https://github.com/lycheeverse/lychee/issues/321#issuecomment-919161530
@fauust is that how you imagined it or is there a better way?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,11 +10,11 @@ on:
 name: Publish Docker Image
 jobs:
   build:
+    if: github.repository_owner == 'lycheeverse'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
       - name: Publish to Registry
-        if: github.repository_owner == 'lycheeverse'
         uses: elgohr/Publish-Docker-Github-Action@3.01
         with:
           name: lycheeverse/lychee
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,6 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@master
       - name: Publish to Registry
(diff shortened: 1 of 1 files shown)


fauust:
Yep, I had prepared something very similar on a local branch:

```diff
commit 66069c1df9e6e9c70c84821ce448197e71462611
Author: Faustin Lammler <faustin@fala.red>
Date:   Wed Sep 15 09:31:08 2021 +0200

    s

diff --git a/.github/workflows/docker.yml b/.github/workflows/docker.yml
index 541c542..a882169 100644
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,6 +8,7 @@ on:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10

 name: Publish Docker Image
+if: github.repository == 'lycheeverse/lychee'
 jobs:
   build:
     runs-on: ubuntu-latest
```

I am not sure where the `if` should go in order to skip the whole job (not sure if the checkout and the start of the VM would be skipped in your case). You can test it by removing `on > push > branch > master`.

mre:
Great idea to move the condition higher up the chain.
The [workflow syntax](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) describes where conditions are allowed to go. I checked and it turns out there are two places:
* `jobs.<job_id>.if`
* `jobs.<job_id>.steps[*].if`

So I moved it up to the jobs level.
Furthermore if you use the online editor for workflow files, it will underline errors, which is very convenient.
I think it should work the way it's written now.

fauust:
@mre I have just seen that there is another step that might be good to disable on forks:
https://github.com/fauust/lychee/actions/runs/1251246990

mre:
Nice one. Let me fix that, too. 😅

PR #345, opened 2021-09-23
Cache `create_uri_from_path` to decrease allocations

mre:
While profiling local file handling, I noticed that resolving paths was taking a
significant amount of time. It also caused quite a few allocations.
By caching the path and using a constant value for the current
directory, we can reduce the number of allocs by quite a lot.
For example, when testing on the sentry documentation, we do 50,4%
less allocations in total now. That's just a single test-case of course,
but it's probably also helping in many other cases as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -420,6 +420,40 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "631ae5198c9be5e753e5cc215e1bd73c2b466a3565173db433f52bb9d3e66dba"
 
 [[package]]
+name = "cached"
+version = "0.25.0"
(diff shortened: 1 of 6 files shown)


mre:
```
cargo instruments --release -t alloc -- --offline --base '/tmp/sentry-docs-master/public' --no-progress '/tmp/sentry-docs-master/public/**/*.html' --timeout 1 --exclude '\.png'
```

Screenshot 2021-09-23 at 15 06 27

mre:
Weirdly enough I can't reproduce the improvements anymore. Must have triggered the run with different arguments, because now the branch is about as fast and uses about as many allocations as `master`... 🤷‍♀️
Guess that's a dead end.

PR #346, opened 2021-09-26
 Cache `absolute_path` to decrease allocations

mre:
While profiling local file handling, I noticed that resolving paths was taking a
significant amount of time. It also caused quite a few allocations.
By caching the path and using a constant value for the current
directory, we can reduce the number of allocs by quite a lot (5%).

This is my second attempt at reducing allocs. The first one is here #345.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -53,7 +53,11 @@ jobs:
       with:
         crate: cargo-publish-all
         version: latest
-    - run: cargo-publish-all --dry-run
+    # This currently doesn't work.
+    # TODO: Re-enable once https://gitlab.com/torkleyy/cargo-publish-all/-/issues/3 
+    # is resolved.
+    # See also https://github.com/lycheeverse/lychee/pull/346#issuecomment-928355735
+    #- run: cargo-publish-all --dry-run
 
   publish:
     if: ${{ startsWith(github.ref, 'refs/tags/') }}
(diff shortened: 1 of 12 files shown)


dblock:
> I'm running into the same issues as in [idanarye/rust-typed-builder#57](https://github.com/idanarye/rust-typed-builder/issues/57) again.

Did that just reappear and is it reproducible on main?

mre:
Oh I might not have expressed myself very well. Just to be clear, the old bug with rust-typed-builder is fixed and that fix still works. But after adding the `cached` crate which also uses macros, I get another compile error about `proc_macro` missing. It's not an issue with the rust-typed-builder or cached crates but with cargo-publish-all.

dblock:
> Oh I might not have expressed myself very well. Just to be clear, the old bug with rust-typed-builder is fixed and that fix still works. But after adding the `cached` crate which also uses macros, I get another compile error about `proc_macro` missing. It's not an issue with the rust-typed-builder or cached crates but with cargo-publish-all.

I see. You should (re)open an issue in cargo-publish-all, cc: @idanarye

mre:
Good point. I've added a comment to https://gitlab.com/torkleyy/cargo-publish-all/-/issues/3.

mre:
There was no update on the issue yet. I've decided to disable the `publish --dry-run` for now as it is blocking the merge and this PR reduces the amount of allocs by 20% from my measurements with [cargo instruments](https://github.com/cmyr/cargo-instruments). The runtime also improved thanks to this, but there is still a lot of low-hanging fruit around glob handling and link extraction, which we can tackle in a separate PR.

image
image

PR #375, opened 2021-10-30
Colorize and simplify status output

mre:
I got a bit annoyed by the current status output.

While it sort of works, it's not too nice to look at and a little annoying to read and understand.
So I reworked it with a human user in mind. The terminal output should not be parsed by machines; that's what we have the JSON output for. That said, we might introduce a third output in the future that is closer to the old output which allows for easier grepping.

The new summary is more compact:

image

The error output also got a revamp and is closer to the the output of [ripgrep](https://github.com/BurntSushi/ripgrep) now:

image

Let me know what you think. Also a code review would be great. 😊

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Link Checker
         uses: lycheeverse/lychee-action@master
         with:
-          args: --verbose --no-progress --exclude example.org
+          args: --verbose --no-progress --exclude example.org -- README.md
 
       - name: Create Issue From File
         if: github.repository_owner == 'lycheeverse'
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1800,9 +1800,9 @@ dependencies = [
 
 [[package]]
 name = "openssl-sys"
-version = "0.9.67"
+version = "0.9.68"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69df2d8dfc6ce3aaf44b40dec6f487d5a886516cf6879c49e98e0710f310a058"
+checksum = "1c571f25d3f66dd427e417cebf73dbe2361d6125cf6e3a70d143fdf97c9f5150"
 dependencies = [
  "autocfg",
  "cc",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.15"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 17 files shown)


MichaIng:
It is about the summary only, right?

Actually, while this is a question of personal taste, I like the multi-line summary more: I see the numbers much quicker in it compared to the one-liner and e.g. the excludes are IMHO an important value when I look at the summary at all (during testing and debugging mostly, else the red/failed URLs are all I'm interested in).

mre:
> Actually, while this is a question of personal taste, I like the multi-line summary more: I see the numbers much quicker in it compared to the one-liner

It's true that once you get used to it, the old status output was easily readable. However we might be biased because we've been using the tool for quite a while. 😆 
What I don't like is that lychee is pretty much the only cli tool that uses dashes as separators for the headline. 

> the excludes are IMHO an important value when I look at the summary at all

That didn't change, no? We still show the number of excludes (if any). 

Overall I'd say the proposed output is better for newcomers and "casual" users. We could print more info (including the detailed summary) by increasing the verbosity level.

MichaIng:
> That didn't change, no? We still show the number of excludes (if any).

Ah sorry, you're right, I think I expected a "0 Excluded" in the last image, but in the one above it is pretty clear 👍.

About one-line vs multi-line: Having `<icon> <text> <count>` in a single line makes everything pretty clear to me that these three are related to the same info on first view and one knows where to look at for icon text and count on the other lines. Having everything in one line for me makes it more difficult to find and associate these three elements for each info. Could be biased but I look at a lot of different command line outputs throughout the day and less lines usually don't make things better for the human reader, at least not for me 😉.

But this is not a blocker at all from my end. As said, usually you want the failed inputs and URLs, that's all, and those are listed pretty clear. And whether it takes a microsecond longer or shorter for me to verify whether URI numbers are as expected doesn't really matter. Just as you asked for feedback and stated that "it's not too nice to look at and a little annoying to read and understand", for me it isn't/wasn't annoying to look at or hard to understand.

Hard to get feedback from newcomers or generally more users on GitHub, that would indeed help for such visual format decisions which are either affected by accustoming or individuality 🤔.

mre:
## Compact format (default)

image

## Detailed format (--format=detailed)

This one is similar to our old format.

image

## Markdown format (--format=markdown)

image

This one is really useful for error reporting on Github such as through [lychee-action](https://github.com/lycheeverse/lychee-action/). Here's the rendered version:

image

---

@MichaIng I've added `--format=detailed` to keep supporting the old format.
Apart from that, I'd say the change is ready to ship if there are no objections.

mre:
Also just realized that you're trending on Github @MichaIng. 😄 
image

MichaIng:
Oh, `--format=detailed` wasn't necessary just for me, but many thanks for implementing it 🙂. `--format=markdown` is pretty welcome for everyone who creates GitHub issues from the result 👍.

> Also just realized that you're trending on Github @MichaIng. 😄

Nice, I didn't see that one, thanks for sharing. Especially the "POPULAR REPO: DietPi" makes me proud 👏.

PR #424, opened 2021-12-15
Use tokenizer for extraction; add benchmark

mre:
This avoids creating a DOM tree for link extraction and instead uses a `TokenSink` for on-the-fly extraction.
In my hyperfine benchmarks it was about 10-25% faster than the `master`.

Old: `4.557 s ±  0.404 s`
New: `3.832 s ±  0.131 s`

The performance fluctuates a little less as well.

I also added a few more element/attribute pairs which contain links according to the HTML spec. These occur very rarely, but it's good to parse them for completeness' sake.

Furthermore tried to clean up a lot of papercuts around our types. We now differentiate between a `RawUri` (stringy-types) and a `Uri`, which is a properly parsed URI type. 
The extractor now only deals with extracting `RawUri`s while the collector creates the request objects.

If someone wants to look into the code, I'd be happy for a review.
(I'll squash the commits before I merge of course. ^^)


mre:
The [flamegraph](https://gist.github.com/mre/7e3dc9426485f2c6bc4f6e1b4f520141) shows a much more flat call stack. (It's interactive, but you have to download it and open it in a browser.)

Issue #389, opened 2021-11-11
Lychee claims my email address is unreachable


hannobraun:
I checked my website using Lychee:
```
$ lychee https://hanno.braun-odw.eu 
📝 Summary
---------------------
🔍 Total...........28
✅ Successful......26
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
⚠ Unknown..........0
🚫 Errors...........2

Errors in https://hanno.braun-odw.eu/
✗ Hanno%20Braun%20%3Channo@braun-odw.eu%3E (Unreachable mail address: Hanno%20Braun%20%3Channo@braun-odw.eu%3E)
✗ hanno@braun-odw.eu (Unreachable mail address: hanno@braun-odw.eu)
```

It complains that my email address is unreachable, which is not true (I've just confirmed).

mre:
I think it parses the mail address incorrectly. There is a `>` at the end.  Can you post the relevant HTML here as well for troubleshooting please?
The second one looks fine. We use the check-if-email-exists, so it might be a bug over there.

hannobraun:
Here's the HTML:
``` html
<a href="mailto:Hanno%20Braun%20%3Channo@braun-odw.eu%3E">
    hanno@braun-odw.eu
</a>
```

I think I'm doing the right thing here. That's just the URL encoding of `Hanno Braun <hanno@braun-odw.eu>`. At least Thunderbird handles it correctly.

You can test yourself by going to any page on [https://hanno.braun-odw.eu/](https://hanno.braun-odw.eu/) and clicking on the email address near the upper-right corner.

hannobraun:
> We use the check-if-email-exists, so it might be a bug over there.

I think I figured out what the problem is.

[`check-if-email-exists`](https://crates.io/crates/check-if-email-exists) has a [live demo](https://reacher.email/). If I enter my email here, I get this:
```
{
  "input": "hanno@braun-odw.eu",
  "is_reachable": "unknown",
  "misc": {
    "is_disposable": false,
    "is_role_account": false
  },
  "mx": {
    "accepts_mail": true,
    "records": [
      "mxext2.mailbox.org.",
      "mxext1.mailbox.org.",
      "mxext3.mailbox.org."
    ]
  },
  "smtp": {
    "error": {
      "type": "SmtpError",
[...]

PR #463, opened 2022-01-14
Add more information about mail errors

mre:
See #389

### DIFF
--- a/lychee-lib/src/types/mail.rs
+++ b/lychee-lib/src/types/mail.rs
@@ -9,12 +9,11 @@ pub(crate) fn error_from_output(o: &CheckEmailOutput) -> String {
         return format!("{:?}", e);
     } else if let Err(e) = &o.mx {
         return format!("{:?}", e);
-    } else {
-        return match &o.is_reachable {
+    }
+    match &o.is_reachable {
             Reachable::Safe => "Safe: The email is safe to send",
             Reachable::Risky => "Risky: The email address appears to exist, but has quality issues that may result in low engagement or a bounce",
             Reachable::Invalid => "Invalid: Email doesn't exist or is syntactically incorrect",
             Reachable::Unknown => "Unknown: We're unable to get a valid response from the recipient's email server."
-        }.to_string();
-    }
+        }.to_string()
 }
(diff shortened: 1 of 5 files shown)


lychee-lib/src/types/mail.rs
```diff
@@ -0,0 +1,19 @@
...
+        return format!("{:?}", e);
+    } else if let Err(e) = &o.mx {
+        return format!("{:?}", e);
+    }
+    match &o.is_reachable {
+            Reachable::Safe => "Safe: The email is safe to send",
+            Reachable::Risky => "Risky: The email address appears to exist, but has quality issues that may result in low engagement or a bounce",
+            Reachable::Invalid => "Invalid: Email doesn't exist or is syntactically incorrect",
+            Reachable::Unknown => "Unknown: We're unable to get a valid response from the recipient's email server."
+        }.to_string()
```
lebensterben, line 18: doesn't seem like indented correctly

mre: that's what cargo fmt gave me 🤷‍♀️

mre: It's weird, though: when I change the indentation of the match blog, it doesn't change it. :/

lebensterben: you might accidentally find a bug...

mre: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=242e40ce131f604a1b1334b86cbec615

lebensterben: I know rustfmt gives up on really long lines. But I don't know it gives up so easily...

mre: Here's a minimal code example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=80660b19853fe958557723f5e00682dd

```
fn main() {
    match () {
    _ => "foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo",
};
}
```

Could be a rustfmt bug.

mre: Manually changed the formatting to the way it _should_ look like IMHO.

PR #464, opened 2022-01-16
Fix bugs in extractor; reduce allocs

mre:
When URLs couldn't be extracted from a tag,
we ran a plaintext search, but never added the
newly found urls to the vec of extracted urls.

Also tried to make the code a little more idiomatic

### DIFF
--- a/lychee-lib/src/extract/html.rs
+++ b/lychee-lib/src/extract/html.rs
@@ -18,7 +18,7 @@ impl TokenSink for LinkExtractor {
     #[allow(clippy::match_same_arms)]
     fn process_token(&mut self, token: Token, _line_number: u64) -> TokenSinkResult<()> {
         match token {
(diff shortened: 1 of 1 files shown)


lychee-lib/src/extract/html.rs
```diff
@@ -28,25 +28,24 @@ impl TokenSink for LinkExtractor {
...
-                                })
-                                .collect::<Vec<_>>(),
-                        );
-                    }
+                    let new_urls = match urls {
+                        None => extract_plaintext(&attr.value),
+                        Some(urls) => urls
+                            .into_iter()
+                            .map(|url| RawUri {
+                                text: url.to_string(),
```
lebensterben, line 42: to_string() is available for types implementing Display.

There's no convention about whether Display gives same output across versions.

So it's always better to use String::from() when available.

PR #489, opened 2022-02-04
Recursion V3: Using semaphores

mre:
The never ending recursion story part 3...

This time using semaphores as suggested in the Tokio Discord.
It's still locking. 😕


mre:
I'm closing this PR because it will not be merged in its current form. We will try again to add recursion support at a later point. In any case, thanks for the design feedback.

Issue #490, opened 2022-02-05
Mandatory max_cache_age when using configuration file


armandgrillet:
Using `lycheeverse/lychee:latest`:

```
$ lychee -v -c .lychee-markdown-links.toml README.md
[lychee markdown link check:1] Error: missing field `max_cache_age` at line 1 column 1
```

Using `lycheeverse/lychee:20211116124015893dff` with the same `.lychee-markdown-links.toml`:

```
$ lychee -v -c .lychee-markdown-links.toml README.md
... works
```

Using `lycheeverse/lychee:latest` with no config (`$ lychee -v README.md`) works, which makes sense as `--max-cache-age` defaults to 1 day according to https://github.com/lycheeverse/lychee/commit/4630216c3019095bcb5864c19a8a067b6042eac4. What is strange is the need to set `max_cache_age` as soon as a user uses a config now, which I think is a bug as the flag has a default value.

I wouldn't mind if setting `max_cache_age` in the config file was easy but it's not 😄. I tried a few values but I always get this kind of errors: `[lychee markdown link check:1] Error: invalid type: string "1d", expected struct Duration for key `max_cache_age``.

mre:
Yup, that's a bug. Thanks for reporting.

mre:
Should be fixed with #491. If you want to test it, feel free to build lychee from this PR. Otherwise it should land in `master` soon.

PR #491, opened 2022-02-05
Fix default config

mre:
The default configuration was broken since the
introduction of caching and specifically `max_cache_age`.
This fixes deserialization and config merging for
the case where this key is missing from the config.

See #490


armandgrillet (approved):
The fix and tests look good to me, the added .toml files don't have a newline at their end but that's a minor issue.

PR #495, opened 2022-02-07
lychee-bin: replace lazy_static by const_format

lebensterben:
This commit replaced the use of `lazy_static` by
`const_format` in `lychee-bin`.

Currently `lazy_static` is used to generate static
String at runtime. With `const_format` we can instead
make constant String at compile time.


mre:
Nice one, thanks!

PR #496, opened 2022-02-08
update rust-toolchain to .toml format

lebensterben:
rust-toolchain is phasing out in favour of rust-toolchain.toml

See: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file


lebensterben:
I've no idea why the test failed...

mre:
Thanks. The failure is unrelated. Seems to be an issue with one of the links in `TEST_MAIL.md`. I'm assuming a link to example.org. We should probably change the test to not make too many calls to other pages or, even better, switch to the mock server.

PR #497, opened 2022-02-08
Update license files

lebensterben:
- The date in MIT license files have been updated to 2022
- Each of the benchmark and example crates are theoretically
  a separate package in Cargo's sense. So license files are
  added for them as well.


mre:
Thanks!

PR #500, opened 2022-02-09
lychee-lib::client reworked

lebensterben:
This commit mainly added or improved documentation for `lychee-lib::client`
module.

But it also contains a few API changes:

- `ClientBuilder::client()` now consumes itself instead of taking a reference.
  This helps to avoid a few unnecessary clones.
- `ClientBuilder::build_filter()` was a private function and is inlined to avoid
  unnecessary clones.
- Added a new crate-socped function `Uri::set_scheme()`.

### DIFF
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -96,8 +96,15 @@ pub struct ClientBuilder {
     ///
     /// This property is defined in [IETF RFC 4193].
     ///
+    /// ## Note
+    ///
+    /// Unicast site-local network was defined in [IETF RFC 4291], but was fully deprecated in
+    /// [IETF RFC 3879]. So it is **NOT** considered as private on this purpose.
+    ///
     /// [IETF RFC 1918]: https://tools.ietf.org/html/rfc1918
     /// [IETF RFC 4193]: https://tools.ietf.org/html/rfc4193
+    /// [IETF RFC 4291]: https://tools.ietf.org/html/rfc4291
+    /// [IETF RFC 3879]: https://tools.ietf.org/html/rfc3879
     exclude_private_ips: bool,
     /// When it's `true`, exclude link-local IPs.
     ///
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -200,7 +200,7 @@ impl Default for ClientBuilder {
 }
 
 impl ClientBuilder {
-    /// Instantiates n [`Client`].
+    /// Instantiates a [`Client`].
     ///
     /// # Errors
     ///
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -1,3 +1,12 @@
+//! Handler of link checking operations.
+//!
+//! This modules defines two structs, [`Client`] and [`ClientBuilder`].
+//! `Client` handles incoming requests and returns responses.
+//! `ClientBuilder` exposes a finer level of granularity for building
+//! a `Client`.
+//!
+//! For convenience, a free function [`check`] is provided for ad-hoc
+//! link checks.
 #![allow(
     clippy::module_name_repetitions,
     clippy::struct_excessive_bools,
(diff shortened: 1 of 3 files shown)


lebensterben:
For reviewing the documentations, I recommend using
```sh
cargo doc --document-private-items --open
```

lebensterben:
hold on, I'll need to include the deprecated unicast_site_local as another case for private Ipv6 address.

lychee-lib/src/client.rs
```diff
@@ -1,3 +1,12 @@
+//! Handler of link checking operations.
+//!
+//! This modules defines two structs, [`Client`] and [`ClientBuilder`].
```
mre, line 3: modules -> module

lychee-lib/src/client.rs
```diff
@@ -24,132 +33,194 @@ use crate::{
...
+")]
 pub struct ClientBuilder {
-    /// Set an optional Github token.
-    /// This allows for more requests before
-    /// getting rate-limited.
+    /// Optional Github token used for Github links.
+    ///
+    /// This allows much more request before getting rate-limited.
+    ///
+    /// *P.S.* As of Feb 2022, it's 60 per hour without Github token
```
mre, line 64: Nitpick: Can we have a sub-headline for this section? Something like

```
# Rate-limiting Defaults

As of Feb 2022, it's 60 per hour without Github token vs. 5000 per hour with token.
```

lychee-lib/src/client.rs
```diff
@@ -24,132 +33,194 @@ use crate::{
...
+    /// This has higher precedence over [`ClientBuilder::excludes`], **but**
+    /// has lower precedence over any other `exclude_` fields or
+    /// [`ClientBuilder::schemes`] below.
     includes: Option<RegexSet>,
-    /// Exclude links matching this set of regular expressions
+    /// Links matching this set of regular expressions are ignored, **except**
+    /// when a link also matches against [`ClientBuilder::includes`].
     excludes: Option<RegexSet>,
-    /// Exclude all private network addresses
+    /// When it's `true`, exclude all private network addresses.
```
mre, line 76: I'd say "When true" instead of "When it's true", but it's more of a stylistic preference.

lychee-lib/src/client.rs
```diff
@@ -24,132 +33,194 @@ use crate::{
...
-            // including private IPs, link-local IPs and loopback IPs
-            exclude_private_ips: self.exclude_all_private || self.exclude_private_ips,
-            exclude_link_local_ips: self.exclude_all_private || self.exclude_link_local_ips,
-            exclude_loopback_ips: self.exclude_all_private || self.exclude_loopback_ips,
-            exclude_mail: self.exclude_mail,
-        }
-    }
-
-    /// The build method instantiates the client.
+    /// Instantiates an [`Client`].
```
mre, line 201: an -> a

lychee-lib/src/client.rs
```diff
@@ -240,13 +343,15 @@ impl Client {
         Ok(Response::new(uri, status, source))
     }
 
-    /// Check if the given URI is filtered by the client
+    /// Returns whehter the given `uri` should be ignored from checking.
```
mre, line 346: whehter -> whether

mre:
lgtm. Some minor comments.

mre:
The tests are flaky. I don't like that.

```
failures:

---- cli::test_lycheeignore_file stdout ----
thread 'cli::test_lycheeignore_file' panicked at 'Unexpected failure.
code-2
stderr=```""```
command=`"/home/runner/work/lychee/lychee/target/debug/lychee" "TEST.md"`
code=2
stdout="Issues found in 1 input(s). Find details below.\n\n[TEST.md]:\n✗ https://example.org/: Network error (reqwest): HTTP status client error (404 Not Found) for url ([https://example.org/)\n\n🔍](https://example.org/)/n/n%F0%9F%94%8D) 9 Total ✅ 1 OK 🚫 1 Error (HTTP:1) 💤 7 Excluded\n"
stderr=""
', /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
```

We should probably move to the `MockServer` for this.
Alternatively we add a `--dump-inputs` command, which would be nice to have anyway (maybe?). This way we won't call any network endpoint for the test. On the other hand other tests could fail as well for the same reason.

lychee-lib/src/client.rs
lebensterben: Okay I will change that.

lebensterben:
> We should probably move to the MockServer for this.
Alternatively we add a --dump-inputs command, which would be nice to have anyway (maybe?). This way we won't call any network endpoint for the test. On the other hand other tests could fail as well for the same reason.

That's right. I've observed similar false failures when testing locally.

But let's do that in another PR.

mre:
Just fixed one final typo, but the rest looked great. Merged.

PR #501, opened 2022-02-09
Improved IPv6 filtering support

### LINKED ISSUE #498: tracking issue: support local, link-local for ipv6
`lychee` currently supports excluding private IP or link-local IP for IPv4.

They're not supported for Ipv6 addresses because the current implementation relies on `std::net` which doesn't have them on stable channel. See https://docs.rs/ip_network/latest/ip_network/struct.Ipv6Network.html#method.is_unicast_global and https://doc.rust-lang.org/stable/std/net/struct.Ipv6Addr.html#method.is_unicast_link_local

Fortunately, we can temporarily use https://docs.rs/ip_network/latest/ip_network/struct.Ipv6Network.html until https://github.com/rust-lang/rust/issues/27709 is stablized.

- [x] Implementation. (#501).
- [x] Adding tests.

lebensterben:
This commit uses crate `ip_network` to determine whether an IPv6 address is
link-local or unique local.

Note that this extra dependencies can be removed once rust-lang/rust#27709 is
stablized.

Also note that this commit doesn't added any test yet.

closes #498

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1609,6 +1609,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "ip_network"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aa2f047c0a98b2f299aa5d6d7088443570faae494e9ae1305e48be000c9e0eb1"
+
+[[package]]
 name = "ipconfig"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1830,6 +1836,7 @@ dependencies = [
  "html5gum",
  "http",
  "hubcaps",
+ "ip_network",
  "jwalk",
  "lazy_static",
  "linkify",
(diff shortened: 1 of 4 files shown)


lebensterben:
I'm really not good at writing test. They could be included in another PR.

mre:
This PR is incredibly well done. I love that you added such detailed comments and refactored the code like this to make the intent clearer. It's probably pretty easy to write tests now. 😄 

```rust
#[test]
fn test_uri_is_loopback() {
  let uri = Uri::try_from("::1").unwrap();
  assert!(uri.is_loopback());
}
```

Adding a few of these would already be a great start.

Testing `Filter::is_ip_excluded` is a little more tricky.
We could add a `Default` impl and a couple of custom filters, which are only activated for tests:

```rust
#[cfg(test)]
impl Filter {
    fn exclude_loopback_ips() -> Self {
        let filter = Filter::default();
        filter.exclude_loopback_ips = true;
        filter
    }
}

// then in the tests...
#[test]
fn test_uri_is_loopback() {
    let uri = Uri::try_from("::1").unwrap();
    let filter = Filter::exclude_loopback_ips();
    assert!(filter.is_excluded(uri));
}
``` 

(this won't compile, but you get the idea. 😉)

Let me know if you want to add the tests in this PR. I can also help out if you like.

lebensterben:
This should be good to merge now.

PR #502, opened 2022-02-09
replace hubcaps by octocrab

lebensterben:
This commit replaced `hubcaps` by `octocrab`, which has more downloads per month
and receives more frequent release updates.

The caveats are:

1. When instantiating the API client, `octocrab` doesn't offer you a way to
specify custom user-agent. But I would argue that, at least presently, this
doesn't seem to cause issues.
2. `octocrab` doesn't export as much details of its error types as `hubcaps`
does. So we will have fewer control on the display of the error message. But I
would also argue that this is not really important. Though we should do more
tests to make sure the error looks good enough.


lebensterben:
Note: this should be merged after my other commits. And I will rebase it.

Cargo.toml
```diff
@@ -7,9 +7,5 @@ members = [
 ]
 resolver = "2"
 
-[patch.crates-io]
-# Switch back to version on crates.io after 0.6.3+ is released
-hubcaps = { git="https://github.com/softprops/hubcaps.git" }
```
mre, line 12: Nice! We can finally get rid of the patch. 😃

lychee-lib/src/types/error.rs
```diff
@@ -27,19 +27,8 @@ pub enum ErrorKind {
...
-            hubcaps::Error::Codec(e) => e.to_string(),
-            hubcaps::Error::Reqwest(e) => e.to_string(),
-            hubcaps::Error::Url(e) => e.to_string(),
-            hubcaps::Error::IO(e) => e.to_string(),
-            hubcaps::Error::JWT(e) => e.to_string(),
-        }
-        None => "".to_string(),
-    })]
-    GithubError(#[from] Option<hubcaps::Error>),
+    #[error("Network error (hubcaps) {}", .0.as_ref().map_or(String::new(), std::string::ToString::to_string))]
```
mre, line 30: this line should probably say `octocrab` or perhaps not even mention the client at all.

mre:
Just one minor comment. Apart from that it's good to go.
I'd argue not being able to configure the user agent isn't a big issue; Github will discern the users based on the token anyway (I'm guessing).

lychee-lib/src/types/error.rs
lebensterben: Right. Let's just say it's a "GitHub Client" error, hiding the implementation details.

lebensterben:
It should be good to go.

mre:
Tested and works great. Thanks!

PR #503, opened 2022-02-10
Add Docker semver tagging for new multiarch image

mre:
Based on https://github.com/docker/build-push-action/blob/master/docs/advanced/tags-labels.md
Fixes #326


mre:
Tested `lycheeverse/lychee-dev:latest` locally on both Intel and arm64 and they work as expected. The size is also quite reasonable (~35MB).

fauust (commented):
This is a lot more simple, so that's maybe good.
The main problem I see:
- what happen on forks (https://github.com/fauust/lychee/runs/5143705225?check_suite_focus=true). I am not sure this is wanted?
- you are skipping a step that would verify that the container works locally.

Regarding the later, building and pushing locally before pushing on docker is maybe overkill but I thing it would still be good to test the image that you are pushing before making it available to the world.

mre:
Good points. I'll work on them.

* [ ] Skip build for forks
* [ ] Add local testing before push

mre:
I tried testing the image with

```yaml
      - name: Build Image
        uses: docker/build-push-action@v2
        with:
          context: .
          file: ${{ env.DOCKERFILE }}
          platforms: ${{ env.DOCKER_PLATFORMS }}
          load: true
          tags: ${{ env.IMAGE_NAME }}:test

      - name: Test
        run: |
          for platform in ${DOCKER_PLATFORMS/,/ }; do
            docker run --rm ${{ env.IMAGE_NAME }}:test --version
          done
```

but I hit https://github.com/docker/buildx/issues/59. 
Given that this is a blocker for pushing multiarch images to Dockerhub at the moment, I'll skip the tests for now. I've tested the images locally and they work fine and we'll find out very quickly if we break something thanks to lychee-action.
It's not ideal, but it's a tradeoff I'm willing to make to move forward.

mre:
Yeah, I tried to add back the tests, but it seems like there are too many blockers at the moment to reach all goals. My main goal was to unblock the pipeline as we're planning to release 0.9.0, so I wanted to get the pipeline ready for that.
I hope we can revisit this topic again at some point when the open issues are resolved.

PR #506, opened 2022-02-11
Multiarch for main Docker workflow

mre:


mre:
Tested it on an intel mac and an m1 and it looks good. It uses the correct architecture, so I'll go ahead and merge this right away.

PR #509, opened 2022-02-13
Replace plain String with SecretString for GitHub token

### LINKED ISSUE #508: tracking issue: fortify lychee-bin and lychee-lib with `secrecy`
We can use [secrecy](https://lib.rs/crates/secrecy) to make sure GitHub token and anything alike would never accidentally be collected in log or printed.

lebensterben:
This commit changed the type of `lychee-lib::ClientBuilder::github_token` from
`String` to `secrecy::SecretString` to fortify the secret management within our
program.

Note that this won't affect TOML configuration of `lychee-bin` because
`serde::Deserialize` is still implemented for `SecretString`.

closes #508

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1818,6 +1818,7 @@ dependencies = [
  "regex",
  "reqwest",
  "ring",
+ "secrecy",
  "serde",
  "serde_json",
  "structopt",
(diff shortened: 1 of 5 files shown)


mre:
Thanks!

PR #514, opened 2022-02-18
Try to prevent flaky tests

mre:
Moves from example.org to example.com
to see if that makes testing results more consistent


mre:
Looks like example.com is slightly more permissive than example.org or at least the two domains are configured differently at least. I'm okay with merging this.

mre:
In the long run we should still use a mock server, but this is good enough to unblock some PRs.

lebensterben:
You may also be interested in https://github.com/nextest-rs/nextest to figure out flaky tests.

mre:
@lebensterben feel free to rebase your pull requests on `master` to see if this resolves our issues. 😃

PR #537, opened 2022-03-02
Print Github token error once at the end

mre:
Print original reqwest error for every Github link.
It contains more information about the underlying error.

Only print a message about the Github token at the
end if it's not set and there were Github errors.

### DIFF
--- a/lychee-lib/src/types/error.rs
+++ b/lychee-lib/src/types/error.rs
@@ -34,9 +34,12 @@ pub enum ErrorKind {
     /// The network client required for making requests cannot be created
     #[error("Error creating request client")]
     BuildRequestClient(#[source] reqwest::Error),
+    /// The Github client required for making requests cannot be created
+    #[error("Error creating Github client")]
+    BuildGithubClient(#[source] octocrab::Error),
     /// Network error while using Github API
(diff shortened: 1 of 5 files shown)


lychee-lib/src/types/error.rs
```diff
@@ -34,9 +34,12 @@ pub enum ErrorKind {
     /// The network client required for making requests cannot be created
     #[error("Error creating request client")]
     BuildRequestClient(#[source] reqwest::Error),
+    /// The Github client required for making requests cannot be created
+    #[error("Error creating Github client")]
+    BuildGithubClient(#[source] octocrab::Error),
     /// Network error while using Github API
     #[error("Network error (GitHub client)")]
-    GithubRequest(#[from] octocrab::Error),
+    GithubRequest(#[source] octocrab::Error),
```
lebensterben, line 42: why is `from` replaced by `source`?

mre: That was a mistake. Nice catch.
During refactoring I had a case where I copied this error and converted it to `BuildGithubClient`. Then I had to `from` impls and I accidentially removed it from both. Rerverted now.

PR #540, opened 2022-03-03
Add support for collecting links without a scheme (fixes #408)

### LINKED ISSUE #408: Add support for URLs without scheme/protocol
linkify 0.8.0 was just released with an exciting new feature:

> New option url_must_have_scheme on LinkFinder that can be set to false to allow URLs without scheme/protocol such as example.com.

Changelog [here](https://github.com/robinst/linkify/blob/main/CHANGELOG.md#080---2021-11-26).

This would allow us to check even more links. 😃 
We should add support for that behind a feature flag:

* [x] Add flag `--no-scheme`, which is off by default.
* [x] If `--no-scheme` is enabled, create linkify with [the new feature](https://github.com/robinst/linkify/blob/bac4a6412a16fa44e4836a09f5dbdfba9a2da21a/demo/src/lib.rs#L5):
  ```rust
  let mut link_finder = LinkFinder::new();
  link_finder.url_must_have_scheme(!no_scheme);
  ```
* [x] Add tests

We could eventually even make this the default, although we have to do a lot of thorough testing before that.

Thanks to @robinst and @timvisee for working on this on the linkify side.

mre:
It works, but there are a lot of false positives.
We should consider (optionally?) filtering by known TLDs.
Something like https://github.com/pagescrape/tld.rs maybe, although the list there is quite long. I'd be fine with a list of the top 10 domains to begin with, maybe. Is there a crate we can use or shall we hardcode the list?

In the long run, we should consider creating a PR to linkify, because the broader Rust community might benefit from that allow-list.


lebensterben:
well this won't be neatly implemented given our current design of extractor.

basically, our extractor treats markdown, html, and text differently. This wasn't a big problem becauae
- In html and markdown, links are expected to appear in a handful of context.
- In text, we use linkify to aggressively matches possible URLs.

Now this becomes a problem. Consider a example similar to the one I saw in a test

[foo.com] lorem ipsum [bar](bar.com) lorem ipsum baz.com

We now should expect to have three links, i.e. foo.com bar.com and baz.com

But the test only expects to find bar.com

mre:
You are right. My current opinion is that URLs without a scheme are more trouble than they are worth. I tried to be as conservative as possible with the implementation, but there are too many constraints for making it useful. I'll close this PR and will probably open a new PR for passing a URL without a scheme as an *input*, so this will work:

```
lychee google.com
```

I also like this part and will probably make a separate PR for that: [lychee-lib/src/extract/plaintext.rs](https://github.com/lycheeverse/lychee/pull/540/files#diff-71d1b3c88bc139a694d64b36c2f5d5792ef6554fc64181971724da3c92101e8cL2-L3)

Issue #539, opened 2022-03-03
Lychee don't panic, invalid URL's are fine.


vipulgupta2048:
From the pleasant discussion with @mre here https://github.com/lycheeverse/lychee-action/issues/82, I intend to work on trying to parse the input URL before passing it to `reqwest` in order to prevent lychee from panicking, skip the invalid URL and move on to the next one.  

Bonus: 
- If we can track these invalidated URL's and show them to users at the end so that they are able to fix or ignore them for future runs

Looking for feedback and @vipulgupta2048 will work on this soon (The email will remind me to work on this)

mre:
Go for it! Just post questions here and we'll try to help. I like the bonus task, but don't stress it too much if it takes longer than expected. We can always to a follow-up PR. 😉

lebensterben:
If the cause of panic is known (reqwest), we can use catch-unwind somewhere in the code.

https://doc.rust-lang.org/stable/std/panic/fn.catch_unwind.html

mre:
Oh that's a really good idea! There's probably no performance overhead except in case of panic, which is perfect for our use-case. @vipulgupta2048 I don't know how far you got already, but could you try wrapping the client call in a catch_unwind as @lebensterben mentioned? 🥺

vipulgupta2048:
I don't mean to hold this up, I was expecting to get some time to write a patch for this but haven't been lucky enough. I will unassign this from me in order to unblock it for others to try out till then.

mre:
@vipulgupta2048, no worries and thanks for looking into it.

Originally I thought this was possible after the discussion with @lebensterben above:

```rust
 let result = panic::catch_unwind(|| self.reqwest_client.execute(request));

let fut = match result {
    Ok(s) => s,
    Err(_) => return Status::Error(ErrorKind::Panic),
};

match fut.await {
    Ok(ref response) => Status::new(response, self.accepted.clone()),
    Err(e) => e.into(),
}
```

but I'm getting an error:

```
   --> lychee-lib/src/client.rs:482:22
    |
482 |         let result = panic::catch_unwind(|| self.reqwest_client.execute(request));
    |                      ^^^^^^^^^^^^^^^^^^^ `UnsafeCell<tokio::util::linked_list::Pointers<tokio::time::driver::entry::TimerShared>>` may contain interior mutability and a reference may not be safely transferrable across a catch_unwind boundary
    |
```

[...]

PR #557, opened 2022-03-18
Avoid reqwest panic on invalid URIs

mre:
Temporary workaround for https://github.com/lycheeverse/lychee/issues/539 until the issue gets fixed upstream.


lebensterben:
I still wonder why reqwest cannot accept URLs like https://web.archive.org/web/20131015212820/http://www.deviantart.com/

This can be opened in browser.

PR #565, opened 2022-03-23
Optimize plaintext extractor for small strings

mre:
Immediately return for very small strings which cannot be valid URIs.

The shortest valid URI without a scheme might be g.cn (Google China)
At least I am not aware of a shorter one. We set this as a lower threshold
for parsing URIs from plaintext to avoid false-positives and as a slight
performance optimization, which could add up for big files.
This threshold might be adjusted in the future.

### DIFF
--- a/lychee-lib/src/extract/plaintext.rs
+++ b/lychee-lib/src/extract/plaintext.rs
@@ -1,7 +1,21 @@
 use crate::{helpers::url, types::raw_uri::RawUri};
 
+/// Shortest valid URI that lychee extracts from plaintext.
+///
+/// The shortest valid URI without a scheme might be g.cn (Google China)
+/// At least I am not aware of a shorter one. We set this as a lower threshold
+/// for parsing URIs from plaintext to avoid false-positives and as a slight
+/// performance optimization, which could add up for big files.
+/// This threshold might be adjusted in the future.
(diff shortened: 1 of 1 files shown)


lychee-lib/src/extract/plaintext.rs
```diff
@@ -1,7 +1,21 @@
...
+///
+/// The shortest valid URI without a scheme might be g.cn (Google China)
+/// At least I am not aware of a shorter one. We set this as a lower threshold
+/// for parsing URIs from plaintext to avoid false-positives and as a slight
+/// performance optimization, which could add up for big files.
+/// This threshold might be adjusted in the future.
+const MIN_URI_LENGTH: usize = 4;
+
 /// Extract unparsed URL strings from plaintext
 pub(crate) fn extract_plaintext(input: &str) -> Vec<RawUri> {
```
lebensterben, line 4: Should the return type be `Option<Vec<RawUri>>`

mre: In general I agree with you. We could save some allocations, but it would require that all extractors return `Option<Vec<RawUri>>` and we'd have to add quite a few `if let Some(urls) = extract_plaintext(input)` statements. We can do it in a separate PR if we like.

PR #568, opened 2022-03-24
Allow input URLs without scheme (fixes #567)

### LINKED ISSUE #567: Support input URLs without a scheme
This should "just work":

```
lychee google.com
```

Question is if the default scheme should be `http` or `https`.
[curl still uses http](https://stackoverflow.com/questions/31479263/can-curl-default-to-using-https).

mre:
This requires `Input::new` to return a `Result`, because the URL
parsing could fail when prepending `http://`.

We use http instead of https, because curl does as well:
https://github.com/curl/curl/blob/70ac27604a2abfa809a7b2736506af0da8c3c8a9/lib/urlapi.c#L1104-L1124

### DIFF
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -114,11 +114,19 @@ pub struct Input {
 }
 
 impl Input {
-    #[must_use]
     /// Construct a new `Input` source. In case the input is a `glob` pattern,
     /// `glob_ignore_case` decides whether matching files against the `glob` is
     /// case-insensitive or not
-    pub fn new(value: &str, file_type_hint: Option<FileType>, glob_ignore_case: bool) -> Self {
+    ///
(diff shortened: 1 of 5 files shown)


mre:
Hum... why didn't the unit tests run? 💭

lychee-lib/src/types/input.rs
```diff
@@ -133,13 +141,25 @@ impl Input {
                     ignore_case: glob_ignore_case,
                 }
             } else {
-                InputSource::FsPath(value.into())
+                let path = PathBuf::from(value);
+                if path.exists() {
+                    InputSource::FsPath(value.into())
```
lebensterben, line 146: instead of value.into(), can you use path directly?

mre: Thanks, changed.

PR #572, opened 2022-03-27
Add support for nofollow

### LINKED ISSUE #548: Add support for skipping `nofollow` links
Especially in relation to #78 (but this should apply independently of recursion), it would be nice to have a flag that would instruct `lychee` to just skip any `<a>` links that are marked with `rel="nofollow"`, `rel="...,nofollow,..."`, etc., as specified in the HTML standard (<https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel>).

The reasoning behind this, is say one has a site with lots of user-generated-content (wikis, comments, blogs, etc.), and the code already marks these "untrusted" links with `nofollow`.  At the moment if one runs `lychee` on such a site, at best it will get loads of errors, and at worst it could actually be used to DOS the targeted sites.  (Granted one could use exclusion, but that list would be so long and so often changing, that perhaps an `--include-only` option would be better.)

However with `nofollow`, the site administrator can still use `lychee` to check that important (and trusted) resources and links are still functional, while skipping those untrusted links in the user-generated-content.

mre:
Adds support for skipping HTML links in the following format:

```html
<a rel="nofollow" href="https://foo.com">do not follow me</a> 
```

Fixes #548.

@untitaker, does my implementation for html5gum make sense or is there a more elegant way to do it?

### DIFF
--- a/lychee-lib/src/extract/html5gum.rs
+++ b/lychee-lib/src/extract/html5gum.rs
@@ -11,6 +11,7 @@ struct LinkExtractor {
     current_string: Vec<u8>,
     current_element_name: Vec<u8>,
     current_element_is_closing: bool,
+    current_element_nofollow: bool,
     current_attribute_name: Vec<u8>,
     current_attribute_value: Vec<u8>,
     last_start_element: Vec<u8>,
@@ -31,6 +32,7 @@ impl LinkExtractor {
             current_string: Vec::new(),
             current_element_name: Vec::new(),
             current_element_is_closing: false,
+            current_element_nofollow: false,
             current_attribute_name: Vec::new(),
             current_attribute_value: Vec::new(),
             last_start_element: Vec::new(),
@@ -112,6 +114,18 @@ impl LinkExtractor {
             let attr = unsafe { from_utf8_unchecked(&self.current_attribute_name) };
             let value = unsafe { from_utf8_unchecked(&self.current_attribute_value) };
 
+            // Ignore links with rel=nofollow
+            // This may be set on a different iteration on the same element/tag before,
+            // so we check the boolean separately right after
+            if attr == "rel" && value.contains("nofollow") {
+                self.current_element_nofollow = true;
+            }
+            if self.current_element_nofollow {
+                self.current_attribute_name.clear();
+                self.current_attribute_value.clear();
+                return;
+            }
+
             let urls = LinkExtractor::extract_urls_from_elem_attr(attr, name, value);
 
             let new_urls = match urls {
@@ -173,6 +187,7 @@ impl Emitter for &mut LinkExtractor {
 
     fn emit_current_tag(&mut self) {
         self.flush_old_attribute();
+        self.current_element_nofollow = false;
     }
 
     fn emit_current_doctype(&mut self) {}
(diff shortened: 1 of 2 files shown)


lychee-lib/src/extract/html5gum.rs
```diff
@@ -173,6 +187,7 @@ impl Emitter for &mut LinkExtractor {
 
     fn emit_current_tag(&mut self) {
         self.flush_old_attribute();
+        self.current_element_nofollow = false;
```
untitaker, line 190: I would reset this in `init_start_tag/init_end_tag`

untitaker (commented):
html5gum looks good to me

lychee-lib/src/extract/html5gum.rs
mre: I tried that initially, but it didn't work. It would not properly reset the state and instead found the links which should be excluded. If you have the time you can test yourself. Could be that the error is on my end.

untitaker: will try, is there a testcase for this?

mre: Yup, the one from this PR.

untitaker: hmm this patch works for me:

```diff
diff --git a/lychee-lib/src/extract/html5gum.rs b/lychee-lib/src/extract/html5gum.rs
index 55d2947..92e13ec 100644
--- a/lychee-lib/src/extract/html5gum.rs
+++ b/lychee-lib/src/extract/html5gum.rs
@@ -172,13 +172,12 @@ impl Emitter for &mut LinkExtractor {
     fn init_start_tag(&mut self) {
         self.flush_current_characters();
         self.current_element_name.clear();
+        self.current_element_nofollow = false;
         self.current_element_is_closing = false;
     }

     fn init_end_tag(&mut self) {
-        self.flush_current_characters();
-        self.current_element_name.clear();
-        self.current_element_is_closing = true;
+        self.init_start_tag();
     }

     fn init_comment(&mut self) {
@@ -187,7 +186,6 @@ impl Emitter for &mut LinkExtractor {

     fn emit_current_tag(&mut self) {
         self.flush_old_attribute();
-        self.current_element_nofollow = false;
     }

     fn emit_current_doctype(&mut self) {}
```

mre:
@untitaker I went ahead and merged this to add general nofollow support. Code improvements on that part are still welcome of course, so feel free to open another PR after testing `init_start_tag/init_end_tag`.

untitaker:
I forgot to submit the review! Here's what I wrote:

----

hmm this patch works for me: 

```diff
diff --git a/lychee-lib/src/extract/html5gum.rs b/lychee-lib/src/extract/html5gum.rs
index 55d2947..92e13ec 100644
--- a/lychee-lib/src/extract/html5gum.rs
+++ b/lychee-lib/src/extract/html5gum.rs
@@ -172,13 +172,12 @@ impl Emitter for &mut LinkExtractor {
     fn init_start_tag(&mut self) {
         self.flush_current_characters();
         self.current_element_name.clear();
+        self.current_element_nofollow = false;
         self.current_element_is_closing = false;
     }

     fn init_end_tag(&mut self) {
-        self.flush_current_characters();
-        self.current_element_name.clear();
-        self.current_element_is_closing = true;
+        self.init_start_tag();
     }

     fn init_comment(&mut self) {
@@ -187,7 +186,6 @@ impl Emitter for &mut LinkExtractor {

     fn emit_current_tag(&mut self) {
         self.flush_old_attribute();
-        self.current_element_nofollow = false;
     }

     fn emit_current_doctype(&mut self) {}
```

to be clear the only thing i did for verification is to run `cargo test --all-features`

mre:
Thanks, I opened https://github.com/lycheeverse/lychee/pull/584 for that.

PR #576, opened 2022-03-30
Update  scheme to sequence in example conf

itamarhaber:


mre:
Thanks! You're right, that was an oversight from extending the `schemes` feature. The PR checks fail on the Docker build now, but this is a consequence of a recent refactoring of the Docker build pipeline that we can look into in isolation. Merging. ✌️

PR #577, opened 2022-03-30
Extract more status information from reqwest

mre:
Recently we cleaned up the commandline output to trim away redundant
information like the URL, which occured twice.
Unfortunately we also removed helpful information from reqwest, which
could support the user in troubleshooting unexpected errors.

This commit reverts that.
We now extract the meaningful information from reqwest, without being
too verbose. For that we have to depend on the string output for the
reqwest error, but it's better than hiding that information from the user.
It is fragile as it depends on the reqwest internals, but in the worst case
we simply return the full error text in case our parsing won't work.

### DIFF
--- a/lychee-lib/src/helpers/reqwest.rs
+++ b/lychee-lib/src/helpers/reqwest.rs
@@ -1,0 +1,39 @@
+/// Extract the most relevant parts from a reqwest error
+///
+/// The reqwest `Error` fields aren't public as they are an implementation
+/// detail. Instead, a human-readable error message can be obtained. However,
+/// the error message is quite verbose and the information is redundant. For
+/// example it contains the `URL`, which is already part of our `ResponseBody`.
+/// Therefore we try to trim away the redundant parts so that the `ResponseBody`
+/// output is cleaner.
+pub(crate) fn trim_error_output(e: &reqwest::Error) -> String {
+    // Defer to separate function for easier testability.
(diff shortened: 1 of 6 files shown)


mre:
As of today we only get "no status code" for errors like the ones below.
This PR changes that.

```
✗ [ERR] http://via1buynow.com/ | Network error: error following redirect for url (http://www.ee-magazine.com/): too many redirects
✗ [ERR] http://superadspro.com/ | Network error: The certificate was not trusted.
✗ [ERR] http://78.al/ | Network error: error sending request for url (https://78.al/): connection closed before message completed
✗ [ERR] http://alledufree13.ml/ | Network error: dns error: no record found for name: alledufree13.ml.fritz.box. type: AAAA class: IN
```

lychee-lib/src/helpers/reqwest.rs
```diff
@@ -0,0 +1,46 @@
+use once_cell::sync::Lazy;
+use regex::Regex;
+
+pub(crate) static REQWEST_ERROR_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"error trying to connect:\s?(.*)").unwrap());
```
lebensterben, line 5: I don't think it's necessary to use regex pattern. This can be done by matching the fixed string.

mre: You're right. I'll change it when I get the chance.

PR #620, opened 2022-05-15
Add support for URI remapping

### LINKED ISSUE #547: Add support for domain remapping URLs to custom endpoints (schema-IP-port)
`lychee` would also be great to test web servers prior to production deployment.

Say one has a HTTP server that runs on `127.0.0.1:8080`, but when it will be put into production it would actually serve `https://example.com/`.  Also say that, due to developer laziness, the web server, even in development mode, has links that use the production domain (i.e. `example.com`).

Thus it would be useful to instruct `lychee` that whenever it sees `https://example.com` it should instead make a HTTP (without TLS) call to `127.0.0.1:8080`.

To generalize, there should be the possibility to remap any HTTP/HTTPS endpoint to another HTTP/HTTPS endpoint, like for example:

~~~~
--remap-url 'https://example.com http://127.0.0.1:8080'
--remap-url 'https://example.com https://staging.example.com'
~~~~

Perhaps the first part (i.e. `https://example.com`) could be a regex to support a more generic use-case.

Also, this feature could also be used to proxy certain requests through certain proxies.

mre:
Allow users to remap URIs. Regular expressions are supported. Remappings get evaluated in order of definition. Multiple remappings are allowed.  
The format is similar to that of `markdown-link-check`.

## How it works ⚙️

Consider the following input:

```markdown
file://../../issues
https://example.com
https://example.org
https://example.net
```

Let's remap these inputs using the `--remap` parameter:

```sh
lychee --remap 'https://example.com http://127.0.0.1:8080' \
       --remap 'https://example.org https://staging.example.com' \
       --remap '../../issues https://github.com/usnistgov/OSCAL/issues' \
       -- input.md
```

This would check the following URIs:

```
https://github.com/usnistgov/OSCAL/issues
http://127.0.0.1:8080/
https://staging.example.com/
https://example.net/
```

## Caveats 🚩

* Be careful when using this feature because checking every link against a large set of regular expressions has a performance impact even though the Rust regex engine is very fast.
* There are no constraints on the URI mapping, so the rules might contradict each other. It's the user's responsibility to make sure this does not happen.
* File remappings might be a bit surprising. Files are handled as URIs internally and get prefixed with a `file://` scheme. As a consequence, using `^../../issues` as a remap pattern would never match. Instead `../../issues` (without the `^`) would work.

## Open Questions 🤔  (please provide feedback)

* Shall we first remap a URI and then check if it is excluded or vice versa? 🤔 

Fixes #547

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,7 @@ OPTIONS:
         --max-retries <max-retries>            Maximum number of retries per request [default: 3]
     -X, --method <method>                      Request method [default: get]
     -o, --output <output>                      Output file of status report
+        --remap <remap>...                     Remap URI matching pattern to different URI
     -r, --retry-wait-time <retry-wait-time>    Minimum wait time in seconds between retries of failed requests [default:
                                                1]
     -s, --scheme <scheme>...                   Only test links with the given schemes (e.g. http and https)
--- a/lychee-bin/src/client.rs
+++ b/lychee-bin/src/client.rs
@@ -1,5 +1,7 @@
 use crate::options::Config;
-use crate::parse::{parse_basic_auth, parse_duration_secs, parse_headers, parse_statuscodes};
+use crate::parse::{
+    parse_basic_auth, parse_duration_secs, parse_headers, parse_remaps, parse_statuscodes,
+};
 use anyhow::{Context, Result};
 use headers::HeaderMapExt;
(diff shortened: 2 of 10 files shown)


mre:
Any comments before this gets merged?

PR #631, opened 2022-05-31
Improve configuration example

### LINKED ISSUE #630: Outdated lychee.example.toml and hard to use
Hi and props on this great project :)

I'm trying to use a configuration file with lychee-action to improve the readability, as in, to not pollute the action with many arguments and I'm struggling to define a valid configuration file.

Having an updated lychee.example.toml would make it easier for everyone to use it.

Issues I found so far:
* Many command line parameters are missing
   * It seems to me (not a rust developer) that all command line parameters can be defined through the configuration file. Still it's not obvious what the variable names should be.
* The commented example `#accept = "text/html"` does not match the description. The example should be `#accept = "503,403"`
* The `exclude_file = []` should have an example as it's not obvious (to me) that file paths must be quoted, as in, must be strings.

vpereira01:
Fixes #630 

Easy test `./lychee -c lychee.example.toml -- README.md`

### DIFF
--- a/lychee.example.toml
+++ b/lychee.example.toml
@@ -1,43 +1,61 @@
-###
-### Display
-###
+#############################  Display  #############################
+
 # Verbose program output
 verbose = false
 
-# Show progress
-progress = false
+# TODO
+no_progress = false
 
+# TODO
+output = "report.md"
 
-###
-### Runtime
-###
+#############################  Cache  ###############################
+
+# TODO
+cache = true
+
+# TODO
+# Error: invalid type: string "2d", expected struct Duration for key `max_cache_age`
+# max_cache_age = "2d"
+
+#############################  Runtime  #############################
+
 # Number of threads to utilize.
 # Defaults to number of cores available to the system if omitted.
-#threads = 2
+threads = 2
 
 # Maximum number of allowed redirects
 max_redirects = 10
 
+# TODO
+max_retries = 2
 
-###
-### Requests
-###
+# TODO
+max_concurrency = 14
+
+#############################  Requests  ############################
+
 # User agent to send with each request
 user_agent = "curl/7.71.1"
 
 # Website timeout from connect to response finished
 timeout = 20
 
-# Comma-separated list of accepted status codes for valid links.
-# Omit to accept all response types.
-#accept = "text/html"
+# TODO
+retry_wait_time = 2
+
+# TODO
+accept = "403, 403"
 
 # Proceed for server connections considered insecure (invalid TLS)
 insecure = false
 
 # Only test links with the given schemes (e.g. https)
 # Omit to check links with any scheme
-#scheme = [ "https" ]
+scheme = [ "https" ]
+
+# TODO
+require_https = false
 
 # Request method
 method = "get"
@@ -45,15 +63,35 @@ method = "get"
 # Custom request headers
 headers = []
 
+# TODO # Remap URI matching pattern to different URI
+remap = [ "https://example.com http://example.invalid" ]
 
-###
-### Exclusions
-###
+# TODO
+# Error: invalid type: unit variant, expected newtype variant for key `base`
+# base = "https://example.com"
+
+# TODO
+basic_auth = "user:pwd"
+
+#############################  Exclusions  ##########################
+
+# TODO
+skip_missing = false
+
+# TODO
+include_verbatim = false
+
+# TODO
+glob_ignore_case = false
+
 # Exclude URLs from checking (supports regex)
-exclude = []
+exclude = [ "\\.github\\.com.*" ]
 
-# Exclude URLs contained in a file from checking
-# If a file named `.lycheeignore` exists in the current working directory,
+# TODO
+exclude_path = ["file/path/to/Ignore", "./other/file/path/to/Ignore"]
+
+# Deprecated: Exclude URLs contained in a file from checking
+# If a file named `.lycheeignore` exists in the current working directory, 
 # its contents will be excluded as well.
 exclude_file = []
 
--- a/lychee.example.toml
+++ b/lychee.example.toml
@@ -3,18 +3,18 @@
 # Verbose program output
 verbose = false
 
-# TODO
+# Don't show interactive progress bar while checking links
 no_progress = false
 
-# TODO
+# Path to summary output file
 output = "report.md"
 
 #############################  Cache  ###############################
 
-# TODO
+# Enable link caching. This can be helpful to avoid checking the same links on multiple runs.
 cache = true
 
-# TODO
+# Discard all cached requests older than this duration.
 # Error: invalid type: string "2d", expected struct Duration for key `max_cache_age`
 # max_cache_age = "2d"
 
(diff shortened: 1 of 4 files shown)


lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
-### Display
-###
+#############################  Display  #############################
+
 # Verbose program output
 verbose = false
 
-# Show progress
-progress = false
+# TODO
```
mre, line 6: Comment suggestion:
```
Don't show interactive progress bar while checking links
```

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+no_progress = false
+
+# TODO
```
mre, line 9: Comment suggestion:
```
Path to summary output file
```

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+output = "report.md"
+
+#############################  Cache  ###############################
+
+# TODO
+cache = true
 
+# TODO
+# Error: invalid type: string "2d", expected struct Duration for key `max_cache_age`
```
mre, line 18: Ouch. Yeah that's something we should fix. I think `max_cache_age` should be `Into<Duration>` instead of `Duration`. And maybe we need an `impl From<String> for Duration` to handle that case or use`human_time`. I can look into this, but for now it's fine to keep it in there.

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
-###
 # Number of threads to utilize.
 # Defaults to number of cores available to the system if omitted.
-#threads = 2
+threads = 2
 
 # Maximum number of allowed redirects
 max_redirects = 10
 
+# TODO
```
mre, line 30: How about this?
```
# Maximum number of allowed retries before a link is declared dead
```

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+max_retries = 2
+
+# TODO
```
mre, line 33: Comment suggestion:
```
# Maximum number of concurrent link checks
```

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
 # Website timeout from connect to response finished
 timeout = 20
 
-# Comma-separated list of accepted status codes for valid links.
-# Omit to accept all response types.
-#accept = "text/html"
+# TODO
+retry_wait_time = 2
+
+# TODO
```
mre, line 47:
```
Comma-separated list of accepted status codes for valid links
```

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
 
 # Proceed for server connections considered insecure (invalid TLS)
 insecure = false
 
 # Only test links with the given schemes (e.g. https)
 # Omit to check links with any scheme
-#scheme = [ "https" ]
+scheme = [ "https" ]
+
+# TODO
```
mre, line 57:
```
When links are available using HTTPS, treat HTTP links as errors
```

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+require_https = false
 
 # Request method
 method = "get"
 
 # Custom request headers
 headers = []
 
+# TODO # Remap URI matching pattern to different URI
+remap = [ "https://example.com http://example.invalid" ]
```
mre, line 67: I think we can support a nicer syntax for remapping similar to https://github.com/lycheeverse/lychee/issues/547#issuecomment-1122690070:


```toml
[[remap]]
pattern = "^../../issues"
replacement = "https://github.com/usnistgov/OSCAL/issues"

[[remap]]
pattern = "https://example.com"
replacement = "https://example.net"
```

Let's tackle that once this PR got merged.

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+
+# TODO
+# Error: invalid type: unit variant, expected newtype variant for key `base`
+# base = "https://example.com"
```
mre, line 71: Heh, don't know what's going on but it has something to do with the variants here: https://github.com/lycheeverse/lychee/blob/c4e004bdf8ee19c4e5945628255851d5498dd400/lychee-lib/src/types/base.rs#L12-L17
There is a `TryFrom<&str>`, which I though would take care of it, but that's incorrect.
We can also tackle that later.

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+
+# TODO
```
mre, line 73:
```
HTTP basic auth support. This will be the username and password passed to the authorization HTTP header. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
```

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+basic_auth = "user:pwd"
+
+#############################  Exclusions  ##########################
+
+# TODO
+skip_missing = false
+
+# TODO
```
mre, line 81: Check links inside `<code>` and `<pre>` blocks as well as Markdown code blocks.

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+glob_ignore_case = false
 
-###
-### Exclusions
-###
 # Exclude URLs from checking (supports regex)
-exclude = []
+exclude = [ "\\.github\\.com.*" ]
```
mre, line 88: I assume the double-backslashes are necessary right now? If so, can you try a literal string (e.g. single quotes)? I think the escaping ain't needed there.
```toml
exclude = [ '\.github\.com.*' ]
```

mre:
Thanks for the draft. Added a few comments. 👍

lychee.example.toml
vpereira01: Sounds good.

lychee.example.toml
vpereira01: Yes, single quoted is much better, changed.

lychee.example.toml
```diff
@@ -1,59 +1,97 @@
...
+
+# TODO
+exclude_path = ["file/path/to/Ignore", "./other/file/path/to/Ignore"]
 
-# Exclude URLs contained in a file from checking
-# If a file named `.lycheeignore` exists in the current working directory,
+# Deprecated: Exclude URLs contained in a file from checking
```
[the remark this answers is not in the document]
vpereira01: Yeah, sounds better, removed.

mre:
Added a smoketest which tests all config options (except `output`, which would create a nasty temp file). Not perfect, but it's a step forward.
Once this pipeline passes this should be good to merge.

I'm not sure if we should keep all options in the example config enabled. The idea is that users would jump in and enable whatever they need at that moment and leave the rest to default. It's fine for now but we might change it in the future.

lychee.example.toml
```diff
@@ -1,63 +1,98 @@
...
-###
+#############################  Display  #############################
+
 # Verbose program output
 verbose = false
 
-# Show progress
-progress = false
+# Don't show interactive progress bar while checking links
+no_progress = false
```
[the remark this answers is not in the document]
vpereira01: Never mind, this works as expected.

Just realized my mistake, previously it was `progress = false` to disable the interactive progress and now it's `true` as in `no_progress = true`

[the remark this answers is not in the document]
mre: Yeah just realized that too. Glad you figured it out.

lychee.example.toml
aj-stein-nist: I like the example! :-p

I did try this re the linked CI/CD work I did on the repo (where https://github.com/lycheeverse/lychee/issues/547#issuecomment-1122690070 came from) and still had some issues. Will follow up later and try to work through understanding how to do the first example with the current syntax, or even the proposed syntax, in a follow-up issue.

Sorry for the delay.

PR #656, opened 2022-06-18
Replace lazy hashset with matches!

untitaker:
llvm will typically create much faster code than accessing a hashset at
runtime

~source: trust me bro~ actually ran benchmarks now

### DIFF
--- a/lychee-lib/src/extract/mod.rs
+++ b/lychee-lib/src/extract/mod.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+
 
 use crate::types::{uri::raw::RawUri, FileType, InputContent};
 
--- a/lychee-lib/src/extract/mod.rs
+++ b/lychee-lib/src/extract/mod.rs
@@ -8,26 +8,23 @@ mod markdown;
 mod plaintext;
 
 use markdown::extract_markdown;
-use once_cell::sync::Lazy;
 use plaintext::extract_plaintext;
 
-/// HTML elements that are deemed verbatim (i.e. preformatted).
-/// These will be excluded from link checking by default.
-static VERBATIM_ELEMENTS: Lazy<HashSet<String>> = Lazy::new(|| {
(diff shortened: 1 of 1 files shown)


untitaker:
ok

```
extract from large docs time:   [52.441 ms 53.355 ms 55.164 ms]
                        change: [-27.284% -23.363% -19.249%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe
```

lychee-lib/src/extract/mod.rs
```diff
@@ -8,26 +6,17 @@ mod markdown;
...
-        "samp".into(),
-        "xmp".into(),
-        "plaintext".into(),
-        "listing".into(),
-    ])
-});
-
 /// Check if the given element is in the list of preformatted tags
+///
+/// preformatted tags are HTML elements that are deemed verbatim (i.e. preformatted).  These will
```
mre, line 13: The wording is a little weird. Reads a bit tautological: "preformatted tags are preformatted". But I get your point. 😅

mre (approved):

mre:
Wait, am I reading this correctly? This gives us a 20% perf improvement on extracting tags!? That's insane.

untitaker:
yeah. it's a bit of guesswork. there's other hashsets in the code where i expected a similar perf gain, (e.g. example_domains), but i'm unable to replicate any sort of performance improvement by replacing those

lychee-lib/src/extract/mod.rs
untitaker: yeah that's a copypaste error. didn't read it well enough

mre:
It hadn't crossed my mind to use `matches` actually. If not for anything else it will at least reduce the code complexity a bit, so if you want, feel free to change the other occurrences as well. But we can do that in another pull request.

Issue #671, opened 2022-07-01
Relative links aren't detected from STDIN or local files


nacnudus:
A similar issue was reported in #183 and addressed by #217.  However, this doesn't work when files are passed into `STDIN`, or for local files.

```sh
lychee --dump --base https://www.example.com example.html # works
lychee --dump --base https://www.example.com example # doesn't work
cat example.html | lychee --dump --base https://www.example.com # doesn't work
```

Could this be fixed by having `--base` force HTML mode?

mre:
That's a tricky one.
I thought about assuming html input if `--base` was a valid URL, but Markdown files can also contain relative links and be prefixed with a URL. So that alone is not a decision criteria.

lychee has an internal file-type hint, that we could expose on the CLI, even though it sounds like a workaround.
https://github.com/lycheeverse/lychee/blob/363b95fe5f748a72a452ed0a9748e3a3d7e8d783/lychee-lib/src/types/input.rs#L107-L116

I'm open for suggestions on how to make the file type explicit in an ergonomic way. Alternatively maybe I'm overlooking a possibility to make it work without any file type hints.

nacnudus:
Thank you for responding so quickly.  I've attempted to address this issue in #673.

katrinafyi:
Fixed by these? 
- https://github.com/lycheeverse/lychee/issues/1665
- https://github.com/lycheeverse/lychee/pull/1842

mre:
@nacnudus, not sure if you saw the notification.

We'd still be interested in your feedback, but I'll close the issue somewhat preemptively to keep our task tracking manageable. Feel free to comment here after trying out the fixes mentioned by @katrinafyi.

PR #673, opened 2022-07-01
Add flag to force html mode

nacnudus:
This is an attempt to address #671 by adding a flag `--html` to parse
the input as HTML.  Otherwise, STDIN and local files without the `.html`
suffix are parsed as plain text.

### DIFF
--- a/fixtures/configs/smoketest.toml
+++ b/fixtures/configs/smoketest.toml
@@ -87,6 +87,9 @@ include_verbatim = false
 # Ignore case of paths when matching glob patterns.
 glob_ignore_case = false
 
+# Treat input as HTML
+html = false
+
 # Exclude URLs from checking (supports regex).
 exclude = [ '.*\.github.com\.*' ]
 
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -2,8 +2,8 @@ use crate::parse::{parse_base, parse_statuscodes};
 use anyhow::{anyhow, Context, Error, Result};
 use const_format::{concatcp, formatcp};
 use lychee_lib::{
-    Base, Input, DEFAULT_MAX_REDIRECTS, DEFAULT_MAX_RETRIES, DEFAULT_RETRY_WAIT_TIME_SECS,
-    DEFAULT_TIMEOUT_SECS, DEFAULT_USER_AGENT,
+    Base, Input, FileType, DEFAULT_MAX_REDIRECTS, DEFAULT_MAX_RETRIES,
+    DEFAULT_RETRY_WAIT_TIME_SECS, DEFAULT_TIMEOUT_SECS, DEFAULT_USER_AGENT,
 };
 use secrecy::{ExposeSecret, SecretString};
 use serde::Deserialize;
@@ -122,6 +122,11 @@ impl LycheeOptions {
     // but we'd get no access to `glob_ignore_case`.
     /// Get parsed inputs from options.
     pub(crate) fn inputs(&self) -> Result<Vec<Input>> {
+        let file_type_hint = if self.config.html {
+            Some(FileType::Html)
+        } else {
+            None
+        };
         let excluded = if self.config.exclude_path.is_empty() {
             None
         } else {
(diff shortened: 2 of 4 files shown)


mre:
Thanks for pushing this forward @nacnudus.

Some questions:

* What if a user wants to check different file types in a single run? E.g. one Markdown file and one HTML file. Would the `--html` flag always override the file extension?
* Are there any other tools that have a similar feature? I know that ripgrep has a `--type` option, which allows a user to filter input files by type (but not parse/interpret the input files differently). That's the closest I could find. (A precedent would help us make the best design decision for lychee.)

We could also change the option to `--input-type <format>`, which is a bit more flexible. E.g. `--input-type markdown` would enforce the Markdown parser instead. It still would not work for different input types, though, but that might be fine.

Alternatively, we could just add a section to the `TROUBLESHOOTING.md` file with a workaround on how to enforce the correct file type by storing the input to a temporary file with the `html` extension. Not pretty, but we avoid maintaining the option.

Looking forward to your thoughts.

nacnudus:
Hi, thanks for your reply, and apologies for my slowness responding.

> - What if a user wants to check different file types in a single run? E.g. one Markdown file and one HTML file. Would the --html flag always override the file extension?

Yes, `--html` would always override the file extension.  I really have unix pipes in mind, which don't have file extensions.

> - Are there any other tools that have a similar feature?

Imagemagick [requires an image format specifier](https://imagemagick.org/script/formats.php) to read an APNG image sequence, otherwise it assumes that it is PNG and reads only the first frame.

> We could also change the option to `--input-type <format>`

I agree, that would be better.

> Alternatively, we could just add a section to the TROUBLESHOOTING.md file with a workaround on how to enforce the correct file type by storing the input to a temporary file with the html extension.

That wouldn't work for STDIN, but perhaps lychee isn't the best tool for that any, because pipes can't benefit from lychee's multithreading, as far as I know.

lebensterben:
currently we relies on extension to determine the file format, and non supported format are treated as plaintext where we simply grab urls.

this process is VERY inefficient due to the underlying implementation of linkify.

instead, I think we can TRY to parse plaintext as html first, should that fail, we fallback to linkify.

this is helpful given the most likely use case of lychee is to validate links in html rather than any other formats.

mre:
> this is helpful given the most likely use case of lychee is to validate links in html rather than any other formats.

I'm not sure if that's actually true. Most users use lychee-action and I wouldn't be surprised if the majority of links are checked in Markdown files.

> this process is VERY inefficient due to the underlying implementation of linkify.

Not sure if parsing HTML is faster, either. We'd have to run a benchmark to see if trying HTML first would give us any significant performance gains.

lebensterben:
plain text file are `text/plain` and markdown files are `text/markdown` so 
> I think we can TRY to parse plaintext as html first

doesn't apply to mark down files and as such 

>  I wouldn't be surprised if the majority of links are checked in Markdown files

this is irrelevant.

lebensterben:
> Not sure if parsing HTML is faster, either. We'd have to run a benchmark to see if trying HTML first would give us any significant performance gains.

Parsing HTML is definitely faster than using linkify on normal sized html files.
This is because a parsed html file is a structured data and links will only appear in certain tags.

HU90m:
My preference would be a [`ripgrep`](https://github.com/BurntSushi/ripgrep)/[`fd`](https://github.com/sharkdp/fd) style type option, specifically a `--default-type` option which allows the user to override the behaviour for a file without an extension, e.g. `--default-type html`.

mre:
I like that.

mre:
We've finally merged a fix for this with https://github.com/lycheeverse/lychee/issues/1665.
It is an extension of this PR, which covers the use-case.

```sh
# Treat index as HTML
lychee --default-extension html index  

# Works with stdin
echo '<a href="https://example.com/">link</a>' | lychee --default-extension html -
```

For more info, see https://github.com/lycheeverse/lychee/pull/1842.

Thanks, @nacnudus, for the initial discussion and creating the foundation for this functionality. The feature will be released with the next version. 🎉

PR #685, opened 2022-07-11
Fix deserialize of lycheecache status codes

### LINKED ISSUE #647: Cache expecting strings, not status codes
When using lychee with `--cache`, the second run (after the cache is instantiated) results in:

> Error while loading cache: CSV deserialize error: record 1 (line: 2, byte: 54): unknown variant `404`, expected one of `Ok`, `Error`, `Excluded`, `Unsupported`. Continuing without.

This is lychee 0.10.0 (packaged in homebrew).

walterbm:
# Fix #647

Add a custom serde deserializer for `CacheStatus` to properly classify status codes in a `.lycheecache` file. 

The linked issue showed that currently when using the `--cache` flag the cli can show some confusing console errors:
> Error while loading cache: CSV deserialize error: record 1 (line: 2, byte: 54): unknown variant 404, expected one of Ok, Error, Excluded, Unsupported. Continuing without.

These errors are thrown by serde because it cannot properly deserialize a `CacheStatus` from the `.lycheecache` file using the default `Deserialize` behavior. 

In the linked issue there was a [very helpful discussion](https://github.com/lycheeverse/lychee/issues/647#issuecomment-1170711699) that walked through how this deserialization error could potentially be resolved through clever use of serde's various macro flags.

Unfortunately, I could not get it to work as expected using only serde flags. 

Instead I decided to remove the autogenerated serde deserializer and manually implement `Deserialize` for `CacheStatus`. This approach is a bit more complicated but helped guarantee that the new logic does not introduce any breaking changes with existing `.lycheecache` files and also ensures that status codes are properly categorized into the correct `CacheStatus` variant (e.g. `200` becomes a  `CacheStatus::Ok(200)` and `404` becomes `CacheStatus::Error(Some(404))`)

To help verify this functionality I also added some unit & integration tests to ensure the cache deserialization is working as expected.

Let me know what you think! This is a bit more involved than my last PR so I'm happy to make any changes and updates!

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -5,6 +5,7 @@ mod cli {
         fs::{self, File},
         io::Write,
         path::{Path, PathBuf},
+        time::SystemTime,
     };
 
     use assert_cmd::Command;
@@ -34,6 +35,18 @@ mod cli {
             .parent()
             .unwrap()
             .join("fixtures")
+    }
+
+    fn generate_lycheecache_file(data: &str) {
+        fs::write(fixtures_path().join("cache").join(".lycheecache"), data)
+            .expect("Unable to write to .lycheecache file");
+    }
+
+    fn timestamp() -> u64 {
+        SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("SystemTime before UNIX EPOCH!")
+            .as_secs()
     }
 
     #[derive(Default)]
@@ -602,6 +615,49 @@ mod cli {
             .success()
             .stdout(contains("8 Total"))
             .stdout(contains("6 Excluded"));
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_lycheecache_file() -> Result<()> {
+        // populate the .lycheecache file
+        generate_lycheecache_file(&format!(
+            r#"
+http://example.org/,200,{0}
+http://foo.example.org/,,{0}
+http://example.com/,Excluded,{0}
+https://www.rust-lang.org/,200,{0}
+https://github.com/rust-lang/rust,200,{0}
+http://example.net/foo/bar,404,{0}"#,
+            timestamp()
+        ));
+
+        let mut cmd = main_command();
+        let test_base_path = fixtures_path().join("cache");
+        let test_file_path = fixtures_path().join("TEST_EXAMPLE_DOMAINS.md");
+
+        cmd.current_dir(test_base_path)
+            .arg("--verbose")
+            .arg("--cache")
+            .arg("--exclude")
+            .arg("example.com")
+            .arg("--")
+            .arg(test_file_path)
+            .assert()
+            .stderr(contains("[200] http://example.org/ | OK (cached)"))
+            .stderr(contains("[ERR] http://foo.example.org/ | Error (cached)"))
+            .stderr(contains("[EXCLUDED] http://example.com/ | Excluded"))
+            .stderr(contains("[200] https://www.rust-lang.org/ | OK (cached)"))
+            .stderr(contains(
+                "[200] https://github.com/rust-lang/rust | OK (cached)",
+            ))
+            .stderr(contains(
+                "[404] http://example.net/foo/bar | Error (cached)",
+            ));
+
+        // reset the .lycheecache file
+        generate_lycheecache_file("");
 
         Ok(())
     }
--- a/lychee-lib/src/types/cache.rs
+++ b/lychee-lib/src/types/cache.rs
@@ -1,13 +1,13 @@
 use std::fmt::Display;
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::{ErrorKind, Status};
 
 /// Representation of the status of a cached request. This is kept simple on
 /// purpose because the type gets serialized to a cache file and might need to
 /// be parsed by other tools or edited by humans.
-#[derive(Debug, Serialize, Deserialize, Hash, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+#[derive(Debug, Serialize, Hash, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
 pub enum CacheStatus {
     /// The cached request delivered a valid response
     Ok(u16),
@@ -17,6 +17,28 @@ pub enum CacheStatus {
     Excluded,
     /// The protocol is not yet supported
     Unsupported,
+}
+
+impl<'de> Deserialize<'de> for CacheStatus {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let status = <&str as Deserialize<'de>>::deserialize(deserializer)?;
+        match status {
+            "Excluded" => Ok(CacheStatus::Excluded),
+            "Unsupported" => Ok(CacheStatus::Unsupported),
+            other => match other.parse::<u16>() {
+                Ok(code) => match code {
+                    // classify successful status codes as cache status success
+                    200..=299 => Ok(CacheStatus::Ok(code)),
+                    // classify redirects, client errors, & server errors as cache status errors
+                    _ => Ok(CacheStatus::Error(Some(code))),
+                },
+                Err(_) => Ok(CacheStatus::Error(None)),
+            },
+        }
+    }
 }
 
 impl Display for CacheStatus {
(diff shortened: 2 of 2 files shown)


lychee-lib/src/types/cache.rs
```diff
@@ -54,3 +76,61 @@ impl From<&Status> for CacheStatus {
...
+#[cfg(test)]
+mod tests {
+    use serde::de::value::{BorrowedStrDeserializer, Error as DeserializerError};
+    use serde::Deserialize;
+
+    use crate::CacheStatus;
+
+    #[test]
+    fn test_deserialize_cache_status_success_code() {
+        let deserializer: BorrowedStrDeserializer<DeserializerError> =
```
mre, line 89: Looks like this deserializer gets initiated for every test. How about creating a helper function for it?
```rust
fn deserialize_cache_status(s: &str) -> CacheStatus {
    let deserializer: BorrowedStrDeserializer<DeserializerError> =
        BorrowedStrDeserializer::new(s);
    CacheStatus::deserialize(deserializer)
}

assert_eq!(
    deserialize_cache_status("200"),
    Ok(CacheStatus::Ok(200))
);
```

mre:
@polarathene fyi

polarathene:
I should be able to review within 12 hours :+1:

lychee-lib/src/types/cache.rs
walterbm: Will do!

lychee-bin/tests/cli.rs
```diff
@@ -606,6 +619,49 @@ mod cli {
...
+            .stderr(contains("[200] https://www.rust-lang.org/ | OK (cached)"))
+            .stderr(contains(
+                "[200] https://github.com/rust-lang/rust | OK (cached)",
+            ))
+            .stderr(contains(
+                "[404] http://example.net/foo/bar | Error (cached)",
+            ));
+
+        // reset the .lycheecache file
+        generate_lycheecache_file("");
```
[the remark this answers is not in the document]
walterbm: Sounds good! I decided to leave it and ensure it is gitignored to make troubleshooting easier

lychee-lib/src/types/cache.rs
```diff
@@ -19,6 +19,28 @@ pub enum CacheStatus {
...
+        D: Deserializer<'de>,
+    {
+        let status = <&str as Deserialize<'de>>::deserialize(deserializer)?;
+        match status {
+            "Excluded" => Ok(CacheStatus::Excluded),
+            "Unsupported" => Ok(CacheStatus::Unsupported),
+            other => match other.parse::<u16>() {
+                Ok(code) => match code {
+                    // classify successful status codes as cache status success
+                    200..=299 => Ok(CacheStatus::Ok(code)),
```
polarathene, line 34: Does the `--accept` option affect caching a status like 404 from an `Error` to `Ok`? If so should it be deserialized as such when present, or does it not have relevance as it's a cached value that won't be requested?

lychee-lib/src/types/cache.rs
```diff
@@ -1,13 +1,13 @@
...
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::{ErrorKind, Status};
 
 /// Representation of the status of a cached request. This is kept simple on
 /// purpose because the type gets serialized to a cache file and might need to
 /// be parsed by other tools or edited by humans.
-#[derive(Debug, Serialize, Deserialize, Hash, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+#[derive(Debug, Serialize, Hash, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
 pub enum CacheStatus {
```
polarathene, line 11: If using serde attributes like my comment you referenced showed:

```rust
#[serde(tag = "t", content = "c")]
pub enum CacheStatus {
```

This would include the enum type/tag, followed by the value (if any). Although I noted at the end of my comment that the default deserializer had trouble for some reason.

lychee-lib/src/types/cache.rs
```diff
@@ -19,6 +19,28 @@ pub enum CacheStatus {
...
 
+impl<'de> Deserialize<'de> for CacheStatus {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let status = <&str as Deserialize<'de>>::deserialize(deserializer)?;
+        match status {
+            "Excluded" => Ok(CacheStatus::Excluded),
+            "Unsupported" => Ok(CacheStatus::Unsupported),
```
polarathene, line 30: I know that you've covered these with the tests, but do those values actually get serialized?

I've discarded my test container since, but looking at my comment, I shared two `.lycheecache` files. One was with my modifications to use serde attributes and write the enum type followed by the content/value.

Then in a [follow-up comment I shared another cache file](https://github.com/lycheeverse/lychee/issues/647#issuecomment-1170773449) which noted that `Excluded` and `Unsupported` were not written for the serde attribute "untagged".

---

Has something changed since then? Or am I mistaken, and `.lycheecache` does get these enums stored properly? (_perhaps the default for serde doesn't encounter the problem?_)

polarathene (commented):
Looks good so far, just a few concerns to verify :sweat_smile: 

---

`fixtures/cache/.keep` - What is the reason to add this? Is it just for `cli.rs:generate_lycheecache_file()` test? 

Is there a need for it, or is it better to just provide a `.lycheecache` file that has a unix timestamp far into the future?

Whatever @mre prefers is fine of course :) I'm not too familiar with Rust or writing tests myself.

lychee-lib/src/types/cache.rs
walterbm: I grabbed a baseline `.lycheecache` file by generating it from what is currently on master right now. Based on that it seems like both `Excluded` and `Unsupported` are written as expected to the file even with serde's defaults
```
 ~/lychee (master) cargo run -- --verbose --cache fixtures/TEST_EXAMPLE_DOMAINS.md                                                                                  
🔍 7 Total ✅ 2 OK 🚫 0 Errors 💤 5 Excluded

 ~/lychee (master) bat .lycheecache                                                                                                                                     
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: .lycheecache
       │ Size: 303 B
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ https://github.com/rust-lang/rust,200,1657687413
   2   │ http://example.net/foo/bar,Excluded,1657687413
   3   │ http://example.org/,Excluded,1657687413
   4   │ http://foo.example.org/,Excluded,1657687413
   5   │ https://www.rust-lang.org/,200,1657687413
   6   │ http://example.com/,Excluded,1657687413
   7   │ https://example.com/,Excluded,1657687413
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

That being said I will definitely add some test to ensure the new deserialization logic works with the default serialization

lychee-lib/src/types/cache.rs
walterbm: That's a really good question — my changes do not take the `--accept` option into account. @mre what's the expected behavior when the `--accept` and the `--cache` flags are passed at the same time?

walterbm:
> 
> `fixtures/cache/.keep` - What is the reason to add this? Is it just for `cli.rs:generate_lycheecache_file()` test?
> 
@polarathene yeah the `fixtures/cache/.keep` file is just to ensure the `cli.rs:generate_lycheecache_file()` test works as expected and to isolate the cache fixtures in their own directory. 

I'm happy to replace it with a hardcoded `.lycheecache` file if that makes more sense to you and @mre

lychee-lib/src/types/cache.rs
walterbm: I improved the CLI test to be a full integration test of the caching behavior: https://github.com/lycheeverse/lychee/pull/685/commits/e55b8239028217334b2e4a7301845022dca2ca9f

That should hopefully validate the serialization and deserialization of `CacheStatus`

lychee-lib/src/types/cache.rs
walterbm: Unfortunately I also could not get that to work with the default deserializer and I was also weary of making any "breaking changes" to the `.lycheecache` file format

lychee-lib/src/types/cache.rs
mre: Good question; I don't know.

I think the cache should reflect whatever the current settings were on execution to avoid confusion. The tricky part are consecutive runs with different settings. We could think about not caching those, but it would get complicated really quickly.

The alternative is to just keep the caching behavior sort of naive and see if users make any change requests.

lychee-bin/tests/cli.rs
```diff
@@ -606,6 +606,80 @@ mod cli {
...
 
+    #[tokio::test]
+    async fn test_lycheecache_file() -> Result<()> {
+        let mut cmd = main_command();
+
+        let test_base_path = fixtures_path().join("cache");
+        assert!(
+            fs::metadata(test_base_path.join(".lycheecache")).is_err(),
+            ".lycheecache file should not exist before this test"
+        );
```
mre, line 617: There's also https://doc.rust-lang.org/std/path/struct.Path.html#method.exists

```rust
assert!(
    !test_base_path.join(".lycheecache").exists(),
    ".lycheecache file should not exist before this test"
);
```

mre: Under the hood it does exactly the same, but it's arguably a bit easier to read.

lychee-bin/tests/cli.rs
```diff
@@ -606,6 +606,80 @@ mod cli {
...
+        writeln!(file, "{}", mock_server_err.uri().as_str())?;
+        writeln!(file, "{}", mock_server_exclude.uri().as_str())?;
+
+        // run first without cache to generate the .lycheecache file
+        cmd.current_dir(&test_base_path)
+            .arg(dir.path().join("c.md"))
+            .arg("--verbose")
+            .arg("--cache")
+            .arg("--exclude")
+            .arg(mock_server_exclude.uri())
```
mre, line 636: Realized you're constructing this command with the same arguments below.
I think you could construct it once and assign it to a variable?

```rust
let cache_cmd = cmd
        .current_dir(&test_base_path)
        .arg(dir.path().join("c.md"))
        .arg("--verbose")
        .arg("--cache")
        .arg("--exclude")
        .arg(mock_server_exclude.uri());
}

// and then
// cache_cmd.assert().stderr(...);
```

lychee-bin/tests/cli.rs
```diff
@@ -606,6 +606,80 @@ mod cli {
...
+                "[404] {}/ | Error (cached)\n",
+                mock_server_err.uri()
+            )))
+            .stderr(contains(format!(
+                "[EXCLUDED] {}/ | Excluded\n",
+                mock_server_exclude.uri()
+            )));
+
+        // clear the .lycheecache file
+        fs::remove_file(fixtures_path().join("cache").join(".lycheecache"))?;
```
mre, line 678: Same here, `fixtures_path().join("cache")` could be replaced with `test_base_path().join(".lycheecache")` for consistency. 
Alternatively you could also introduce a `cache_file` variable in the beginning and reuse that throughout the test. 😉 
```rust
let cache_file = test_base_path().join(".lycheecache");
```

mre:
Added a few more comments about the integration test. These are all very minor things, so they shouldn't be a blocker for merging. Let me know if you want to address those.

I also added a comment about `--accept` and `--cache` behavior. It feels like we can postpone the decision because the behavior wasn't changed here and there might not be an optimal solution. Let's wait for some user feedback before we move forward.

Apart from that we're good. 🎉

lychee-bin/tests/cli.rs
walterbm: That's much nicer! I'll update it to use `exists`

lychee-bin/tests/cli.rs
walterbm: Good idea! I'll make that update

lychee-bin/tests/cli.rs
walterbm: done!

lychee-lib/src/types/cache.rs
walterbm: Makes sense to me!

This is a challenging technical problem. I've been trying to find a clean way to pass the context of the `--accept` option into the `CacheStatus` deserialization and so far have not found any easy solutions.

For now I added a comment with a `TODO` explaining that the cache status does not account for status code overrides passed through the `--accept` flag. I'm happy to revisit this in the future if needed.

walterbm:
@mre and @polarathene thank you for all the great feedback! Really helped me learn this project (and Rust) a lot better and made the fix + tests a lot stronger!

mre:
LGTM. Tested locally and works as expected. Thanks for your contribution @walterbm. 🥳

tooomm:
Awesome, thanks for this fix @walterbm! ❤️

PR #691, opened 2022-07-15
Update to gracefully handle nonexistent relative paths

### LINKED ISSUE #666: Inputs prefixed with `./` take 30 seconds to timeout when the file does not exist
When prefixing a relative path with `./` as a hint that this is a relative path, not a DNS label/hostname to lookup (_see Network error in reference below_), `lychee` shows progress for attempting link extraction for the non-existent local (or remote) input, until it times out at 30 seconds:

```console
$ lychee './relative-path/to/file.md'
⠒ 0/0 [00:00:29] █████████████████████████ Extracting links

# At 30 seconds, lychee times out with output:
Error: Network error

Caused by:
    0: error sending request for url (http://./relative-path/to/file.md): error trying to connect: dns error: request timed out
    1: error trying to connect: dns error: request timed out
    2: dns error: request timed out
    3: request timed out
```

- Without the `./` prefix, the resolution that the input does not exist is instant.
- If the file does exist, the response is instant with the `./` prefix. Only non-existent files are affected.
- `--skip-missing` makes no difference, other than ignoring the network error shown above and implying no problems:

```console
$ lychee --skip-missing './relative-path/to/file.md'
⠒ 0/0 [00:00:29] █████████████████████████ Extracting links

# At 30 seconds, lychee times out with output:
🔍 0 Total ✅ 0 OK 🚫 0 Errors
```

---

**For reference (below works as expected):**

```
--skip-missing           Skip missing input files (default is to error if they don't exist)
```

Providing a relative local filepath, when the file does not exist results in a network error (_an implicit fallback to a `http://` request?_):

```console
$ lychee 'relative-path/to/file.md'
Error: Network error

Caused by:
    0: error sending request for url (http://relative-path/to/file.md): error trying to connect: dns error: no record found for name: relative-path. type: AAAA class: IN
    1: error trying to connect: dns error: no record found for name: relative-path. type: AAAA class: IN
    2: dns error: no record found for name: relative-path. type: AAAA class: IN
    3: no record found for name: relative-path. type: AAAA class: IN
```

`--skip-missing` hides the network failure completely (_related, it might be useful to [report skipped/empty inputs](https://github.com/lycheeverse/lychee/issues/665)_)

```console
$ lychee --skip-missing 'relative-path/to/file.md'

🔍 0 Total ✅ 0 OK 🚫 0 Errors
```

walterbm:
# Fix #666

Update `Input::new` to immediately exit when the input is a relative path that does not exist. The early return helps guarantee the input is not coerced into a URL that will take 30 seconds to resolve.

Also adds a small test to validate the new behavior.

I tried to follow the guidance from this comment when resolving the issue: https://github.com/lycheeverse/lychee/issues/666#issuecomment-1172172797

### DIFF
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -148,7 +148,7 @@ impl Input {
                 if path.exists() {
                     InputSource::FsPath(path)
                 } else if path.is_relative() {
-                    // If the file does not exist and it is a relative path exit immediately
+                    // If the file does not exist and it is a relative path, exit immediately
                     return Err(ErrorKind::FileNotFound(path));
                 } else {
                     // Invalid path; check if a valid URL can be constructed from the input
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -147,6 +147,8 @@ impl Input {
                 let path = PathBuf::from(value);
                 if path.exists() {
                     InputSource::FsPath(path)
+                } else if path.is_relative() {
+                    return Err(ErrorKind::FileNotFound(path));
                 } else {
                     // Invalid path; check if a valid URL can be constructed from the input
                     // by prefixing it with a `http://` scheme.
@@ -373,6 +375,13 @@ fn is_excluded_path(excluded_paths: &[PathBuf], path: &PathBuf) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn test_input_gracefully_handles_nonexistent_relative_paths() {
+        let input = Input::new("./relative/path", None, false, None);
+        assert!(input.is_err());
+        assert!(matches!(input, Err(ErrorKind::FileNotFound(..))));
+    }
 
     #[test]
     fn test_valid_extension() {
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -399,7 +399,13 @@ mod tests {
 
     #[test]
     fn test_input_handles_nonexistent_relative_paths() {
-        let input = Input::new("./nonexistent/relative/path", None, false, None);
+        let test_file = "./nonexistent/relative/path";
+        let path = Path::new(test_file);
+
+        assert!(!path.exists());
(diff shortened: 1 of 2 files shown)


mre:
Yeah a relative path that exists is still fine.

walterbm:
@lebensterben @mre this still works as expected for relative paths since the first part of the conditional checks that the `path.exists`. The `path.is_relative()` early return is only triggered when the file does not exist. 

Is that not the correct logic?

Here's a quick test run on this branch with a relative path showing that those still work as expected:
```
 ~/lychee (handle-nonexistent-relative-paths) cargo run ./fixtures/TEST_EXAMPLE_DOMAINS.md                                                                                  
    Finished dev [unoptimized + debuginfo] target(s) in 0.72s
     Running `target/debug/lychee ./fixtures/TEST_EXAMPLE_DOMAINS.md`
🔍 7 Total ✅ 2 OK 🚫 0 Errors 💤 5 Excluded
```

I can also add some tests to further validate this behavior

mre:
Sounds good. Do you think we should also add an integration test, which validates the non-blocking behavior for non-existent files?

mre:
> Do you have any guidance on how I could verify the non-blocking behavior?

I guess the reason why it was blocking in the first place was that it was incorrectly identified as a URL. With that you can just check whether a non-existent relative path does indeed get recognized as a missing path and not converted to URL?

walterbm:
@mre updated with some integration tests validating this behavior in `lychee-bin`. Let me know what you think!

lychee-lib/src/types/input.rs
```diff
@@ -147,6 +147,9 @@ impl Input {
                 let path = PathBuf::from(value);
                 if path.exists() {
                     InputSource::FsPath(path)
+                } else if path.is_relative() {
+                    // If the file does not exist and it is a relative path exit immediately
```
mre, line 151:
```suggestion
                    // If the file does not exist and it is a relative path, exit immediately
```

mre:
Tested locally and works as expected. Thanks @walterbm. 🥳

PR #692, opened 2022-07-15
Skip caching unsupported and excluded URLs

mre:
As discussed in https://github.com/lycheeverse/lychee/issues/647#issuecomment-1170773449, it does not make much sense to cache unsupported
and excluded URLs.
Unsupported URLs might be supported in the future and caching them
would mean they won't get checked then. Excluded URLs were
excluded for a reason and should not appear in the cache.
Furthermore they might not be excluded
in a consecutive run, leading to a false-positive.

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -135,10 +135,14 @@ async fn handle(client: &Client, cache: Arc<Cache>, request: Request) -> Respons
     // TODO: Handle error as soon as https://github.com/seanmonstar/reqwest/pull/1399 got merged
     let response = client.check(request).await.expect("cannot check URI");
 
-    // Never cache filesystem access as it is fast already so caching has no
-    // benefit
-    if !uri.is_file() {
-        cache.insert(uri, response.status().into());
+    // - Never cache filesystem access as it is fast already so caching has no
+    //   benefit.
+    // - Skip caching unsupported URLs as they might be supported in a
+    //   future run.
+    // - Skip caching excluded links; they might not be excluded in the next run
+    let status = response.status();
+    if !uri.is_file() && !status.is_excluded() && !status.is_unsupported() {
+        cache.insert(uri, status.into());
     }
     response
 }
--- a/lychee-lib/src/types/cache.rs
+++ b/lychee-lib/src/types/cache.rs
@@ -16,6 +16,11 @@ pub enum CacheStatus {
     /// The request was excluded (skipped)
     Excluded,
     /// The protocol is not yet supported
+    // We no longer cache unsupported files as they might be supported in future
+    // versions.
+    // Nevertheless, keep for compatibility when deserializing older cache
+    // files, even though this no longer gets serialized. Can be removed at a
+    // later point in time.
     Unsupported,
 }
 
@@ -27,6 +32,9 @@ impl<'de> Deserialize<'de> for CacheStatus {
         let status = <&str as Deserialize<'de>>::deserialize(deserializer)?;
         match status {
             "Excluded" => Ok(CacheStatus::Excluded),
+            // Keep for compatibility with older cache files, even though this
+            // no longer gets serialized. Can be removed at a later point in
+            // time.
             "Unsupported" => Ok(CacheStatus::Unsupported),
             other => match other.parse::<u16>() {
                 Ok(code) => match code {
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -14,6 +14,11 @@ mod cli {
 
     type Result<T> = std::result::Result<T, Box<dyn Error>>;
 
+    // The lychee cache file name is used for some tests.
+    // Since it is currently static and can't be overwritten, declare it as a
+    // constant.
+    const LYCHEE_CACHE_FILE: &str = ".lycheecache";
+
     macro_rules! mock_server {
         ($status:expr $(, $func:tt ($($arg:expr),*))*) => {{
             let mock_server = wiremock::MockServer::start().await;
@@ -609,7 +614,10 @@ mod cli {
     #[tokio::test]
     async fn test_lycheecache_file() -> Result<()> {
         let base_path = fixtures_path().join("cache");
-        let cache_file = base_path.join(".lycheecache");
+        let cache_file = base_path.join(LYCHEE_CACHE_FILE);
+
+        // Unconditionally remove cache file if it exists
+        let _ = fs::remove_file(&cache_file);
 
         let mock_server_ok = mock_server!(StatusCode::OK);
         let mock_server_err = mock_server!(StatusCode::NOT_FOUND);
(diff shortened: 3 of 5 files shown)


mre:
@walterbm, @polarathene can you have a look? We discussed that earlier.

polarathene (approved):
LGTM :+1: 

Thanks for tackling that :)

lychee-bin/src/commands/check.rs
```diff
@@ -135,10 +135,14 @@ async fn handle(client: &Client, cache: Arc<Cache>, request: Request) -> Respons
...
-    // benefit
-    if !uri.is_file() {
-        cache.insert(uri, response.status().into());
+    // - Never cache filesystem access as it is fast already so caching has no
+    //   benefit.
+    // - Skip caching unsupported URLs as they might be supported in a
+    //   future run.
+    // - Skip caching excluded links; they might not be excluded in the next run
+    let status = response.status();
+    if !uri.is_file() && !status.is_excluded() && !status.is_unsupported() {
```
walterbm, line 144: Could similar logic be used to disable caching when the `Status` has been overridden through the `--accept` flag? Or is that the wrong way to solve the conflict between `--accept` and `--cache` raised here:  https://github.com/lycheeverse/lychee/pull/685#discussion_r919501382 ?

lychee-lib/src/types/cache.rs
```diff
@@ -113,6 +121,7 @@ mod tests {
     }
 
     #[test]
+
```
walterbm, line 124: Does this extra space affect here affect the `#[test]` macro in anyway?

lychee-lib/src/types/cache.rs
```diff
@@ -16,6 +16,11 @@ pub enum CacheStatus {
     /// The request was excluded (skipped)
     Excluded,
     /// The protocol is not yet supported
+    // We no longer cache unsupported files as they might be supported in future
+    // versions.
+    // Nevertheless, keep for compatibility when deserializing older cache
+    // files, even though this no longer gets serialized. Can be removed at a
+    // later point in time.
```
walterbm, line 23: This is a tangential discussion but have you all considered versioning the `.lycheecache` file to allow for "breaking" changes in the file format? 

I thought about that while working on the cache deserialization fix but it seemed unwise to break the expected format for a simple bug fix

walterbm (approved):
Looks good! Thank you for cleaning up some of the tests!

Left a few questions because (selfishly) I wanted to pick your brain to better understand the project.

lychee-bin/src/commands/check.rs
lebensterben: TBH I don't know what's the use case of --accept..

lychee-lib/src/types/cache.rs
lebensterben: this is a good idea.

lychee-bin/src/commands/check.rs
mre: It was mostly a workaround for the dreaded 429 status code because we don't support rate limiting yet.
That said, some websites send special status codes (e.g. LinkedIn uses 999), that we don't know how to interprete. In such cases it's best to let the user decide by overwriting the default accepted status codes.

lychee-lib/src/types/cache.rs
mre: I don't think so, but it's a formatting issue that I should fix. Thanks!

lychee-lib/src/types/cache.rs
mre: Agree, it's a solution. How can csv files be versioned? Would we just add a header to the file? In that case, would that still be a valid csv file?

lebensterben: @mre 

just add comment lines, for example

#Cache file generated by lychee
#Do NOT modify it by hand
#Timestamp=xxxxxxx
#Version=1

lebensterben: the drawback is that CSV standard doesn't specify comment syntax

so some application like Excel see them as actual data.

but decent parsers usually allows you to specify comment character.

polarathene: How would the version metadata be handled by `lychee` when encountering an older version?

If you're not providing backward-compatibility, it would fail either way?

Alternatively you could add a version suffix to the filename itself, or just switch away from CSV? Is there anyone requiring CSV? Presumably `.lycheecache` would only ever be intended to be parsed by `lychee`.

I think we had some simpler approaches to handle the recent serde issues for handling the cache, but ended up with a custom deserializer because of support issues with serde for CSV with enums. If you're going to introduce versioning that would make existing format redundant, might as well consider a different format for `.lycheecache` too.

lebensterben: we can have backward-compatibility.
mostly likely the breaking changes are:
- Changing cache format
- Adding/removing keys or re-assigning meaning of keys
- Including/excluding queries in/from cache

---

`.lycheecache` doesn't even't need to human readable. binary format would be smaller and faster to load.

lychee-bin/src/commands/check.rs
mre: @walterbm, I had another look. Our caching takes the accepted status codes into account. Consider this constructor for the status code:
https://github.com/lycheeverse/lychee/blob/a557cba0b4fcf63c496b9dc202787e899d290e08/lychee-lib/src/types/status.rs#L72-L85
The status only gets overridden if `accepted.map(|a| a.contains(&code))`. The remaining status codes remain untouched. So for example if a user sets `--accept 123,456`, then these status codes will be set to `Status::Ok(code)` on top of the normally accepted ones.

To answer your question, yes I think we could add the check for `--accept` here and maybe it's fine if we just skip caching the user-specified status codes. These could change from run to run.

lychee-lib/src/types/cache.rs
mre: The reason why I like CSV is because it's simple.
The manual editability is a nice benefit: users could retry a few URLs by removing lines from the file without invalidating the rest of the cache. I make use of that trick sometimes.

The fact that it's human-readable also means that other Unix tools can work with the data:
```bash
# How many non-200 status codes were cached?
grep -v "200" .lycheecache | wc -l
```

It also compresses really well with GZIP, so file size isn't a major concern. For instance, a cache for the top 10k websites is 358K and 88K gzipped (25% of original size). With brotli it's 72K.

We could change the cache filetype, but I don't know of a better alternative that still retains these properties. 
I'd be okay with adding a version as a comment to CSV &mdash; at least it seems to be a relatively common way to add metadata. Then again I wonder if we should postpone the decision until we hit our first breaking change with the existing format.
In any case, the review section is probably not the right place for making a decision, so maybe someone wants to open up an issue and summarize the discussion? 😅

mre:
Thanks for the review and the comments. ✌️

lychee-bin/src/commands/check.rs
walterbm: @mre ah that makes a lot of sense! Thank you for this context!

PR #697, opened 2022-07-19
Update description with token now supported

fauust:
See: https://github.com/peter-evans/dockerhub-description/issues/10

Bump GH actions versions.

As a reminder, we already discussed this in https://github.com/lycheeverse/lychee/pull/325

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,6 +23,9 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'lycheeverse' &&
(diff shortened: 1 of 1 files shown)


fauust:
Yep, thanks. I am not sure if we should skip the whole build step or just those that push to dockerhub.

mre:
The other steps are just login and setup steps, so I think it's fine if we skip the entire pipeline in this case unless I'm missing something.
In any case, thanks a lot for the addition @fauust. 😃

fauust:
Yes you are right! On other CI I build the image locally and push it in a local registry. But that's not the case here. So skipping is perfectly fine.

PR #718, opened 2022-08-03
Change progress bar style

mre:
* Use pink for spinner
* Show ETA instead of elapsed
* dim progress bar and adjust size to terminal width


mre:
What I don't understand is how why the progress line sometimes overflows on large URLs.

image

This is only happening since indicatif 0.17.0. Unless it's an obvious mistake on my end (which is likely), I will open an issue with indicatif to reach out for help.

lychee-bin/src/commands/check.rs
```diff
@@ -51,9 +51,11 @@ where
     let pb = if params.cfg.no_progress {
         None
     } else {
-        let bar = ProgressBar::new_spinner().with_style(ProgressStyle::default_bar().template(
-            "{spinner:.red.bright} {pos}/{len:.dim} [{elapsed_precise}] {bar:25} {wide_msg}",
-        ).expect("Valid progress bar"));
+        let bar = ProgressBar::new_spinner().with_style(
+            ProgressStyle::default_bar()
+                .template("{spinner:.197.bright} {pos}/{len:.dim} ETA {eta} {bar:.dim} {wide_msg}")
```
lebensterben, line 56: can you replace default_bar() and template() by with_template()?

mre: Good idea, thanks.

lebensterben:
> This is only happening since indicatif 0.17.0. Unless it's an obvious mistake on my end (which is likely), I will open an issue with indicatif to reach out for help.

probably because of https://github.com/console-rs/indicatif/commit/d7851c7a467313571d71133b1729af9c5dd3b2ae ?

what if we turn off this feature?

mre:
Do you mean like this?

```toml
[dependencies.indicatif]
version = "0.17.0"
default-features = false
features  = []
```

Just tried that, but the result is still the same.

lebensterben:
`git-bisect` helped me pin down the commit that breaks it.
https://github.com/console-rs/indicatif/commit/5038ba8

mre:
I just tried it again locally after a rebase on top of `master` and it works now. No more overflows. Maybe a recent update of indicativ fixed it. 🤔 In any case, I'll take it. If there's still an issue it most likely precedes this change, so I'll go ahead and merge it.

PR #730, opened 2022-08-12
Listing InnerSourceCommons/InnerSourcePatterns as user

spier:
The repo [InnerSourceCommons/InnerSourcePatterns](https://github.com/InnerSourceCommons/InnerSourcePatterns) is using lychee as a link checker since a couple of days.
(we found the recommendation for it in the liche repo)

I wasn't clear if we should add the new user/repo at top or bottom of the list.
Went for the top :)

Thank you for this great tool!

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,6 +400,7 @@ We collect a list of common workarounds for various websites in our [troubleshoo
 
 ## Users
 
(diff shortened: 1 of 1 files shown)


mre:
Hey, thanks a lot for using lychee and for the pull request. 👏☺️

PR #732, opened 2022-08-12
Move from structopt to clap

mre:
Structopt was subsumed by clap. See
https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating

I think the help output looks much cleaner now (and has colors! ✨).
image

### TODO
* [ ] Check if config merging works as expected. We should probably write an integration test for that at some point(?).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,15 +27,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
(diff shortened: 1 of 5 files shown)


lebensterben (approved):

mre:
Writing an integration test for config merging was trickier than I expected. It would require some refactoring and I don't want to spend much time on this. 😆 In my manual tests it did still work, though, so I'm willing to merge it in as is.

PR #737, opened 2022-08-16
Merge all updates as long as they pass CI

mre:


lebensterben:
we also need https://github.com/obi1kenobi/cargo-semver-checks-action

mre:
Should be something like this:

```yaml
- name: Check semver
  uses: obi1kenobi/cargo-semver-checks-action@v1
  with:
    crate-name: lychee-lib
```

We could enable it once 1.0 is reached.

PR #792, opened 2022-10-12
Fix syntax highlighting in README

jwilk:
The help message was highlighted as if it was a gitignore file.
Let's disable this spurious syntax highlighting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ token with no extra permissions is enough to be able to check public repos links
 There is an extensive list of commandline parameters to customize the behavior.
 See below for a full list.
 
-```ignore
+```text
 USAGE:
     lychee [OPTIONS] <inputs>...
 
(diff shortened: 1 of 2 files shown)


lebensterben:
replacing ignore by text should work.

jwilk:
> replacing ignore by text should work.

Indeed, it did the trick. Thanks.

lebensterben (approved):

mre:
While it technically isn't the correct file format, I liked the syntax highlighting of the `[default]` arguments. Is there any way to keep those?

lebensterben:
@mre use inline html and you can format it freely. but that'll be very tedious...

mre:
Guess that's too much work. I think it's better to merge it as is. Thanks for the input @lebensterben and thanks for fixing the syntax highlighting @jwilk. 👍

PR #801, opened 2022-10-21
Extend description of "--exclude" to also exclude email addresses, not only URLs

andygrunwald:
Having this help available:
```
$ lychee --help
[...]
--exclude <exclude>...                 Exclude URLs from checking (supports regex)
[...]
--exclude-mail           Exclude all mail addresses from checking
[...]
```

makes me think "How to I exclude single email addresses?".
The answer is: `--exclude` does this as well.
For me, this was not clear from the start.

This pull requests changes the help text for the `--exclude` argument from

> Exclude URLs from checking (supports regex)

to

> Exclude URLs and mail addresses from checking (supports regex)

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ OPTIONS:
             Equivalent to `--exclude-private --exclude-link-local --exclude-loopback`
 
         --exclude <EXCLUDE>
-            Exclude URLs from checking (supports regex)
+            Exclude URLs and mail addresses from checking (supports regex)
 
         --exclude-file <EXCLUDE_FILE>
             Deprecated; use `--exclude-path` instead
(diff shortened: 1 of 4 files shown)


lebensterben:
strictly speaking URL includes email address:

    mailto:username@example

So excluding URL already implies it can exclude email address.

andygrunwald:
@lebensterben that is true. For this, you need to know that email addresses fall into the category of URLs. This was not obvious to me.
I also did not find a note in the repo/docs that indicate this.

The pull request try to make this more explicit for people that don't know this (like me).
This might help others to understand the `--exclude` flag better.

PS: Fixed unit tests in https://github.com/lycheeverse/lychee/pull/801/commits/d92f65f1ce9aaa071a6f2dcb61ed8811afff966b

mre:
Thanks for your contribution @andygrunwald. ⭐

PR #809, opened 2022-11-05
Fix parsing error of email addresses with query params

### LINKED ISSUE #653: Mailto urls with query parameters incorrectly converted to file urls
Lychee seems to incorrectly convert `mailto:` links with query parameters into `file://` links.

`mailto:user@example.com?subject=Hello` becomes `file:///LOCAL/PATH/mailto:user@example.com`

The actual link to check (when checking emails) is `mailto:user@example.com`.

mre:
Email addresses with query parameters often get used in
contact forms on websites. They can also be found in
other documents like Markdown.

A common use-case is to add a subject line to the email
as a parameter e.g. `mailto:mail@example.com?subject="Hello"`.

Previously we handled such cases incorrectly by interpreting
them as files. The reason was that our email parsing was too strict
to allow for query parameters at the end of an email address.
With [email_address] we switched to a more permissive parser.

Note that this does not affect the actual address email _checking_,
as this is still done with [check-if-email-exists], which has more strict
check functionality. For the check we pass the path to `check-if-email-exists`
and ignore the query params because they aren't part the email address
but rather get passed to the mail client.

This change modifies the URI parsing for all links.
All tests pass, so I hope I didn't break anything, but I'd appreciate some
review/testing before merging this.
As an added benefit I optimized the common path, which is a non-email
URL, so parsing should overall be _slightly_ faster, even though I haven't
done any benchmarks and the performance benefit is probably negligible. 

Fixes #653.
@egrieco, @Stargator, @wolf99 fyi

[check-if-email-exists]: https://github.com/reacherhq/check-if-email-exists
[email_address]: https://github.com/johnstonskj/rust-email_address

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -982,6 +982,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e78d4f1cc4ae33bbfc157ed5d5a5ef3bc29227303d595861deb238fcec4e9457"
 
 [[package]]
+name = "email_address"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e2153bd83ebc09db15bcbdc3e2194d901804952e3dc96967e1cd3b0c5c32d112"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "encode_unicode"
 version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1800,7 +1809,7 @@ dependencies = [
  "cached",
  "check-if-email-exists",
  "doc-comment",
- "fast_chemail",
+ "email_address",
  "futures",
  "glob",
  "html5ever",
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -55,21 +55,21 @@ Or, you can accept all content/MIME types: `--headers "accept=*/*"`.
 See more info about the Accept header
 [over at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept).
 
-
 ## Unreachable Mail Address
 
 We use https://github.com/reacherhq/check-if-email-exists for email checking.
 You can test your mail address with curl:
-```
+
+```bash
  curl -X POST \
   'https://api.reacher.email/v0/check_email' \
   -H 'content-type: application/json' \
   -H 'authorization: test_api_token' \
   -d '{"to_email": "box@domain.test"}'
 ```
-Some settings on your mail server (such as SPF Policy, DNSBL) may prevent your email from being verified.
-If you have an error with checking a working email, you can disable this check using the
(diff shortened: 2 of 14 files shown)


Stargator:
Thank you, @mre. Nothing stands out from a code review, but Ruby is not my strongest language.

I'll be able to run the tests later and double check all is good.

mre:
> Ruby is not my strongest language

Mine neither. 🤣 But I know you meant Rust. 

> I'll be able to run the tests later and double check all is good.

Thanks for the support on this one. 👍

Stargator:
> Mine neither. 🤣 But I know you meant Rust.

Oh...yeah... I didn't even look at the file extension. I just thought the syntax looked ruby-like. Thanks

Stargator:
@mre When I run the two new tests the output does not seem to state the tests were actually executed.

Below is an example. All I see is "running 0 tests", am I doing anything wrong?

```bash
~/Workspace/GitHub/lychee (mailto *)$ cargo test test_email_html_with_subject
    Finished test [unoptimized + debuginfo] target(s) in 0.44s
     Running unittests src/main.rs (target/debug/deps/lychee-27966ae4bc81e0dc)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 12 filtered out; finished in 0.00s

     Running tests/example_domains.rs (target/debug/deps/example_domains-8d3a6e088dd0d97d)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s

     Running tests/local_files.rs (target/debug/deps/local_files-88fa86338aa53c8d)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s

     Running tests/usage.rs (target/debug/deps/usage-6b8f80821d17bf05)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/lychee_lib-4364c2b43d456ab2)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 112 filtered out; finished in 0.00s
```

mre:
This would be the full command:

```bash
cargo test --package lychee --test cli --features check_example_domains -- cli::test_email_html_with_subject --exact --nocapture
```

mre:
I did test it with the dart repo and it looks fine. Anything else I should test it with?

Stargator:
@mre Could you test it with the exercism/problem-specifications repo?

Stargator:
> This would be the full command

Thank you. I played around with the test, changing the expected values as well as the TEST_EMAIL_QUERY_PARAMS html and md files, just to verify it accepts net and org domains as well as verifying it works with no query params.

Outside of the problem-specifications repo, I think that would be it.

mre:
Very nice. Thanks for taking a look. 
The exercism/problem-specifications works, too:

```
problem-specifications ❯❯❯ lychee .
🔍 94 Total ✅ 94 OK 🚫 0 Errors
```

Merging.

PR #810, opened 2022-11-05
Add support for dumping links to file

### LINKED ISSUE #698: No output for `--dump` / Allow to `--dump` extracted links to file
It looks like the dump flag will not create the expected output.

`cat /tmp/lychee/out.md` gives `No output. Check pipeline run to see if lychee panicked.`
The dumped output should have the found links, no? 🤔

mre:
Fixes #698


mre:
@tooomm, this should fix your use-case described in #698.

examples/builder/builder.rs
```diff
@@ -9,9 +9,9 @@ use std::{collections::HashSet, time::Duration};
 #[allow(clippy::trivial_regex)]
 async fn main() -> Result<()> {
     // Excludes
-    let excludes = Some(RegexSet::new(&[r"example"]).unwrap());
+    let excludes = Some(RegexSet::new([r"example"]).unwrap());
```
mre, line 12: Note: these changes are unrelated to the feature, but I've added them to make clippy happy after upgrading to Rust 1.65.

tooomm:
Thanks for this addition, @mre! I just tried it and it works. :)

However, if I combine it with a custom path for the output, I can not find the output file with the extracted/dumped links. I'm using [lychee Action](https://github.com/lycheeverse/lychee-action) and the docs there did not help me. It's probably some relative path thing?
It is mentioned [here](https://github.com/lycheeverse/lychee-action/blob/master/action.yml#L28), that the default location is `lychee/out.md`. And I can access that output file at `/tmp/lychee/out.md` if I do not set `output`.

But if I specify `output: lychee/foo.txt` similar to one of the examples in the lychee-action Readme (or only `foo.txt`), `cat /tmp/lychee/foo.txt` returns `cat: /tmp/lychee/foo.txt: No such file or directory`.

Any idea?

mre:
According to the `Dockerfile`, the working directory should be `/lychee`. If I'm not mistaken, if you set `output: lychee/foo.txt` then the file should be at `/lychee/lychee/foo.txt`. Can you check?

Issue #709, opened 2022-07-30
Cache usage not logged


tooomm:
On cache expiry, a helpful message is print:
``` 
Cache is too old (age: 9days 23h 47m 11s 823ms 371us 801ns, max age: 8h). Discarding
 ✗ [404] https://*URL | Network error: Not Found
```

If cache is used, nothing shows up. I just realized that it probably works because I had an error cached and that popped up.
```
 ↻ [404] https://*URL | Error (cached)
```

<br>

I think some similar information would be very helpful there as well:
``` 
Cache is recent (age: 2h 47m 11s 823ms 371us 801ns, max age: 8h). Using
 ↻ [404] https://*URL | Error (cached)
```

lebensterben:
it should be hidden by default but be turned on with -v or -vv flag.

mre:
@lebensterben is probably right.

tooomm:
> it should be hidden by default but be turned on with -v or -vv flag.

Hmm. Considering that `--verbose`/ `-v` can print hundreds or thousands of lines below that single `Cache is used` line we are talking about, I'm not so sure if it would be that useful at all then.

It looks like `-vv` is not even documented? 🤔

lebensterben:
@tooomm 

When everything is working fine the program should not emit any noise except the simple status of success. That's why I think it should not report a good cache.

mre:
> What about a simple else here? https://github.com/lycheeverse/lychee/blob/master/lychee-bin/src/main.rs#L179-L186

Kinda leaning towards what @lebensterben said: if it works as expected let's not make the output more verbose.

> It looks like -vv is not even documented? 🤔

That's because it doesn't exist yet. 😆
We should add that and print the message.

mre:
Fixed with #824.

Issue #704, opened 2022-07-24
List excluded URLS?


wolf99:
Is it possible to list excluded URLs?

I am adding this action to a repo in https://github.com/GitCredentialManager/git-credential-manager/pull/700, where it reports as follows:

```console
| Status        | Count |
|---------------|-------|
| 🔍 Total      | 236   |
| ✅ Successful | 158   |
| ⏳ Timeouts   | 0     |
| 🔀 Redirected | 0     |
| 👻 Excluded   | 78    |
| ❓ Unknown    | 0     |
| 🚫 Errors     | 0     |
```

Though there are some items in a `.lycheeignore` file, I would not have expected as many as 78 excludeds. Therefore I would like to understand in more detail exactly what is excluded so as to troubleshoot my configuration of Lychee.
Is it possible to configure the action to list the excludeds in the output in some way?

mre:
At least in verbose mode we print the excluded URLs now when we check them.
```
lychee -vv .
```

Note: this will be part of the **next release**.
With that I consider this done. 🥳  Thanks for the report.

Issue #661, opened 2022-06-23
Is there a valid reason to check `data:` URIs?


polarathene:
Not sure if this is a bug. `data:` URIs in an HTML document are identified as links (ok), but only ignored when checking (`--dump` does not ignore, which is fine), yet the error is "Unsupported The given URI is invalid".

I'm not sure when it would be valid to check? `data:` URI would be local data AFAIK, so always valid. The contents might contain additional links that could be parsed, similar to a linked document (which presently I don't think is supported). In this case a `w3.org` link is present which seems to still get extracted from it's contents and excluded). I assume that extraction is unrelated to being part of the URI, and that other encodings such as base64 would not be parsed to check links?

Just raising an issue if it actually makes sense to parse these URI values for checking. I can understand that it may still be useful with `--dump` to collect such URIs from a document, and that one can opt out with `--exclude 'data:.*'` :+1: 

Repeating the same content/URI that was just listed for the error also seems redundant?

---

## Reproduction

I tested `lychee` against [this 404 page](https://dms-doc-previews.netlify.app/) ([Source on Github](https://github.com/docker-mailserver/docker-mailserver/blob/master/docs/overrides/404.html)). That is intended as a 404 page (with response) however, so I wasn't sure if `lychee` would parse it (it does for inputs, which is fine). Below I tested with a local copy and webserver at `localhost` instead:

```console
$ lychee localhost/index.html

🔍 6 Total ✅ 5 OK 🚫 0 Errors 💤 1 Excluded

# Verbose:
$ lychee -v localhost/index.html

? [IGNORED] data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 147 40' /%3e | Unsupported The given URI is invalid: data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 147 40' /%3e
[...]

PR #824, opened 2022-11-11
Add different verbosity levels

mre:
More granular verbosity levels have been asked
for repeatedly.
To enable that we're moving to [env_logger] and [clap-verbosity-flag]
to provide more flexible verbosity settings.

Also tackles #661, #709.
Lays the groundwork for tackling #268 and #704.

https://github.com/rust-cli/env_logger
https://github.com/clap-rs/clap-verbosity-flag

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1096,6 +1096,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "env_logger"
+version = "0.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a12e6657c4c97ebab115a42dcee77225f7f482cdd841cf7088c657a42e9e00e7"
+dependencies = [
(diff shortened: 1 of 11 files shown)


mre:
If someone finds the time I'd love to get a review to avoid regressions.

tooomm:
Is there some brief explanation somewhere? Which information do the various level add (v) or remove (q)?

mre:
At the moment it's pretty basic:

* Verbose mode (`-v`) prints `data` URLs and false-positive URLs (e.g. example.com, which is an IANA reserved domain).
* Normal mode (without any `-v`) skips those.

That's it.
We could add something to the docs, but I don't know if it's worth it. Suggestions welcome.

tooomm:
Hmm, but what will `-vv` print then? Or `-vvv`?

PR #825, opened 2022-11-11
Add support for dumping inputs

mre:
Dumps all detected inputs to stdout without checking them
This is useful for debugging purposes.

The implementation is suboptimal at the moment.
- The collected inputs get read from disk before being written to stdout.
- If inputs don't contain any links, they are ignored.

It would be better to just stream the inputs and not load them into memory.
This would require some refactoring in the `lychee-lib` crate,
so it is left as a TODO for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,9 @@ OPTIONS:
             Don't perform any link checking. Instead, dump all the links extracted from inputs that
             would be checked
 
+        --dump-inputs
+            Don't perform any link checking. Instead, dump all the inputs that would be checked
+
     -E, --exclude-all-private
             Exclude all private IPs from checking.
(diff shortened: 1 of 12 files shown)


mre:
Doesn't work as expected. It omits inputs in the output and I don't know why. Maybe another pair of 👀 would help.

mre:
Closing this PR as I don't have any ideas on how to fix the issues with missing inputs at this point. We can try again after some refactoring.

lebensterben:
instead of using a CLI option to dump input, we should instead log *every* unique link for verbosity level >= debug. (in other word we need an overhaul for the entire crate and adding a bunch of logging with help of https://docs.rs/env_logger/latest/env_logger/)

PR #847, opened 2022-11-28
Recursively skip verbatim elements

mre:
This handles cases like

```html
<code>
    <pre>
        <span>https://example.org</span>
    </pre>
</code>
```

where a link was embedded in a verbatim block, but embedded in a non-verbatim element (`<span>`).
This was correctly handled in html5ever, but not html5gum (which is our default HTML parser in lychee)

Fixes https://github.com/lycheeverse/lychee/issues/259#issuecomment-1328292006.

@untitaker, is there a better way to handle nested tags in html5gum or would you have resolved the issue in a similar way?

### DIFF
--- a/lychee-lib/src/extract/html5ever.rs
+++ b/lychee-lib/src/extract/html5ever.rs
@@ -214,6 +214,27 @@ mod tests {
     }
 
     #[test]
+    fn test_include_verbatim_recursive() {
+        const HTML_INPUT: &str = r#"
+        <a href="https://example.com/">valid link</a>
+        <code>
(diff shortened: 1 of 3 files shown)


mre:
Oh I see. Would you have the time to create a pull request to change it to the new structure? As the author of the library it's probably way easier to do for you than for me. 😅 Maybe you want to ignore what I did in this PR and base it off of the last version of `master` just before this commit?

mre:
Probably not the cleanest implementation, but yeah. Thanks for the feedback. 👍

PR #898, opened 2023-01-03
Add missing test- and linting flags

mre:
Noticed that we don't test all targets in CI.


lebensterben:
`--all-features` flag is potentially problematic if some features are not supposed to be enabled at the same time.

PR #905, opened 2023-01-05
chore(ci)!: update github actions

lebensterben:
Most changes are merely on version changes, log messages and comments, and structuring.

The only significant change is that `actions/upload-release-asset@v1` is deprecated and its replacement, i.d. `softprops/upload-release-asset@v1` has rather different interface.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
           override: true
 
       - name: Run cargo check
-        uses: actions-rs/cargo@v2
+        uses: actions-rs/cargo@v1
         with:
           command: check
 
@@ -48,7 +48,7 @@ jobs:
           override: true
 
       - name: Run cargo test
-        uses: actions-rs/cargo@v2
+        uses: actions-rs/cargo@v1
         with:
           command: test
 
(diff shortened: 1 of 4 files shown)


mre:
Thanks, looks good.
CI seems to be failing because `cargo@v2` doesn't exist. Guess you have to change that back to v1.

lebensterben:
caught a bug in actions-rs/toolchain...
it only recognizes rust-toolchain, not rust-toolchain.toml

lebensterben:
seems like action-rs/toolchain is not actively maintained. I'm going to add back the explicit toolchain argument for the moment.

lebensterben:
is there anyway to test the "Release" action? That's where the major changes occured.

mre:
I don't know of a good way to test the release action. (Which doesn't mean that there is no good way.)
The only way I can think of is to comment out `upload-release-asset` and [replace it with a `tmate` session](https://github.com/mxschmitt/action-tmate).
Speaking of which, where is the source code for `softprops/upload-release-asset`? I thought it would be at https://github.com/softprops/upload-release-asset, but it isn't. Also can't find it on the marketplace.

lebensterben:
sorry, it should be https://github.com/softprops/action-gh-release

mre:
@lebensterben can you rebase on top of `master`? I might some CI improvements to speed up the builds.

lebensterben:
I spotted some errors when rebasing. I will address that first

lebensterben:
I will divide this to smaller PRs

mre:
Thanks so much.

PR #913, opened 2023-01-12
chore!: improve client and remap modules

lebensterben:
`lychee_lib::client`:

- Improved documentation.
- Added an log message in `ClientBuilder::client()` when provied user-agent
  overrides the one defined in provied custom header.
- Removed unnecessary error handling in `Client::check()` when setting HTTPS
  scheme because all failure cases should occur when checking this URL the first
  time already.
- Removed unnecessary error handling in `Client::remap()` since
  `lychee-lib::remap::Remaps::remap()` doesn't returns a `Result` anymore.
- Fixed potential integer overflow in `Client::check_website()` when the wait
  time between retries doubles, by using `std::time::Duration::saturating_mul`
  instead.
- Renamed `invalid()` to `validate_url()`.

`lychee_lib::remap`:

- Improved documentation, in particular, clarified (in the comment) that it's
  URLs not URIs being remapped.
- Changed `Remaps::remap()` so it takes `&mut Url` instead of `Uri` as its
  argument, and doesn't return a `Result` as a result.
    - Using `Url` instead of `Uri` because it aligns with the concept of
      remapping locations rather than identifiers.
    - Mutating the URL directly instead of returning a new one for it's more
      straightforward.
    - There is no error handling because we don't convert from URL to URI
      anymore. Furthermore, this always succeed in the first place so we never
      needed error handling.
- Added implementation of `IntoIterator` for `&'a Remaps` and convenience method
  of `Remaps::iter`. (Their mutable or moving counterparts are deliberately
  avoided because we don't want library users to modify all consume the
  remapping rules after its instantiation.)

`lychee_lib::error`:

- Renamed `ErrorKind::InvalidUriRemap` to `InvalidUrlRemap` and improved
  its error message.

Changes to other modules are minor and only serves to accompany aforementioned
changes.

### DIFF
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -91,7 +91,7 @@ pub struct ClientBuilder {
     /// performance issues.
     ///
     /// Furthermore rules are executed sequentially and multiple mappings for
-    /// a same URI are allowed, so it is up to the library user's discretion to
+    /// the same URI are allowed, so it is up to the library user's discretion to
     /// make sure rules don't conflict with each other.
     remaps: Option<Remaps>,
 
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -284,7 +284,7 @@ impl ClientBuilder {
     /// Returns an `Err` if:
     /// - The user-agent contains characters other than ASCII 32-127.
     /// - The reqwest client cannot be instantiated. This occurs if a TLS
-    ///   backend cannot be initialized or if the resolver fails to load system
+    ///   backend cannot be initialized or the resolver fails to load the system
     ///   configuration. See [here].
     /// - The Github client cannot be created. Since the implementation also
     ///   uses reqwest under the hood, this error in the same circumstances as
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -282,7 +282,7 @@ impl ClientBuilder {
     /// # Errors
     ///
     /// Returns an `Err` if:
-    /// - The user-agent is contains charachters other than ASCII 32-127.
+    /// - The user-agent contains characters other than ASCII 32-127.
     /// - The reqwest client cannot be instantiated. This occurs if a TLS
     ///   backend cannot be initialized or if the resolver fails to load system
     ///   configuration. See [here].
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -450,9 +450,12 @@ impl Client {
             match self.check_website(uri).await {
                 Status::Ok(code) if self.require_https && uri.scheme() == "http" => {
                     let mut https_uri = uri.clone();
-                    // here `uri` must be valid, otherwise `check_website` won't
-                    // return `Ok`, thus `set_scheme` won't fail
-                    https_uri.set_scheme("https").unwrap();
+                    {
+                        // here `uri` must be valid, otherwise `check_website` won't
+                        // return `Ok`, thus `set_scheme` won't fail
+                        debug_assert!(!https_uri.url.cannot_be_a_base());
+                        https_uri.set_scheme("https").unwrap();
+                    }
                     if self.check_website(&https_uri).await.is_success() {
                         Status::Error(ErrorKind::InsecureURL(https_uri))
                     } else {
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -98,7 +98,7 @@ pub struct ClientBuilder {
     /// Links matching this set of regular expressions are **always** checked.
     ///
     /// This has higher precedence over [`ClientBuilder::excludes`], **but**
-    /// has lower precedence comparing to any other `exclude_` fields or
+    /// has lower precedence compared to any other `exclude_` fields or
     /// [`ClientBuilder::schemes`] below.
     includes: Option<RegexSet>,
 
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -287,7 +287,7 @@ impl ClientBuilder {
     ///   backend cannot be initialized or the resolver fails to load the system
     ///   configuration. See [here].
     /// - The Github client cannot be created. Since the implementation also
-    ///   uses reqwest under the hood, this error in the same circumstances as
+    ///   uses reqwest under the hood, this errors in the same circumstances as
     ///   the last one.
     ///
     /// [here]: https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#errors
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -212,7 +212,7 @@ pub struct ClientBuilder {
     /// # Warning
     ///
     /// You should think very carefully before allowing invalid SSL
-    /// certificates. It will accepts any certificate for any site to be trusted
+    /// certificates. It will accept any certificate for any site to be trusted
     /// including expired certificates. This introduces significant
     /// vulnerabilities, and should only be used as a last resort.
     // TODO: We should add a warning message in CLI. (Lucius, Jan 2023)
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -39,7 +39,7 @@ use crate::{
 pub const DEFAULT_MAX_REDIRECTS: usize = 5;
 /// Default number of retries before a request is deemed as failed, 3.
 pub const DEFAULT_MAX_RETRIES: u64 = 3;
-/// Default wait time in seconds between requests, 1.
+/// Default wait time in seconds between retries, 1.
 pub const DEFAULT_RETRY_WAIT_TIME_SECS: usize = 1;
 /// Default timeout in seconds before a request is deemed as failed, 20.
 pub const DEFAULT_TIMEOUT_SECS: usize = 20;
@@ -47,10 +47,12 @@ pub const DEFAULT_TIMEOUT_SECS: usize = 20;
 pub const DEFAULT_USER_AGENT: &str = concat!("lychee/", env!("CARGO_PKG_VERSION"));
 
 // Constants currently not configurable by the user.
-/// A timeout for only the connect phase of a Client.
+/// A timeout for only the connect phase of a [`Client`].
 const CONNECT_TIMEOUT: u64 = 10;
-/// TCP keepalive
-/// See <https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html> for more info
+/// TCP keepalive.
+///
+/// See <https://tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html> for more
+/// infomation.
 const TCP_KEEPALIVE: u64 = 60;
 
 /// Builder for [`Client`].
@@ -61,7 +63,8 @@ const TCP_KEEPALIVE: u64 = 60;
 #[builder(builder_method_doc = "
 Create a builder for building `ClientBuilder`.
 
-On the builder call, call methods with same name as its fields to set their values.
+On the builder call, call methods with the same names as its fields to set their
+ values.
 
 Finally, call `.build()` to create the instance of `ClientBuilder`.
 ")]
@@ -70,28 +73,32 @@ pub struct ClientBuilder {
     ///
     /// This allows much more request before getting rate-limited.
     ///
-    /// ## Rate-limiting Defaults
+    /// # Rate-limiting Defaults
     ///
     /// As of Feb 2022, it's 60 per hour without GitHub token v.s.
     /// 5000 per hour with token.
     github_token: Option<SecretString>,
 
-    /// Remap URIs matching a pattern to a different URI
+    /// Remap URIs matching a pattern to a different URI.
     ///
     /// This makes it possible to remap any HTTP/HTTPS endpoint to a different
-    /// HTTP/HTTPS endpoint. This feature could also be used to proxy
+    /// HTTP/HTTPS one. This feature could also be used to proxy
     /// certain requests.
     ///
+    /// # Usage Notes
+    ///
     /// Use with caution because a large set of remapping rules may cause
-    /// performance issues. Furthermore rules are executed in order and multiple
-    /// mappings for the same URI are allowed, so there is no guarantee that the
-    /// rules may not conflict with each other.
+    /// performance issues.
+    ///
+    /// Furthermore rules are executed sequentially and multiple mappings for
+    /// a same URI are allowed, so it is up to the library user's discretion to
+    /// make sure rules don't conflict with each other.
     remaps: Option<Remaps>,
 
     /// Links matching this set of regular expressions are **always** checked.
     ///
     /// This has higher precedence over [`ClientBuilder::excludes`], **but**
-    /// has lower precedence over any other `exclude_` fields or
+    /// has lower precedence comparing to any other `exclude_` fields or
     /// [`ClientBuilder::schemes`] below.
     includes: Option<RegexSet>,
 
@@ -109,7 +116,7 @@ pub struct ClientBuilder {
 
     /// When `true`, exclude private IP addresses.
     ///
-    /// ## IPv4
+    /// # IPv4
     ///
     /// The private address ranges are defined in [IETF RFC 1918] and include:
     ///
@@ -117,16 +124,17 @@ pub struct ClientBuilder {
     ///  - `172.16.0.0/12`
     ///  - `192.168.0.0/16`
     ///
-    /// ## IPv6
+    /// # IPv6
     ///
     /// The address is a unique local address (`fc00::/7`).
     ///
     /// This property is defined in [IETF RFC 4193].
     ///
-    /// ## Note
+    /// # Note
     ///
-    /// Unicast site-local network was defined in [IETF RFC 4291], but was fully deprecated in
-    /// [IETF RFC 3879]. So it is **NOT** considered as private on this purpose.
+    /// Unicast site-local network was defined in [IETF RFC 4291], but was fully
+    /// deprecated in [IETF RFC 3879]. So it is **NOT** considered as private on
+    /// this purpose.
     ///
     /// [IETF RFC 1918]: https://tools.ietf.org/html/rfc1918
     /// [IETF RFC 4193]: https://tools.ietf.org/html/rfc4193
@@ -136,17 +144,19 @@ pub struct ClientBuilder {
 
     /// When `true`, exclude link-local IPs.
     ///
-    /// ## IPv4
+    /// # IPv4
     ///
     /// The address is `169.254.0.0/16`.
     ///
     /// This property is defined by [IETF RFC 3927].
     ///
-    /// ## IPv6
+    /// # IPv6
     ///
-    /// The address is a unicast address with link-local scope,  as defined in [RFC 4291].
+    /// The address is a unicast address with link-local scope,  as defined in
+    /// [RFC 4291].
     ///
-    /// A unicast address has link-local scope if it has the prefix `fe80::/10`, as per [RFC 4291 section 2.4].
+    /// A unicast address has link-local scope if it has the prefix `fe80::/10`,
+    /// as per [RFC 4291 section 2.4].
     ///
     /// [IETF RFC 3927]: https://tools.ietf.org/html/rfc3927
     /// [RFC 4291]: https://tools.ietf.org/html/rfc4291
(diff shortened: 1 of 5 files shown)


lychee-lib/src/client.rs
```diff
@@ -70,28 +73,32 @@ pub struct ClientBuilder {
...
+    /// # Usage Notes
+    ///
     /// Use with caution because a large set of remapping rules may cause
-    /// performance issues. Furthermore rules are executed in order and multiple
-    /// mappings for the same URI are allowed, so there is no guarantee that the
-    /// rules may not conflict with each other.
+    /// performance issues.
+    ///
+    /// Furthermore rules are executed sequentially and multiple mappings for
+    /// a same URI are allowed, so it is up to the library user's discretion to
```
mre, line 94:
```suggestion
    /// the same URI are allowed, so it is up to the library user's discretion to
```

lychee-lib/src/client.rs
```diff
@@ -70,28 +73,32 @@ pub struct ClientBuilder {
...
+    /// make sure rules don't conflict with each other.
     remaps: Option<Remaps>,
 
     /// Links matching this set of regular expressions are **always** checked.
     ///
     /// This has higher precedence over [`ClientBuilder::excludes`], **but**
-    /// has lower precedence over any other `exclude_` fields or
+    /// has lower precedence comparing to any other `exclude_` fields or
```
mre, line 101:
```suggestion
    /// has lower precedence compared to any other `exclude_` fields or
```

lychee-lib/src/client.rs
```diff
@@ -173,39 +184,50 @@ pub struct ClientBuilder {
...
-    /// ## Warning
+    /// # Warning
     ///
-    /// You should think very carefully before using this method. If
-    /// invalid certificates are trusted, any certificate for any site
-    /// will be trusted for use. This includes expired certificates. This
-    /// introduces significant vulnerabilities, and should only be used
-    /// as a last resort.
+    /// You should think very carefully before allowing invalid SSL
+    /// certificates. It will accepts any certificate for any site to be trusted
```
mre, line 215:
```suggestion
    /// certificates. It will accept any certificate for any site to be trusted
```

lychee-lib/src/client.rs
```diff
@@ -173,39 +184,50 @@ pub struct ClientBuilder {
...
+    /// including expired certificates. This introduces significant
+    /// vulnerabilities, and should only be used as a last resort.
+    // TODO: We should add a warning message in CLI. (Lucius, Jan 2023)
```
mre, line 218: Not a big fan of adding usernames to `TODO`s.

lychee-lib/src/client.rs
```diff
@@ -220,17 +242,29 @@ pub struct ClientBuilder {
...
+    /// # Notes
+    ///
+    /// For each request, the wait time increases using an exponential backoff
+    /// mechanism. For example, if the value is 1 second, then it waits for
+    /// 2 ^ (N-1) seconds before the N-th retry.
     ///
-    /// The wait time will increase using an exponential backoff mechanism
-    retry_wait_time: Option<Duration>,
+    /// This prevents spending too much system resources on slow responders and
+    /// prioritize on other requests.
```
mre, line 259:
```suggestion
    /// prioritizes other requests.
```

lychee-lib/src/client.rs
```diff
@@ -248,25 +282,31 @@ impl ClientBuilder {
     /// # Errors
     ///
     /// Returns an `Err` if:
-    /// - The user-agent is invalid.
-    /// - The request client cannot be created.
-    ///   See [here](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#errors).
-    /// - The Github client cannot be created.
+    /// - The user-agent is contains charachters other than ASCII 32-127.
```
mre, line 285:
```suggestion
    /// - The user-agent contains characters other than ASCII 32-127.
```

lychee-lib/src/client.rs
```diff
@@ -248,25 +282,31 @@ impl ClientBuilder {
...
+    /// - The reqwest client cannot be instantiated. This occurs if a TLS
+    ///   backend cannot be initialized or if the resolver fails to load system
```
mre, line 287:
```suggestion
    ///   backend cannot be initialized or if the resolver fails to load the system
```

lychee-lib/src/client.rs
```diff
@@ -248,25 +282,31 @@ impl ClientBuilder {
...
+    ///   configuration. See [here].
+    /// - The Github client cannot be created. Since the implementation also
+    ///   uses reqwest under the hood, this error in the same circumstances as
```
mre, line 290:
```suggestion
    ///   uses reqwest under the hood, this errors in the same circumstances as
```

lychee-lib/src/client.rs
lebensterben: it helps to track who added the idea without the need to use git blame.

lychee-lib/src/client.rs
```diff
@@ -374,38 +414,42 @@ pub struct Client {
...
-            match self.check_website(&uri).await {
+            match self.check_website(uri).await {
                 Status::Ok(code) if self.require_https && uri.scheme() == "http" => {
                     let mut https_uri = uri.clone();
-                    https_uri
-                        .set_scheme("https")
-                        .map_err(|_| ErrorKind::InvalidURI(uri.clone()))?;
+                    // here `uri` must be valid, otherwise `check_website` won't
+                    // return `Ok`, thus `set_scheme` won't fail
+                    https_uri.set_scheme("https").unwrap();
```
mre, line 452: I'd prefer if we kept the check. While it's true that the `uri` must be valid, this could change if we refactor the code. The HTTPS rewrite code might move into a separate function or even module and there's no guarantee that `set_scheme` will always return an `Ok` in that case.

lebensterben: I think if we can add the error handling back when it's refactored. The comment here serves as the reminder.

lychee-lib/src/client.rs
```diff
@@ -456,21 +496,22 @@ impl Client {
...
         let mut status = self.check_default(uri).await;
         while retries < self.max_retries {
             if status.is_success() {
                 return status;
             }
-            sleep(wait).await;
+            sleep(wait_time).await;
             retries += 1;
-            wait *= 2;
+            wait_time = wait_time.saturating_mul(2);
```
mre, line 508: Very nice

lychee-lib/src/client.rs
```diff
@@ -456,21 +496,22 @@ impl Client {
...
             status = self.check_default(uri).await;
         }
 
         // Pull out the heavy machinery in case of a failed normal request.
         // This could be a GitHub URL and we ran into the rate limiter.
+        // TODO: We should first try to parse the URI as GitHub URI first (Lucius, Jan 2023)
```
mre, line 514: Why is that? We should mention the reason here.
It's certainly a common use-case to check Github URLs, but only because many people use lychee as a Github action. Then again the additional parsing is probably negligible, so feel free to move it up.

lychee-lib/src/client.rs
```diff
@@ -494,10 +535,7 @@ impl Client {
     /// A better approach would be to download the file through the API or
     /// clone the repo, but we chose the pragmatic approach.
     async fn check_github(&self, uri: GithubUri) -> Status {
-        let client = match &self.github_client {
-            Some(client) => client,
-            None => return ErrorKind::MissingGitHubToken.into(),
-        };
+        let Some(client) = &self.github_client else { return ErrorKind::MissingGitHubToken.into() };
```
mre, line 538: Oh, I didn't know that this pattern is valid.

lychee-lib/src/client.rs
```diff
@@ -633,9 +673,9 @@ mod tests {
 
         assert!(res.status().is_failure());
 
-        // on slow connections, this might take a bit longer than nominal backed-off timeout (7 secs)
-        assert!(end.as_secs() >= 7);
-        assert!(end.as_secs() <= 8);
+        // on slow connections, this might take a bit longer than nominal
+        // backed-off timeout (7 secs)
+        assert!((7..=8).contains(&end.as_secs()));
```
mre, line 678: cool

lychee-lib/src/remap.rs
```diff
@@ -1,20 +1,39 @@
+//! Remapping rules which allow to map URLs matching a pattern to a different
+//! URL.
+//!
+//! # Notes
+//! Use in moderation as there are no sanity or performance guarantees.
+//!
+//! - There is no constraints on remapping rules upon instantiation or during
```
mre, line 7:
```suggestion
//! - There are no constraints on remapping rules upon instantiation or during
```

lychee-lib/src/remap.rs
```diff
@@ -1,20 +1,39 @@
...
+//!   remapping. In particular, rules are checked sequentially so later rules
+//!   might contradict with earlier ones if they both match a URL.
+//! - A large rule set has a performance impact because the client needs to
+//!   match every link against all rules.
+
+// Notes on terminology:
+// The major difference between URI(Uniform Resource Identifier) and
```
mre, line 14:
```suggestion
// The major difference between URI (Uniform Resource Identifier) and
```

lychee-lib/src/remap.rs
```diff
@@ -1,20 +1,39 @@
...
+// URL(Uniform Resource Locator) is that the former is an indentifier for
```
mre, line 15:
```suggestion
// URL (Uniform Resource Locator) is that the former is an indentifier for
```

lychee-lib/src/remap.rs
```diff
@@ -1,20 +1,39 @@
...
+// resources and the later is a locator.
```
mre, line 16:
```suggestion
// resources and the latter is a locator.
```

lychee-lib/src/remap.rs
```diff
@@ -1,20 +1,39 @@
...
+// We are not interested in differentiate resources by names and the purpose of
```
mre, line 17:
```suggestion
// We are not interested in differentiating resources by names and the purpose of
```

lychee-lib/src/remap.rs
```diff
@@ -1,20 +1,39 @@
...
 
-use crate::{ErrorKind, Result};
 use regex::Regex;
 use reqwest::Url;
 
-use crate::Uri;
+use crate::ErrorKind;
 
-/// Remaps allow mapping from a URI pattern to a specified URI
+/// Rules that remaps URL matching patterns.
```
mre, line 29:
```suggestion
/// Rules that remap matching URL patterns.
```

lychee-lib/src/remap.rs
```diff
@@ -1,20 +1,39 @@
...
-/// - Testing URIs behind a proxy
+/// Some use-cases are:
+/// - Testing URLs prior to production deployment.
+/// - Testing URLs behind a proxy.
 ///
-/// Be careful when using this feature because checking every link against a
-/// large set of regular expressions has a performance impact. Also there are no
-/// constraints on the URI mapping, so the rules might contradict each other.
+/// # Notes
+/// See module level documentation of usage notes.
```
mre, line 36:
```suggestion
/// See module-level documentation of usage notes.
```

lychee-lib/src/remap.rs
```diff
@@ -64,13 +83,24 @@ impl Index<usize> for Remaps {
 impl TryFrom<&[String]> for Remaps {
     type Error = ErrorKind;
 
+    /// Try to convert a slice of `String`s to remapping rules.
+    ///
+    /// Each string should contains a Regex pattern and a URL, separated by
```
mre, line 88:
```suggestion
    /// Each string should contain a Regex pattern and a URL, separated by
```

lychee-lib/src/remap.rs
```diff
@@ -83,44 +113,56 @@ impl TryFrom<&[String]> for Remaps {
     }
 }
 
+// implementation for mutable iterator and moving iterator are deliberately
```
mre, line 116:
```suggestion
// Implementation for mutable iterator and moving iterator are deliberately
```

lychee-lib/src/remap.rs
lebensterben: https://github.com/search?q=repo%3Arust-lang%2Frust+%22module+level%22&type=code

they are two words. there are plenty of "module level documentation" everywhere.

lychee-lib/src/client.rs
lebensterben: ideally we can treat GitHub urls as others (suppose there is no rate limit).

we need octocrab client to deal with GitHub urls only when we hit the rate limit. but in this case the rate limit is not going away at time soon, so any subsequent GitHub urls always results in at least two checks.

lychee-lib/src/client.rs
lebensterben: this is very new.

lychee-lib/src/remap.rs
lebensterben: it should be singular case  "is no constraint"

lebensterben, line 7:
```suggestion
//! - There is no constraint on remapping rules upon instantiation or during
```

lebensterben:
Updated

lebensterben:
the latest commit should fix it.

lychee-lib/src/client.rs
mre: Admittedly, it's a rather minor thing but I'd still prefer if we could remove it.   
`git blame` works well and the usernames in comments cause some mental overhead with little benefit, at least to me. I also don't know of any bigger Rust project that uses them. Besides that, it shouldn't matter who added a `TODO` as long as it's relevant to the section of the code.  We should track ideas in Github discussions and issues in Github issues.

lychee-lib/src/client.rs
mre: So you mean the original request to a Github URL using `reqwest` also counts into the Github rate limit, which is why it makes no sense to check with `reqwest` first and only then use `octocrab`, right? I guess I understand now.
Yeah you might be right. But in that case, why not move the Github URL check up in this pull request? Wouldn't that immediately resolve the `TODO`? 🤔

lychee-lib/src/client.rs
mre: On the other hand, `map_err` has very little (probably zero?) overhead if it's not triggered, so why not just keep it? Then the potential error condition stays a part of the code and gets checked by the compiler, which makes refactoring less risky. If we didn't have our own `ErrorKind` for it, I'd probably opt for an `expect` with an explanation, but in our case the error variant is already there and was used before, so why not just keep it? There are quite a few error conditions in the [`set_scheme` code](https://docs.rs/url/latest/src/url/lib.rs.html#2242-2291) that are specific to the `scheme` as well and I don't think we covered that with the initial `uri` parsing, so there are cases where this code could fail where the normal parsing would succeed.

lychee-lib/src/client.rs
lebensterben: https://github.com/rust-lang/rust/blob/4a04f252f91b2ccef6fc16817b3e7a3aeda703b6/compiler/rustc_codegen_gcc/src/lib.rs#L2-L6

this is a well known convention and it's even in our compiler.

lychee-lib/src/client.rs
lebensterben: > There are quite a few error conditions in the set_scheme code that are specific to the scheme as well and I don't think we covered that with the initial uri parsing, so there are cases where this code could fail where the normal parsing would succeed.

I've checked all cases and none of that will happen. it's not possible to fail at this point.

lychee-lib/src/client.rs
mre: Still not convinced (for the reasons above), but I'm fine with keeping it in.

lychee-lib/src/client.rs
mre: Guess we can also do that later at your preference.

lychee-lib/src/client.rs
mre, line 452:
```suggestion
                    https_uri.set_scheme("https").map_err(|_| ErrorKind::InvalidURI(uri.clone()))?;
```

lychee-lib/src/client.rs
lebensterben: > So you mean the original request to a Github URL using reqwest also counts into the Github rate limit, which is why it makes no sense to check with reqwest first and only then use octocrab, right? I guess I understand now.

Exactly.

> why not move the Github URL check up in this pull request?

The initial intent of this PR is to improve ONLY the documentation.
Specifically, we have not explained the error scenarios very well before.

The by-product is, during this process I found some errors never happens.

lebensterben:
Below is a detailed analysis why calling `set_scheme` won't cause any error.

---

https://docs.rs/url/latest/src/url/lib.rs.html#2244

This fails if the new scheme is invalid. `HTTPS` is certainly valid.

---

https://docs.rs/url/latest/src/url/lib.rs.html#2248-2253

The new scheme is `HTTPS` and the old one is `HTTP`, which are both [`SchemeType::SpecialNotFile`](https://docs.rs/url/latest/src/url/parser.rs.html#148).

So none of these three conditions is satisfied and so it doesn't fail here.

---

https://docs.rs/url/latest/src/url/lib.rs.html#2258

Here `remaining` is the part of the provided new scheme stripped of any `[[:alnum:]+-.]` and optionally a `:`. We supplied "https" as the new scheme and so the `remaining` is certainly empty.

Meanwhile, our original link is special but it must contain a host. Otherwise the first `check_website` on the `http` link would not succeed. (Note that `check_website` only takes full URL not relative URL`.)

Therefore these two conditionals are never true and it won't fail here.

---

In summary it never fails.

lebensterben:
Since URL specification is not likely to change the parser is not likely to have breaking change as well.

The only possible cases that `set_scheme` would fail are:

- `set_scheme` was called before the success of `check_website` on the original link. But this is logically unsound. Because if a user provided a HTTP url and it fails, we should immediately report it as a failure without needing to check HTTPS url.
- `check_website` accepts relative URLs (and therefore also takes a base URL). This is possible so I'm willing to add additional notes for this case as a notice/warning for future development.

What's your thought?

mre:
> What's your thought?

Your arguments why this can't fail are all valid. Thanks for doing such a deep analysis of the possible error cases. I'm convinced that the current implementation works and that the `map_err` is unnecessary.

My concern however was not about the validity of the code but with the future implications for this project. As I said, if we refactor the code and forget to add back the check when it becomes necessary, we introduce a bug. Troubleshooting will get harder and it's unnecessary because we already _have_ error handling in place which tells us what went wrong.

What's your concern with just keeping `map_err`? Are you afraid that it will have unnecessary overhead? If so, we should do a benchmark to check. Or don't you like the fact that it's unnecessary right now and an `unwrap` would do? In that case I'd argue that it's just another safeguard and that we want all the checks we can possibly get from the compiler (especially if they are free). In any case I fail to see why we should remove a check that's already in place and replace it with documentation on why it cannot fail instead. What's the gain?

Sorry for being so direct on this matter, but I feel like this is a fundamental difference in perspective and I feel the urge to set some guidelines for future development. If I let this slide, I'm setting a precedent for future changes of this nature, and that's not the direction I want to take the project.

In summary, I would very much like to merge the changes in this pull request except for the `unwrap` unless there is hard evidence that it will have a dramatic performance impact or otherwise has any big advantage that I am unable to see right now.

lebensterben:
> Are you afraid that it will have unnecessary overhead? 

Nope. As I commented earlier, the initial motivation of this PR is just to improve the documentation.

Specifically, I find it's not very clear what are the error scenarios. Consider the `# Errors` section of `Client::check()` before this PR:

```rust
    /// # Errors
    ///
    /// This returns an `Err` if
    /// - `request` is invalid.
    /// - The URI of the request is invalid.
    /// - Encrypted connection for a HTTP URL is available but unused.
    ///   (Only checked when `Client::require_https` is `true`.)
    /// Returns an `Err` if `request` does not represent a valid URI.
```

The first and second cases are not very clear. What do they mean by invalid?
So I checked the source of the errors, including the upstream crates, and summarized the error scenarios as

```rust
    /// # Errors
    ///
    /// Returns an `Err` if:
    /// - `request` does not represent a valid URI.
    /// - Encrypted connection for a HTTP URL is available but unused.  (Only
    ///   checked when `Client::require_https` is `true`.)
```

The third one is unchanged.

The first point translates to: The argument `request` must represent a valid URI. Or put it in a more verbose way, it must be parsable as a valid URI according to <https://www.w3.org/Addressing/URL/uri-spec.html>.

But then I found the second one doesn't make any sense as I explained [above](https://github.com/lycheeverse/lychee/pull/913#issuecomment-1382502809).

If we leave the `ErrorKind::InvalidURI` as is, how are we going to document this error scenario that is not going to happen at all?

lebensterben:
How about
```rust
{
    debug_assert!(!https_uri.url.cannot_be_a_base());
    // here `uri` must be valid, otherwise `check_website` won't
    // return `Ok`, thus `set_scheme` won't fail
    https_uri.set_scheme("https").unwrap();
}
```

there is no way you will miss it.

mre:
Okay I finally got it. That makes sense. Yes, let's add the debug_assert, which enforces the invariant as code while still being unobtrusive and not misleading as was map_err. 👍

lebensterben:
Updated.

mre:
Nicely done! Thanks very much @lebensterben 🎉

PR #919, opened 2023-01-17
Experiment with Github actions cache

mre:
Test if it speeds up compile times


mre:
🐇 **This branch**: 5m 24s
🐢 **`master`**: 8m 46s 

So that's a 60% build time reduction with two additional lines. Not bad.
We've also seen much slower runs (e.g. 14m 20s), which makes the gap even bigger.

PR #921, opened 2023-01-17
Try nextest for CI builds

mre:


mre:
Very nice 20-40 second improvement. The output looks nicer, too. That's a keeper!

PR #922, opened 2023-01-17
Run publish-check in parallel with other jobs

mre:
I don't see the reason why publish-check has to wait for the other jobs to complete. After all, it's yet another "lint". Running it in parallel might shave of a significant amount of time from our CI builds.


mre:
Dang! It works.
So we shave off another ~3 minutes from the build!
We went from 6m 15s in `master` to 3m 1s in this branch. 🚀

PR #959, opened 2023-02-23
Update name of Arch package

polyzen:
https://archlinux.org/packages/community/x86_64/lychee/


mre:
Perfect. Nice that we could simplify the package name. Thanks!

PR #981, opened 2023-03-09
Better retry handling

mre:
Previously, lychee would blindly retry all requests,
no matter if the request error was transient or fatal.

Taking a lesson from https://github.com/TrueLayer/reqwest-middleware,
we can be more granular about the error behavior.
This PR adds their retry logic to lychee, reducing the number of
unnecessary requests significantly.

I also made some ergonomic changes to the client, which should not
affect its behavior.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1643,9 +1643,9 @@ dependencies = [
 
 [[package]]
 name = "hyper"
-version = "0.14.23"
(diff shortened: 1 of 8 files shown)


lebensterben:
btw there's a crate specialises on retry strategies: <https://github.com/Xuanwo/backon>.

Issue #986, opened 2023-03-10
Incorrect URL extraction for links in `<pre>` tags


djahandarie:
lychee has issues with links in `<pre>` tags:

```html
<!DOCTYPE html>
<html lang="en">

<head>
    <meta charset="UTF-8">
    <title>Test</title>
</head>

<body>
    <pre>
    This package uses the <a href="https://www.edrdg.org/jmdict/edict.html" target="_blank" rel="noopener">EDICT</a> and <a href="https://www.edrdg.org/wiki/index.php/KANJIDIC_Project" target="_blank" rel="noopener">KANJIDIC</a> dictionary files. These files are
    the property of the <a href="https://www.edrdg.org/" target="_blank" rel="noopener">Electronic Dictionary Research and Development Group</a>,
    and are used in conformance with the Group's <a href="https://www.edrdg.org/edrdg/licence.html" target="_blank" rel="noopener">licence</a>.
    </pre>
</body>

</html>
```

results in 

```
| * [404] [https://www.edrdg.org/jmdict/edict.html_blanknoopenerhttps://www.edrdg.org/wiki/index.php/KANJIDIC_Project_blanknoopenerhttps://www.edrdg.org/_blanknoopenerhttps://www.edrdg.org/edrdg/licence.html_blanknoopener](https://www.edrdg.org/jmdict/edict.html_blanknoopenerhttps://www.edrdg.org/wiki/index.php/KANJIDIC_Project_blanknoopenerhttps://www.edrdg.org/_blanknoopenerhttps://www.edrdg.org/edrdg/licence.html_blanknoopener) | Failed: Network error: Not Found
| 
```

You can see that the links have all been merged into one link, and contents of the attributes have been included as well, resulting in a failure.

PR #988, opened 2023-03-10
Fix nested URL extraction in verbatim elements

mre:
Skipping URLs in verbatim elements didn't take nested
elements into consideration, which were not verbatim.

For instance, the following HTML snippet would yield
`https://example.com` in non-verbatim mode, even if
it is nested inside a verbatim `<pre>` element:

```html
<pre><a href="https://example.com">link</a></pre>
```

This commit fixes the behavior for both `html5gum` and
`html5ever`.

Note that nested verbatim elements of the same kind
still are not handled correctly.

For instance,  the following HTML snippet would still yield
`https://example.com`:

```html
<pre>
  <pre></pre>
  <a href="https://example.com">link</a>
</pre>
```

The reason is that we currently only keep track of a single
verbatim element and not a stack of elements, which we
would need to unwind and resolve the situation.

Fixes https://github.com/lycheeverse/lychee/issues/986.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,9 @@ lint: ## Run linter
 
 .PHONY: test
 test: ## Run tests
-	cargo nextest run --all-targets
+	cargo nextest run --all-targets --all-features --filter-expr '!test(test_exclude_example_domains)'
+	cargo nextest run --filter-expr 'test(test_exclude_example_domains)'
+	cargo test --doc
 
 .PHONY: doc
 doc: ## Open documentation
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,12 +25,8 @@ jobs:
           toolchain: stable
       - uses: taiki-e/install-action@nextest
       - uses: Swatinem/rust-cache@v2
-      - name: Run cargo test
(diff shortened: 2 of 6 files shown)


mre:
This fixes the bug and I've added tests to avoid future regressions, but I'm not entirely happy with how the html extraction is evolving. The logic is hard to follow and there are many branches in the code. Ideally we should refactor this at some point. Thankfully this part is encapsulated in the extractor modules, so at least it doesn't spill over to the other parts of the codebase. Parsing HTML is not easy as there are many subtle edge-cases, so I wonder how much of it is accidental complexity vs essential complexity.

I will keep this open for a day in case someone wants to take a look.

Makefile
```diff
@@ -40,7 +40,9 @@ lint: ## Run linter
 
 .PHONY: test
 test: ## Run tests
-	cargo nextest run --all-targets
+	cargo nextest run --all-targets --all-features --filter-expr '!test(test_exclude_example_domains)'
+	cargo nextest run --filter-expr 'test(test_exclude_example_domains)'
+	cargo test --doc
```
mre, line 45: Moved the test calls here to keep the local dev env in sync with CI.

Issue #984, opened 2023-03-09
Write errors to stderr


thomas-zahner:
Currently some errors are written to stdout instead of stderr.

Steps to reproduce:
1. Create a file `a.md` with the following content: `My URL: https://a-very-invalid-and-non-existing-domain.org`
2. Run `lychee a.md --format json --no-progress | cat`

The output now contains not only the expected JSON, but also the error message `✗ [ERR] https://a-very-invalid-and-non-existing-domain.org/ | Failed: Network error: dns error: no record found for Query { name: Name("a-very-invalid-and-non-existing-domain.org.ost.ch."), query_type: AAAA, query_class: IN }`

Expected behaviour:

It should behave the same as with `lychee a.md --format json --no-progress -o out`. But instead of writing to the specified file it should write to stdout. This way we could pipe the output to other programs as expected. For example piping into `jq` ([jq](https://stedolan.github.io/jq/manual/)) doesn't work because of this bug: `lychee a.md --format json --no-progress | jq`

mre:
Thanks for reporting. Attempted to fix that in #990. I might have missed a few cases, so please let me know if it works for you.

mre:
## Test Command

```bash
echo 'https://a-very-invalid-and-non-existing-domain.org' | cargo run -- --format json --no-progress - | jq
```

## Before

```
parse error: Invalid numeric literal at line 1, column 4
```

## This PR

```
✗ [ERR] https://a-very-invalid-and-non-existing-domain.org/ | Failed: Network error: dns error: no record found for Query { name: Name("a-very-invalid-and-non-existing-domain.org.fritz.box."), query_type: AAAA, query_class: IN }
{
  "detailed_stats": false,
  "total": 1,
  "successful": 0,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 0,
  "excludes": 0,
  "errors": 1,
  "cached": 0,
  "success_map": {},
  "fail_map": {
    "stdin": [
      {
        "url": "https://a-very-invalid-and-non-existing-domain.org/",
        "status": "Failed: Network error"
      }
    ]
  },
  "excluded_map": {}
}
```

Note that `jq` properly handles the input, which means that `✗ [ERR] ...` is not printed to `stdout` anymore.

PR #990, opened 2023-03-11
Use standard error for error output

mre:
Fixes https://github.com/lycheeverse/lychee/issues/984

From https://doc.rust-lang.org/book/ch12-06-writing-to-stderr-instead-of-stdout.html:

> Command line programs are expected to send error messages to the standard error stream so we can still see error messages on the screen even if we redirect the standard output stream to a file. Our program is not currently well-behaved: we’re about to see that it saves the error message output to a file instead!

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -772,8 +772,8 @@ mod cli {
         // run first without cache to generate the cache file
         test_cmd
             .assert()
-            .stdout(contains(format!("[200] {}/\n", mock_server_ok.uri())))
-            .stdout(contains(format!(
+            .stderr(contains(format!("[200] {}/\n", mock_server_ok.uri())))
+            .stderr(contains(format!(
                 "[404] {}/ | Failed: Network error: Not Found\n",
                 mock_server_err.uri()
             )));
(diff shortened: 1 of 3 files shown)


mre:
@thomas-zahner fyi. Can you compile lychee from this branch and report if everything works as expected? 🙏

thomas-zahner:
@mre Yes, my use-case now works as expected, thank you for addressing it so quickly. I assume you will also take a look at the failing tests.

lychee-bin/tests/cli.rs
```diff
@@ -786,11 +786,11 @@ mod cli {
         // run again to verify cache behavior
         test_cmd
             .assert()
-            .stdout(contains(format!(
+            .stderr(contains(format!(
                 "[200] {}/ | Cached: OK (cached)\n",
                 mock_server_ok.uri()
```
mre, line 791: I would like to get a second opinion on this part:
does it make sense to output `[200]` to `stderr`?
@thomas-zahner?

mre:
@thomas-zahner, sorry that this wasn't merged yet. I have a question before I merge. Can you have a look? 🙏

lychee-bin/tests/cli.rs
thomas-zahner: Yes I think it makes sense, as these are just the status/progress report messages.
It's just the final result (JSON, MarkDown, etc.) that should be printeted to stdout, when lychee finished checking the links. (even if the result contains broken links)

This way the output can be used in much more useful ways, e.g. piping the JSON result into `jq`, writing or appending the MarkDown result to files with `>` or `>>` while the prigress is still separate.

I was also a bit unsure at some point whether this is really best-practice but this stackexchange discussion makes it quite clear: https://unix.stackexchange.com/questions/331611/do-progress-reports-logging-information-belong-on-stderr-or-stdout

mre: Perfect. Thanks a lot for your feedback and the research.

Issue #283, opened 2021-07-13
Suggest possible fixes for broken URLs to archive sites?


kaihendry:
As mentioned upon https://github.com/lycheeverse/lychee-action/issues/37#issuecomment-878248088

It would be good to have a `--fix` flag that could perhaps rewrite dead URLs into working archived URLs like that provided by

* http://web.archive.org/

mre:
This can be done with the [Wayback Machine API](https://archive.org/help/wayback_api.php) if anyone wants to give it a try.
Here is a sample request:  
https://archive.org/wayback/available?url=example.com

PR #1003, opened 2023-03-18
Wayback integration

thomas-zahner:
Closes https://github.com/lycheeverse/lychee/issues/283

## Summary

@mre This is a first working version of the wayback machine integration. Feedback is greatly appreciated.

We have added the `--suggest` flag for now, which expects the archive as value, with "wayback" as the only option for now. To test the functionality:

```bash
echo 'https://semat.org/essence-1.2' | cargo run -- --suggest wayback --format json -
```

## Questions

* Is the code and the structure ok? For now we have added the module `archive` to lychee-bin. Is the idea to add it to lychee-lib?
* When checking the archive, should an additional progress bar be shown?
* Since we might add `--fix` in the future (in my opinion not necessarily in this PR) it probably makes sense to split up the falgs as follows, do you agree?
  * `--archive wayback`
  * `--suggest` (requires `--archive`)
  * `--fix` (requires `--archive`)


mre:
From a first test this works like a charm. Thanks for looking into it! 👍 

About your questions:

**Is the code and the structure ok? For now we have added the module archive to lychee-bin. Is the idea to add it to lychee-lib?**

I will have a closer look in the next days.
Generally yes, let's keep the `archive` module in the binary. Eventually we might move it into a separate crate in the same project or even a plugin to isolate those responsibilities. The idea is that the library is only concerned about the mechanics of link checking itself.

**When checking the archive, should an additional progress bar be shown?**

Not sure. On one side I'd imagine it should be part of the normal link checking process from a user's perspective . So a dedicated progress bar might not be necessary.
On the other side I gave it a spin and there was a moment where it didn't show any progress and it felt like it was "stuck" for a second. We probably want to show some output there. Could be as simple as printing the link to the archive URL that we check. This would help with troubleshooting, too, in the end once we support multiple archive pages. A user would see which archive link was checked. And yes, it feels like that's another link check, so it would fit well into the existing progress bar.
At the end of the day we have to experiment a bit I think. Whatever is easiest for the time being I'd say. If a new progress bar is easier, go for that. Otherwise add it to the current progress bar and maybe add an `info` log for verbose mode.


**Since we might add --fix in the future (in my opinion not necessarily in this PR) it probably makes sense to split up the falgs as follows, do you agree?**
* `--archive wayback`
* `--suggest` (requires `--archive`)
* `--fix` (requires `--archive`)

From my experience, a combination of required options can be tedious for users.
I.e. as a user I'd be more happy to have a single option to enable a feature and use sane defaults for the rest.

Here's how I think about it right now:

```bash
# Automatically fix all broken links by replacing with archived versions.
# WaybackMachine is the default archive provider.
lychee --fix

# Show fixes, but don't a apply them. With `--format json` it would print the `suggestion_map`.
# With default formatting I'd guess we just log the suggestions to `stdout` and 
# show them in the summary. (We could also add a `--format diff` in the future for 
# showing the file diffs for review; not sure.)
lychee --fix --dry-run

# Overwrite the archive provider. Does nothing if `--fix` is not activated.
lychee --archive <provider> ..
```

Open for "suggestions", though (hah).

The reason why I prefer `--dry-run` is that it aligns well with other tools like `cargo publish`. In general, we should aim to introduce as little new vocabulary as possible and stick to the existing naming in other CLI tools.
On top of that, it allows us to follow the separation of concerns described in [this article](https://www.gresearch.co.uk/blog/article/in-praise-of-dry-run/): I would really like us to keep the separation between gathering the replacement links and applying the list of changes.

mre:
Code looks nice!

I suggest to use strum to simplify archive provider parsing.
This makes it future-proof for future additions. We can't easily
forget to `match` for them in `FromStr`. 
Also it avoids some boilerplate. 
Furthermore it would be nice to list of supported providers if the user
specified an invalid option (e.g. in case they made a typo) without having to consult the
documentation.

lychee-bin/src/archive/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+use serde::{Deserialize, Deserializer, Serialize};
+
+#[derive(Debug, Serialize, Eq, Hash, PartialEq)]
+pub(crate) struct Suggestion {
+    pub(crate) original: Url,
+    pub(crate) suggestion: Url,
+}
+
+#[derive(Debug, Deserialize, Default, Clone)]
+pub(crate) enum Archive {
```
lebensterben, line 15: should mark this as non_exhaustive

lychee-bin/src/archive/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        let function = match self {
+            Archive::WaybackMachine => get_wayback_link,
+        };
+
+        function(original).await
+    }
+}
+
+async fn get_wayback_link(url: &Url) -> Result<Option<Url>, Error> {
+    let mut archive_url = Url::parse("https://archive.org/wayback/available").unwrap();
```
lebensterben, line 41: can we parse this as a static variable? (when it's inside this function, we can clone the static variable and make the clone mutable)

lychee-bin/src/archive/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        .await?;
+
+    Ok(response
+        .archived_snapshots
+        .closest
+        .map(|closest| closest.url))
+}
+
+#[derive(Debug, Deserialize, Eq, PartialEq)]
+pub(crate) struct InternetArchiveResponse {
```
lebensterben, line 56: is this and next two structs used for wayback machine only?

they can be organized in a sub module.

lychee-bin/src/archive/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+where
+    D: Deserializer<'d>,
+{
+    let value: &str = Deserialize::deserialize(deserializer)?;
+    let result = value.parse::<u16>().unwrap();
+    Ok(StatusCode::from_u16(result).unwrap())
+}
+
+#[tokio::test]
+async fn wayback_suggestion() -> Result<(), Error> {
```
lebensterben, line 85: the two test functions should be wrapped in a tests module.

lychee-bin/src/archive/mod.rs
mre: Agree. @thomas-zahner, here's how you can do that: https://github.com/lycheeverse/lychee/blob/cd45f9db0745ed58e728984ce4c2d4407582dde8/lychee-lib/src/helpers/path.rs#L9-L10

lychee-bin/src/options.rs
```diff
@@ -179,6 +180,11 @@ pub(crate) struct Config {
     #[serde(default)]
     pub(crate) dump: bool,
 
+    /// Suggest link replacements for broken links, using a web archive. Only supports "wayback"
+    /// for the moment
+    #[arg(long, default_value = None)]
```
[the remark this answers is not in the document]
mre: Good point. Also made this change in https://github.com/thomas-zahner/lychee/pull/1.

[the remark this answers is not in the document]
thomas-zahner: Thank you very much for the feedback the PR. I totally agree and have merged it.
I have update the flags like you suggested.
The default value is now "wayback" for the `--archive` flag.
Because we now use a value_parser for parsing the archive option, we must provide a string for `default_value`.

I've tried the following, but it doesn't work, because `clap::builder::OsStr` is expected:

`#[arg(long, value_parser = parse_archive_provider, default_value = Archive::default().to_string())]`

I couldn't figure out how to do it, so for now I just used `default_value = "wayback"`. Do you know how this could be done?

lychee-bin/src/archive/mod.rs
thomas-zahner: Thank you for the suggestion I did:

```rust
static WAYBACK_URL: Lazy<Url> =
    Lazy::new(|| Url::parse("https://archive.org/wayback/available").unwrap());
```

thomas-zahner:
@mre I have now updated the pull request to show a second progress bar. I didn't extend the existing progress bar because it is impossible to know the size of the progress bar before having checked all URLs since we only create suggestions for the failed URLs / the fail map. If we used only one we would have to increase its size while checking, which I think is not intuitive from a user's perspective.

There are still 3 failing tests concerning config files. I'm not quite sure what the problem is because the error message is not very specific. Do you have an idea how to fix them?

After having fixed the failing tests it would be nice to wrap up the pull request. Is it ok to add the `--fix` flag in a separate pull request? Is there something you would like to change before merging?

The only thing left to improve I can think of, is to make the suggestion querying async or parallel. Currently it is done sequentially. Do you think it is necessary? If so, should we also do it with `futures::StreamExt::for_each_concurrent`?

mre:
Thanks. We already extend the existing progress bar while running because we can't know the number of links on a page either. As such it would be a huge problem, but let's keep the separate one for now.

It would be nice to query asynchronously, and I would use `for_each_concurrent` or https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html, but it can also be done in a separate pull request. Up to you.

The failing config tests might be because a new option was added and there is no default for it yet in case it's missing in the config file. You might have to add `suggest = true` to make the tests pass. If that fixes the problem you can add a default to the `merge` macro. It's somewhere in the config parser. (I'm on mobile right now, so please check.)

The only remaining thing is the name of the config option. My initial thought was `--dry-run`, which only works in combination with `--fix`. Since `--fix` isn't implemented yet, let's keep `--suggest`. It's just a heads up that we might adjust the naming in the future.

Apart from that we're good to go.

mre:
Thanks for your contribution. 😃 
`#[serde(default)]` is totally fine; no need to add the default values, I think. Changed looked fine, so I went ahead and merged them.

PR #1014, opened 2023-03-24
Status codes in maps

vilben:
This is the PR regarding to the previous discussion about the status code in the fail_map and success_map (https://github.com/lycheeverse/lychee/discussions/1013).

`cargo run -- www.example.com --format json -v` now prints a success_map, containing a status object with a text and a code field. 
The code field is not included, when the status doesn't contain a code.
```
..., 
 "success_map": {
    "http://www.example.com/": [
      {
        "url": "https://www.iana.org/domains/example",
        "status": {
          "text": "OK (200 OK)",
          "code": 200
        }
      }
    ]
  },
...
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2038,6 +2038,7 @@ dependencies = [
  "ring",
  "secrecy",
  "serde",
+ "serde_json",
  "shellexpand",
  "tempfile",
  "thiserror",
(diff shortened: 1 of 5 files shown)


mre:
Code looks good to me. Guess you just have to run `cargo fmt` once more.

mre:
Thanks for your contribution @vilben. 😃

PR #1049, opened 2023-04-21
fix(test): perform a warm up request in test_exponential_backoff

kxxt:
Perform a warm-up request to ensure the lazy regexes in `lychee-lib/src/quirks/mod.rs` are compiled. On some platforms, this can take some time(approx. 110ms), which should not be counted in the test.

Additional context: I am packaging lychee for RISC-V archlinux and this test fails.


kxxt:
The CI is failing because of some network error not related to this PR.

mre:
Nice. Thanks for the fix.

PR #1054, opened 2023-04-25
Create on-demand binaries (useful for debugging)

mre:


mre:
I can't really test the workflow-dispatch with merging, but I think it should be syntactically correct.

PR #1064, opened 2023-05-05
Add check duration (in seconds) to report

mre:
This adds a new field, `duration_secs` to the JSON report:

```json
{
  "total": 55,
  "successful": 55,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 0,
  "excludes": 0,
  "errors": 0,
  "cached": 0,
  "success_map": {},
  "fail_map": {},
  "suggestion_map": {},
  "excluded_map": {},
  "duration_secs": 1,
  "detailed_stats": false
}
```

It's helpful for troubleshooting slow runs and for measuring the performance over time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2013,6 +2013,7 @@ name = "lychee"
 version = "0.12.0"
 dependencies = [
  "anyhow",
+ "assert-json-diff",
  "assert_cmd",
  "clap 4.1.11",
  "console",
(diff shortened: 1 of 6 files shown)


lebensterben:
should we use https://docs.rs/tokio/latest/tokio/time/index.html instead?

mre:
The wayback suggestions are currently broken, but that's an unrelated issue.

PR #1079, opened 2023-05-26
test(client): make exponential_backoff better

kxxt:
This test is still flaky on riscv64 boards after #1049.

It turns out that building the client might take 59ms, which should not be counted.


mre:
thanks a lot

PR #1088, opened 2023-06-01
Remove inaccurate details in compact view

mre:
The error details in the compact report are computed incorrectly.
This leads to weird output, as shown in the screenshot.

image

Since the information is misleading, and we show a list of errors anyway, the safest way to fix this bug is to get rid of the detailed stats.

### DIFF
--- a/lychee-bin/src/formatters/stats/compact.rs
+++ b/lychee-bin/src/formatters/stats/compact.rs
@@ -1,10 +1,7 @@
-use std::{
-    collections::HashMap,
-    fmt::{self, Display},
-};
+use std::fmt::{self, Display};
 
 use crate::{
-    color::{color, BOLD_GREEN, BOLD_PINK, BOLD_YELLOW, NORMAL, PINK},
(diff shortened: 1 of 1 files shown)


lychee-bin/src/formatters/stats/compact.rs
```diff
@@ -15,27 +12,6 @@ use anyhow::Result;
 
 struct CompactResponseStats(ResponseStats);
 
-// Helper function, which prints the detailed list of errors
-pub(crate) fn print_errors(stats: &ResponseStats) -> String {
-    let mut errors = HashMap::new();
-    errors.insert(
-        "HTTP",
-        stats.errors - stats.redirects - stats.timeouts - stats.unknown,
```
mre, line 23: That's the reason for the bug. Redirects, timeouts, and unknown link statuses aren't counted as errors, but we still substract them from the total number of errors to calculate the HTTP errors. We might also just remove the HTTP errors in the output, but the counts of all individual "errors" would not sum up to the total error count and be even more confusing to the user. Safer to remove that than show inaccurate information.

Issue #594, opened 2022-04-17
Licensing issue: AGPL derivative


Seirdy:
Just noticed that Lychee uses code from https://github.com/reacherhq/check-if-email-exists, which is licensed under the copyleft GNU AGPLv3 for open source projects and a commercial license for proprietary software. Lychee therefore would need to use a GNU AGPLv3 license to remain compliant, or drop the dependency.

Just thought you should know.

lebensterben:
let's drop email capability.

Seirdy:
Honestly, email features were one of the best parts of Lychee IMO. Would it be possible to switch to AGPL, or to do something similar to projects like ffmpeg and support optional copyleft features?

lebensterben:
it's possible but requires some changes.

lychee-lib can be licensed under MIT/Apache 2.0 without email feature.

The email feature can be moved to a new crate lychee-email under AGPL 3.0 license.

Then lychee-bin can be licensed under one of these licenses depending on whether the email feature is enabled.

lebensterben:
also need to mention that, we cannot easily change the license. 

when people contributed their work to the codebase, their intent was to license their code under MIT/Apache 2.0.

you'll need everyone to agree to change the license.

lebensterben:
to prevent any license issues in the future, we should use https://lib.rs/crates/cargo-deny to check license compatibility in CI workflow.

Seirdy:
You do not need permission to relicense from MIT/Apache to AGPL. You *would* need permission to do the reverse. MIT and Apache are AGPLv3-compatible, but AGPLv3 is not MIT/Apache-compatible.

lebensterben:
@Seirdy 

Nope. that's a misunderstanding.

As for license compatibility: MIT/Apache 2.0 code can be used in AGPL 3.0 project and the reverse is not true.

But what I'm saying is that you can change the license of this project without the agreement from the authors. This is not about license compatibility.

lebensterben:
changing license is nearly impossible unless you get agreement from ALL contributors.

Seirdy:
If I were to fork an MIT project and make my fork AGPLv3, that would be allowed. I'd be re-using contributors' code that was licensed under the MIT license, because using MIT code in an AGPLv3 project is allowed.

If a project were to switch from MIT to AGPLv3, that would be allowed because using contributors' MIT code in an AGPLv3 project is allowed.

[...]

PR #1099, opened 2023-06-07
Add optional Rustls support

stefankreutz:
Hello. I would like to discuss adding optional [Rustls](https://github.com/rustls/rustls) support to Lychee. I've prepared a pull request to show that's it's mostly a question of passing feature flags down to existing dependencies. What do you think?

---

This commit adds a non-default feature flag to use Rustls instead of OpenSSL.

My personal motivation is to use Lychee on OpenBSD -current, where the `openssl` crate frequently fails to link against the unreleased system LibreSSL. Using the `vendored-openssl` feature helps with compilation, but segfaults at runtime.

The commit adds three feature flags to the library, binary, benchmark, and all examples:

- The `native-tls` feature flag toggles the `openssl` crate.
- The `rustls-tls` feature flag toggles the `rustls` crate.
- The `email-check` feature flag toggles the `check-if-email-exists` crate, which is the only existing functionality currently incompatible with Rustls.

By default, `native-tls` and `email-check` are enabled. Thus, Lychee (bin and lib) can be used as before unless default features are disabled.

To use the Rustls feature, pass `--no-default-features --features rustls` to cargo check/build/test/..., e.g.,

    $ cargo clippy --workspace --all-targets --no-default-features --features rustls-tls -- --deny warnings

Checking email addresses requires both, `native-tls` and `email-check`, to be enabled. Otherwise, email addresses are excluded.

The `email-check` feature flag is technically not necessary. I preferred it over `not(rustls-tls)` because it's clearer and it addresses the AGPL license issue #594. As far as I understand, a Lychee binary compiled without the `email-check` feature could be distributed with file-based copyleft for the MPL-licensed dependencies only. But that's out of scope here.

The benchmark shows a performance regression varying between 2% and 4.4% when using Rustls instead of OpenSSL on my machine.

PS: The `ring` crate needs to be patched on OpenBSD 7.3 and later until the new xonly patches have been upstreamed, see the `rust-ring` port.

### DIFF
--- a/lychee-bin/Cargo.toml
+++ b/lychee-bin/Cargo.toml
@@ -76,11 +76,22 @@ tracing-subscriber = { version = "0.3.17", default-features = false, features = 
 
 [features]
 #tokio-console = ["console-subscriber", "tracing-subscriber/registry"]
+
+# Compile and statically link a copy of OpenSSL.
 vendored-openssl = ["openssl-sys/vendored"]
+
+# Allow checking example domains such as example.com.
 check_example_domains = ["lychee-lib/check_example_domains"]
+
+# Enable checking email addresses. Requires the native-tls feature.
 email-check = ["lychee-lib/email-check"]
+
+# Use platform-native TLS.
 native-tls = ["lychee-lib/native-tls", "openssl-sys", "reqwest/native-tls"]
+
+# Use Rustls TLS.
 rustls-tls = ["lychee-lib/rustls-tls", "reqwest/rustls-tls-native-roots"]
+
 default = ["native-tls", "email-check"]
 
 # Unfortunately, it's not possible to automatically enable features
--- a/lychee-bin/Cargo.toml
+++ b/lychee-bin/Cargo.toml
@@ -24,10 +24,10 @@ const_format = "0.2.31"
 headers = "0.3.8"
 http = "0.2.9"
 indicatif = "0.17.5"
-openssl-sys = "0.9.88"
+openssl-sys = { version = "0.9.88", optional = true }
 pad = "0.1.6"
 regex = "1.8.4"
-reqwest = { version = "0.11.18", features = ["gzip"] }
+reqwest = { version = "0.11.18", default-features = false, features = ["gzip", "json"] }
 # Make build work on Apple Silicon.
 # See https://github.com/briansmith/ring/issues/1163
 # This is necessary for the homebrew build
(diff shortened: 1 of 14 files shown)


mre:
Couldn't take a look at the code yet, but I like the idea and I'm all for supporting more platforms. Thanks for the efforts so far.

lychee-bin/Cargo.toml
```diff
@@ -24,10 +24,10 @@ const_format = "0.2.31"
 headers = "0.3.8"
 http = "0.2.9"
 indicatif = "0.17.5"
-openssl-sys = "0.9.88"
+openssl-sys = { version = "0.9.88", optional = true }
 pad = "0.1.6"
 regex = "1.8.4"
-reqwest = { version = "0.11.18", features = ["gzip"] }
+reqwest = { version = "0.11.18", default-features = false, features = ["gzip", "json"] }
```
mre, line 30: Interesting. Was the json feature missing before? Was it somehow implicitly activated?

lychee-lib/Cargo.toml
```diff
@@ -17,17 +17,17 @@ repository = "https://github.com/lycheeverse/lychee"
...
 http = "0.2.9"
 linkify = "0.9.0"
-openssl-sys = "0.9.88"
+openssl-sys = { version = "0.9.88", optional = true }
 pulldown-cmark = "0.9.3"
 regex = "1.8.4"
 # Use trust-dns to avoid lookup failures on high concurrency
 # https://github.com/seanmonstar/reqwest/issues/296
-reqwest = { version = "0.11.18", features = ["gzip", "trust-dns"] }
+reqwest = { version = "0.11.18", default-features = false, features = ["gzip", "trust-dns"] }
```
mre, line 30: This could potentially save a bit of compilation time.

mre (approved):
Changes look good. I wonder if we should mention something in the docs (maybe a line or two in the `README.md` somewhere)?

lychee-bin/Cargo.toml
stefankreutz: Good question. I assumed "json" was a default feature of reqwest, but it is not. In fact, I cannot see how it worked before. There is no other reqwest feature that enables the "json" feature, and it is necessary for the `lychee::archive::wayback::get_wayback_link` function.

mre (approved):
Awesome, thanks 🙏

PR #1110, opened 2023-06-14
feat: add support for basic auth per URI

Techassi:
This PR adds support to pass basic auth credentials (username and password) based on a specific URI. The new basic auth syntax looks like this:

```toml
basic_auth = ["https://example.com* foo:bar"]
```

```shell
lychee --basic-auth "https://example.com* foo:bar" README.md
```

The selector supports regular expressions. The basic auth credentials are formatted like before: `<username>:<password>`. Supporting passing credentials per URI allows users to specify multiple different `<username>:<password>` combinations. Additionally, we do not leak credentials to other URIs.


lychee-lib/src/client.rs
```diff
@@ -454,6 +472,21 @@ impl Client {
         Ok(Response::new(uri.clone(), status, source))
     }
 
+    /// Returns optional basic auth based on if the URI matches any of the
+    /// selectors.
+    pub fn select_basic_auth_by_uri<'a>(
+        &'a self,
+        uri: &'a Uri,
+    ) -> Option<&'a Authorization<Basic>> {
+        for selector in &self.basic_auth {
```
mre, line 481: It's not a big deal, but that's O(n). If there's a lot of links or auth headers, this could slow down the checking process. We could consider using a [`RegexSet`](https://docs.rs/regex/latest/regex/struct.RegexSet.html) to achieve better performance.

lychee-lib/src/types/basic_auth.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    #[error("Missing basic auth password")]
+    MissingPassword,
+
+    #[error("Missing basic auth username")]
+    MissingUsername,
+
+    #[error("Too many values separated by colon. Expected 2, got {0}. Valid form is '<username>:<password>'")]
+    TooManyParts(usize),
+}
```
mre, line 22: I like that you put the auth-related errors into the respective module. We should do the same with the other errors at some point.

lychee-lib/src/types/basic_auth.rs
```diff
@@ -0,0 +1,174 @@
...
+
+        if parts.len() == 1 {
+            return Err(BasicAuthCredentialsParseError::InvalidSyntax);
+        }
+
+        if parts.len() != 2 {
+            return Err(BasicAuthCredentialsParseError::TooManyParts(parts.len()));
+        }
+
+        if parts[0].is_empty() {
```
mre, line 49: A single space ` ` would not be empty. Would it be a valid basic auth username, though? Shall we apply `trim` before `is_empty`?

lychee-lib/src/types/basic_auth.rs
```diff
@@ -0,0 +1,174 @@
...
+/// [`BasicAuthSelector`] provides basic auth credentials for URLs which match
+/// the specified regex. This allows users to set different credentials based
+/// on the URLs they want to target.
+#[derive(Debug, Clone, DeserializeFromStr, PartialEq)]
+pub struct RawBasicAuthSelector {
+    /// The basic auth credentials made up of username and password
+    pub credentials: BasicAuthCredentials,
+
+    /// This regex matches URLs which will use the basic auth credentials
+    pub url: String,
```
mre, line 88: Since this is an unparsed URL regex, shall we call it `url_regex`?

lychee-lib/src/types/basic_auth.rs
```diff
@@ -0,0 +1,174 @@
...
+            url: Regex::new(&selector.url)?,
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
```
mre, line 160: We should add some negative tests as well.

mre (commented):
Great PR and nice syntax!
I've added a few comments.
We should probably add a CLI test as well to see if the basic auth headers get applied correctly for different servers. If you want to tackle that, you could take a look at the mock server we use for some tests in the tests folder.

Techassi:
> Great PR and nice syntax! I've added a few comments. We should probably add a CLI test as well to see if the basic auth headers get applied correctly for different servers. If you want to tackle that, you could take a look at the mock server we use for some tests in the tests folder.

Sure thing! I will add some tests :)

lychee-lib/src/types/basic_auth.rs
Techassi: Yes, agreed. But I guess this is out-of-scope for this PR.

lychee-lib/src/types/basic_auth.rs
Techassi: I would even consider `unparsed_url_regex` or `raw_url_regex`.

lychee-lib/src/types/basic_auth.rs
Techassi: Yes, you are right. I will add those.

lychee-lib/src/types/basic_auth.rs
mre: Yes, indeed 👍

lychee-lib/src/types/basic_auth.rs
mre: Yup, good idea. I like both. `raw` is a bit shorter, so I'd vote for that.

lychee-lib/src/client.rs
Techassi: I moved the complete URI matching to the link collection phase. Now we do the heavy work upfront and not when making the requests. See 1083855

lychee-lib/src/types/basic_auth.rs
Techassi: I added multiple negative tests. This also resolves the above comment.

Techassi:
The next step will be to add some tests for the mock server. I will add this in the follow-up commits.

lychee-lib/src/collector.rs
```diff
@@ -65,16 +86,24 @@ impl Collector {
             })
             .flatten();
 
+        // FIXME (Techassi): This error should be returned. But I don't know
+        // how we can return it inside the stream.
+        let basic_auth_extractor = BasicAuthExtractor::new(self.basic_auth_selectors).unwrap();
+
```
Techassi, line 92: This needs to be fixed. I'm unsure how tho. Maybe you have some input here @mre

mre: Yeah, that's a tricky one.
The challenge will be to find a single type that will be returned from that function. The Result + Future + Stream combination won't make that easy.

But I wonder why you construct the extractor in there?
Why not pass it into the collector instead?
Here's how you could do that:
https://github.com/Techassi/lychee/compare/feat/basic-auth-per-domain...lycheeverse:lychee:feat/basic-auth-per-domain-fixes?expand=1
If you want, you can directly merge that branch into yours; it should merge cleanly.

mre:
I've added two integration tests in this branch:
https://github.com/Techassi/lychee/compare/feat/basic-auth-per-domain...lycheeverse:lychee:feat/basic-auth-per-domain-tests?expand=1

The first one with a single BasicAuth server is working, but the second one with two separate servers doesn't.
Still trying to wrap my head around the problem. Maybe someone has an idea.
It seems like the URL gets bodged somewhere.

```rust
running 1 test
thread 'cli::test_multi_basic_auth' panicked at 'Unexpected failure.
code-1
stderr=``````
Error: Network error

Caused by:
    0: error sending request for url (http://http//127.0.0.1:53560%20admin_user:admin_pw): error trying to connect: dns error: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }
    1: error trying to connect: dns error: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }
    2: dns error: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }
    3: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }


command=`"target/debug/lychee" "--verbose" "--basic-auth" "http://127.0.0.1:53558 username:password123" "http://127.0.0.1:53560 admin_user:admin_pw" "-"`
stdin=````
http://127.0.0.1:53558
http://127.0.0.1:53560```
`
code=1
stdout=""
stderr=```
Error: Network error

Caused by:
    0: error sending request for url (http://http//127.0.0.1:53560%20admin_user:admin_pw): error trying to connect: dns error: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }
    1: error trying to connect: dns error: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }
    2: dns error: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }
    3: no record found for Query { name: Name(\"http.\"), query_type: AAAA, query_class: IN }
```

Techassi:
Okay, I will look into this!

Techassi:
I figured it out. If you want to provide multiple basic auth selectors using the CLI, you need to specify `--basic-auth` multiple times. The correct test looks like this:

```rust
#[tokio::test]
async fn test_multi_basic_auth() -> Result<()> {
    let username1 = "username";
    let password1 = "password123";
    let mock_server1 = wiremock::MockServer::start().await;
    Mock::given(basic_auth(username1, password1))
        .respond_with(ResponseTemplate::new(200))
        .mount(&mock_server1)
        .await;

    let username2 = "admin_user";
    let password2 = "admin_pw";
    let mock_server2 = wiremock::MockServer::start().await;

    Mock::given(basic_auth(username2, password2))
        .respond_with(ResponseTemplate::new(200))
        .mount(&mock_server2)
        .await;

    // Configure the command to use the BasicAuthExtractor
    main_command()
        .arg("--verbose")
        .arg("--basic-auth")
        .arg(format!("{} {username1}:{password1}", mock_server1.uri()))
        .arg("--basic-auth")
        .arg(format!("{} {username2}:{password2}", mock_server2.uri()))
        .arg("-")
        .write_stdin(format!("{}\n{}", mock_server1.uri(), mock_server2.uri()))
        .assert()
        .success()
        .stdout(contains("2 Total"))
        .stdout(contains("2 OK"));

    Ok(())
}
```

---

However, the test `cli::test_exclude_example_domains` somehow fails while being unaffected by the changes. When running the CLI command directly, it produces the correct output.

lychee-lib/src/collector.rs
Techassi: Merged in 92522c2

mre:
> If you want to provide multiple basic auth selectors using the CLI, specify --basic-auth multiple times (That's the default behaviour of clap)

Ah! Yeah, I should have known that; a footgun I trip over regularly. Thanks for looking into that and finding the issue. 🧙 

> However, the test `cli::test_exclude_example_domains` somehow fails while being unaffected by the changes. When running the CLI command directly, it produces the correct output.

CI is the source of truth, so we're good, I think. That said, it's weird that you get a different result locally. Example domains get excluded for the CLI only and are totally valid domains in our tests. So it depends on how you ran the tests. If you just ran `make test` it should work as expected. That's what CI executes, at least.

Apart from that, we should be good, right?

Techassi:
> If you just ran make test it should work as expected. That's what CI executes, at least.

Ah well never mind then, I ran the tests with `cargo test -p lychee cli`. Running them with `make test` succeeds.

> Apart from that, we should be good, right?

Yes, we can merge!

lychee-lib/src/collector.rs
```diff
@@ -50,13 +53,27 @@ impl Collector {
...
     /// All relative URLs get prefixed with `base` (if given).
     /// (This can be a directory or a base URL)
     ///
     /// # Errors
     ///
     /// Will return `Err` if links cannot be extracted from an input
+    ///
+    /// # Panics
+    ///
+    /// Will panic if the [`BasicAuthExtractor`] cannot be constructed.
```
Techassi, line 76: I guess this isn't needed anymore. We should remove it again.

mre: done

Techassi: lgtm!

mre (approved):

mre:
FYI: used `AsRef` to avoid having to clone the selectors: https://github.com/lycheeverse/lychee/pull/1110/commits/0440af4a1a356cacd254521b016a27066af133f6
Added some tests for the matching behavior in case there are multiple selectors. https://github.com/lycheeverse/lychee/pull/1110/commits/f938b242bf791ca1350defebb5b0784e55d5f36d

mre:
Alright, it's a wrap! 🥳 Thanks a lot for your contribution @Techassi. This brings lychee to a new level in terms of authentication and avoids leaking secrets to unaffected websites.

Techassi:
Thanks for the nice words. Happy to help out!

PR #1123, opened 2023-06-26
Avoid false positives when checking email addresses in HTML input

### LINKED ISSUE #1111: URL misinterpreted as email
When checking https://wikipedia.org for broken links with: `lychee https://wikipedia.org`
The result looks as follows:

```

[https://wikipedia.org/]:
✗ [ERR] portal/wikipedia.org/assets/img/Wikipedia-logo-v2@2x.png | Failed: Unreachable mail address: portal/wikipedia.org/assets/img/Wikipedia-logo-v2@2x.png: Invalid: Email doesn't exist or is syntactically incorrect
✗ [ERR] portal/wikipedia.org/assets/img/Wikipedia-logo-v2@1.5x.png | Failed: Unreachable mail address: portal/wikipedia.org/assets/img/Wikipedia-logo-v2@1.5x.png: Invalid: Email doesn't exist or is syntactically incorrect
✗ [ERR] portal/wikipedia.org/assets/img/Wikinews-logo_sister@2x.png | Failed: Unreachable mail address: portal/wikipedia.org/assets/img/Wikinews-logo_sister@2x.png: Invalid: Email doesn't exist or is syntactically incorrect

🔍 334 Total ✅ 330 OK 🚫 3 Errors (HTTP:3) 💤 1 Excluded
```

So `portal/wikipedia.org/assets/img/Wikipedia-logo-v2@2x.png` is erroneously detected as email address.
The problem can be isolated by creating the file "test.html" and then running `lychee test.html`.

test.html:

```html
<!DOCTYPE html>
<html lang="en-US">
  <head>
    <meta charset="utf-8">
    <title>Test</title>
  </head>
  <body>
    <img srcset="v2@1.5x.png" alt="Wikipedia" width="200" height="183">
  </body>
</html>
```

So the same thing happens with a value of `v2@1.5x.png` for `srcset`. Probably the same happens for just `src` in `img` tags.
Does it make sense to check `src` and `srcset` for email addresses?

mre:
Fixes #1111

### DIFF
--- a/lychee-lib/src/extract/mod.rs
+++ b/lychee-lib/src/extract/mod.rs
@@ -5,8 +5,35 @@ mod html5gum;
 mod markdown;
 mod plaintext;
 
+use linkify::{LinkFinder, LinkKind};
 use markdown::extract_markdown;
 use plaintext::extract_plaintext;
+
+/// Check if the given URL is an email link.
+///
+/// This operates on the raw URL strings, not the linkified version because it
+/// gets used in the HTML extractors, which parse the HTML attributes directly
+/// and return the raw strings.
+///
+/// Note that LinkFinder::links() is lazy, so this will only check the first
+/// email in the input.
+pub(crate) fn is_email_link(input: &str) -> bool {
+    let mut emails = LinkFinder::new().kinds(&[LinkKind::Email]).links(input);
+
+    // If the input contains more than one email, it's not a valid email link
+    let (first, second) = (emails.next(), emails.next());
+    if first.is_none() || second.is_some() {
+        return false;
+    }
+
+    // Unwrap is safe here because we checked that there is exactly one email
+    // in the input.
+    let link_text = first.unwrap().as_str();
+
+    // Email needs to match the full string.
+    // Strip the "mailto:" prefix if it exists.
+    input.strip_prefix("mailto:").unwrap_or(input) == link_text
+}
 
 /// Check if the given element is in the list of preformatted ("verbatim") tags.
 ///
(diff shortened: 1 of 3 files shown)


mre:
@thomas-zahner, can you have a look and see if that looks correct to you?

lychee-lib/src/extract/mod.rs
```diff
@@ -341,4 +362,12 @@ mod tests {
 
         assert_eq!(links, expected_links);
     }
+
+    #[test]
+    fn test_is_email_link() {
```
thomas-zahner, line 367: maybe you could also add:

```
assert!(!is_email_link("mailto:steve@apple.com in a sentence"));
```

thomas-zahner:
@mre Looks good, don't see a problem with the implementation but I could be missing something.

The only improvement I could think of is the code duplication in the tests.
They are exactly the same for both implementations.
Also the HTML in the tests could either be extracted to a function where the email link is a parameter or the HTML could be merged to a single HTML file as a fixture with a single test expecting multiple email links

lychee-lib/src/extract/mod.rs
mre: done

mre:
> The only improvement I could think of is the code duplication in the tests.
> They are exactly the same for both implementations.

Yes, that is correct. My goal was to test one behavior per unit test. Even though that's a bit redundant, it helps me troubleshoot issues faster, as I can read the name of the unit test to get the context of what's failing.

> Also the HTML in the tests could either be extracted to a function where the email link is a parameter or the HTML could be merged to a single HTML file as a fixture with a single test expecting multiple email links

Also true. We could refactor the tests at some point to avoid the duplication, but the goal is to use one HTML engine and remove the other eventually. Given that, there's no urge to refactor at the moment.

I appreciate the code review and creating the issue in the first place. 👍  Added the test you suggested. Merging.

PR #1126, opened 2023-06-27
Introduce fragment checking for links to markdown files.

HU90m:
This covers part of what has been asked for in https://github.com/lycheeverse/lychee/issues/185, specifically checking fragments for links to markdown files.

I've introduced the concept of a *fragment generator* because there doesn't appear to be a standard way in which fragments/anchors are generated from markdown files. `HeadingAttributes` uses [the heading attribute syntax](https://github.com/raphlinus/pulldown-cmark/blob/master/specs/heading_attrs.txt). `UniqueKebabCase` uses the method used by GitHub and mdBook to generate fragments.

At the moment, all fragment generators ignore `id` attributes within inline HTML. These could be included using html5ever in the same way as is done with links. It may make sense to add support for fragment checks for html pages at the same time. NB: [this comment](https://github.com/lycheeverse/lychee/issues/185#issuecomment-1259588945).

This probably needs a little polishing. Opening a draft PR because I wanted some feedback before doing any more.

One change that is definitely needed is some unit tests for `extract_markdown_fragments`.

The general approach is simple but as a result the file being linked to is parsed for fragment link for each link. This is probably worth the simplicity because parsing a file isn't very slow compared to a network access.

### DIFF
--- a/lychee-lib/src/types/error.rs
+++ b/lychee-lib/src/types/error.rs
@@ -91,7 +91,7 @@ pub enum ErrorKind {
 
     /// The given path does not resolve to a valid file
     #[error("Cannot find local file {0}")]
-    FileNotFound(PathBuf),
+    InvalidFile(PathBuf),
 
     /// Error while traversing an input directory
     #[error("Cannot traverse input directory: {0}")]
@@ -140,14 +140,6 @@ pub enum ErrorKind {
     /// Cannot load cookies
     #[error("Cannot load cookies")]
     Cookies(String),
-
-    /// File doesn't have a file name.
-    #[error("File `{0}` doesn't have a file name.")]
-    NoFileName(PathBuf),
-
-    /// Filename is invalid unicode.
-    #[error("Filename of file `{0}` is invalid unicode.")]
-    InvalidFilenameUnicode(PathBuf),
 }
 
 impl ErrorKind {
@@ -245,8 +237,7 @@ impl PartialEq for ErrorKind {
                 e1.to_string() == e2.to_string()
             }
             (Self::Cookies(e1), Self::Cookies(e2)) => e1 == e2,
-            (Self::NoFileName(p1), Self::NoFileName(p2)) => p1 == p2,
-            (Self::InvalidFilenameUnicode(p1), Self::InvalidFilenameUnicode(p2)) => p1 == p2,
+            (Self::InvalidFile(p1), Self::InvalidFile(p2)) => p1 == p2,
             _ => false,
         }
     }
@@ -271,7 +262,7 @@ impl Hash for ErrorKind {
             Self::GithubRequest(e) => e.to_string().hash(state),
             Self::InvalidGithubUrl(s) => s.hash(state),
             Self::DirTraversal(e) => e.to_string().hash(state),
-            Self::FileNotFound(e) => e.to_string_lossy().hash(state),
+            Self::InvalidFile(e) => e.to_string_lossy().hash(state),
             Self::EmptyUrl => "Empty URL".hash(state),
             Self::ParseUrl(e, s) => (e.to_string(), s).hash(state),
             Self::InvalidURI(u) => u.hash(state),
@@ -294,8 +285,6 @@ impl Hash for ErrorKind {
             Self::TooManyRedirects(e) => e.to_string().hash(state),
             Self::BasicAuthExtractorError(e) => e.to_string().hash(state),
             Self::Cookies(e) => e.to_string().hash(state),
-            Self::NoFileName(p) => p.hash(state),
-            Self::InvalidFilenameUnicode(p) => p.hash(state),
         }
     }
 }
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -74,9 +74,9 @@ pub(crate) fn create(
                     match root.file_name() {
                         Some(file_name) => match file_name.to_str() {
                             Some(valid_str) => valid_str.to_string() + &text,
-                            None => return Err(ErrorKind::InvalidFilenameUnicode(root.clone())),
+                            None => return Err(ErrorKind::InvalidFile(root.clone())),
                         },
-                        None => return Err(ErrorKind::NoFileName(root.clone())),
+                        None => return Err(ErrorKind::InvalidFile(root.clone())),
                     }
                 } else {
                     text
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -153,7 +153,7 @@ impl Input {
                     // and exit early if it does
                     // This check might not be sufficient to cover all cases
                     // but it catches the most common ones
-                    return Err(ErrorKind::FileNotFound(path));
+                    return Err(ErrorKind::InvalidFile(path));
                 } else {
                     // Invalid path; check if a valid URL can be constructed from the input
                     // by prefixing it with a `http://` scheme.
@@ -442,10 +442,7 @@ mod tests {
 
         let input = Input::new(test_file, None, false, None);
         assert!(input.is_err());
-        assert!(matches!(
-            input,
-            Err(ErrorKind::FileNotFound(PathBuf { .. }))
-        ));
+        assert!(matches!(input, Err(ErrorKind::InvalidFile(PathBuf { .. }))));
     }
 
     #[test]
--- a/lychee-lib/src/utils/path.rs
+++ b/lychee-lib/src/utils/path.rs
@@ -45,7 +45,7 @@ pub(crate) fn resolve(src: &Path, dst: &Path, base: &Option<Base>) -> Result<Opt
         relative if dst.is_relative() => {
             // Find `dst` in the parent directory of `src`
             let Some(parent) = src.parent() else {
-                return Err(ErrorKind::FileNotFound(relative.to_path_buf()))
+                return Err(ErrorKind::InvalidFile(relative.to_path_buf()))
             };
             parent.join(relative)
         }
@@ -62,7 +62,7 @@ pub(crate) fn resolve(src: &Path, dst: &Path, base: &Option<Base>) -> Result<Opt
             };
             join(dir.to_path_buf(), absolute)
         }
-        _ => return Err(ErrorKind::FileNotFound(dst.to_path_buf())),
+        _ => return Err(ErrorKind::InvalidFile(dst.to_path_buf())),
     };
     Ok(Some(absolute_path(resolved)))
 }


lychee-lib/src/client.rs
```diff
@@ -624,13 +636,33 @@ impl Client {
...
+            .map_or(false, |ext| ext.eq_ignore_ascii_case("md"));
+
+        if let (true, Some(frag), Some(frag_gen)) = (is_md, uri.url.fragment(), check_md_frag) {
+            let mut content = String::new();
+            File::open(path)
+                .await
+                .expect("Couldn't open a file expected to exist.")
+                .read_to_string(&mut content)
+                .await
+                .expect("Couldn't read a linked file.");
```
mre, line 657: As you mentioned, this is the key part where we might lose a lot of performance, because we potentially parse the file many times. That's an `O(n^2)` algorithm, which scales badly with the number of input files.

As an alternative, we could introduce a `FragementCache`.
It would be defined as

```rust
type FragmentCache = Arc<Mutex<HashMap<String, HashSet<String>>>>;
```

This way it could be independent of the file extension, so the algorithm becomes

```rust
if let Some(fragement) = uri.url.fragment() {
  self.fragment_cache.contains(fragment)
}
```

The `FragmentCache` would be initialized when the client starts (i.e. in the client constructor and be a member of the client, so that it can be accessed with `self.fragment_cache.contains(fragment)` as seen above.

mre:
Awesome progress and a nice PR! Thanks for looking into this @HU90m.

I like your approach and that you created a draft first, to discuss the overall design.
The first draft looks good, and I'd love to move forward with this.
My main concern is the repeated file parsing, which you already mentioned; but I guess there is a workaround, which is relatively easy and performant by using a `FragmentCache` (see my comments for more info).

lychee-lib/src/extract/markdown.rs
```diff
@@ -77,6 +82,93 @@ pub(crate) fn extract_markdown(input: &str, include_verbatim: bool) -> Vec<RawUr
...
+/// from a markdown document.
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq)]
+pub enum FragmentGen {
+    /// Only take fragments specified in heading attributes.
+    HeadingAttribute,
+    /// Convert Headings into Kebab Case for use as ids.
+    KebabCase,
+    /// KebabCase but append a number if duplicates exist.
+    UniqueKebabCase,
+}
```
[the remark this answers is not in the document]
HU90m: `UniqueKebabCase` currently ignores heading attributes, but I agree using heading attributes with a fallback to unique kebab case would make a great default.

Also note, fragments don't appear to be covered in the [CommonMark spec](https://spec.commonmark.org/0.30/). The unique kebab case method appears to be used by mdBook because it is used in GitHub.

lychee-lib/src/extract/markdown.rs
```diff
@@ -77,6 +82,93 @@ pub(crate) fn extract_markdown(input: &str, include_verbatim: bool) -> Vec<RawUr
...
+            if ch.is_alphanumeric() || ch == '_' || ch == '-' {
+                Some(ch.to_ascii_lowercase())
+            } else if ch.is_whitespace() {
+                Some('-')
+            } else {
+                None
+            }
+        })
+        .collect::<String>()
+}
```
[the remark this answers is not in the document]
HU90m: Haha, I guess so. That crate looks good.

lychee-lib/src/extract/markdown.rs
```diff
@@ -77,6 +82,93 @@ pub(crate) fn extract_markdown(input: &str, include_verbatim: bool) -> Vec<RawUr
         .collect()
 }
 
+/// Extract unparsed URL strings from a Markdown string.
+pub(crate) fn extract_markdown_fragments(input: &str, frag_t: FragmentGen) -> Vec<String> {
```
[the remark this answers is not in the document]
HU90m: Brill! I'll swap this for a HashSet for now.

lychee-lib/src/client.rs
HU90m: That sounds sensible. I'll add a fragment cache.

I was toying with the idea of finding the fragments in the same pass as links for the files given by the user to be checked, avoiding one having to open and parse these files twice.  I'd guess a common use case is to check many files linking mainly to one another.

It would be nice to pre-populate the cache with the given files in this way, but it would require some larger changes. Do you think it would be worth it, if so how would you recommend going about it?

HU90m:
Hey @MichaIng , thanks for sharing your thoughts.

Yep this is to check fragments if the "target" of a link is a Markdown file, right

> So the use case is when you run lychee on the Markdown source code of e.g. an MkDocs site where links are defined to other Markdown files, while building the actual website translates them to HTML page links, like the website itself?

Spot on, but I would like to be able to check html directly as well in the future.

I tend to put fragment generation in one of two categories: explicit and automatic. *Heading attributes* seems by far the most popular explicit method. It is what is an option in this PR at the moment (`HeadingAttributes`). This is just using the implementation in pulldown-cmark, which is compatible with the https://python-markdown.github.io/extensions/attr_list/ so long as you don't use the optional `:`, e.g. `{ #thing }` not `{: #thing }`.

It looks like https://python-markdown.github.io/extensions/toc/ gives a lot of configuration options. I couldn't quite work out the default settings. Could you describe them?

I've laid out the big open questions (as I see them) in https://github.com/lycheeverse/lychee/issues/185 . I thought I would move higher level discussions back to the issue. Then when we have a rough consensus of what is wanted, we can review implementation details in this (or possibly a different) PR.

mre:
@HU90m, I just found this old branch, which contains some test cases for HTML fragment extraction. Feel free to copy some of that stuff over into your PR if you like. https://github.com/lycheeverse/lychee/compare/master...anchors

HU90m:
- Removed fragment generator options. Both unque kebab case and heading attribute fragments are always used.
- Now using convert_case for kebab case conversion.
- Added fragment cache.

Will add some tests tomorrow.

> @HU90m, I just found this old branch, which contains some test cases for HTML fragment extraction. Feel free to copy some of that stuff over into your PR if you like.

Thank you! This looks useful.

HU90m:
Rebased PR

mre:
Awesome work! This will be extremely helpful once merged. I will do a review as soon as I get around.

lychee-lib/src/extract/markdown.rs
```diff
@@ -77,6 +81,49 @@ pub(crate) fn extract_markdown(input: &str, include_verbatim: bool) -> Vec<RawUr
...
+                    out.insert(frag.to_string());
+                }
+
+                let mut id = heading.to_case(Case::Kebab);
+
+                let k_id_count = kebab_id_counter.entry(id.clone()).or_insert(0);
+                if *k_id_count != 0 {
+                    id = format!("{id}-{k_id_count}");
+                }
+                *k_id_count += 1;
```
mre, line 108: Maybe we can extract that id-generation logic into a separate struct:

```rust
struct HeadingIdGenerator {
    counter: HashMap<String, usize>,
}

impl HeadingIdGenerator {
    fn new() -> Self {
        Self {
            counter: HashMap::new(),
        }
    }

    fn generate(&mut self, heading: &mut String) -> String {
        let mut id = heading.to_case(Case::Kebab);
        let count = self.counter.entry(id.clone()).or_insert(0);
        if *count != 0 {
            id = format!("{}-{}", id, *count);
        }
        *count += 1;

        id
    }
}
```

Might look a bit like overkill, but it's easier to test, separates the logic, and keeps track of state.

You can then initialize it at the beginning of this function like so:

```rust
let mut id_generator = HeadingIdGenerator::new();
```

and use it like

```rust
let id = id_generator.generate(&mut heading);
out.insert(id);
```

This gives us full control over how IDs get generated. What do you think?

lychee-lib/src/utils/request.rs
```diff
@@ -70,10 +70,19 @@ pub(crate) fn create(
...
-                } else if let Some(url) = create_uri_from_path(root, &text, base)? {
+                let path = if is_anchor {
+                    // If the link is just an anchor,
+                    // prepend the filename of the file from which the link came.
+                    root.file_name()
+                        .expect("File doesn't have a file name.")
+                        .to_str()
+                        .expect("Filename is invalid unicode")
+                        .to_string()
+                        + &text
```
mre, line 81: Those `expect`s would cause the program to panic. I think a better way to handle this would be to log a warning (with `warn!`) and just `return Ok(None)` as before?

lychee-lib/src/client.rs
```diff
@@ -650,13 +668,52 @@ impl Client {
...
+        if self.include_fragments {
+            self.check_fragment(&path, uri).await
+        } else {
+            Status::Ok(StatusCode::OK)
+        }
+    }
+
+    /// Checks a `file` URI's fragment.
+    pub async fn check_fragment(&self, path: &Path, uri: &Uri) -> Status {
+        let is_md = path
```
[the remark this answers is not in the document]
HU90m: Oh nice!

lychee-lib/src/extract/markdown.rs
HU90m: I have no strong opinion either way, so it now exists.

lychee-lib/src/utils/request.rs
HU90m: Agreed, it would be nice to avoid the panic.

lychee-lib/src/client.rs
```diff
@@ -650,13 +668,52 @@ impl Client {
...
+            return ErrorKind::InvalidFilePath(uri.clone()).into();
+        }
+        if self.include_fragments {
+            self.check_fragment(&path, uri).await
+        } else {
+            Status::Ok(StatusCode::OK)
+        }
+    }
+
+    /// Checks a `file` URI's fragment.
```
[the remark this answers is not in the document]
HU90m: Encapsulating this logic sounds is a great idea. I renamed the `FragmentCache` to `FragmentChecker` because it isn't just a cache; it also parses files and populates the cache if it is entry is vacant.

I've left `extract_markdown_fragments` in `extract::markdown` because I like the idea of all the markdown parsing logic in one module. This would also be nice if we ever want a function that can parse for both links and fragments at the same time in the future.

HU90m:
`FragmentChecker` is lives!

HU90m:
Rebased. Also squashed a lot of my commits to make the rebase simpler.

HU90m:
Fixed the failing test

mre:
I made a few adjustments to (hopefully) make the code a bit more idiomatic in https://github.com/lycheeverse/lychee/pull/1126/commits/1001ccda646d422c504302ddc26821f8e59a9c23.
I tried to write an integration test in https://github.com/lycheeverse/lychee/pull/1126/commits/52b6dbf404c16ce065385f81d090dedb48364824, but it's failing right now; don't know why.

mre:
Sorry for the additional commits. We can squash them in the end.

HU90m:
Thanks for adding some tests! This saved me quite a bit of time getting my head around the testing infrastructure.

The heading attributes test was failing because:
1. I hadn't set the parser up for the heading attributes extension
2. The `pulldown-cmark` parser doesn't parse a heading attribute if it's on the last line. This is probably a bug. To fix it, I moved the heading with an attributes so it wasn't on the last line.

HU90m:
Rebased on master

HU90m:
The `convert_case`'s doesn't kebab-ify in the same way as mdBook, GitHub and GitLab. It replaces `_` with `-` whereas the others leave `_` be. The algorithm we had before is the same as that used by mdBook. Should I revert the change to using the `convert_case` crate.

mre:
Oh. Nice catch. Yes, maybe we should revert it then. We could still take over some of those tests for edge cases from the `convert_case` crate. If you find the time, you could also open an upstream issue or maybe even contribute your algorithm there.

lychee-lib/src/types/error.rs
```diff
@@ -225,6 +238,15 @@ impl PartialEq for ErrorKind {
             (Self::Regex(e1), Self::Regex(e2)) => e1.to_string() == e2.to_string(),
             (Self::DirTraversal(e1), Self::DirTraversal(e2)) => e1.to_string() == e2.to_string(),
             (Self::Channel(_), Self::Channel(_)) => true,
+            (Self::TooManyRedirects(e1), Self::TooManyRedirects(e2)) => {
+                e1.to_string() == e2.to_string()
+            }
+            (Self::BasicAuthExtractorError(e1), Self::BasicAuthExtractorError(e2)) => {
+                e1.to_string() == e2.to_string()
+            }
+            (Self::Cookies(e1), Self::Cookies(e2)) => e1 == e2,
```
mre, line 247: These changes aren't related to fragment checking, but I added the missing branches anyway while I was at it. I keep forgetting to do that in a separate PR. 😅

mre (commented):
fyi @HU90m: I went ahead and changed `expect` to `ErrorKind`. This way, we can handle errors a little more up the call chain. We don't differentiate between the errors at the moment, but technically we could do so in the future. Right now, I've added to separate variants, `InvalidFilenameUnicode` and `NoFileName`, but we can also merge it into one and call it `InvalidFile` or so.

mre:
Looks like you need to rebase on top of `master` one more time. Other than that, this should be good to go, right? Or do you plan any additional changes?

Oh! And maybe we can mention fragments in the feature comparison table in the `README`. ✨  It would be required to check if the other link checkers mentioned in the table have this feature, but it might be quick?

HU90m:
> I went ahead and changed `expect` to `ErrorKind`.

I never said. Thanks for doing this!

> Looks like you need to rebase on top of `master` one more time. Other than that, this should be good to go, right? Or do you plan any additional changes?

Just rebased. Yep it's all good to go :smiley: 

> Oh! And maybe we can mention fragments in the feature comparison table in the `README`. sparkles It would be required to check if the other link checkers mentioned in the table have this feature, but it might be quick?

Yeah, it shouldn't take too long. I started doing it, but stopped because we don't have fragment checks up and running for html and inline html yet. I feel like we should wait until we've got those PRs in before advertising this feature.

Issue #1108, opened 2023-06-13
Twitter link sharing URL check returns 404


craig:
What I did:
```
$ docker run --init -it --rm -w /input -v $(pwd):/input lycheeverse/lychee https://www.auswaertiges-amt.de/de/
  216/216 ETA 0s ████████████████████ Finished extracting links                                                                                                                                                                                                                            Issues found in 1 input. Find details below.

[https://www.auswaertiges-amt.de/de/]:
✗ [404] http://twitter.com/share?url=https://www.auswaertiges-amt.de/de&text=Ausw%C3%A4rtiges+Amt | Failed: Network error: Not Found
✗ [ERR] ?subject=Empfehlung%20aus%20dem%20Internetauftritt%20des%20Ausw%C3%A4rtigen%20Amtes&body=Der%20nachfolgende%20Artikel%20k%C3%B6nnte%20f%C3%BCr%20Sie%20interessant%20sein:%20https://www.auswaertiges-amt.de/de | Failed: Unreachable mail address: ?subject=Empfehlung%20aus%20dem%20Internetauftritt%20des%20Ausw%C3%A4rtigen%20Amtes&body=Der%20nachfolgende%20Artikel%20k%C3%B6nnte%20f%C3%BCr%20Sie%20interessant%20sein:%20https://www.auswaertiges-amt.de/de: Invalid: Email doesn't exist or is syntactically incorrect

🔍 216 Total ✅ 213 OK 🚫 2 Errors (HTTP:2)

```
The URL http://twitter.com/share?url=https://www.auswaertiges-amt.de/de&text=Ausw%C3%A4rtiges+Amt is valid and opening a popup to share the URL when logged in. When not logged in, it redirects to:
`location: https://twitter.com/share?url=https://www.auswaertiges-amt.de/de&text=Ausw%C3%A4rtiges+Amt`
and then again to itself, while setting a cookie:
```
location: /share?url=https://www.auswaertiges-amt.de/de&text=Ausw%C3%A4rtiges+Amt
set-cookie: guest_id=v1%3A168664601832869523; Max-Age=34214400; Expires=Sat, 13 Jul 2024 08:46:58 GMT; Path=/; Domain=.twitter.com; Secure; SameSite=None

```

This might possibly require a quirk implementation or the current one is broken?

DariuszPorowski:
@craig try:

1. change `http` to `https` in your url
[...]

Issue #645, opened 2022-06-08
Add cookie support?


mre:
Over at https://github.com/lycheeverse/lychee/issues/574 I noted that checking https://sms-digital.com would not work without cookie support. curl has the same problem, but they have cookie support with `--cookie`. We could do the same in lychee. reqwest supports that.

We'd have to enable the "cookies" crate feature, construct a Client with `cookie_store(true)` and then place a cookie into its cookie jar using https://docs.rs/reqwest/latest/reqwest/cookie/struct.Jar.html#method.add_cookie_str ([source](https://www.reddit.com/r/rust/comments/p0vcm2/how_to_use_browser_cookies_in_reqwest/)).

Should this feature be enabled by default to avoid confusion on failed requests?

egrieco:
I don't know about enabled by default, but maybe provide some guidance to the user that `--cookie` is an option if they encounter authorization errors.

Probably a good idea to give them some guidance on how to get cookies from their browser into the lychee cookie store as well. There are some extensions that can make this easier.

mre:
To test if it works we can try it on `https://sms-digital.com`.

egrieco:
Yes. I think trying to handle auth and such within lychee, especially with modern all JS sites will be far more trouble than it's worth.

Having the ability to extract the necessary cookies from the browser and pass them to lychee via a file or config option should be adequate.

lebensterben:
getting cookies from browser is a very bad idea.

There might not be browser, for example in CI.

How do you deal with multiple browser or multiple user accounts of a browser?

How do you prevent accessing potential private information in cookies?

mre:
Where the cookies come from might be irrelevant to lychee. The scenarios you mention are out of scope for lychee itself and be the user's responsibility. 

At least curl has ways to set cookies through the cli and their [cookiejar file](https://curl.se/docs/http-cookies.html). We could support the same behavior and even the same file format.

We also need to separate between storing cookies we receive and setting cookies provided by a user. Right now we don't handle cookies we receive. There could be a setting for allowing server-provided cookies, e.g. `--accept-cookies` or `--store-cookies`. To set a cookie we could provide `--cookie` similar to curl.

lebensterben:
the way `curl` handles cookies has nothing to d with browser. It simply takes a given cookie file.

what I commented above is regarding

> Having the ability to extract the necessary cookies from the browser 

I don't think `lychee` should ever need to talk to the browser.

mre:
[...]

Issue #715, opened 2022-08-03
Parameter `--max-redirects` did not take effect


Jiahui-Gu:
I set `--max-redirects`  is 30 and tested a 7 redirects example lychee reports too many redirects.

My CMD is:
`lychee /path/to/links.html --max-redirects 30 --format detailed`

The result is:
```
📝 Summary
---------------------
🔍 Total............1
✅ Successful.......0
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
❓ Unknown..........0
🚫 Errors...........1

Errors in /path/to/links.html
✗ [ERR] https://account.microsoft.com/family | Network error: error following redirect for url (https://account.microsoft.com/auth/complete-silent-signin?ru=https%3A%2F%2Faccount.microsoft.com%2Ffamily%3Frefd%3Daccount.microsoft.com): too many redirects
```

`links.html` content is:
`<a href="https://account.microsoft.com/family"></a>`

lebensterben:
```
𝛌> lychee test.html --max-redirects 10000
Issues found in 1 input. Find details below.

[test.html]:
⧖ [TIMEOUT] https://account.microsoft.com/family | Timeout

🔍 1 Total ✅ 0 OK 🚫 0 Errors
```

I don't think this is lychee's issue.

You can also try this with curl

```
𝛌> curl -iL https://account.microsoft.com/family --verbose
```

Jiahui-Gu:
But in fact this url does not time out, lychee still gives an error result, you can click on [it](https://account.microsoft.com/family) to see

lebensterben:
> the actual URL is only redirected 7 times

How do you know this?

lebensterben:
There are definitely more than 30 redirects.

I've managed to get TIMEOUT with 100 redirects and firefox's user-agent
```
𝛌> lychee test.html --max-redirects 100 --user-agent "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/81.0"
Issues found in 1 input. Find details below.

[test.html]:
⧖ [TIMEOUT] https://account.microsoft.com/family | Timeout

🔍 1 Total ✅ 0 OK 🚫 0 Errors
```

lebensterben:
It turns out that if you don't provide a user-agent it returns too many redirects error. With user-agent it times out.

Jiahui-Gu:
> > the actual URL is only redirected 7 times
> 
> How do you know this?

[...]

PR #1146, opened 2023-07-08
Cookie Support

mre:
This is a very conservative and limited implementation of cookie support.

The goal is to ship an MVP, which covers 80% of the use-cases.
When you run lychee with `--cookie-jar cookies.json`, all cookies will be stored in `cookies.json`, one cookie per line.
This makes cookies easy to edit by hand if needed, although this is an advanced use-case and the API for the format is **not** guaranteed to be stable.

My initial goal was to use the same cookie file format that curl uses, but I backed away from this thought for a few reasons:
* The cookie file format is an implementation detail: people should not rely on the format to never change.
* There is on Rust implementation for the curl format out there, so we'd have to roll our own, test it, and maintain it, which sounds like unnecessary work.
So, at least for now, we use [reqwest_cookie_store](https://github.com/pfernie/reqwest_cookie_store) under the hood.

Fixes: https://github.com/lycheeverse/lychee/issues/645, https://github.com/lycheeverse/lychee/issues/715
Partially fixes: https://github.com/lycheeverse/lychee/issues/1108#issuecomment-1591356601

Reviews/feedback welcome!

----

Note: I had to change an unrelated unit test for invalid status codes, because it was flaky, and I wanted the CI pipeline to be green before merging.

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -46,7 +46,7 @@ where
     let cache = params.cache;
     let accept = params.cfg.accept;
 
-    let pb = if params.cfg.no_progress {
+    let pb = if params.cfg.no_progress || params.cfg.verbose.log_level() >= log::Level::Info {
         None
     } else {
         Some(init_progress_bar("Extracting links"))
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -68,7 +68,7 @@ use clap::Parser;
 use color::YELLOW;
 use commands::CommandParams;
 use formatters::response::ResponseFormatter;
-use log::{debug, error, info, warn};
+use log::{error, info, warn};
 
 #[cfg(feature = "native-tls")]
 use openssl_sys as _; // required for vendored-openssl feature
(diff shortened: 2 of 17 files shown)


lychee-bin/src/commands/check.rs
```diff
@@ -46,7 +46,7 @@ where
     let cache = params.cache;
     let accept = params.cfg.accept;
 
-    let pb = if params.cfg.no_progress {
+    let pb = if params.cfg.no_progress || params.cfg.verbose.log_level() >= log::Level::Info {
```
mre, line 49: This is here, because showing the progress bar in verbose mode makes the output harder to read. It's not related to the other changes in this PR, but I got annoyed by the output while troubleshooting, so I fixed it in one go.

lychee-bin/src/main.rs
```diff
@@ -188,6 +191,15 @@ fn load_config() -> Result<LycheeOptions> {
     Ok(opts)
 }
 
+/// Load cookie jar from path (if exists)
+/// Load cookie jar from path (if exists)
```
mre, line 195:
```suggestion

```

mre:
I am wondering if we should enable cookies by default. So if the user didn't set a custom cookie jar file, we still enable cookie support for this "link check session". After that, we throw the jar away. This would handle more cases, like https://github.com/lycheeverse/lychee/issues/715. Opinions?

PR #1147, opened 2023-07-09
Don't check Twitter URLs

mre:
Twitter completely locked down and requires
a login to read tweets. (Temporarily) disable all
Twitter URLs to avoid false-positives.

For context:
https://github.com/zedeus/nitter/issues/919
https://news.ycombinator.com/item?id=36540957
https://techcrunch.com/2023/06/30/twitter-now-requires-an-account-to-view-tweets/

Fixes https://github.com/lycheeverse/lychee/issues/1108

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -241,8 +241,10 @@ mod cli {
     }
 
     #[test]
-    #[ignore = "Twitter quirk works locally but is flaky on Github (timeout)"]
-    fn test_twitter_quirk() {
+    // Exclude Twitter links because they require login to view tweets.
+    // https://techcrunch.com/2023/06/30/twitter-now-requires-an-account-to-view-tweets/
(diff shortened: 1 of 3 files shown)


szepeviktor:
What cookies should one add to `headers = []` to get content from Twitter?

szepeviktor:
Surprising fact: twitter.com responds with 200 OK for a non-existent Tweet!

mre:
> What cookies should one add to headers = [] to get content from Twitter?

I honestly never tried. I can see a few cookies in my Firefox storage tab inside the browser tools. Stuff like `auth_multi` and `auth_token` and `twid`. I would probably start with these. 🤷‍♀️

mre:
> Surprising fact: twitter.com responds with 200 OK for a non-existent Tweet!

Oh yeah, it's completely unpredictable what you might get at times. It leads to so many false-positives that it will be completely disabled by default in the next release.

PR #1154, opened 2023-07-12
fix exclude pattern in example

woolfg:
I think the regex was wrong and it should be like that. it is also used in `include = [ 'gist\.github\.com.*' ]` some lines below.


mre:
Makes sense. Thanks a lot!

PR #1159, opened 2023-07-15
feat: Add support for --dump-inputs

### LINKED ISSUE #684: Add `--dump-input` flag to print all files, which would be searched for links
This is mostly helpful for debugging purposes. For example if a regular expression for file matching doesn't work as expected.
Example: `lychee --dump-input './**/*.md'` would print all file names that the glob expression matches and exit.

Techassi:
Fixes #684

This PR adds support for the CLI flag `--dump-inputs` which prints out input sources without collecting and checking links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,9 @@ Options:
       --dump
           Don't perform any link checking. Instead, dump all the links extracted from inputs that would be checked
 
+      --dump-inputs
+          Don't perform any link collection and checking. Instead, dump all input sources from which links would be collected
+
       --archive <ARCHIVE>
           Specify the use of a specific web archive. Can be used in combination with `--suggest`
           
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ Options:
           Don't perform any link checking. Instead, dump all the links extracted from inputs that would be checked
 
       --dump-inputs
-          Don't perform any link collection and checking. Instead, dump all input sources from which links would be extracted
+          Don't perform any link extraction and checking. Instead, dump all input sources from which links would be collected
 
       --archive <ARCHIVE>
           Specify the use of a specific web archive. Can be used in combination with `--suggest`
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ Options:
           Don't perform any link checking. Instead, dump all the links extracted from inputs that would be checked
 
       --dump-inputs
-          Don't perform any link collection and checking. Instead, dump all input sources from which links would be collected
+          Don't perform any link collection and checking. Instead, dump all input sources from which links would be extracted
 
       --archive <ARCHIVE>
           Specify the use of a specific web archive. Can be used in combination with `--suggest`
--- a/lychee-bin/src/commands/dump.rs
+++ b/lychee-bin/src/commands/dump.rs
@@ -37,8 +37,8 @@ where
     let requests = params.requests;
     tokio::pin!(requests);
 
-    if let Some(outfile) = &params.cfg.output {
-        fs::File::create(outfile)?;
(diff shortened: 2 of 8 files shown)


lychee-lib/src/collector.rs
```diff
@@ -63,6 +63,13 @@ impl Collector {
         self
     }
 
+    /// TODO
```
Techassi, line 66: This needs documentation. Will add it asap!

Techassi: Added in 3282463

README.md
```diff
@@ -285,6 +285,9 @@ Options:
       --dump
           Don't perform any link checking. Instead, dump all the links extracted from inputs that would be checked
 
+      --dump-inputs
+          Don't perform any link collection and checking. Instead, dump all input sources from which links would be collected
```
mre, line 289:
```suggestion
          Don't perform any link collection and checking. Instead, dump all input sources from which links would be extracted
```

...just to stay close to the nomenclature in `--dump`

lychee-bin/src/commands/dump.rs
```diff
@@ -70,6 +70,30 @@ where
     Ok(ExitCode::Success)
 }
 
+/// Dump all input sources to stdout without detecting any links and checking
```
mre, line 73:
```suggestion
/// Dump all input sources to stdout without extracting any links and checking
```

lychee-bin/src/options.rs
```diff
@@ -181,6 +181,12 @@ pub(crate) struct Config {
     #[serde(default)]
     pub(crate) dump: bool,
 
+    /// Don't perform any link collection and checking.
```
mre, line 184:
```suggestion
    /// Don't perform any link extraction and checking.
```

lychee-lib/src/types/input.rs
```diff
@@ -260,6 +260,40 @@ impl Input {
         }
     }
 
+    /// Retrieve all sources from this input. The output depends on the type of
+    /// input:
+    ///
+    /// - Remote URLs are returned as is, in their full form
+    /// - Filepath Glob Patterns are expanded and each matched entry is returned
+    /// - Absolute or relative filepaths are returned as is
+    /// - All other input types are not returned
```
mre, line 269: Maybe we could print `Stdin` for standard input and `Raw String` for strings? This could avoid some confusion when people try to troubleshoot their inputs. I don't have a strong opinion on this, though, so feel free to ignore.

mre (commented):
Thanks for tackling this! Very nice and clean implementation.
The only thing I'd add would be a unit and/or integration test, to make sure it produces the expected output. Let me know if you need some support with that, but I guess you got this.

lychee-lib/src/types/input.rs
Techassi: Yes, I was also thinking about this. I don't have a super strong opinion on this as well, but I can print out `Stdin` and `Raw String` as you suggested.

Techassi: Added in 07fe746

Techassi:
I added tests in ac3fa41. I think this PR is now good to go!

mre:
Great! I made a minor change to run the integration tests from a newly-created `fixtures/dump_input` subfolder. The reason being, that it's easier to keep the test fixtures isolated this way. Otherwise, we'd have to update the `dump_input` tests if we moved files around for other tests.

Hope that's fine. Looks good to me now. 😃

Techassi:
Yes, that's fine! Makes total sense to move these test files into their own folder for the obvious reasons you mentioned :smile: Thank you for merging this super quickly!

PR #1160, opened 2023-07-16
Improve `srcset` parsing

mre:
Our current `srcset` parsing is pretty basic.

We split on comma and then on whitespace and take the first part, which is the image source URL.
However, we don't handle URLs containing unencoded commas like
`/cdn-cgi/image/format=webp,width=640/https://img.youtube.com/vi/hVBl8_pgQf0/maxresdefault.jpg`, which leads to false-positives.

According to the spec, [commas in strings should be encoded](https://developers.google.com/maps/url-encoding), but in practice, there are some websites which don't do that. To handle these cases, too, I propose to extend the `srcset` parsing to make use of a small "state machine", which detects if a comma is within the image source or outside of it while parsing.

This is part of an effort to reduce false-positives during link checking.

### DIFF
--- a/lychee-lib/src/extract/html/srcset.rs
+++ b/lychee-lib/src/extract/html/srcset.rs
@@ -57,7 +57,7 @@ pub(crate) fn parse(input: &str) -> Vec<&str> {
             info!("srcset parse Error");
             return vec![];
         }
-        index += start.char_indices().count();
+        index += start.chars().count();
 
         if remaining.is_empty() {
             return candidates;
--- a/lychee-lib/src/extract/html/srcset.rs
+++ b/lychee-lib/src/extract/html/srcset.rs
@@ -33,7 +33,7 @@ fn split_at<F>(input: &str, predicate: F) -> (&str, &str)
 where
     F: Fn(&char) -> bool,
 {
-    for (i, ch) in input.chars().enumerate() {
+    for (i, ch) in input.char_indices() {
         if !predicate(&ch) {
             return (&input[0..i], &input[i..]);
         }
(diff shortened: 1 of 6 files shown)


mre:
@Techassi, @HU90m: maybe you got any feedback. 😉

Techassi:
State machines are definitely the way to go when writing parsers. I will take a look at the PR asap.

lychee-lib/src/extract/html/srcset.rs
```diff
@@ -0,0 +1,218 @@
...
+    InsideParens,
+}
+
+/// Split an input string at the first character for which
+/// the predicate returns false.
+fn split_at<F>(input: &str, predicate: F) -> (&str, &str)
+where
+    F: Fn(&char) -> bool,
+{
+    for (i, ch) in input.chars().enumerate() {
```
HU90m, line 36:
```suggestion
    for (i, ch) in input.char_indices() {
```

lychee-lib/src/extract/html/srcset.rs
```diff
@@ -0,0 +1,218 @@
...
+        if !predicate(&ch) {
+            return (&input[0..i], &input[i..]);
```
HU90m, line 38:
```suggestion
            return input.split_at(i);
```

lychee-lib/src/extract/html/srcset.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    while index < input.len() {
+        let position = &input[index..];
+        let (start, remaining) = split_at(position, |c| *c == ',' || c.is_whitespace());
+
+        if start.find(',').is_some() {
+            info!("srcset parse Error");
+            return vec![];
+        }
+        index += start.char_indices().count();
```
HU90m, line 60:
```suggestion
        index += start.chars().count();
```

HU90m (approved):
Looks great!

Just a few suggestions:

mre:
Loved the suggestions. Makes the code quite a bit more idiomatic. Thanks!

Issue #1157, opened 2023-07-13
Don't treat sites with 403 status codes as broken links?


mre:
In a recent test of the top 1000 websites, I found the following 403 status errors:

```
✗ [403] https://cell.com/ | Failed: Network error: Forbidden
✗ [403] https://ticketmaster.com/ | Failed: Network error: Forbidden
✗ [403] https://kraken.com/ | Failed: Network error: Forbidden
✗ [403] https://media.giphy.com/ | Failed: Network error: Forbidden
✗ [403] https://yummly.com/ | Failed: Network error: Forbidden
✗ [403] https://bhphotovideo.com/ | Failed: Network error: Forbidden
✗ [403] https://inc.com/ | Failed: Network error: Forbidden
✗ [403] https://science.sciencemag.org/ | Failed: Network error: Forbidden
✗ [403] https://support.cloudflare.com/ | Failed: Network error: Forbidden
✗ [403] https://docs.wixstatic.com/ | Failed: Network error: Forbidden
✗ [403] https://deviantart.com/ | Failed: Network error: Forbidden
✗ [403] https://use.fontawesome.com/ | Failed: Network error: Forbidden
✗ [403] https://yoursite.com/ | Failed: Network error: Forbidden
✗ [403] https://upwork.com/ | Failed: Network error: Forbidden
✗ [403] https://gleam.io/ | Failed: Network error: Forbidden
✗ [403] https://kickstarter.com/ | Failed: Network error: Forbidden
✗ [403] https://static.wixstatic.com/ | Failed: Network error: Forbidden
✗ [403] https://onlinelibrary.wiley.com/ | Failed: Network error: Forbidden
✗ [403] https://ericsson.com/ | Failed: Network error: Forbidden
✗ [403] https://journals.sagepub.com/ | Failed: Network error: Forbidden
✗ [403] https://kstatic.googleusercontent.com/ | Failed: Network error: Forbidden
✗ [403] https://coinbase.com/ | Failed: Network error: Forbidden
✗ [403] https://zazzle.com/ | Failed: Network error: Forbidden
✗ [403] https://thelancet.com/ | Failed: Network error: Forbidden
✗ [403] https://bluehost.com/ | Failed: Network error: Forbidden
✗ [403] https://tandfonline.com/ | Failed: Network error: Forbidden
✗ [403] https://event.on24.com/ | Failed: Network error: Forbidden
✗ [403] https://hostgator.com/ | Failed: Network error: Forbidden
[...]

PR #1167, opened 2023-07-17
feat: Add support for ranges in the `--accept` option / config field

Techassi:
Adds support for accept ranges discussed in #1157. This allows the user to specify custom HTTP status codes accepted during checking and thus will report as valid (not broken). The accept option only supports specifying status codes as a comma-separated list. With this PR, the option will accept a list of status code ranges formatted like this:

```toml
accept = ["100..=103", "200..=299", "403"]
```

These combinations will be supported: `..<end>`, `..=<end>`, `<start>..<end>` and `<start>..=<end>`. The behaviour is copied from the Rust `Range` like concepts:

- `..<end>`, includes 0 to `<end>` (exclusive)
- `..=<end>`, includes 0 to `<end>` (inclusive)
- `<start>..<end>`, includes `<start>` to `<end>` (exclusive)
- `<start>..=<end>`, includes `<start>` to `<end>` (inclusive)

---

This draft PR only implements some foundation work for this feature. Once all functionality is added, the PR will be ready to review.

### DIFF
--- a/fixtures/configs/smoketest.toml
+++ b/fixtures/configs/smoketest.toml
@@ -46,7 +46,13 @@ timeout = 20
 retry_wait_time = 2
 
 # Comma-separated list of accepted status codes for valid links.
-accept = "200, 429"
+# Supported values are:
+#
+# accept = ["200..=204", "429"]
+# accept = "200..=204, 429"
+# accept = ["200", "429"]
+# accept = "200, 429"
+accept = ["200", "429"]
 
 # Proceed for server connections considered insecure (invalid TLS).
 insecure = false
--- a/fixtures/configs/smoketest.toml
+++ b/fixtures/configs/smoketest.toml
@@ -46,14 +46,14 @@ timeout = 20
 retry_wait_time = 2
 
 # Comma-separated list of accepted status codes for valid links.
-accept = [200, 429]
+accept = "200, 429"
 
 # Proceed for server connections considered insecure (invalid TLS).
 insecure = false
 
 # Only test links with the given schemes (e.g. https).
 # Omit to check links with any scheme.
-scheme = [ "https" ]
+scheme = ["https"]
 
 # When links are available using HTTPS, treat HTTP links as errors.
 require_https = false
@@ -66,7 +66,7 @@ headers = []
 
 # Remap URI matching pattern to different URI.
 # This also supports (named) capturing groups.
-remap = [ 
+remap = [
     "https://example.com http://example.invalid",
     "https://example.com/(.*) http://example.org/$1",
     "https://github.com/(?P<org>.*)/(?P<repo>.*) https://gitlab.com/$org/$repo",
@@ -93,13 +93,13 @@ include_verbatim = false
 glob_ignore_case = false
 
 # Exclude URLs and mail addresses from checking (supports regex).
-exclude = [ '.*\.github.com\.*' ]
+exclude = ['.*\.github.com\.*']
 
 # Exclude these filesystem paths from getting checked.
 exclude_path = ["file/path/to/Ignore", "./other/file/path/to/Ignore"]
 
 # URLs to check (supports regex). Has preference over all excludes.
-include = [ 'gist\.github\.com.*' ]
+include = ['gist\.github\.com.*']
 
 # Exclude all private IPs from checking.
 # Equivalent to setting `exclude_private`, `exclude_link_local`, and
--- a/fixtures/configs/smoketest.toml
+++ b/fixtures/configs/smoketest.toml
@@ -58,7 +58,8 @@ accept = ["200", "429"]
 insecure = false
 
 # Only test links with the given schemes (e.g. https).
-# Omit to check links with any scheme.
+# Omit to check links with any other scheme.
+# At the moment, we support http, https, file, and mailto.
 scheme = ["https"]
 
 # When links are available using HTTPS, treat HTTP links as errors.
--- a/lychee-lib/src/types/accept/range.rs
+++ b/lychee-lib/src/types/accept/range.rs
@@ -69,7 +69,7 @@ impl AcceptRange {
         self.0.end()
     }
 
-    /// Returns wether this [`AcceptRange`] contains `value`.
+    /// Returns whether this [`AcceptRange`] contains `value`.
     pub fn contains(&self, value: usize) -> bool {
         self.0.contains(&value)
     }
--- a/lychee-lib/src/types/accept/range.rs
+++ b/lychee-lib/src/types/accept/range.rs
@@ -92,7 +92,7 @@ impl AcceptRange {
 #[cfg(test)]
 mod test {
     use super::*;
-    use rstest::*;
+    use rstest::rstest;
 
     #[rstest]
     #[case("0..=10", vec![0, 1, 4, 5, 10], vec![11, 12])]
--- a/lychee-lib/src/types/accept/range.rs
+++ b/lychee-lib/src/types/accept/range.rs
@@ -1,0 +1,133 @@
+use std::{num::ParseIntError, ops::RangeInclusive, str::FromStr};
+
+use once_cell::sync::Lazy;
+use regex::Regex;
+use thiserror::Error;
+
+static RANGE_PATTERN: Lazy<Regex> = Lazy::new(|| Regex::new(r"([0-9]*)?..(=?)([0-9]+)+").unwrap());
+
+/// The [`AcceptRangeParseError`] indicates that the parsing process of an
+/// [`AcceptRange`]  from a string failed due to various underlying reasons.
+#[derive(Debug, Error, PartialEq)]
+pub enum AcceptRangeParseError {
+    /// The string input didn't contain any range pattern.
+    #[error("No range pattern found")]
+    NoRangePattern,
+
+    /// The start or end index could not be parsed as an integer.
+    #[error("Failed to parse str as integer")]
+    ParseIntError(#[from] ParseIntError),
+
+    /// The start index is larger than the end index.
+    #[error("Invalid range indices, only start < end supported")]
+    InvalidRangeIndices,
+}
+
+/// [`AcceptRange`] specifies which HTTP status codes are accepted and
+/// considered successful when checking a remote URL.
+#[derive(Debug, PartialEq)]
+pub struct AcceptRange(RangeInclusive<usize>);
+
+impl FromStr for AcceptRange {
+    type Err = AcceptRangeParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let captures = RANGE_PATTERN
+            .captures(s)
+            .ok_or(AcceptRangeParseError::NoRangePattern)?;
+
+        let inclusive = !captures[2].is_empty();
+
+        let start: usize = captures[1].parse().unwrap_or_default();
+        let end: usize = captures[3].parse()?;
+
+        if start >= end {
+            return Err(AcceptRangeParseError::InvalidRangeIndices);
+        }
+
+        match inclusive {
+            true => Ok(Self::new(start, end)),
+            false => Ok(Self::new(start, end - 1)),
+        }
+    }
+}
+
+impl AcceptRange {
+    /// Creates a new [`AcceptRange`] which matches values between `start` and
+    /// `end` (both inclusive).
+    pub fn new(start: usize, end: usize) -> Self {
+        Self(RangeInclusive::new(start, end))
+    }
+
+    /// Returns the `start` value of this [`AcceptRange`].
+    pub fn start(&self) -> &usize {
+        self.0.start()
+    }
+
+    /// Returns the `end` value of this [`AcceptRange`].
+    pub fn end(&self) -> &usize {
+        self.0.end()
+    }
+
+    /// Returns wether this [`AcceptRange`] contains `value`.
+    pub fn contains(&self, value: usize) -> bool {
+        self.0.contains(&value)
+    }
+
+    pub(crate) fn update_start(&mut self, new_start: usize) {
+        self.0 = RangeInclusive::new(new_start, *self.end());
+    }
+
+    pub(crate) fn update_end(&mut self, new_end: usize) {
+        self.0 = RangeInclusive::new(*self.start(), new_end);
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_from_str() {
+        let range = AcceptRange::from_str("0..10").unwrap();
+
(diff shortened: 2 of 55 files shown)


lychee-lib/src/types/accept/range.rs
```diff
@@ -0,0 +1,133 @@
...
+        self.0.end()
+    }
+
+    /// Returns wether this [`AcceptRange`] contains `value`.
+    pub fn contains(&self, value: usize) -> bool {
+        self.0.contains(&value)
+    }
+
+    pub(crate) fn update_start(&mut self, new_start: usize) {
+        self.0 = RangeInclusive::new(new_start, *self.end());
```
mre, line 78: This would need the same sanity checks as in `from_str` I guess?
Namely, that start is not after end.
Maybe you want to extract the logic into a `new` method (which returns `Result`) or something and use it here and in `from_str`?

lychee-lib/src/types/accept/range.rs
```diff
@@ -0,0 +1,133 @@
...
+        assert!(range.contains(9));
+        assert!(range.contains(10));
+        assert!(!range.contains(11));
+    }
+
+    #[test]
+    fn test_from_str_invalid_indices() {
+        let range = AcceptRange::from_str("10..=0");
+        assert_eq!(range, Err(AcceptRangeParseError::InvalidRangeIndices))
+    }
```
mre, line 132: Some more test ideas:
* `0..0`
* `0..-1`
* `abcd`
* `0`
* `-1`

lychee-lib/src/types/accept/selector.rs
```diff
@@ -0,0 +1,115 @@
+use crate::types::accept::AcceptRange;
+
+/// A [`AcceptSelector`] determines if a returned HTTP status code should be
```
mre, line 3:
```suggestion
/// An [`AcceptSelector`] determines if a returned HTTP status code should be
```

lychee-lib/src/types/accept/selector.rs
```diff
@@ -0,0 +1,115 @@
...
+            }
+        }
+
+        // If neither is the case, the ranges have no overlap at all. Just add
+        // to the list of ranges.
+        self.ranges.push(range);
+        self
+    }
+
+    /// Returns wether this [`AcceptSelector`] contains `value`.
```
mre, line 43:
```suggestion
    /// Returns whether this [`AcceptSelector`] contains `value`.
```

mre (commented):
This is still a draft, but I thought I might add some early feedback. Keep it up! 😃

lychee-lib/src/types/accept/selector.rs
```diff
@@ -0,0 +1,115 @@
...
+            if last.start() <= range.end() && range.start() <= last.start() {
+                last.update_start(*range.start());
+                return self;
+            }
+        }
+
+        // If neither is the case, the ranges have no overlap at all. Just add
+        // to the list of ranges.
+        self.ranges.push(range);
+        self
```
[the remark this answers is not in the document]
Techassi: Yes, I also thought about that. This will be added in the next iteration of this PR.

lychee-lib/src/types/accept/range.rs
Techassi: True. Currently, this is only used internally. That's why I opted to skip the validation here. But I can extract the validation into a standalone method/function.

lychee-lib/src/types/accept/range.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn start(&self) -> &usize {
+        self.0.start()
+    }
+
+    /// Returns the `end` value of this [`AcceptRange`].
+    pub fn end(&self) -> &usize {
+        self.0.end()
+    }
+
+    /// Returns wether this [`AcceptRange`] contains `value`.
```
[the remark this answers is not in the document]
Techassi: Man. Typos be strong right now...

lychee-lib/src/types/accept/selector.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Returns wether this [`AcceptSelector`] contains `value`.
+    pub fn contains(&self, value: usize) -> bool {
+        for range in &self.ranges {
+            if range.contains(value) {
+                return true;
+            }
+        }
+
+        false
```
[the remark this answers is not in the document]
Techassi: I would have used this pattern eventually (I wanted to get the basics down first). So, let's add it :smile:

lychee-lib/src/types/accept/range.rs
Techassi: Added in 7866393

lychee-lib/src/types/accept/range.rs
Techassi: Fixed in 6645927

lychee-lib/src/types/accept/selector.rs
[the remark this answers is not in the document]
Techassi: This feature is much more complex than one might think: it must be recursive. Merging a newly added range with an existing range at index 0 requires us to re-check (and maybe merge) ranges with index >= 1.

Techassi:
The new accept syntax is now included in the CLI. I opted to use a string of comma-separated ranges instead of a Vec of ranges. This allows users to specify ranges like `--accept "100..=199,200,404"` instead of

```
--accept "100..=199" --accept 200 --accept 404
```

---

One of the tests keeps failing, which boggles my mind because the test case should be unaffected by my changes. Maybe you can take a look @mre

fixtures/configs/smoketest.toml
```diff
@@ -46,14 +46,14 @@ timeout = 20
 retry_wait_time = 2
 
 # Comma-separated list of accepted status codes for valid links.
-accept = [200, 429]
+accept = "200, 429"
```
mre, line 49: This change would break existing configs. I wonder if we should keep the array syntax here (and use the string syntax for the CLI). There's likely also a way to support both. At the end, it should work as long as the TOML library knows how to handle it and convert it to our internal format, no? Wanna look into that?

Techassi: I can try to make both syntax work. However, I'm not quite sure if it is possible to have a different format in the config compared to the CLI, as both use the same underlying struct... Let me know if you have any insights on this.

mre: Yeah, not 100% sure either. I think you can impl `From<String>` and `From<[&str]>` for the struct, but I'm not certain. Something to investigate.

mre: At least this works:

```rust
use serde::Deserialize;

#[derive(Deserialize)]
struct Config {
    accept: Accept,
}

// Accept can be either a string or a list of strings
#[derive(Deserialize, Debug, PartialEq)]
#[serde(untagged)]
enum Accept {
    One(String),
    Many(Vec<String>),
}

fn main() {
    let config1: Config = toml::from_str(r"accept = '200, 429'").unwrap();
    assert_eq!(config1.accept, Accept::One("200, 429".to_string()));

    let config2: Config = toml::from_str(r"accept = ['200', '429']").unwrap();
    assert_eq!(
        config2.accept,
        Accept::Many(vec!["200".to_string(), "429".to_string()])
    );
}
```

mre: And this also works:

```rust
use serde::{
    de::{self, Deserializer, SeqAccess, Visitor},
    Deserialize,
};
use std::fmt;

#[derive(Deserialize)]
struct Config {
    accept: Accept,
}

#[derive(Debug, PartialEq)]
struct Accept(Vec<String>);

struct AcceptVisitor;

impl<'de> Visitor<'de> for AcceptVisitor {
    type Value = Accept;

    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
        formatter.write_str("a string or sequence of strings")
    }

    fn visit_str<E: de::Error>(self, value: &str) -> Result<Accept, E> {
        Ok(Accept(
            value.split(',').map(|s| s.trim().to_string()).collect(),
        ))
    }

    fn visit_seq<S: SeqAccess<'de>>(self, mut seq: S) -> Result<Accept, S::Error> {
        let mut vec = Vec::new();
        while let Some(value) = seq.next_element()? {
            vec.push(value);
        }
        Ok(Accept(vec))
    }
}

impl<'de> Deserialize<'de> for Accept {
    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Accept, D::Error> {
        deserializer.deserialize_any(AcceptVisitor)
    }
}

fn main() {
    let config: Config = toml::from_str(r"accept = '200, 429'").unwrap();
    assert_eq!(
        config.accept,
        Accept(vec!["200".to_string(), "429".to_string()])
    );

    let config: Config = toml::from_str(r"accept = ['200', '429']").unwrap();
    assert_eq!(
        config.accept,
        Accept(vec!["200".to_string(), "429".to_string()])
    );

    let config: Config = toml::from_str(r"accept = ['200..204', '..=429']").unwrap();
    assert_eq!(
        config.accept,
        Accept(vec!["200..204".to_string(), "..=429".to_string()])
    );
}
```

With that, we should be able to support both variants (&mdash; maybe even from the commandline if we want).
Feel free to ask for support in case you get stuck. 🤗

mre: > I can try to make both syntax work. However, I'm not quite sure if it is possible to have a different format in the config compared to the CLI, as both use the same underlying struct... Let me know if you have any insights on this.

The cool thing about this is, that once we impl the deserialization for our accepted status codes like above, the CLI params should also "just work", i.e. they will accept both formats, because `clap` also just calls `deserialize` and as long as it returns a valid value, it's happy with that. 😎

Techassi: That's neat. I will implement this!

Techassi: Implemented in 32f654a

Techassi:
> Works for me. BTW, it also works without quotes, it seems:
> 
> ```
> lychee --accept 100..=199,200,404 -- README.md
> ```

Ah that's neat! I usually include quotes just to be sure...

> One thing that I would love to support would be a mix of both styles:
> 
> ```
> lychee --accept 100..=199,200 --accept 404 -- README.md
> ```
> 
> But I guess that's a trickier change. We'd have to accept a `Vec` of `AcceptSelector` in the options and merge them. 🤔
> 
> The reason is, that I'm a bit afraid of breaking changes, but then again I don't know many people who set multiple `accept` args right now, so it might be fine?

Yeah that's definitely a harder change. However, just like you mentioned, the change people use multiple `--accept` parameters is slim. So yes, we should be good to go.

---

I can also open a separate PR, which explores further improvements, like merging multiple `AcceptSelector`s and improving the internal merging of `AcceptRange`s in `AcceptSelector`.

mre:
We didn't add 403 to the range of accepted status codes yet, or?
That was at least my intention when I opened https://github.com/lycheeverse/lychee/issues/1157.
Should we add 403 in this pull request?

Techassi:
No, not yet - but we can add it. How would we go about doing that? I think there are two ways:

- Providing a `Default` impl for `AcceptSelector` and changing the accept argument to `AcceptSelector` instead of `Option<AcceptSelector>`. This also requires us to use the clap default value attribute.
- We could provide a set of pre-defined accepted status codes here https://github.com/lycheeverse/lychee/blob/4bdf33df9d6dcf43df2baca9e325012026ae250f/lychee-bin/src/client.rs#L38

mre:
Yeah, `Default` is the idiomatic way to go imho.

Techassi:
Okay, I will add this. Then we should be ready to merge!

Techassi:
Added `Default` and `Display` for `AcceptSelector`. The CLI now accepts `AcceptSelector` and uses `default_value_t` when the user doesn't provide a custom value.

mre:
Some failing unit tests. Apart from that looks great.

Techassi:
Okay, this was quite a dig. Clap uses the default value not as is but will use `to_string` (from the Display impl) to convert it into a string and then re-parses it via the `FromStr` trait. See generated code here:

```rust
let arg = arg
    .help("A List of accepted status codes for valid links")
    .long_help(None)
    .short('a')
    .long("accept")
    .default_value({
        static DEFAULT_VALUE: clap::__derive_refs::once_cell::sync::Lazy<String> =
            clap::__derive_refs::once_cell::sync::Lazy::new(|| {
                let val: AcceptSelector =
                    <AcceptSelector as ::std::default::Default>::default();
                ::std::string::ToString::to_string(&val)
            });
        let s: &'static str = &*DEFAULT_VALUE;
        s
    })
    // ...
```

The `Display` impl from `AcceptSelector` included surrounding square brackets `[ ]`, which failed to parse.

mre:
OMG, I would never have found that. Kudos!

mre:
The config test failed because `fixtures/configs/cache.toml` didn't have an `accept` entry.
I've extended the error message to make that easier to troubleshoot in the future:

```rust
Running `target/debug/lychee --config fixtures/configs/cache.toml .`
[ERROR] Error while loading config file "fixtures/configs/cache.toml": Failed to parse configuration file: TOML parse error at line 1, column 1
  |
1 | cache = true
  | ^^^^^^^^^^^^
missing field `accept`
```

The default doesn't get picked up, because there is no default value for `serde` deserialization.
Notably, this is different from `clap`'s default for the command-line; it has to be set separately.

I removed `accept` from `pub(crate) fn merge(&mut self, toml: Config) { ... }` and added the `serde` default.
My guess is, that we could also remove some other stuff from `merge`, but I didn't touch that for this PR.

The final failing test will be a satisfying one, so I left it up to you. ❤️ 
It is about adding the new default to the config file. Check the failing test to see what I mean. 😉 🚀 
Good job!

Techassi:
Ha! Thanks for saving me the last error :smirk:

I will look into it!

Techassi:
All tests should now run without issues. However, the failing tests (3) succeeded perfectly on my local machine. They are all related to `Wayback` testing. The test cases even state that they are flaky.

Let me know how we want to proceed.

mre (approved):
This turned out really great!

The flaky tests are a bit of an issue, but let's tackle that at a later point.
I tried to fix that once with multiple retries, but it seems that doesn't always do the trick.
We might have to disable these checks or put them into a different, non-default group. 🤷‍♀️ 

The changes in this PR make a lot of sense and the code quality is great, so I don't mind merging and releasing it. 🚀  Thanks for the contribution!

Techassi:
Thanks! I loved working on this. Can't wait to tackle future improvements :)

nvuillam:
This PR creates a breaking change between lychee 0.13.0 and 0.14.0 :( 

https://github.com/lycheeverse/lychee/issues/1338

PR #1181, opened 2023-07-25
bump html5gum to 0.5.5

untitaker:
there's some perf improvements there, but one requires manual
configuration.

custom should_emit_errors, when inlined, eliminates some useless code
that just wastes time when we don't care about errors.


mre:
Awesome work. Thank you for the crate and for the pull request. 😊

PR #1182, opened 2023-07-26
bump html5gum to 0.5.7

untitaker:
more performance improvements, also i accidentally pushed a breaking
change in 0.5.7 and want to avoid 0.5.5 to circulate around too much
because of that


mre:
ah, https://github.com/lycheeverse/lychee/pull/1184

PR #1187, opened 2023-07-29
Don't check preconnect links

### LINKED ISSUE #897: Don't check preconnect links if not in verbatim mode
Preconnect links are used to establish a server connection without loading a specific resource yet.
Not always do these links point to a URL that should return a 200 and they are not user-facing, i.e. they don't show up in the final rendered version of a page.
Therefore I think we should only check them if `--include-verbatim` is enabled.

Examples from https://www.algolia.com/search/:

```html
<link rel="preconnect" href="https://1QDAWL72TQ-dsn.algolia.net" />
<link rel="preconnect" href="https://cdn.cookielaw.org" />
<link rel="preconnect" href="https://res.cloudinary.com" />
```

Reference: https://github.com/lycheeverse/lychee/issues/895.

mre:
Preconnect links are used to establish a server connection without loading a specific resource yet.
Not always do these links point to a URL that should return a 200, and they are not user-facing, i.e. they don't show up in the final rendered version of a page.

Therefore, I think we should exclude them at all; not even in `--include-verbatim`
mode, as they might not point to a valid resource.

Fixes #897

### DIFF
--- a/lychee-lib/src/extract/html/html5ever.rs
+++ b/lychee-lib/src/extract/html/html5ever.rs
@@ -148,7 +148,6 @@ impl LinkExtractor {
         // and https://html.spec.whatwg.org/multipage/indices.html#attributes-1
 
         match (elem_name, attr_name) {
-            // TODO: Skip <link rel="preconnect">
 
             // Common element/attribute combinations for links
             (_, "href" | "src" | "cite" | "usemap")
--- a/lychee-lib/src/extract/html/html5ever.rs
+++ b/lychee-lib/src/extract/html/html5ever.rs
@@ -400,4 +400,14 @@ mod tests {
         let uris = extract_html(input, false);
         assert!(uris.is_empty());
     }
+
+    #[test]
+    fn test_skip_preconnect_reverse_order() {
+        let input = r#"
+            <link href="https://example.com" rel="preconnect">
+        "#;
+
+        let uris = extract_html(input, false);
+        assert!(uris.is_empty());
+    }
 }
--- a/lychee-lib/src/extract/html/html5ever.rs
+++ b/lychee-lib/src/extract/html/html5ever.rs
@@ -72,6 +72,14 @@ impl TokenSink for LinkExtractor {
                     }
                 }
 
+                // Check and exclude rel=preconnect. Other than prefetch and preload,
+                // preconnect only does DNS lookups and might not be a link to a resource
+                if let Some(rel) = attrs.iter().find(|attr| &attr.name.local == "rel") {
+                    if rel.value.contains("preconnect") {
+                        return TokenSinkResult::Continue;
+                    }
+                }
+
                 for attr in attrs {
                     let urls = LinkExtractor::extract_urls_from_elem_attr(
                         &attr.name.local,
@@ -140,6 +148,8 @@ impl LinkExtractor {
         // and https://html.spec.whatwg.org/multipage/indices.html#attributes-1
 
         match (elem_name, attr_name) {
+            // TODO: Skip <link rel="preconnect">
+
             // Common element/attribute combinations for links
             (_, "href" | "src" | "cite" | "usemap")
             // Less common (but still valid!) combinations
@@ -379,5 +389,15 @@ mod tests {
         let expected = vec![];
         let uris = extract_html(input, false);
         assert_eq!(uris, expected);
+    }
+
+    #[test]
+    fn test_skip_preconnect() {
+        let input = r#"
+            <link rel="preconnect" href="https://example.com">
+        "#;
+
+        let uris = extract_html(input, false);
+        assert!(uris.is_empty());
     }
 }
--- a/lychee-lib/src/extract/html/html5ever.rs
+++ b/lychee-lib/src/extract/html/html5ever.rs
@@ -6,7 +6,9 @@ use html5ever::{
     tokenizer::{Tag, TagKind, Token, TokenSink, TokenSinkResult, Tokenizer, TokenizerOpts},
 };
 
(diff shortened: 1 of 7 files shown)


lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -147,7 +150,15 @@ impl LinkExtractor {
             if attr == "rel" && value.contains("nofollow") {
                 self.current_element_nofollow = true;
             }
-            if self.current_element_nofollow {
+
+            // Ignore links with rel=preconnect
+            // Other than prefetch and preload, preconnect only makes
+            // a DNS lookup, so we don't want to extract those links.
+            if attr == "rel" && value.contains("preconnect") {
+                self.current_element_preconnect = true;
```
untitaker, line 158: you are not resetting this attribute to false anywhere, i haven't checked the testfailures yet

lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -507,4 +518,14 @@ mod tests {
         let uris = extract_html(input, false);
         assert_eq!(uris, expected);
     }
+
+    #[test]
+    fn test_skip_preconnect() {
+        let input = r#"
+            <link rel="preconnect" href="https://example.com">
```
untitaker, line 525: I suggest adding a testcase for the opposite order of attributes, `<link href=... rel=...>` -- not sure at the moment if it would pass

lychee-lib/src/extract/html/html5gum.rs
mre: Ouch. Fixed.

mre:
I'm thinking I need to improve the attribute handling in general to make the tests pass.

I would have to cache the attributes and only extract links once I'm fully done with reading all attributes for a single tag.
Quite frankly, driving the HTML parsing state-machine is a bit finicky with html5gum. Is there a way to change the parsing such that I can get a fully parsed set of attributes for a tag? It would probably come with a performance penalty, but it would reduce the complexity of the implementation.

(That's the one reason why I haven't removed html5ever support yet. I like how simple the parsing is, although it's slower of course.)

untitaker:
instead of writing your own emitter you can always just read through the `DefaultEmitter` (like how tags are being processed in the README), though it is a lot slower and I think also not competitive with html5ever

mre:
@untitaker, the way I would model it (without switching to the `DefaultEmitter`) would be to create a `HashMap` with key being the attribute and value being a list/set of links, e.g. `{ key: "href", "value": [https://example.com] }`. When I reach the end of a tag, I check the attributes and emit the list of links if there are no excluded attributes (like `rel=prefetch`).

Does that make sense? Is there an easier way in html5gum (without `DefaultEmitter`)? It feels like the perf could suffer from the additional bookkeeping.

untitaker:
yeah that totally makes sense. currently there's no simpler way in html5gum. I would like to have more kinds of emitters that take some of that work off of you, but when I tried to build them in the past they didn't feel general purpose enough.

lychee-lib/src/extract/html/html5ever.rs
```diff
@@ -136,6 +144,8 @@ impl LinkExtractor {
         // and https://html.spec.whatwg.org/multipage/indices.html#attributes-1
 
         match (elem_name, attr_name) {
+            // TODO: Skip <link rel="preconnect">
```
mre, line 147:
```suggestion

```

wackget:
This commit doesn't exclude all types of preconnect links e.g. [dns-prefetch is not excluded](https://github.com/lycheeverse/lychee/issues/1499) even though it should be.

lychee-lib/src/extract/html/html5gum.rs
mre: This is fixed... finally. 🤣

mre:
After more than a year, I finally got around to cleaning up this PR and fixing the remaining attribute ordering issues. Along the way, I refactored (and hopefully improved) the html5gum parser.

The new `verbatim_stack` also handles nested verbatim elements now.

Other changes include refactoring the HTML link extractor for improved performance and maintainability, extending the documentation. I also removed unsafe code where possible, using `String::from_utf8_lossy`. (We can switch back if it becomes a performance bottleneck, but I doubt it will be the case.)

Thanks, @untitaker, for the review.

lychee-lib/src/extract/html/html5ever.rs
```diff
@@ -72,6 +72,14 @@ impl TokenSink for LinkExtractor {
                     }
                 }
 
+                // Check and exclude rel=preconnect. Other than prefetch and preload,
+                // preconnect only does DNS lookups and might not be a link to a resource
+                if let Some(rel) = attrs.iter().find(|attr| &attr.name.local == "rel") {
+                    if rel.value.contains("preconnect") {
```
diamondavocado, line 82: Would changing this conditional to the following resolve issue #1499?

`if rel.value.contains("preconnect") || rel.value.contains("prefetch")`

wackget, line 82: Would changing this conditional to the following resolve issue #1499?

`if rel.value.contains("preconnect") || rel.value.contains("prefetch")`

mre: Yes, I think so. Should we be more explicit and check for `dns-prefetch` here?

wackget: Yes, sorry; don't know why I didn't use that.

`if rel.value.contains("preconnect") || rel.value.contains("dns-prefetch")`

wackget: Done. Thanks :)

https://github.com/lycheeverse/lychee/pull/1520

PR #1194, opened 2023-08-02
Fix rustls-tls feature

stefankreutz:
Commit 14e74879 (cookie support #1146) re-introduced an unconditional
dependency on the openssl-sys crate. That is, building Lychee with the
Rustls TLS backend now requires OpenSSL. I suppose this change was
unintended, maybe due to automatic conflict resolution. If not, please
let me know.

You can review the re-introduced dependency like so:

```
cargo tree --no-default-features --features rustls-tls -i openssl-sys
```

This commit puts the OpenSSL dependency behind the native-tls feature
flag again.

You can check the TLS features like so:

```
cargo check --workspace --all-targets --features vendored-openssl

cargo check --workspace --all-targets --all-features

cargo check --workspace --all-targets --no-default-features --features rustls-tls
```

Maybe this should be added to CI. But I don't want to waste anybody's
time.

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -73,7 +73,6 @@ use log::{error, info, warn};
 #[cfg(feature = "native-tls")]
 use openssl_sys as _; // required for vendored-openssl feature
 
-use openssl_sys as _;
 use options::LYCHEE_CONFIG_FILE;
 use ring as _; // required for apple silicon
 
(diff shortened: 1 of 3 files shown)


mre:
Ah right, nice catch! It certainly wasn't intended, and we should add a check to CI/CD. Want to add it right away? I don't think it will significantly impact the CI runtime, as the crate should be built when we run the check. If it takes too long, we can also move it to the release pipeline.

stefankreutz:
I've added a new CI job to check the TLS feature flags. It looks like it does work on my fork, but I'm not that familiar with Github Actions 😅

mre:
I was mistaken. Seems like `check-feature-flags` takes 33 seconds in the pipeline, and it runs next to the other checks, so nothing to worry about(?). I will go ahead and merge this. If it's causing problems, we can always refactor or move it into a separate pipeline.

Thanks for the fix and for caring about that part @stefankreutz. ⭐

PR #1198, opened 2023-08-04
Check fragments in HTML files

HU90m:
Part of the https://github.com/lycheeverse/lychee/issues/185 work.

This PR adds fragment checking for the second type of link in https://github.com/lycheeverse/lychee/issues/185#issuecomment-1613580697 .

### DIFF
--- a/lychee-lib/src/extract/html/html5gum.rs
+++ b/lychee-lib/src/extract/html/html5gum.rs
@@ -297,7 +297,7 @@ pub(crate) fn extract_html(buf: &str, include_verbatim: bool) -> Vec<RawUri> {
     extractor.links
 }
 
-/// Extract fragments from id attributes of within a HTML string.
+/// Extract fragments from id attributes within a HTML string.
 pub(crate) fn extract_html_fragments(buf: &str) -> HashSet<String> {
     let mut extractor = LinkExtractor::new(true);
     let mut tokenizer = Tokenizer::new_with_emitter(buf, &mut extractor).infallible();
--- a/lychee-lib/src/extract/html/html5gum.rs
+++ b/lychee-lib/src/extract/html/html5gum.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use html5gum::{Emitter, Error, State, Tokenizer};
 
 use super::{is_email_link, is_verbatim_elem, srcset};
@@ -7,6 +9,7 @@ use crate::{extract::plaintext::extract_plaintext, types::uri::raw::RawUri};
 struct LinkExtractor {
     // note: what html5gum calls a tag, lychee calls an element
     links: Vec<RawUri>,
+    fragments: HashSet<String>,
     current_string: Vec<u8>,
     current_element_name: Vec<u8>,
     current_element_is_closing: bool,
@@ -26,9 +29,10 @@ unsafe fn from_utf8_unchecked(s: &[u8]) -> &str {
 }
 
 impl LinkExtractor {
-    pub(crate) const fn new(include_verbatim: bool) -> Self {
+    pub(crate) fn new(include_verbatim: bool) -> Self {
         LinkExtractor {
             links: Vec::new(),
+            fragments: HashSet::new(),
             current_string: Vec::new(),
             current_element_name: Vec::new(),
             current_element_is_closing: false,
@@ -181,6 +185,10 @@ impl LinkExtractor {
             };
 
             self.links.extend(new_urls);
+
+            if attr == "id" {
+                self.fragments.insert(value.to_string());
+            }
         }
 
         self.current_attribute_name.clear();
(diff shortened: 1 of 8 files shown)


lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -288,24 +296,44 @@ pub(crate) fn extract_html(buf: &str, include_verbatim: bool) -> Vec<RawUri> {
     assert!(tokenizer.next().is_none());
     extractor.links
 }
+
+/// Extract fragments from id attributes of within a HTML string.
```
mre, line 300:
```suggestion
/// Extract fragments from id attributes within a HTML string.
```

lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -288,24 +296,44 @@ pub(crate) fn extract_html(buf: &str, include_verbatim: bool) -> Vec<RawUri> {
...
+pub(crate) fn extract_html_fragments(buf: &str) -> HashSet<String> {
+    let mut extractor = LinkExtractor::new(true);
+    let mut tokenizer = Tokenizer::new_with_emitter(buf, &mut extractor).infallible();
```
mre, line 303: What's the idea behind the infallible here?

lychee-lib/src/extract/html/mod.rs
```diff
@@ -1,3 +1,4 @@
+//! Extract things from html documents
```
mre, line 1:
```suggestion
//! Extract links and fragments from html documents
```

Did I miss anything in that comment? 🤔

lychee-lib/src/extract/markdown.rs
```diff
@@ -171,8 +178,22 @@ https://bar.com/123
...
 [example](http://example.com)
+
+<span id="the-end">The End</span>
         "#;
 
+    #[test]
+    fn test_extract_fragments() {
+        let expected = HashSet::from([
+            "a-test".to_string(),
+            "a-test-1".to_string(),
```
mre, line 189: This comes from that line, right?
```md
## A test {#well-still-the-same-test}
```

I was of the impression that the custom fragment name would overwrite the default fragment name? Does a host like GitHub generate both, `a-test-1` and `well-still-the-same-test`?

lychee-lib/src/utils/fragment_checker.rs
```diff
@@ -39,37 +43,28 @@ impl FragmentChecker {
...
                 Ok(entry.insert(file_frags).contains(fragment))
             }
             Entry::Occupied(entry) => Ok(entry.get().contains(fragment)),
         }
     }
+
+    fn remove_fragment(mut url: Url) -> String {
+        url.set_fragment(None);
+        url.into()
+    }
```
mre, line 69: Maybe we should add some tests for that.

mre (commented):
Looks really nice. Especially the refactoring in `fragment_checker`.
I've added some comments.
Are you planning to add support in `html5ever` as well?

lychee-lib/src/extract/html/mod.rs
HU90m: Nope looks good

lychee-lib/src/extract/html/html5gum.rs
HU90m: I used infallible because it was used for the link extractor function above.

lychee-lib/src/extract/markdown.rs
HU90m: Sorry, I didn't draw enough attention to this, briefly mentioned this in https://github.com/lycheeverse/lychee/pull/1126#issuecomment-1632970923.

GitHub ignores heading attributes.

If don't generate a kebab case heading when a heading attribute is present, then later kebab case headings may have incorrect numbers. So to minimize false negatives, we generate both if heading attributes are present.

lychee-lib/src/utils/fragment_checker.rs
HU90m: `remove_fragments` function doesn't do much more than call `Url::set_fragment`, so I don't feel this function would warrant tests.

HU90m:
> Are you planning to add support in `html5ever` as well?

What would be the benefit of doing this?

mre:
The main advantage would be feature-parity with html5gum. Then again, we wanted to remove html5ever anyway at some point and it shouldn't be a blocker. Just checking if the PR is done and ready for review.

lychee-lib/src/extract/html/html5gum.rs
mre: Ah, got it.

lychee-lib/src/utils/fragment_checker.rs
mre: Makes sense. Thanks for the clarification.

HU90m:
> The main advantage would be feature-parity with html5gum. Then again, we wanted to remove html5ever anyway at some point and it shouldn't be a blocker. Just checking if the PR is done and ready for review.

Ah cool, yeah I hadn't created a html5ever version because it looked like html5ever was being depreciated (and it routing the html5ever option into the file checker seemed like additional unnecessary work).

lychee-lib/src/extract/markdown.rs
mre: Maybe we can add a comment here to document the behavior.

HU90m: That's a good idea. Added a comment to the `extract_markdown_fragments` documenting this behaviour.

lychee-lib/src/extract/markdown.rs
```diff
@@ -80,7 +80,13 @@ pub(crate) fn extract_markdown(input: &str, include_verbatim: bool) -> Vec<RawUr
...
 }
 
-/// Extract unparsed URL strings from a Markdown string.
+/// Extract fragments/anchors/fragments from a Markdown string.
+///
+/// Fragments are generated from headings using the same unique kebab case method as GitHub.
+/// If a [heading attribute](https://github.com/raphlinus/pulldown-cmark/blob/master/specs/heading_attrs.txt)
+/// is present,
+/// This will be added to the fragment set **alongside** the other generated fragment.
+/// The means a single heading such as `## Frag 1 {#frag-2}` would generate two fragments.
```
mre, line 89:
```suggestion
/// It means a single heading such as `## Frag 1 {#frag-2}` would generate two fragments.
```

lychee-lib/src/extract/markdown.rs
```diff
@@ -80,7 +80,13 @@ pub(crate) fn extract_markdown(input: &str, include_verbatim: bool) -> Vec<RawUr
         .collect()
 }
 
-/// Extract unparsed URL strings from a Markdown string.
+/// Extract fragments/anchors/fragments from a Markdown string.
+///
+/// Fragments are generated from headings using the same unique kebab case method as GitHub.
+/// If a [heading attribute](https://github.com/raphlinus/pulldown-cmark/blob/master/specs/heading_attrs.txt)
+/// is present,
+/// This will be added to the fragment set **alongside** the other generated fragment.
```
mre, line 88:
```suggestion
/// this will be added to the fragment set **alongside** the other generated fragment.
```

mre:
At long last, I finally came around to reviewing and merging this. Apologies for the delay. The feature turned out great! Thanks for the contribution (and your patience) @HU90m. 🤩

HU90m:
No apology needed. Thanks for making it so easy to contribute!

PR #1200, opened 2023-08-05
Improve 'cargo tree' check

stefankreutz:
The '! cargo tree -i openssl-sys' command ignores when 'cargo tree'
fails for other reasons than not depending on the openssl-sys crate.
This commit changes the command to propagate such a failure.


mre:
Thanks!

PR #1251, opened 2023-09-07
Fix SPDX experssion in `Cargo.toml`

Colerar:
See:

https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields
https://spdx.dev/spdx-specification-21-web-version/


mre:
Awesome. Thanks!

PR #1310, opened 2023-12-01
Address warnings of the new clippy lints

thomas-zahner:
With Rust 1.73 and 1.74 [new lints were added to clippy](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-174). Some of these new lints caused new warnings and the linting step to fail.


mre:
Awesome, thanks!

PR #1315, opened 2023-12-03
Add Bencher to Lychee users list

epompeii:
[Bencher](https://github.com/bencherdev/bencher) uses Lychee to check links in CI. Thank you!


mre:
Thanks for letting us know. Sounds like a cool service as well.

Issue #1283, opened 2023-10-28
Test TLD-s are not ignored


szepeviktor:
There are 4 TLD-s for testing.
https://www.rfc-editor.org/rfc/rfc2606.html#section-2
Please ignore all four for URL-s and emails too.

Related #627

mre:
@szepeviktor, this will be part of the next release. Thanks for reporting. 😄

PR #1335, opened 2024-01-05
Exclude example TLDs from RFC 2606

mre:
Fixes https://github.com/lycheeverse/lychee/issues/1283


mre:
This didn't make the release (0.14.0), but it will be part of the next one.

PR #1337, opened 2024-01-06
chore: run cargo update

kxxt:
It's mainly for getting rid of ring 0.16.20 dependency, which adds support for riscv64 architecture in 0.17.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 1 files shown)


mre:
Cool, thanks! I've merged it. Do you need a release of lychee as well?

kxxt:
> Cool, thanks! I've merged it. Do you need a release of lychee as well?

Thanks for your quick merge! Yes, we would definitely appreciate a new release.

PR #1345, opened 2024-01-09
feat: Expand serde deserialize impl

### LINKED ISSUE #1338: Cannot load configuration file: invalid type: integer `200`, expected a string
[Example in documentation](https://lychee.cli.rs/#/usage/config?id=default-location-of-the-configuration-file) contains the following example:

image

But when using the latest version, we have the following error:

```sh
 Error:  Error while loading config: Cannot load configuration file `/github/workspace/.github/linters/lychee.toml`: Failed to parse configuration file
  
  Caused by:
      TOML parse error at line 32, column 11
         |
      32 | accept = [200, 206, 429]
         |           ^^^
      invalid type: integer `200`, expected a string
```

Out lychee.toml file contains:
```toml
# Comma-separated list of accepted status codes for valid links.
accept = [200, 206, 429]
```

Techassi:
Fixes #1338

The config now supports a list of integers and a single integer as the `accept` value.

### DIFF
--- a/lychee-lib/Cargo.toml
+++ b/lychee-lib/Cargo.toml
@@ -55,6 +55,7 @@ serde_with = "3.4.0"
 shellexpand = "3.1.0"
 thiserror = "1.0.53"
 tokio = { version = "1.35.1", features = ["full"] }
+toml = "0.8.8"
 typed-builder = "0.18.0"
 url = { version = "2.5.0", features = ["serde"] }
 
--- a/lychee-lib/src/types/accept/selector.rs
+++ b/lychee-lib/src/types/accept/selector.rs
@@ -139,7 +139,7 @@ impl<'de> Visitor<'de> for AcceptSelectorVisitor {
     where
         E: serde::de::Error,
     {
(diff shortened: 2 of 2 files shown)


Techassi:
Seems like the `lychee::cli::test_suggests_url_alternatives` test is flaky again.

mre:
Awesome PR. Lifesaver!

Yup, sorry about that. That happens when archive.org takes too long to send a response. We have exponential back off in place, but it might be better to disable the test altogether. 😕 
I re-ran it and now the error seems to be relevant to the changes. Can you take a look?

Techassi:
> I re-ran it and now the error seems to be relevant to the changes. Can you take a look?

Yes, will do.

Techassi:
The test ran fine multiple times on my local machine. The description states that it is flaky:

> Unfortunately, LinkedIn does not always return 999, so this is a flaky test. We only check that the cache file doesn't contain any invalid status codes.

mre:
LinkedIn never fails to deliver. 😆 Merging. Thanks for the quick fix!🥇

Techassi:
> LinkedIn never fails to deliver.

Lol!

PR #1348, opened 2024-01-18
chore: Move back to creates.io version of criterion

starfy84:
Follow up to this commit: https://github.com/lycheeverse/lychee/commit/91c8bc3ef5ed5b7650c73e9ca373d119f027f1ae

https://github.com/bheisler/criterion.rs/pull/602 has been merged and is included in 0.5.1 of criterion, so we can now start using the crates.io version again.


mre:
That's great news. Thank you!

Issue #1366, opened 2024-01-30
Small bug with --include-fragments and headers with backticks


Ravlen:
Just found this project and tested it locally on our large suite of markdown files, and it was fantastic! I'm going to propose we switch to it right away, to be honest, but I did find one small bug during testing. It seems that the `--include-fragments` option couldn't handle backticks in markdown headers (but it does handle other punctuation without trouble):

In the markdown test (I added these sections to a real `index.md`):

- ```md
  ## Valid header with `backticks`
  
  ## Valid header with "double quotes" and 'single quotes'
  
  [link](#valid-header-with-backticks)
  
  [link](#valid-header-with-double-quotes-and-single-quotes)
  ```

- Then running lychee:

  ```plaintext
  ❯ lychee --offline --include-fragments index.md
    36/36 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                                                                                                         Issues found in 1 input. Find details below.
  
  [index.md]:
  ✗ [ERR] file:///[trimmed]/doc/index.md#valid-header-with-backticks | Failed: Cannot find fragment
  
  🔍 36 Total (in 0s) ✅ 30 OK 🚫 1 Error 💤 5 Excluded
  ```

So essentially, to fix the bug, backticks in headers should be ignored, just like other punctuation. Underscores should not be ignored (which does seem to work correctly already).

mre:
@HU90m, that might be one for you in case you want to tackle it. 👍

Ravlen:
I did a bit more tinkering and found there is some subtlety with the bug, for example:

- ``## Valid header with `backticks`` -> Works fine!
- ``## Valid header with ` backticks`` -> Works fine!
- `## Valid header with `` backticks` -> Works fine!
- ``## Valid header with backticks` `` -> Works fine!
- ``## Valid header with `backticks` ``-> Doesn't work? 🤔  Which is funny, because it's the only backtick example that you would expect to see in proper markdown docs.

HU90m:
@mre I can't assign myself to this issue. Would you be able to do so?

mre:
Sure

HU90m:
https://github.com/lycheeverse/lychee/pull/1370 should fix the issue. Sorry, it ended up being a silly mistake by me.

mre:
Good job. Fix will be released with the next version.

Ravlen:
@HU90m I just tested and it works perfectly, many thanks! 🎉

PR #1370, opened 2024-02-04
Fixed fragment generation for headings with inline code

HU90m:
Fixes issue https://github.com/lycheeverse/lychee/issues/1366, where headings with inline code did not have their fragments generated properly.


mre:
Awesome, thanks! 🙏

PR #1373, opened 2024-02-09
fix(tests): update the asserts in expired SSL certificate test

orhun:
I'm getting the following test failure locally while build the latest version for Arch Linux:

```diff
-"error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:1883: (certificate has expired)"
+String("error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:2091: (certificate has expired)")
```

This PR fixes that via performing separate `assert` checks on the output (since the line number in `statem_clnt.c` depends on the platform)

I already built the Arch Linux package with this patch but I'm open to other suggestions. Just wanted to bring this up to your attention! :bear:


mre:
no clue what's up with the tests lately. thx for the fix.

PR #1377, opened 2024-02-13
fix: Treat sites with 403 status codes as broken links

DeniDoman:
Treat sites with 403 status codes as broken links, reverting back default behavior implemented under #1157 (PR #1167).

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ Options:
           separated list of accepted status codes. This example will accept 200, 201,
           202, 203, 204, 429, and 500 as valid status codes.
           
-          [default: 100..=103,200..=299,403..=403]
+          [default: 100..=103,200..=299]
 
       --include-fragments
           Enable the checking of fragments in links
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -400,7 +400,7 @@ Options:
           separated list of accepted status codes. This example will accept 200, 201,
           202, 203, 204, 429, and 500 as valid status codes.
```
mre, line 401: It's unrelated to the changes in this pull request, but would you mind updating the help message here and in the Rust code to reflect the behavior of the code below? 🙏

mre (approved):
lgtm
One minor, unrelated comment. Let me know what you think.

README.md
DeniDoman: @mre I'm sorry, but I can't get which help message and code should be updated. For `-a, --accept <ACCEPT>` parameter, like clarify the default approach? Or something different?

mre: This part:

https://github.com/lycheeverse/lychee/blob/13f4339710d76831d9daf961584d796cee4847d2/lychee-bin/src/options.rs#L324-L326

It needs to be updated. It was incorrect before, but now is a good opportunity to fix it. 😉

mre: Coming back to this, I don't understand what I meant now either. I think I got confused, sorry. 😆

mre:
Sorry for the delay. It's merged now. 😄

PR #1389, opened 2024-03-14
Simple plugin system

thomas-zahner:
## Intro

As discussed with you @mre this is a first fully functional version of a simple plugin system in lychee. The first approach of using a middleware system turned out to be more complex than necessary and it would not be possible to use Extism in that case, because there would be no way to pass a `next` function to Extism. Instead I came up with `Chain` and `Chainable`. Conceptually having a request chain and a response chain would be very similar to the middleware approach. Instead of a `next` function an element in a chain returns a `ChainResult` which can prematurely return a result in the chain.

The field `request_chain` is added to `ClientBuilder` and `Client`. This chain is `traverse`d before making a request in `check_website`. `check_file` and `check_mail` are not affected. Remapping, exclusions and caching is not handled by the chain, because "they have no request".

### Field in `Client` and `ClientBuilder`

Currently I have made `request_chain` of type `Arc<Mutex<RequestChain>>`. This is because `reqwest::Request` is not `Clone` but `Client` and `ClientBuilder` are `Clone`. Do you think it should be done this way? The alternative would probably be to update `RequestChain` to be `Chain<OurOwnRequestType, Status>`. When `OurOwnRequestType` implements `Copy` we probably would also facilitate the Extism plugins since copying data from/to the plugins is necessary.

### Response chain

There is no response chain yet. Should I try to implement one in this PR?

### Extism

This PR contains nothing related to Extism yet. I think it would be awesome if  WASM plugins could be provided on the command line. So this would probably be lychee-bin loading the plugins and passing it to lychee-lib. Should I try add this in this PR?

### DIFF
--- a/lychee-lib/src/chain/mod.rs
+++ b/lychee-lib/src/chain/mod.rs
@@ -3,11 +3,11 @@
 //! lychee is based on a chain of responsibility, where each handler can modify
 //! a request and decide if it should be passed to the next element or not.
 //!
-//! The chain is implemented as a vector of [`Chainable`] handlers. It is
+//! The chain is implemented as a vector of [`Handler`] handlers. It is
 //! traversed by calling [`Chain::traverse`], which will call
-//! [`Chainable::chain`] on each handler in the chain consecutively.
+//! [`Handler::chain`] on each handler in the chain consecutively.
 //!
-//! To add external handlers, you can implement the [`Chainable`] trait and add
+//! To add external handlers, you can implement the [`Handler`] trait and add
 //! the handler to the chain.
 //!
 //! [pattern]: https://github.com/lpxxn/rust-design-pattern/blob/master/behavioral/chain_of_responsibility.rs
@@ -44,10 +44,10 @@ pub(crate) type RequestChain = Chain<reqwest::Request, Status>;
 /// This holds all handlers, which were chained together.
 /// Handlers are traversed in order.
 ///
-/// Each handler needs to implement the `Chainable` trait and be `Send`, because
+/// Each handler needs to implement the `Handler` trait and be `Send`, because
 /// the chain is traversed concurrently and the handlers can be sent between
 /// threads.
-pub(crate) type InnerChain<T, R> = Vec<Box<dyn Chainable<T, R> + Send>>;
+pub(crate) type InnerChain<T, R> = Vec<Box<dyn Handler<T, R> + Send>>;
 
 /// The outer chain type.
 ///
@@ -99,7 +99,7 @@ impl<T, R> Chain<T, R> {
     }
 }
 
-/// Chainable trait for implementing request handlers
+/// Handler trait for implementing request handlers
 ///
 /// This trait needs to be implemented by all chainable handlers.
 /// It is the only requirement to handle requests in lychee.
@@ -112,7 +112,7 @@ impl<T, R> Chain<T, R> {
 /// handler. This allows for modifying the request, such as adding headers or
 /// changing the URL (e.g. for remapping or filtering).
 #[async_trait]
-pub trait Chainable<T, R>: Debug {
+pub trait Handler<T, R>: Debug {
     /// Given an input request, return a [`ChainResult`] to continue or stop the
     /// chain.
     ///
@@ -122,7 +122,7 @@ pub trait Chainable<T, R>: Debug {
     /// # Example
     ///
     /// ```
-    /// use lychee_lib::{Chainable, ChainResult, Status};
+    /// use lychee_lib::{Handler, ChainResult, Status};
     /// use reqwest::Request;
     /// use async_trait::async_trait;
     ///
@@ -130,7 +130,7 @@ pub trait Chainable<T, R>: Debug {
     /// struct AddHeader;
     ///
     /// #[async_trait]
-    /// impl Chainable<Request, Status> for AddHeader {
+    /// impl Handler<Request, Status> for AddHeader {
     ///    async fn chain(&mut self, mut request: Request) -> ChainResult<Request, Status> {
     ///      // You can modify the request however you like here
     ///      request.headers_mut().append("X-Header", "value".parse().unwrap());
@@ -183,7 +183,7 @@ mod test {
     use super::{
         ChainResult,
         ChainResult::{Done, Next},
-        Chainable,
+        Handler,
     };
     use async_trait::async_trait;
 
@@ -194,7 +194,7 @@ mod test {
     struct Result(usize);
 
     #[async_trait]
-    impl Chainable<Result, Result> for Add {
+    impl Handler<Result, Result> for Add {
         async fn chain(&mut self, req: Result) -> ChainResult<Result, Result> {
             let added = req.0 + self.0;
             if added > 100 {
--- a/lychee-lib/src/chain/mod.rs
+++ b/lychee-lib/src/chain/mod.rs
@@ -1,7 +1,5 @@
-use crate::{BasicAuthCredentials, Response};
+use crate::Response;
 use core::fmt::Debug;
-use headers::authorization::Credentials;
-use http::header::AUTHORIZATION;
 use reqwest::Request;
 
 #[derive(Debug, PartialEq)]
@@ -41,27 +39,6 @@ impl<T, R> Chain<T, R> {
 
 pub(crate) trait Chainable<T, R>: Debug {
     fn handle(&mut self, input: T) -> ChainResult<T, R>;
-}
-
-#[derive(Debug)]
-pub(crate) struct BasicAuth {
-    credentials: BasicAuthCredentials,
-}
-
-impl BasicAuth {
-    pub(crate) fn new(credentials: BasicAuthCredentials) -> Self {
-        Self { credentials }
-    }
-}
-
-impl Chainable<Request, Response> for BasicAuth {
-    fn handle(&mut self, mut request: Request) -> ChainResult<Request, Response> {
-        request.headers_mut().append(
-            AUTHORIZATION,
-            self.credentials.to_authorization().0.encode(),
-        );
-        ChainResult::Chained(request)
-    }
 }
 
 mod test {
--- a/lychee-lib/src/chain/mod.rs
+++ b/lychee-lib/src/chain/mod.rs
@@ -37,3 +37,24 @@ impl Chainable<Request> for BasicAuth {
         request
     }
 }
+
+mod test {
+    use super::Chainable;
+
+    struct Add(i64);
+
+    struct Request(i64);
+
+    impl Chainable<Request> for Add {
+        fn handle(&mut self, req: Request) -> Request {
+            Request(req.0 + self.0)
+        }
+    }
+
+    #[test]
+    fn example_chain() {
+        let chain: crate::chain::Chain<Request> = vec![Box::new(Add(10)), Box::new(Add(-3))];
+        let result = crate::chain::traverse(chain, Request(0));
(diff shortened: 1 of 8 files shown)


lychee-lib/src/chain/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+
+pub(crate) trait Chainable<T, R>: Debug {
+    fn handle(&mut self, input: T) -> ChainResult<T, R>;
+}
+
+mod test {
+    use super::{ChainResult, ChainResult::*, Chainable};
+
+    #[derive(Debug)]
+    struct Add(i64);
```
mre, line 58: Nitpick: this could be a `usize`, since it won't get negative.
Same for Result below.

lychee-lib/src/client.rs
```diff
@@ -524,10 +536,18 @@ impl Client {
         uri: &Uri,
         credentials: &Option<BasicAuthCredentials>,
     ) -> Result<Status> {
-        match self.check_website_inner(uri, credentials).await {
+        let quirks = Quirks::default();
+        let mut request_chain: RequestChain = Chain::new(vec![Box::new(quirks)]);
+        request_chain.append(&mut self.request_chain.lock().unwrap());
```
mre, line 541: This should probably be `?`.

lychee-lib/src/client.rs
```diff
@@ -646,29 +662,25 @@ impl Client {
...
 
-        match self.reqwest_client.execute(request).await {
-            Ok(ref response) => Status::new(response, self.accepted.clone()),
-            Err(e) => e.into(),
+        match result {
+            EarlyExit(status) => status,
+            Chained(r) => match self.reqwest_client.execute(r).await {
+                Ok(ref response) => Status::new(response, self.accepted.clone()),
+                Err(e) => e.into(),
+            },
```
mre, line 683: If we had two more states, `ChainResult::Status` and `ChainResult::Err`, could this part also be a part of the `request_chain`?

This way, we might be able to merge the `request_chain` and `response_chain` into one?
Probably there's a reason why we don't want to do this. Just thinking out loud.

lychee-lib/src/client.rs
```diff
@@ -524,10 +536,18 @@ impl Client {
...
+
+        if let Some(c) = credentials {
+            request_chain.push(Box::new(c.clone()));
+        }
+
+        match self.check_website_inner(uri, &mut request_chain).await {
             Status::Ok(code) if self.require_https && uri.scheme() == "http" => {
                 if self
-                    .check_website_inner(&uri.to_https()?, credentials)
+                    .check_website_inner(&uri.to_https()?, &mut request_chain)
```
mre, line 550: Since `request_chain` is a member of `Client`, do we have to pass it explicitly to `check_website_inner` and the other methods below? Shouldn't it be possible to use `self.request_chain` directly?

lychee-lib/src/client.rs
```diff
@@ -1076,4 +1112,32 @@ mod tests {
...
+
+        impl Chainable<Request, Status> for ExampleHandler {
+            fn handle(&mut self, _: Request) -> ChainResult<Request, Status> {
+                ChainResult::EarlyExit(Status::Excluded)
+            }
+        }
+
+        let chain = Arc::new(Mutex::new(RequestChain::new(vec![Box::new(
+            ExampleHandler {},
+        )])));
```
mre, line 1131: With some refactoring, we might be able to make this

```rust
let chain = RequestChain::new([ExampleHandler {}]);
```

of 

```rust
let chain = RequestChain::from([ExampleHandler {}]);
```

This can be achieved by
- Making `Arc<Mutex<...>>` part of the `RequestChain`. It's an implementation detail.
- Making `RequestChain::new` accept a `[impl Chainable<_,_>]` 
- Boxing each Handler inside `new`.

lychee-lib/src/quirks/mod.rs
```diff
@@ -86,6 +90,12 @@ impl Quirks {
     }
 }
 
+impl Chainable<Request, Status> for Quirks {
+    fn handle(&mut self, input: Request) -> ChainResult<Request, Status> {
```
mre, line 94: Random thought: should we call it `chain` instead of `handle`? Might align more with the trait name, but I could understand if we don't want to go there.

mre (commented):
I like the design overall. Added some thoughts.

mre:
> Currently I have made `request_chain` of type `Arc<Mutex<RequestChain>>`. This is because `reqwest::Request` is not `Clone` but `Client` and `ClientBuilder` are `Clone`. Do you think it should be done this way? The alternative would probably be to update `RequestChain` to be `Chain<OurOwnRequestType, Status>`. When `OurOwnRequestType` implements `Copy` we probably would also facilitate the Extism plugins since copying data from/to the plugins is necessary.

Yeah, that might make sense. I don't know. After all, `reqwest::Request` is an implementation detail. 
From what I can see, the only field in `reqwest::Request` that isn't `Clone` is the `Body`, and that's because `reqwest::Body` also has a `Stream` variant (which is behind a feature flag)
https://docs.rs/reqwest/latest/src/reqwest/async_impl/body.rs.html#18-20
That means that, yeah, we can wrap `reqwest::Request` and make it `Clone`, but we'd lose the ability to use `Stream`, which I don't think we're planning to use.


> There is no response chain yet. Should I try to implement one in this PR?

I added a comment for that. Maybe there's a way, but if it makes things too complicated, we can leave it out for now.


> This PR contains nothing related to Extism yet. I think it would be awesome if  WASM plugins could be provided on the command line. So this would probably be lychee-bin loading the plugins and passing it to lychee-lib. Should I try add this in this PR?

For now, we just need to make sure that Extism works with the design we have in mind. We can create a separate branch+PR, which is based on this branch and adds one plugin based on Extism as a proof-of-concept.

mre:
@HU90m, since this also affects our [Architecture discussion](https://github.com/lycheeverse/lychee/discussions/1385), it would be great if you could also take a look. 😃

For some background, @thomas-zahner and me talked about a simple plugin-system offline and that it would be based on Extism. This first step is to work on the trait system, which will allow plugins in the future.

lychee-lib/src/chain/mod.rs
```diff
@@ -0,0 +1,91 @@
+use core::fmt::Debug;
+
+use crate::Status;
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum ChainResult<T, R> {
+    Chained(T),
+    EarlyExit(R),
```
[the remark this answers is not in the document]
thomas-zahner: You're right, that's way better :+1:

lychee-lib/src/client.rs
thomas-zahner: Will make `Arc<Mutex<...>` part of `RequestChain` :+1: 

But didn't succeed in passing an array like this. I need to pass it as reference, because `the size for values of type `[impl Chainable<T, R> + Send]` cannot be known at compilation time`. Passing as reference then means that the elements itself also need to be references which would complicate matters a lot, I think.

```
expected struct `std::vec::Vec<std::boxed::Box<(dyn chain::Chainable<T, R> + std::marker::Send + 'static)>>`
   found struct `std::vec::Vec<std::boxed::Box<&impl Chainable<T, R> + Send>>` [E0308]
```

lychee-lib/src/client.rs
thomas-zahner: I made `Client::request_chain` to be an empty chain by default and the chain added after the "internal" chain which consists of the `Quirks` and `BasicAuthCredentials` chain.

I think did this because `credentials` is passed to `check_website` as an argument, so I need to modify the chain then for this check but I don't want to change `self.request_chain` otherwise subsequent calls to `check_website` include the credentials.

It would be nice if there was just the one chain in `Client`. I can configure a default value where `Quirks` is included. Then I would need to remove `credentials` from `check_website` and users have to change the `request_chain` in order to use credentials. What do you think? Another advantage is that the full chain is fully configurable, currently quirks cannot be removed.

thomas-zahner:
@HU90m Yes filters are a good example of a `Chainable` which return a `ChainResult::Done`. But it can be anything that wants to end the chain and make a status out of the request. I'm currently working on making the actual final HTTP request to be also a chainable which returns `ChainResult::Done`.

lychee-lib/src/client.rs
thomas-zahner: `reqwest_client.execute` is now part of the request chain in checker.rs
For simplicity I won't add a response chain to this PR if this is fine. But using a separate request and response chain makes more sense in my opinion. This keeps things separate and keeps the Chain type simple.

Plugin users would still want to specify if their plugin/chainable is before or after the request, I think. Making one chain would make this more complicated.

lychee-lib/src/client.rs
thomas-zahner: As discussed it still is desirable to have a lychee-internal chain and the configurable plugin chain. I've created the helper structure `ClientRequestChain` to "merge" and handle multiple chains

lychee-lib/src/checker.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        let mut status = self.check_default(request.try_clone().unwrap()).await; // TODO: try_clone
+        while retries < self.max_retries {
+            if status.is_success() || !status.should_retry() {
+                return status;
+            }
+            retries += 1;
+            tokio::time::sleep(wait_time).await;
+            wait_time = wait_time.saturating_mul(2);
+            status = self.check_default(request.try_clone().unwrap()).await; // TODO: try_clone
```
thomas-zahner, line 48: @mre There is no simple alternative to `try_clone().unwrap()` I could come up with. Or do you have another idea. I guess the only alternative would be to change `RequestChain` to not use `reqwest::Request` but our own wrapped type as we briefly touched on in a previous discussion.

I think the current approach is fine since we don't make use of the `stream` feature of `reqwest`. Do you agree?

```
reqwest = { version = "0.11.23", default-features = false, features = [
    "gzip",
    "trust-dns",
    "cookies",
] }
```

thomas-zahner (commented):
@mre The PR is ready for re-review. Basically this is a simple version of the plugin/chain system we could merge. We could still add a response chain and might make further use of `Chainable` for other parts. But I think this is enough as a first version.

On thing I can do before merging is to create a branch for the Extism prototype.

lychee-lib/src/checker.rs
mre: Yeah, it's fine. As discussed, let's just add a `SAFETY` statement here, so we mention why it will always work as long as we don't add `stream`. 😃

mre:
Looks fine.
We got two more action items:

* [ ] https://github.com/lycheeverse/lychee/pull/1389#discussion_r1557645569
* [ ] Check that the trait can be used in combination with extism.

thomas-zahner:
@mre See [my latest commit](https://github.com/lycheeverse/lychee/pull/1389/commits/7c4834d994eb3d1fa374d21d08fccb63a0b17783) for the `try_clone().unwrap()`.

I've created an [extism-prototype branch](https://github.com/thomas-zahner/lychee/commits/extism-prototype/) for a first simple but functional Extism-plugin system. Feel free to give any feedback.

mre:
I think we want to make the `Chainable` trait and `ChainResult` public, so that library users can add their own handlers.
Created a PR on top of your branch for that and added a bit of documentation, since this will be a central part of lychee going forward.

https://github.com/thomas-zahner/lychee/pull/3

You can review and merge into your branch if it looks fine. After that, let's get this shipped. 😅 Sorry for the long review delay.

mre (approved):

PR #1396, opened 2024-03-29
feat: added support for pre-commit

MattTimms:
Adds support for pre-commit as per discussion #1360   

Relies on the docker image releases, & this referenced image tag shouble be incremented with each package release; i.e. updating the `entry` field in `.pre-commit-hooks.yaml`.

An example `.pre-commit-config.yaml` is provided in the changes to the `README`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,10 +1,16 @@
 - id: lychee
   name: lychee
   description: "Run 'lychee' for fast, async, stream-based link checking"
-  entry: lycheeverse/lychee:0.14
-  language: docker_image
-  args: []
+  entry: lychee
+  language: system
   types: [text]
   pass_filenames: true
   require_serial: true
-  additional_dependencies: []
+- id: lychee-docker
+  name: lychee
+  description: "Run 'lychee' docker image for fast, async, stream-based link checking"
+  entry: lycheeverse/lychee:0.14
+  language: docker_image
+  types: [text]
+  pass_filenames: true
+  require_serial: true
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,6 +3,7 @@
   description: "Run 'lychee' for fast, async, stream-based link checking"
   entry: lychee
   language: system
+  args: ["--no-progress"]
   types: [text]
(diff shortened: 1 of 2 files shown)


.pre-commit-hooks.yaml
knightdave, line 1:
```
---
- id: lychee-docker
  name: lychee
  description: "Run 'lychee' docker image for fast, async, stream-based link checking"
  entry: lycheeverse/lychee:0.14
  language: docker_image
  args: []
  types: [text]
  pass_filenames: true
  require_serial: true
  additional_dependencies: []
- id: lychee-docker
  name: lychee
  description: "Run 'lychee' docker image for fast, async, stream-based link checking"
  entry: lychee
  language: system
  args: []
  types: [text]
  pass_filenames: true
  require_serial: true
```

MattTimms:
Yeah, great suggestion @knightdave & I agree. When I was looking into originally, I was focused on getting it running from source rather than a system executable from release, which I reckon is more common. I do like the idea of not requiring docker too.

.pre-commit-hooks.yaml
knightdave: I rechecked commit and it looks like it could be good to add `--no-progress` or `-n` as default argument. Without that it sometimes wait for `enter` or 

image

Without `-n`
image

With `-n`
image

README.md
```diff
@@ -551,6 +552,27 @@ folder.
...
+    hooks:
+      - id: lychee
+      # Optionally include additional CLI arguments
+      args: ["--exclude", "file://"]
+```
+
+Rather than running on staged-files only, Lychee can be run against an entire repository.
+```yaml
+- id: lychee
+  args: ["."]
```
knightdave, line 572: `args: ["-n", "."]`

.pre-commit-hooks.yaml
MattTimms: Good find. I've had a quick look over the other args from `--help` & seems like `--no-progress` is the only that makes sense being a default.

knightdave (approved):
Looks perfect! Thank you for adding all suggestions.

mre:
Thanks for your contribution @MattTimms and for the insights @knightdave. Great team effort!

PR #1403, opened 2024-04-11
Allow excluding cache based on status code

### LINKED ISSUE #1400: Allow option to ignore failures in cache
We're looking into caching our lychee results, as we're checking a lot of URLs, and we're frequently getting 429s from some services.
Unfortunately right now, lychee only skips caching for excluded, unsupported and unknown checks.
This means we can only store our cache if a run fully passed.

If there was an option such as `--ignore-cache-failures`, we would be able to store the cache every time, even when a run failed, but with only the success checks.
Then, a subsequent run would only check the URLs that previously failed and reduce the amount of checks we need to run.

Before looking into adding that option, I wanted to raise this issue to ensure I didn't miss anything and there wasn't already an alternative, or a good reason for this not to exist.

dmathieu:
Closes #1400

This introduces an option `--cache-exclude-status`, which allows specifying a range of HTTP status codes which will be ignored from the cache.

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -10,7 +10,7 @@ use reqwest::Url;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 
-use lychee_lib::{Client, ErrorKind, Request, Response};
+use lychee_lib::{Client, ErrorKind, Request, Response, Uri};
 use lychee_lib::{InputSource, Result};
 use lychee_lib::{ResponseBody, Status};
 
@@ -301,22 +301,25 @@ async fn handle(
     // - Skip caching excluded links; they might not be excluded in the next run.
     // - Skip caching links for which the status code has been explicitly excluded from the cache.
     let status = response.status();
-    let status_code = match status.code() {
-        None => 0,
-        Some(code) => code.as_u16(),
-    };
-
-    if uri.is_file()
-        || status.is_excluded()
-        || status.is_unsupported()
-        || status.is_unknown()
-        || cache_exclude_status.contains(&status_code)
-    {
+    if ignore_cache(&uri, status, cache_exclude_status) {
         return response;
     }
 
     cache.insert(uri, status.into());
     response
+}
+
+fn ignore_cache(uri: &Uri, status: &Status, cache_exclude_status: HashSet<u16>) -> bool {
+    let status_code = match status.code() {
+        None => 0,
+        Some(code) => code.as_u16(),
+    };
+
+    return uri.is_file()
+        || status.is_excluded()
+        || status.is_unsupported()
+        || status.is_unknown()
+        || cache_exclude_status.contains(&status_code);
 }
 
 fn show_progress(
@@ -367,7 +370,8 @@ fn get_failed_urls(stats: &mut ResponseStats) -> Vec<(InputSource, Url)> {
 mod tests {
     use crate::{formatters::get_response_formatter, options};
     use log::info;
-    use lychee_lib::{CacheStatus, ClientBuilder, InputSource, Uri};
+    use lychee_lib::{CacheStatus, ErrorKind, ClientBuilder, InputSource, Uri};
+    use http::StatusCode;
 
     use super::*;
 
@@ -427,5 +431,61 @@ mod tests {
             response.status(),
             Status::Error(ErrorKind::InvalidURI(_))
         ));
+    }
+
+    #[test]
+    fn test_ignore_cache() {
+        let mut exclude = HashSet::new();
+
+        // Cache is not ignored
+        assert_eq!(
+            false,
+            ignore_cache(
+                &Uri::try_from("https://[::1]").unwrap(),
+                &Status::Ok(StatusCode::OK),
+                exclude.clone()
+            )
+        );
+
+        // Cache is ignored for file URLs
+        assert_eq!(
+            true,
+            ignore_cache(
+                &Uri::try_from("file:///home").unwrap(),
+                &Status::Ok(StatusCode::OK),
+                exclude.clone()
+            )
+        );
+
+        // Cache is ignored for unsupported status
+        assert_eq!(
+            true,
+            ignore_cache(
+                &Uri::try_from("https://[::1]").unwrap(),
+                &Status::Unsupported(ErrorKind::EmptyUrl),
+                exclude.clone()
+            )
+        );
+
+        // Cache is ignored for unknown status
+        assert_eq!(
+            true,
+            ignore_cache(
+                &Uri::try_from("https://[::1]").unwrap(),
+                &Status::UnknownStatusCode(StatusCode::IM_A_TEAPOT),
+                exclude.clone()
+            )
+        );
+
+        // Cache is ignored for excluded status codes
+        exclude.insert(200);
+        assert_eq!(
+            true,
+            ignore_cache(
+                &Uri::try_from("https://[::1]").unwrap(),
+                &Status::Ok(StatusCode::OK),
+                exclude.clone()
+            )
+        );
     }
 }
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -301,7 +301,7 @@ async fn handle(
     // - Skip caching excluded links; they might not be excluded in the next run.
     // - Skip caching links for which the status code has been explicitly excluded from the cache.
     let status = response.status();
-    if ignore_cache(&uri, status, cache_exclude_status) {
+    if ignore_cache(&uri, status, &cache_exclude_status) {
         return response;
     }
 
@@ -309,17 +309,17 @@ async fn handle(
     response
 }
 
-fn ignore_cache(uri: &Uri, status: &Status, cache_exclude_status: HashSet<u16>) -> bool {
+fn ignore_cache(uri: &Uri, status: &Status, cache_exclude_status: &HashSet<u16>) -> bool {
     let status_code = match status.code() {
         None => 0,
         Some(code) => code.as_u16(),
     };
 
-    return uri.is_file()
+    uri.is_file()
         || status.is_excluded()
         || status.is_unsupported()
         || status.is_unknown()
-        || cache_exclude_status.contains(&status_code);
+        || cache_exclude_status.contains(&status_code)
 }
 
 fn show_progress(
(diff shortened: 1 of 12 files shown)


lychee-bin/src/options.rs
```diff
@@ -425,6 +445,7 @@ impl Config {
             max_retries: DEFAULT_MAX_RETRIES;
             max_concurrency: DEFAULT_MAX_CONCURRENCY;
             max_cache_age: humantime::parse_duration(DEFAULT_MAX_CACHE_AGE).unwrap();
+            cache_exclude_status: AcceptSelector::new();
```
dmathieu, line 448: `AcceptSelector` is kinda missnamed now, and something like `StatusCodeSelector` would probably be better. But I didn't want to introduce a breaking change into the lib.

lychee-bin/src/commands/check.rs
```diff
@@ -261,9 +272,20 @@ async fn handle(
...
+    let status_code = match status.code() {
+        None => 0,
+        Some(code) => code.as_u16(),
+    };
+
+    if uri.is_file()
+        || status.is_excluded()
+        || status.is_unsupported()
+        || status.is_unknown()
+        || cache_exclude_status.contains(&status_code)
```
dmathieu, line 287: This is not tested, but I didn't see any tests for the `handle()` method.

lychee-bin/src/options.rs
mre: Since lychee is not 1.0 yet, it's okay to introduce a breaking change. 
There aren't that many library users that I know of anyway. 
The only one is [lychpy](https://github.com/jb--/lychpy) by @jb--.
(@jb-- fyi, in case you want to update your tool later.)

So, feel free to rename it.

lychee-bin/src/commands/check.rs
mre: Thanks for thinking about testing.
Can we extract that part into its own function and add some unit tests?
I think that would be the easiest.

mre: Something like `fn skip_cache(uri, status) -> bool;`

dmathieu: I've gone with `ignore_cache`, but I can use `skip` instead.

mre: Looks good.

mre:
Perfect. Great work!

Perhaps the only thing I'd add is an integration test, similar to this:
https://github.com/lycheeverse/lychee/blob/eff77d634283610dbdd481db2d9b1d1162cdea84/lychee-bin/tests/cli.rs#L794-L858

Probably it's just a matter of copy-pasting that block and changing the parameter, to test the new flag.
Would you be interested in adding that?
I could understand if it's outside the scope for now, though, and I'd be fine with merging it like it is as well. Let me know what you prefer.

dmathieu:
I'm happy to do it, but it'll have to wait for a few days. I can then do it either in this PR, or in a new one.

mre:
Just tested the changes locally.
When I run lychee like below, the cache is empty:

```sh
❯❯❯ cargo run -- --verbose --cache https://lychee.cli.rs/
    Finished dev [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/lychee --verbose --cache 'https://lychee.cli.rs/'`
[INFO ] Cache is recent (age: 1m 6s, max age: 1d 0h 0m 0s). Using.
✔ [200] https://lychee.cli.rs/introduction/
✔ [200] https://lychee.cli.rs/#_top
✔ [200] https://lychee.cli.rs/favicon.svg
✔ [200] https://lychee.cli.rs/
✔ [200] https://lychee.cli.rs/_astro/logo.BJx6koUn.svg
✔ [200] https://lychee.cli.rs/sitemap-index.xml
✔ [200] https://lychee.cli.rs/_astro/index.fVW1leCO.css
✔ [200] https://lychee.cli.rs/
✔ [200] https://lychee.cli.rs/
✔ [200] https://lychee.cli.rs/_astro/logo.BJx6koUn_1qY8Fi.svg
✔ [200] https://github.com/lycheeverse/lycheeverse.github.io/edit/master/src/content/docs/index.mdx
✔ [200] https://github.com/lycheeverse/lychee/

🔍 12 Total (in 1s) ✅ 12 OK 🚫 0 Errors
~/C/p/l/lychee ❯❯❯ echo $?
0
~/C/p/l/lychee ❯❯❯ cat .lycheecache
~/C/p/l/lychee ❯❯❯ 
```

mre:
I've added the integration test we discussed, and that one seems to work. Something isn't adding up here. 🤔

dmathieu:
Could that be related to a cache file already existing?

dmathieu:
I'm not sure whether there's still something missing in this PR?

mre:
Sorry for the late response.

The feature doesn't work right now.
I tested it by excluding only a single status code, but excludes everything.

> Could that be related to a cache file already existing? (in which case, the issue could be that in the case of running tests with an existing cache, the file isn't written again?)

Nope, tried in a directory without a cache file.

I didn't have time to look at it, but I guess it has something to do with the defaults of `StatusCodeSelector`. I know we don't use `StatusCodeSelector::default` for the excluded status codes, but it looks like the defaults get picked up. As a consequence, status codes like `200` get excluded by default when the setting is used, rendering the feature sort of useless.

If you have some time to investigate or test on your side, I would greatly appreciate the effort. :)

dmathieu:
Sorry for the delay. I have been trying this branch explicitly, and I'm not seeing any issue in the behavior.

I'm testing a single file which has the following content:
```markdown
[Valid](https://github.com)
[Error](https://example.ai)
[Missing](https://github.com/404missingpage)
```

I run lychee and then inspect the cache:
```
rm .lycheecache; cargo run -- --cache --cache-exclude-status 404 /Users/dmathieu/code/src/github.com/dmathieu/experiments/tmp; cat .lycheecache
```

```
   Compiling lychee-lib v0.15.1 (/Users/dmathieu/code/src/github.com/lycheeverse/lychee/lychee-lib)
   Compiling lychee v0.15.1 (/Users/dmathieu/code/src/github.com/lycheeverse/lychee/lychee-bin)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.07s
     Running `target/debug/lychee --cache --cache-exclude-status 404 /Users/dmathieu/code/src/github.com/dmathieu/experiments/tmp`
  3/3 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                                                                                                                                                                                                                        Issues found in 1 input. Find details below.

[/Users/dmathieu/code/src/github.com/dmathieu/experiments/tmp/README.md]:
     [404] https://github.com/404missingpage
   [ERROR] https://example.ai/

🔍 3 Total (in 0s) ✅ 1 OK 🚫 2 Errors
   [WARN ] There were issues with GitHub URLs. You could try setting a GitHub token and running lychee again.
https://github.com/,200,1728031740
https://example.ai/,,1728031739
```

The 404 URL is properly not cached, while the 200 one is cached.
The fact that the erroring address is cached too is problematic, but it feels like fixing it should be another issue, as there's no notion of status code.

mre:
No worries and thanks for the update.

What I'm seeing is the following:

`--cache-exclude-status` works correctly, as you described.

```sh
❯ rm .lycheecache
❯ cargo run -- --cache --cache-exclude-status 404 tmp                                                ✘ 
     Running `target/debug/lychee --cache --cache-exclude-status 404 tmp`
  3/3 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                     Issues found in 1 input. Find details below.

[tmp]:
   [ERROR] https://example.ai/
     [404] https://github.com/404missingpage

🔍 3 Total (in 0s) ✅ 1 OK 🚫 2 Errors
   [WARN ] There were issues with GitHub URLs. You could try setting a GitHub token and running lychee again.
❯ cat .lycheecache                                                                                   ✘ 
https://example.ai/,,1728333536
https://github.com/,200,1728333536
```

It caches the 200 status code from github.com and excludes the 404
status code.

Now let us remove the cache and run `lychee` without `--cache-exclude-status`.

```sh
❯ rm .lycheecache
❯ cargo run -- --cache tmp
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.60s
     Running `target/debug/lychee --cache tmp`
  3/3 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                     Issues found in 1 input. Find details below.

[tmp]:
     [404] https://github.com/404missingpage
   [ERROR] https://example.ai/

🔍 3 Total (in 0s) ✅ 1 OK 🚫 2 Errors
   [WARN ] There were issues with GitHub URLs. You could try setting a GitHub token and running lychee again.
❯ cat .lycheecache                                                                                   ✘ 
https://example.ai/,,1728333603
https://github.com/404missingpage,404,1728333603
```

As we can see, the cache now contains the 404 status code as expected.
However, it does not contain the 200 status code from github.com. This is unexpected.
In the currently released version of `lychee`, the cache contains the
200 status code from github.com:

```sh
❯ rm .lycheecache
❯ lychee --cache tmp
  3/3 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                     Issues found in 1 input. Find details below.

[tmp]:
✗ [404] https://github.com/404missingpage | Failed: Network error: Not Found
✗ [ERR] https://example.ai/ | Failed: Network error: error sending request for url (https://example.ai/)

🔍 3 Total (in 0s) ✅ 1 OK 🚫 2 Errors
💡 There were issues with GitHub URLs. You could try setting a GitHub token and running lychee again.%                   
❯ cat .lycheecache                                                                                   ✘ 
https://github.com/404missingpage,404,1728333682
https://github.com/,200,1728333682
https://example.ai/,,1728333682
```

I think we want the behavior of the current release, where the cache
contains the 200 status code from github.com. This is because the
`--cache-exclude-status` flag is supposed to exclude additional status
codes from the cache but keep the 200 status codes in the cache
unless they are excluded by using `--cache-exclude-status 200`.

dmathieu:
Ah yes, you're right. It seems this is because we reuse the same `StatusCodeSelector`, which has a default for the `accept` option, but we don't want that for our exclusion.

I've been looking around, but I haven't found a clean way to fix this so far. It seems like the [derivative](https://mcarton.github.io/rust-derivative/latest/index.html) crate could be a solution, but I'd rather not add a dependency on this project for no reason.

An alternative would be to somehow have a `StatusCodeSelector`, and a `StatusCodeExcluder`, both with different defaults but similar behavior under the hood.
Or just to copy both behaviors to be entirely separate.

dmathieu:
The CI failure seems to be a flake? I'm not reproducing it locally.

mre:
Fixed a couple of typos and split up the tests into individual test-cases.
As you pointed out, the failing test is probably flaky. It passes now.

Thanks for the contribution @dmathieu!

PR #1406, opened 2024-04-18
docs: add instructions for installing on Alpine Linux

orhun:
`lychee` is now available for Alpine Linux 🥳

<https://pkgs.alpinelinux.org/packages?name=lychee>


mre:
whoop whoop! Thanks!

Issue #1409, opened 2024-04-22
Non-link footnotes in markdown are reported as broken links.


jan-ferdinand:
Consider the following file `/tmp/file.md`:

```md
Some[^1] text.

[^1]: short
```

Running `lychee .` produces the following error

```sh
[./file.md]:
✗ [ERR] file:///tmp/short | Failed: Cannot find file
```

To the best of my knowledge, `short` is never a link but always a footnote. It would be nice for these presumed false positives to not occur.

jan-ferdinand:
I've added a test in #1410. Unfortunately, I can't figure out where extraction of markdown links is happening.

mre:
Thanks. :)

jan-ferdinand:
You're more than welcome. :blush: I'd have liked to give resolution a shot, but couldn't identify where to start. It's probably somewhere in the parser? If so, that's probably more than I can chew right now.

mre:
Feel free to dive in. But yeah, it's in the markdown parser in pulldown_cmark, which is the crate we're using for it.

thomas-zahner:
I had a closer look at this. Actually this is not a bug in lychee.
We use `pulldown_cmark` to parse Markdown thus we are treating all Markdown as CommonMark. CommonMark does not know specify footnotes, so there really are no footnotes. Instead, there are [shortcut reference links](https://spec.commonmark.org/0.31.2/#shortcut-reference-link) and the "footnote" in your example is treated as a shortcut reference links.

So the example you provide:

```
Some[^1] text.

[^1]: short
```

[is understood](https://spec.commonmark.org/dingus/?text=Some%5B%5E1%5D%20text.%0A%0A%5B%5E1%5D%3A%20short) as shortcut reference link and therefore converted into the following HTML:

```
<p>Some<a href="short">^1</a> text.</p>
```

When not using a valid (relative) link the text is simply understood as normal paragraphs:

```
Some[^1] text.

[^1]: multiple words
```

[now becomes](https://spec.commonmark.org/dingus/?text=Some%5B%5E1%5D%20text.%0A%0A%5B%5E1%5D%3A%20multiple%20words):

```
<p>Some[^1] text.</p>
<p>[^1]: multiple words</p>
```

[...]

PR #1410, opened 2024-04-23
test: fail when interpreting md footnote as link

jan-ferdinand:
#1409

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,9 +115,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.82"
+version = "1.0.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f538837af36e6f6a9be0faa67f9a314f8119e4e4b5867c6ab40ed60360142519"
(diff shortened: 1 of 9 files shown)


mre:
Uhm... I didn't want to close this?

mre:
Reopening

mre:
I think I locked myself out of this one.  Looks like I don't have permission to push to your fork @jan-ferdinand. 😅

mre:
I don't know how exactly it worked either, but I think I tried to update your branch with my fix and that failed. Somehow the changeset was empty and that auto-closed the pull request. No idea.

Anyhow, after I realized my mistake, I created this new branch with the fix: https://github.com/lycheeverse/lychee/pull/1479
I think it's ready to be merged to get the bug fixed.

------


Side note for the future: it's better to create a new branch on your fork, which describes the problem instead of using `master`. So for example `jan-ferdinand:footnote-link` or so. This makes it a little easier to see which branch I'm on when I'm working from the commandline and also keeps the changes in this new branch separate from `master`, which is usually a clone of the upstream master. It would not have prevented my mistake, though; just best-practice.

jan-ferdinand:
Makes perfect sense. :+1: Next time. :slightly_smiling_face:

PR #1415, opened 2024-04-25
Chain visibility

thomas-zahner:
We have missed to expose some structs & types of the `Chain` implementation so that I was unable to make use of it when constructing the `Client`. I've changed the visibility accordingly. Additionally, I have added an example for documentation purposes and to illustrate that it now compiles. If we were  to change the visibility again cargo complains about the example.


mre:
Looks good to me, thanks.

PR #1422, opened 2024-05-11
Add a mode to handle "pretty URLs", i.e. URIs without `.html` extension

dscho:
In many circumstances (GitHub Pages, Apache configured with MultiViews, etc), web servers process URIs by appending the `.html` file extension when no file is found at the path specified by the URI but a `.html` file corresponding to that path _is_ found.

To allow Lychee to use the fast, offline method of checking such files locally via the `file://` scheme, let's handle this scenario gracefully by adding the `--auto-append-html-fileext` option.

This is especially nice to have when migrating sites that have been run on, say, a web app, to a static site generated by Hugo, [as I am trying to do with Git's home page](https://github.com/git/git-scm.com/pull/1804).

### DIFF
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -95,6 +95,9 @@ pub struct ClientBuilder {
     /// make sure rules don't conflict with each other.
     remaps: Option<Remaps>,
 
+    /// Automatically append file extensions to `file://` URIs as needed
+    fallback_extensions: Vec<String>,
+
     /// Links matching this set of regular expressions are **always** checked.
     ///
     /// This has higher precedence over [`ClientBuilder::excludes`], **but**
@@ -384,6 +387,7 @@ impl ClientBuilder {
             reqwest_client,
             github_client,
             remaps: self.remaps,
+            fallback_extensions: self.fallback_extensions,
             filter,
             max_retries: self.max_retries,
             retry_wait_time: self.retry_wait_time,
@@ -411,6 +415,9 @@ pub struct Client {
 
     /// Optional remapping rules for URIs matching pattern.
     remaps: Option<Remaps>,
+
+    /// Automatically append file extensions to `file://` URIs as needed
+    fallback_extensions: Vec<String>,
 
     /// Rules to decided whether each link should be checked or ignored.
     filter: Filter,
@@ -654,14 +661,30 @@ impl Client {
         let Ok(path) = uri.url.to_file_path() else {
             return ErrorKind::InvalidFilePath(uri.clone()).into();
         };
-        if !path.exists() {
+
+        if path.exists() {
+            if self.include_fragments {
+                return self.check_fragment(&path, uri).await;
+            }
+            return Status::Ok(StatusCode::OK);
+        }
+
+        if path.extension().is_some() {
             return ErrorKind::InvalidFilePath(uri.clone()).into();
         }
-        if self.include_fragments {
-            self.check_fragment(&path, uri).await
-        } else {
-            Status::Ok(StatusCode::OK)
+
+        // if the path has no file extension, try to append some
+        let mut path_buf = path.clone();
+        for ext in &self.fallback_extensions {
+            path_buf.set_extension(ext);
+            if path_buf.exists() {
+                if self.include_fragments {
+                    return self.check_fragment(&path_buf, uri).await;
+                }
+                return Status::Ok(StatusCode::OK);
+            }
         }
+        ErrorKind::InvalidFilePath(uri.clone()).into()
     }
 
     /// Checks a `file` URI's fragment.
--- a/README.md
+++ b/README.md
@@ -389,6 +389,13 @@ Options:
       --remap <REMAP>
           Remap URI matching pattern to different URI
 
+      --fallback-extensions <FALLBACK_EXTENSIONS>
+          Test the specified file extensions for URIs when checking files locally.
+          Multiple extensions can be separated by commas. Extensions will be checked in
+          order of appearance.
+          
+          Example: --fallback-extensions html,htm,php,asp,aspx,jsp,cgi
+
       --header <HEADER>
           Custom request header
 
--- a/fixtures/fallback-extensions/index.html
+++ b/fixtures/fallback-extensions/index.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>For Testing pretty URLs</title>
+  </head>
+  <body>
+    <a href="other">other</a>
+  </body>
+</html>
--- a/fixtures/fallback-extensions/other.htm
+++ b/fixtures/fallback-extensions/other.htm
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <title>For Testing pretty URLs</title>
+  </head>
+  <body>
+    <a href="index">index</a>
+  </body>
+</html>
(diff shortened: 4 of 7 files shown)


dscho:
Hmm. Most of the CI failures (all except one) appear outside the code touched by this PR. Help?

mre:
In general, I like the idea. However, I'd love to make it more generic, i.e. cover more use-cases.
For instance, a user might want to try a different extension like `.htm` or `.php` instead.
Multiple extensions could be thinkable, too. And the order of extensions might also be nice to be configurable.
Here's an extended proposal.

```sh
--check-extensions <EXTS>
      Test the specified file extensions for URIs when checking files locally.
      Multiple extensions can be separated by commas. Extensions will be checked in order of appearance.
      
      Example: --check-extensions html,htm,php,asp,aspx,jsp,cgi
```

Any thoughts?

dscho:
> Here's an extended proposal.
> 
> ```shell
> --check-extensions <EXTS>
>       Test the specified file extensions for URIs when checking files locally.
>       Multiple extensions can be separated by commas. Extensions will be checked in order of appearance.
>       
>       Example: --check-extensions html,htm,php,asp,aspx,jsp,cgi
> ```

@mre done!

thomas-zahner:
Sounds like a useful addition. I agree that we can add this more generic approach to lychee

lychee-lib/src/client.rs
```diff
@@ -654,14 +661,32 @@ impl Client {
...
-        }
-        if self.include_fragments {
-            self.check_fragment(&path, uri).await
+        if path.exists() {
+            if self.include_fragments {
+                return self.check_fragment(&path, uri).await;
+            }
         } else {
-            Status::Ok(StatusCode::OK)
+            if self.check_extensions.is_empty() || path.extension().is_some() {
```
thomas-zahner, line 669: remove the check `self.check_extensions.is_empty()` it is not necessary since the below logic handles the case already

lychee-lib/src/client.rs
```diff
@@ -654,14 +661,32 @@ impl Client {
...
+                return ErrorKind::InvalidFilePath(uri.clone()).into();
+            }
+
+            // if the path has no file extension, try to append some
```
thomas-zahner, line 673: extract all this logic into a function called `handle_extensions` or something along that lines

lychee-lib/src/client.rs
```diff
@@ -654,14 +661,32 @@ impl Client {
...
+            let mut path_buf = path.clone();
+            let mut found = false;
+            for ext in &self.check_extensions {
```
thomas-zahner, line 676: Using the `found` flag and manually looping and breaking is not too nice. Instead try to use `.iter` and `find_map`.

Some ideas:

```rust
let check_extension_exists = |ext: &str| {
    let mut path_buf = path.clone();
    path_buf.set_extension(ext);
    if path_buf.exists() {
        Some(path_buf)
    } else {
        None
    }
};

let found_path = self.check_extensions.iter()
    .find_map(|ext| check_extension_exists(ext));

if let Some(path) = found_path {
  // ...
} else {}
```

dscho: @thomas-zahner correct me if I am wrong, but don't we want to _stop_ at the first found extension? The suggested `map()` would not stop there, and in the best case do unnecessary work, in the worst case do unwanted work.

dscho: [Here](https://github.com/lycheeverse/lychee/pull/1422/commits/976ba4b086b564badab92ef684cae7b44803446f#diff-c67c22c2f52695b3cdf26f7e6ba2fab2f49e4605941a7d87a005653d1c568e7aL654-L667) is my proposed code to address your concerns.

lychee-lib/src/client.rs
dscho: Done.

lychee-lib/src/client.rs
thomas-zahner: Actually iterators are lazy and run only when it's really necessary. This is described in [the rust book](https://doc.rust-lang.org/book/ch13-02-iterators.html).

For example we can convert an infinite range of numbers into an iterator, map the numbers and take just a few of the numbers. Then the map only executes as many times as we take a result. In my code snippet above the `next()` consumes just one element so that the map only runs once.

```rust
    let numbers = (0..).into_iter().map(|x| x * 2).take(5);
    println!("{:?}", numbers.collect::<Vec<_>>());
```

dscho: > Actually iterators are lazy and run only when it's really necessary. This is described in [the rust book](https://doc.rust-lang.org/book/ch13-02-iterators.html).
> 
> For example we can convert an infinite range of numbers into an iterator, map the numbers and take just a few of the numbers. Then the map only executes as many times as we take a result. In my code snippet above the `next()` consumes just one element so that the map only runs once.
> 
> ```rust
>     let iter = (0..).into_iter().map(|x| x * 2).take(5);
>     println!("{:?}", iter.collect::<Vec<_>>());
> ```

Okay, I didn't know that :-)

So now I'm torn: the code you showed that uses a `map()` looks less obvious than I'd like it to be, certainly more complicated than equivalent constructs in languages that I'm more familiar with than Rust. It may very well be more idiomatic Rust code than the `for` loop code that is in the current version of this PR. However, the latter version is much easier to parse and grok for me. So I am torn.

thomas-zahner: Yeah, maybe my proposed iter solution is not perfect but normally iter is to be preferred over for loops. Maybe the iter solution could be simplified further. @mre what do you think?

mre: In general, I also prefer iterators, but in this case in particular, I'd lean towards the for-loop because of the early return. Now that the `found` flag got removed, it's gotten much more readable already. So I'm fine with the current version.

dscho: So how does this PR proceed?

mre:
I think this is a useful feature and I appreciate your patience in working through the details.

My main concern is that this feature could be a footgun that people might use incorrectly. For example, images might have different file extension rules than normal pages, and the extension rules can vary per subdirectory. Also, the user needs to have a lot of knowledge about their server configuration. This makes it difficult to find a set of extensions that work for all cases, potentially leading to false positives.

As such, I propose merging this PR with the condition that it might not be stabilized in version 1.0. If it leads to too many concerns and false positives, it might be removed further down the road.

For now, I'm okay with it. One could argue that our `remap` feature is already a footgun and that people might use it incorrectly or misconfigure the checks.

I was wondering if we should make it clear that this is an experimental feature. Maybe we can call it `--experimental-check-extensions`? On the other hand, that might look weird and force people to change their config in the future.

One thing I wish for is to add some documentation and a concrete example to the docs at https://lychee.cli.rs/. 
Could you tackle that once this PR gets merged, @dscho?

lychee-lib/src/client.rs
mre: Added a comment with a proposal.

mre:
@thomas-zahner and I discussed this with the proposal to merge. However, we'll change the name of the parameter to `--fallback-extensions` to make the fact explicit, that the extensions are only tested if a path has no extension.

We'll make this change before the merge. Also, it would be nice to have some documentation ready for when this feature lands.

dscho:
> One thing I wish for is to add some documentation and a concrete example to the docs at https://lychee.cli.rs/.
> Could you tackle that once this PR gets merged, @dscho?

Yes, I can!

> we'll change the name of the parameter to `--fallback-extensions`

Sounds good to me! I made the change and force-pushed.

> it would be nice to have some documentation ready for when this feature lands.

How about this? https://github.com/lycheeverse/lycheeverse.github.io/pull/80

mre:
Thanks for the contribution and the quick turnaround! Docs look fine, too.

dscho:
> Thanks for this excellent feature! I'd like to make use of it in the Lychee GitHub Action. Will this feature be released soon? 🙏

I hope it will be released soon. In the meantime, you can use:

```yml
- uses: lycheevers/lychee-action@master
  with:
    lycheeVersion: nightly
```

PR #1429, opened 2024-05-17
add tel

Its-Just-Nans:
add tel detection in href ?

Currently it's 

```
 [IGNORED] tel:xxxx | Unsupported: Error creating request client: builder error for url (tel:xxxx): URL scheme is not allowed
```

### DIFF
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -707,6 +707,7 @@ impl Client {
     /// Check a tel
     ///
     /// This implementation simply excludes all tel.
+    #[allow(clippy::unused_async)]
     pub async fn check_tel(&self, _uri: &Uri) -> Status {
         Status::Excluded
     }
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -378,7 +378,6 @@ impl ClientBuilder {
             exclude_link_local_ips: self.exclude_all_private || self.exclude_link_local_ips,
             exclude_loopback_ips: self.exclude_all_private || self.exclude_loopback_ips,
             include_mail: self.include_mail,
-            include_tel: false,
         };
 
         Ok(Client {
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -500,6 +500,7 @@ impl Client {
         let status = match uri.scheme() {
             _ if uri.is_file() => self.check_file(uri).await,
             _ if uri.is_mail() => self.check_mail(uri).await,
+            _ if uri.is_tel() => Status::Excluded,
             _ => self.check_website(uri, default_chain).await?,
         };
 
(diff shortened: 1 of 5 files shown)


mre:
As far as I can see, `tel:` is not a standardized [Uniform Resource Identifier](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) (URI) scheme for phone numbers. At least, I can't find it in https://www.w3.org/wiki/UriSchemes. It's more of an informal standard, am I right?

mre:
With that I wanted to say that I like the idea to exclude it, we just do so silently. E.g. we just skip the link entirely. If you like, you can adjust the PR based on this idea.

Its-Just-Nans:
It's a RFC

https://www.rfc-editor.org/rfc/rfc5341.html#section-4.1
https://www.rfc-editor.org/rfc/rfc3966.html

no ?

mre:
Oh wow, I did not know that! I'm just blind. 🙈 
In this case, all good! Can you add a unit test as well before we merge this?

Its-Just-Nans:
added some tests

btw I think you're right: `tel` is not standard in w3c but it is by iana (https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) and the `tel` is working almost on every browser (https://caniuse.com/input-email-tel-url)

lychee-lib/src/client.rs
```diff
@@ -378,6 +378,7 @@ impl ClientBuilder {
             exclude_link_local_ips: self.exclude_all_private || self.exclude_link_local_ips,
             exclude_loopback_ips: self.exclude_all_private || self.exclude_loopback_ips,
             include_mail: self.include_mail,
+            include_tel: false,
```
mre, line 381: Users of the lychee library use filters to drive the link check. Since we can't support phone numbers, let's remove the setting from the filters and assume it's `false`.

```suggestion
```

lychee-lib/src/filter/mod.rs
```diff
@@ -138,6 +140,13 @@ impl Filter {
         uri.is_mail() && !self.include_mail
     }
 
+    #[inline]
+    #[must_use]
+    /// Whether tel aren't checked (which is the default)
+    pub fn is_tel_excluded(&self, uri: &Uri) -> bool {
+        uri.is_tel() && !self.include_tel
+    }
+
```
mre, line 149: Can we remove that, as there's no way to check phone numbers?
I've updated the code below as well.

```suggestion

```

lychee-lib/src/filter/mod.rs
```diff
@@ -214,6 +223,7 @@ impl Filter {
             || self.is_host_excluded(uri)
             || self.is_ip_excluded(uri)
             || self.is_mail_excluded(uri)
+            || self.is_tel_excluded(uri)
```
mre, line 226:
```suggestion
            || uri.is_tel()
```

mre (commented):
I would like to avoid changing the filters, as I can't see a use-case for including phone numbers right now. They can't be checked by lychee and users of the library would be required to set the field. 

Apart from that, this is ready to be merged from my side.

Its-Just-Nans:
Thanks

I guess it's mergeable now

mre (approved):
Thanks for the pull request and for being so responsive about the changes!

PR #1435, opened 2024-06-06
Add Execa to list of users

ehmicky:
Thanks for this great library!

https://github.com/sindresorhus/execa/pull/1120


mre:
Thanks for mentioning, and thanks for using lychee. 😊

mre:
The per-host rate-limiting is definitely planned. It's something we'll need urgently.
While it's true that lychee finished in 2 seconds, the reason is that the server's rate limiting immediately replied with a `429` and did so for all parallel requests. As such, lychee is mostly limited by the number of requests a server can handle, and we need to avoid overloading a server.

PR #1438, opened 2024-06-10
Don't trim mailto: prefix when converting a Uri into a string

### LINKED ISSUE #1436: Converting Uri to Url string
I have a use-case where I need to convert lychee's `Uri` into a string. In this use-case I send the converted string to some web-browser and would like to simply use `new Url(stringConvertedFromLycheesUri)`. So I'd like to get a string which is a valid `Url` which I would expect to be simple since the `Uri` struct just wraps a `Url`.

However, the implementation currently looks as follows:

```rust
impl Uri {
    /// Returns the string representation of the `Uri`.
    ///
    /// If it's an email address, returns the string with scheme stripped.
    /// Otherwise returns the string as-is.
    #[inline]
    #[must_use]
    pub fn as_str(&self) -> &str {
        self.url.as_ref().trim_start_matches("mailto:")
    }
}
```

This implementation is then also used for `AsRef<str>` and `Display`.
So in my use-case I need to do the following to get a back a valid URL string:

```rust
match uri.scheme() {
    "mailto" => format!("mailto:{}", uri.as_str()),
    _ => uri.to_string(),
}
```

Could we update the implementation so that we have a possibility to get the URL string directly? Maybe this should also be the default behaviour and the `mailto` should only be trimmed in a special method?

thomas-zahner:
Fixes #1436

So actually I think it's never necessary to remove the `mailto:` prefix. @mre tell me if you disagree. I assume that this was done for cosmetic purposes in the CLI output. But I even prefer to keep the scheme in the CLI output since it feels more correct and terminal emulators then detect the address as correct URL so that you can click on it to send a mail to that address using your email application. (noticed in Alacritty)

### DIFF
--- a/lychee-lib/src/types/uri/valid.rs
+++ b/lychee-lib/src/types/uri/valid.rs
@@ -22,13 +22,10 @@ pub struct Uri {
 
 impl Uri {
     /// Returns the string representation of the `Uri`.
-    ///
-    /// If it's an email address, returns the string with scheme stripped.
-    /// Otherwise returns the string as-is.
     #[inline]
     #[must_use]
     pub fn as_str(&self) -> &str {
-        self.url.as_ref().trim_start_matches("mailto:")
+        self.url.as_ref()
     }
 
     #[inline]


mre:
Fine with me. Apologies if this caused any inconvenience.

Issue #1303, opened 2023-11-21
--offline still requires dns configuration


stephen-huan:
Running
```shell
lychee --offline .
```
errors with the message
```text
Error: Failed to create request client

Caused by:
    0: Network error
    1: builder error: error reading DNS system conf: No such file or directory (os error 2)
    2: error reading DNS system conf: No such file or directory (os error 2)
```
if there is no network configuration. Ideally `--offline` would modify the codepath so dns isn't required at all.

My specific use case is a bit contrived, but basically I want to use `lychee` in a [sandboxed environment](https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-sandbox) without network. It's still useful to lint that local links are valid as cross-linking can easily get out of date when one is re-organizing their webpage.

If it's too much effort to support such a niche use case I understand, thanks!

roberth:
It's not necessarily about _wireless_ configuration, but requiring an `/etc/resolv.conf` to exist is a problem for me too.

It seems that this problem could be solved in the most general way in https://github.com/seanmonstar/reqwest/issues/2046.

mre:
Thanks for creating the issue and also thanks for providing a swift pull request to upstream. Here's hoping that this will be merged soon.

roberth:
Oh that read like I referenced a PR, but I didn't write one (yet?). I'm not productive in Rust yet, so I figured I'd wait for a response first.

mre:
Oh, sorry, I checked on my phone and it looked like a PR.

As a workaround, can you try to create a dummy `/etc/resolv.conf`?
Something like

```
nameserver 8.8.8.8
nameserver 8.8.4.4
```

mre:
@roberth, any update on this? The upstream issue on `reqwest` didn't seem to get much traction. Did you try the workaround above by any chance?

mre:
So looking at the upstream issue, it seems like we can pass our own resolver in that case.
We could define our own dummy resolver, similar to 
https://github.com/seanmonstar/reqwest/blob/e639bdc1264c266a84dae3bae3193e07cae6861b/src/dns/gai.rs#L9
and use that when `--offline` is set. 
If someone wants to give this a shot, please add a comment or create a PR.

mre:
Oh, okay. Thanks for the update. 

Invalid configs should definitely not be silently ignored. Perhaps we should switch to https://github.com/bnjjj/twelf
We used it for another project, and it worked quite fine.

roberth:
[...]

PR #1448, opened 2024-06-16
README: Add Nixpkgs to Installation

roberth:
Hi again :wave:

This adds the new `testers.lycheeLinkCheck` to the Installation section, as a follow-up to the conversation with @mre starting at https://github.com/lycheeverse/lychee/issues/1303#issuecomment-2016857155.

I think it's brief enough to fit the Installation section, and with infrastructure as code, putting something in file is the act of installing something, so I think it makes sense to put it here despite not being a shell command.

The text itself is sufficient to get users familiar with Nix started, and the links provide more details.

I can change it up if we need to, but here's the ~first~ second version, [rendered](https://github.com/roberth/lychee/blob/e25ef83/README.md#nixos).

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,8 @@ nix-env -iA nixos.lychee
 
 - [`lychee` package](https://search.nixos.org/packages?show=lychee&query=lychee) for configurations, Nix shells, etc.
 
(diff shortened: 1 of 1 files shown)


mre:
Sounds good. Thanks!

PR #1458, opened 2024-06-25
feat(clap): show help menu on no args as well

pwnwriter:
__This pr adds support for help menu when the end user doesn't provide any argument__ 

POC:

before: 

old

now:

new


mre:
Oh, very nice. Thanks a lot @pwnwriter!

Issue #1459, opened 2024-06-29
Windows Filepaths With a Drive Specifier Incorrectly Identified as Urls


toadslop:
Hello, this is the first of the other issues that I mentioned in https://github.com/lycheeverse/lychee/issues/1454.

Many tests in the test suite are failing because the Url crate will erroneously parse a windows filebath with a drive specifier as a url. An example of the test results are included below. Note in particular the Network Errors caused by 'URL Scheme is not allowed'.

```
FAIL [   0.565s] lychee::example_domains cli::test_do_not_exclude_false_positive_example_domains

--- STDOUT:              lychee::example_domains cli::test_do_not_exclude_false_positive_example_domains ---

running 1 test
test cli::test_do_not_exclude_false_positive_example_domains ... FAILED

failures:

failures:
    cli::test_do_not_exclude_false_positive_example_domains

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.55s


--- STDERR:              lychee::example_domains cli::test_do_not_exclude_false_positive_example_domains ---
thread 'cli::test_do_not_exclude_false_positive_example_domains' panicked at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081\library\core\src\ops\function.rs:250:5:
Unexpected failure.
code=1
stderr=
Error: Network error

Caused by:
    0: builder error for url (c:\\Users\\bnhei\\source\\repos\\lychee\\fixtures\\TEST_EXAMPLE_DOMAINS_FALSE_POSITIVES.md)
    1: URL scheme is not allowed


Command and error output:

command="C:\\Users\\bnhei\\source\\repos\\lychee\\target\\debug\\lychee.exe" "C:\\Users\\bnhei\\source\\repos\\lychee\\fixtures\\TEST_EXAMPLE_DOMAINS_FALSE_POSITIVES.md" "--include-mail" "--dump"
code=1
stdout=""
stderr=
Error: Network error

Caused by:
    0: builder error for url (c:\\Users\\bnhei\\source\\repos\\lychee\\fixtures\\TEST_EXAMPLE_DOMAINS_FALSE_POSITIVES.md)
    1: URL scheme is not allowed

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

Canceling due to test failure: 7 tests still running

FAIL [   0.574s] lychee::data_uris cli::test_dump_data_uris_in_verbose_mode

[...]

PR #1460, opened 2024-06-29
Fix: Windows drive paths misidentified as URLs

toadslop:
Attempts to fix this [issue](https://github.com/lycheeverse/lychee/issues/1459).

This has resulted in some failing tests on Linux. Will revise.

Edit:

Fixed Linux tests. Let me know if you have any suggestions for improvement.

### DIFF
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -132,37 +132,61 @@ impl Input {
     ) -> Result<Self> {
         let source = if value == STDIN {
             InputSource::Stdin
-        } else if let Ok(url) = Url::parse(value) {
-            InputSource::RemoteUrl(Box::new(url))
         } else {
-            // this seems to be the only way to determine if this is a glob pattern
-            let is_glob = glob::Pattern::escape(value) != value;
-
-            if is_glob {
-                InputSource::FsGlob {
-                    pattern: value.to_owned(),
-                    ignore_case: glob_ignore_case,
-                }
-            } else {
-                let path = PathBuf::from(value);
-                if path.exists() {
-                    InputSource::FsPath(path)
-                } else if value.starts_with('~') || value.starts_with('.') {
-                    // The path is not valid, but it might be a valid URL
-                    // Check if the path starts with a tilde or a dot
-                    // and exit early if it does
-                    // This check might not be sufficient to cover all cases
-                    // but it catches the most common ones
-                    return Err(ErrorKind::InvalidFile(path));
-                } else {
-                    // Invalid path; check if a valid URL can be constructed from the input
-                    // by prefixing it with a `http://` scheme.
-                    // Curl also uses http (i.e. not https), see
-                    // https://github.com/curl/curl/blob/70ac27604a2abfa809a7b2736506af0da8c3c8a9/lib/urlapi.c#L1104-L1124
-                    let url = Url::parse(&format!("http://{value}")).map_err(|e| {
-                        ErrorKind::ParseUrl(e, "Input is not a valid URL".to_string())
-                    })?;
+            // We use [`reqwest::Url::parse`] because it catches some other edge cases that [`http::Request:builder`] does not
+            // This could be improved with further refinement.
+            match Url::parse(value) {
+                // Weed out non-http schemes, including Windows drive specifiers, which will be successfully parsed by the Url crate
+                Ok(url) if url.scheme() == "http" || url.scheme() == "https" => {
                     InputSource::RemoteUrl(Box::new(url))
+                }
+                Ok(_) => {
+                    // URL parsed successfully, but it's not http or https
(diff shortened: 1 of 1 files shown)


lychee-lib/src/types/input.rs
```diff
@@ -132,7 +133,8 @@ impl Input {
     ) -> Result<Self> {
         let source = if value == STDIN {
             InputSource::Stdin
-        } else if let Ok(url) = Url::parse(value) {
+        } else if let (Ok(_), Ok(url)) = (Request::builder().uri(value).body(()), Url::parse(value))
```
mre, line 136: Your version from the issue description was a bit more readable to me, but that's probably just personal taste. Both works, so it's up to you to decide if you like to change it back.

Apart from that, can we add a comment here why this is needed? Essentially the main gist from the issue to let people know that it's a workaround for Windows.

mre:
Good job! Just a minor ask before we can merge this.

lychee-lib/src/types/input.rs
toadslop: I switched to this because the other version caused some tests to break on Linux. Apparently there are cases that RequestBuilder catches but Url doesn't, and vice versa.

toadslop:
I made an adjustment to be more explicit. Relying on the behavior of RequestBuilder hides the intent a bit, so instead I explicitly say that I'm rejecting Urls that don't have an http (or https) scheme.

I also adjusted the last part of the function to handle Windows behavior better. The function originally assumed that missing a scheme could have been a mistake and adds https, but on Windows this can never happen because the separators in Windows filepaths are not valid for Urls.

Finally, I added some Windows-only tests to validate that the behavior is correct.

toadslop:
After the changes, though, some tests are failing that don't fail on Windows. I will need to investigate to understand the errors. They're related to something called "wayback", which I have never heard of and so I have no idea what the behavior is supposed to be.

lychee-lib/src/types/input.rs
mre: That makes sense. It would also sounds like an important fact to document here.

mre:
Oh, the `wayback` you're referring to is short for "wayback machine", which we use to suggest archived versions in case of broken links. Unfortunately, their API is pretty flaky and lookups randomly fail, which makes these tests unreliable as well.

I suggest to ignore these particular errors for the moment. Perhaps we can even disable them entirely.

toadslop:
Yup, looks like you're right. I forced a test rerun and it passed.

lychee-lib/src/types/input.rs
```diff
@@ -132,38 +132,59 @@ impl Input {
     ) -> Result<Self> {
         let source = if value == STDIN {
             InputSource::Stdin
-        } else if let Ok(url) = Url::parse(value) {
-            InputSource::RemoteUrl(Box::new(url))
+        // The extra use of Request::builder is to weed out Windows filepaths with a drive specifier, which Url will erroneously parse successfully
+        // We still use Url::parse because it catches some other edge cases that Request::builder does not
+        // This could be improved with further refinement.
```
mre, line 137: I'm guessing this comment is no longer needed? At least it looks like we don't use `Request::builder` anymore.

lychee-lib/src/types/input.rs
```diff
@@ -132,38 +132,59 @@ impl Input {
...
-                    // Invalid path; check if a valid URL can be constructed from the input
-                    // by prefixing it with a `http://` scheme.
-                    // Curl also uses http (i.e. not https), see
-                    // https://github.com/curl/curl/blob/70ac27604a2abfa809a7b2736506af0da8c3c8a9/lib/urlapi.c#L1104-L1124
-                    let url = Url::parse(&format!("http://{value}")).map_err(|e| {
-                        ErrorKind::ParseUrl(e, "Input is not a valid URL".to_string())
-                    })?;
+            match Url::parse(value) {
+                // Weed out non-http schemes, including Windows drive specifiers, which will be successfully parsed by the Url crate
+                Ok(url) if url.scheme().starts_with("http") => {
```
mre, line 141: I wonder if we should be a bit more cautious here.
Perhaps something like this?

```rust
Ok(url) if url.scheme() == "http" || url.scheme() == "https" => {
```

mre: (This is just to weed out some false positives like `httpinvalid`. Maybe we could also add a test for that?)

mre: done and added some tests.

lychee-lib/src/types/input.rs
mre: removed the comment. let me know if it's still needed.

mre:
I cleaned up the PR a bit. I think it's ready to be shipped.
Thanks for your contribution @toadslop. 😊

mre:
This is now released!

PR #1470, opened 2024-07-26
Add openSUSE Tumbleweed to the installation instructions

dottorblaster:
:tada: since some days `lychee` is also available as a package into openSUSE Tumbleweed's official repositories! I just felt free to also add it to the README :smile:


mre:
Very cool. Thanks a lot!

Issue #1197, opened 2023-08-04
Lychee trys to parse math equation leading to erroneous error message.


etpalmer63:
For example:

```
$$
[\psi](\mathbf{L}) 
$$
```

Is interpreted as a link and checked, giving an erroneous error message. 

Is there anyway to workaround this?  -- Thank you!

mre:
I'm assuming that's inside a Markdown file?

> Is there anyway to workaround this? -- Thank you!

Well, you can exclude it with `--exclude mathbf` until we fixed the bug.

etpalmer63:
Ah thank you @mre ! I didn't realize the exclude would be that easy so I hadn't even looked yet. Thanks!

HU90m:
The solution to this bug is to support the maths syntax. We wouldn't really want to [work out a spec](https://github.com/raphlinus/pulldown-cmark/pull/733) and add custom parsers atop `pulldown-cmark`, so we are blocked on this being supported upstream.

It looks like there is some interest: https://github.com/raphlinus/pulldown-cmark/pull/622 https://github.com/rust-lang/mdBook/issues/2222 

As this bug has a workaround, I think it would be ok to close this issue with the stance that we will support maths syntax when `pulldown-cmark` does. Of course, discussion can still take place on the closed issue and it can be re-opened by anyone who wishes.

@mre I will leave closing this to you if you agree.

mre:
Good call. Seems like upstream process on this feature is slow, so might not make much sense to hold our breath before it lands. Closing. In case this gets merged in the future, I'd be thankful if someone could reopen the issue.

mre:
Thanks for the notification.

mre:
👋 It got fixed by Hugo.

etpalmer63:
Thank you both for getting this done! 🙏

PR #1473, opened 2024-07-30
Updated pulldown-cmark dependency and fixed maths parsing

HU90m:
Fixes https://github.com/lycheeverse/lychee/issues/1197

In the future, we may want a way in which a user can specify their preferred extensions. This option may be related to https://github.com/lycheeverse/lychee/issues/185#issuecomment-1658581397 , e.g. a github flavoured markdown option.


mre:
This is huge.

It unblocks us, and we can finally update our dependencies and cut a new release.
Thanks, @HU90m!

P.S. I've disabled the Wayback Machine test as it was flaky. We can fix that properly later.

PR #1479, opened 2024-08-06
test: fail when interpreting md footnote as link

mre:
WIP for https://github.com/lycheeverse/lychee/issues/1409.

Accidentally closed #1410 by rebasing on top of master, then pushing mid-rebase and don't have permissions to push there. 😅 So reopening here.

@jan-ferdinand fyi

The underlying issue is still not fixed. Just rebased.
@thomas-zahner let me know if you're still looking into this one.

### DIFF
--- a/lychee-lib/src/extract/markdown.rs
+++ b/lychee-lib/src/extract/markdown.rs
@@ -308,4 +308,12 @@ $$
         let uris = extract_markdown(input, true);
         assert!(uris.is_empty());
     }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


mre:
@thomas-zahner, what do you think? I followed your advice and parsed the plaintext if `link_type` isn't `LinkedType::Inline`. Turned out nicely. Did I miss anything?

mre:
I went ahead and merged this. The tests pass and the changes just extended the existing logic. If we find a bug with it, we can easily fix it in-place.

Issue #1480, opened 2024-08-07
Bug: lychee can not detect error relative url


awang-01:
for this site: https://awang-01.github.io/testing/, there is an image with `src="testing/images/lychee.png"` that I was expecting to fail, but
```
lychee -v https://awang-01.github.io/testing/
✔ [200] https://awang-01.github.io/testing/images/lychee.png

🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors

```
the `https://awang-01.github.iotesting/images/lychee.png` should fail, but lychee automatically add a `/` before the src

mre:
Just checked out your sample page, and it seems to work as expected.

When clickinging on the link, it brings me to https://awang-01.github.io/testing/images/lychee.png, which seems to be the correct URL.

The page responds with a 404, though.

However, when I open it on the command-line with curl, it returns a 200:

```
 curl -vvv https://awang-01.github.io/testing/images/lychee.png
```

It gives me:

``` 
 < HTTP/2 200
< server: GitHub.com
< content-type: image/png
< permissions-policy: interest-cohort=()
< last-modified: Wed, 07 Aug 2024 00:08:14 GMT
< access-control-allow-origin: *
< strict-transport-security: max-age=31556952
< etag: "66b2baee-176be"
< expires: Wed, 07 Aug 2024 00:33:29 GMT
< cache-control: max-age=600
< x-proxy-cache: MISS
< x-github-request-id: C8F3:2D8599:2BDA2D6:2D0C67A:66B2BE81
< accept-ranges: bytes
< age: 0
< date: Wed, 07 Aug 2024 00:23:29 GMT
< via: 1.1 varnish
< x-served-by: cache-fra-etou8220027-FRA
< x-cache: MISS
< x-cache-hits: 0
< x-timer: S1722990209.287670,VS0,VE101
< vary: Accept-Encoding
< x-fastly-request-id: 68cf52bf04fa7335da5ee09db44282dbdfce6794
< content-length: 95934
<
Warning: Binary output can mess up your terminal. Use "--output -" to tell
Warning: curl to output it to your terminal anyway, or consider "--output
Warning: <FILE>" to save to a file.
* Failure writing output to destination
* Connection #0 to host awang-01.github.io left intact
 ```

That's very similar to what lychee sees, I guess.
Note that lychee doesn't differentiate between images and other content when resolving links.
For lychee, all that matters is the response code. In this case, the page returns a 200.
[...]

Issue #1296, opened 2023-11-13
Issues with relative links


mre:
A user was trying out lychee for NLnet and found that version 0.13 misreports many links.
E.g.

```
lychee https://nlnet.nl/thema/OpenDocumentFormat.html

Issues found in 1 input. Find details below.

[https://nlnet.nl/thema/OpenDocumentFormat.html]:
✗ [404] https://nlnet.nl/BinaryAnalysisFund.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Services+Applications.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/InternetInfrastructure.html | Failed: Network error: Not Found
✗ [ERR] http://www.viewerjs.org/examples | Failed: Network error: The certificate was not trusted.
✗ [404] https://nlnet.nl/GetEduroam.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Deployability.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Decentralisedsolutions.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Conferences.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/NetworkApplications.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Applicationprotocols.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/EducationalPrograms.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/InternetHardeningFund.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Reportsandstudies.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/NGIZeroCore.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/NGIZeroDiscovery.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/InformationRetrieval.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Verticals+Search.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/User-operatedInternetFund.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/NGI0Entrust.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Measurement.html | Failed: Network error: Not Found
✗ [404] https://nlnet.nl/Middlewareandidentity.html | Failed: Network error: Not Found
[...]

PR #1481, opened 2024-08-07
Handle relative links

mre:
WIP for #1296 and #1480

This just adds a failing test based on @nobkd's minimal repro.
I didn't fix the underlying issue.

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1569,4 +1569,46 @@ mod cli {
             .success()
             .stdout(contains("0 Errors"));
     }
+
+    /// Test relative paths
+    ///
+    /// Imagine a web server hosting a site with the following structure:
+    /// root
+    /// └── test
+    ///     ├── index.html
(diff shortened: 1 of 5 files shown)


mre:
It doesn't run the failing test yet. See #1483.
The Docker build failure is unrelated.

nobkd:
Might also be interesting. Not sure if this is maybe already used by lychee. :shrug:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

mre:
> Might also be interesting/relevant. Not sure if this is maybe already used by lychee. :man_shrugging:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base

Very interesting; I never heard of it.
The current implementation is pretty basic, and we don't use `<base>`. Looking at the docs, there seem to be quite a few different cases to be handled, so I'd like to postpone handling that for now. But thanks for the tip.

mre:
Superseded by https://github.com/lycheeverse/lychee/pull/1489.

PR #1483, opened 2024-08-07
Run cargo nextest in CI and enable all tests

mre:
I don't know why this was previously enabled.
Perhaps an oversight or it was done on purpose.
At least the remaining tests in `cli` didn't run, apparently.


mre:
Ah, I think it was because of that: https://github.com/lycheeverse/lychee/commit/35580315271b093920f200eae3b7c951cd91048e

The issue here is that this switched to `cargo test` and didn't run all tests.
It might have been faster, but also incorrect.

PR #1484, opened 2024-08-07
ci: fix wrong cpu-arch tag on macos release assets

### LINKED ISSUE #1475: [macos] release asset lychee-v0.15.1-macos-x86_64.dmg contains only ARM64 executable
I expected that a disk image named  "lychee-v0.15.1-macos-x86_64.dmg" would contain an x86_64 executable (or a Universal one) but instead it contains only an ARM64 file.  Did you mean that to be the case? If so, perhaps you could rename the release asset to reflect the architecture it supports.

axel-kah:
closes #1475


mre:
Very good. Thanks!

Issue #1477, opened 2024-08-06
Confusing exit 2, although no error visible


guettli:
We get exit status `2`, although no error is visible:

```
lychee --config .lychee.toml ./*.md  ./docs/**/*.md
  110/110 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links
🔍 110 Total (in 9s) ✅ 96 OK 🚫 0 Errors 💤 13 Excluded

Exit 2
```

Version: 0.15.1

If you look carefully: 96+13 = 109, but 110 were processed.

With `--verbose` I found the issue:

```
? [999] https://www.linkedin.com/company/example.com/ | Unknown status (999 <unknown status code>)
```

Please provide a better error message in the future. It took us some time to find the root cause.

Nevertheless, thank you very much for Lychee!

thomas-zahner:
Thanks for reporting. The result lychee outputs by default if non-standard status codes are encountered may indeed be counterintuitive. If you use a different format (`--format`) it gets clearer as it shows up as `unknown`:

`$ echo "https://httpstat.us/999" | lychee --format detailed -`

```
📝 Summary
---------------------
🔍 Total............1
✅ Successful.......0
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
❓ Unknown..........1
🚫 Errors...........0
```

`999` is not an accepted status code by default, but you could change that with `--accept 999`
So as explained in the README.md the exit code is correct behaviour:

> `2` for link check failures (if any non-excluded link failed the check).

So this is not a bug. But I think what we could do is to show `Unknown` (and same with `Timeouts`?) when `--format` is `compact` and the count is not zero. What do you think @mre?

mre:
Yup, good idea. We could change that. 👍

mre:
If you find the time, feel free to create a PR for that. Happy to get it merged.

mre:
@thomas-zahner fixed it in `master`. @guettli, in case you want to try, please check out lychee from the sources or wait until the next release.

guettli:
@mre thank you very much. It is hard to reproduce, we get the useless http status code 999 only sometimes. I trust you. Thank you!

mre:
This is now part of the latest release, 0.16.1.

PR #1497, opened 2024-09-07
Extend compact format

thomas-zahner:
Addresses https://github.com/lycheeverse/lychee/issues/1477

### DIFF
--- a/lychee-bin/src/formatters/stats/compact.rs
+++ b/lychee-bin/src/formatters/stats/compact.rs
@@ -1,4 +1,6 @@
 use anyhow::Result;
+use console::Style;
+use once_cell::sync::Lazy;
 use std::{
     fmt::{self, Display},
     time::Duration,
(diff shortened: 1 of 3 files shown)


lychee-bin/src/formatters/stats/compact.rs
```diff
@@ -51,25 +53,40 @@ impl Display for CompactResponseStats {
             writeln!(f)?;
         }
 
-        color!(f, NORMAL, "\u{1F50D} {} Total", stats.total)?;
+        color!(f, NORMAL, "🔍 {} Total", stats.total)?;
```
thomas-zahner, line 56: @mre I'd personally prefer this inlined version as it is much more readable compared to the code and a comment at the end of the line (if any). Was there a reason you don't use the emoji directly?

mre: Talked to @m3t0r and we both don't see a reason why using the emoji directly is an issue. So, sure, let's change it back. Easier to read anyway.

mre:
Thanks! 🥳

lychee-bin/src/formatters/stats/compact.rs
thomas-zahner: Sounds great :+1:

Issue #1331, opened 2024-01-04
Read .gitignore


Ekleog-NEAR:
Currently, lychee has its own `.lycheeignore` format.

It’d be very nice if it could read the `.gitignore`, in order to avoid having to duplicate all information there.

See also the discussion we had about our lychee usage at https://github.com/near/nearcore :)

mre:
Currently, we use [`jwalk`](https://github.com/Byron/jwalk) for file traversal, which has limited support for `.gitignore` files.
@Byron, is that correct? Should I create an issue to add `.gitignore` support to `jwalk`?
Alternatively, how hard would it be to add `.gitignore` support to lychee with `jwalk`?

There's also [`walkdir`](https://github.com/BurntSushi/walkdir), which has `.gitignore` support.

Byron:
> There's also [`walkdir`](https://github.com/BurntSushi/walkdir), which has `.gitignore` support.

Can it be that you mean the [`ignore`](https://crates.io/crates/ignore) crate? I just checked and `walkdir` definitely doesn't have `.gitignore` support or hides it from prying eyes.

If I were you, I'd probably try to use the `ignore` crate on the largest input I have and see if the performance is acceptable.

> Should I create an issue to add `.gitignore` support to `jwalk`?
> Alternatively, how hard would it be to add `.gitignore` support to lychee with `jwalk`?

You can create an issue but I wouldn't implement it for the sole reason that it wouldn't be a quick & fun thing for me to do - I didn't write `jwalk` and I am glad it's still working 😅. Thus it might be easier to [use `gix` for that](https://docs.rs/gix/latest/gix/struct.Repository.html#method.excludes), probably after checking if a directory contains a repository, [this code does exactly that](https://github.com/Byron/gitoxide/blob/f1b33ba7518881adfe31a03c366000e1a6e0d7d4/gitoxide-core/src/organize.rs#L31) - you'd definitely want to avoid to call [`gix::open(…)`](https://docs.rs/gix/latest/gix/fn.open.html) on any directory you encounter.

Once a repo is available and paths within the repo are traversed, you should be able to [prime the `AttributeStack`](https://docs.rs/gix/latest/gix/struct.AttributeStack.html#method.at_path) to the path at hand and then [check if it's excluded()](https://docs.rs/gix/latest/gix/worktree/stack/struct.Platform.html#method.is_excluded).

I hope that helps.

mre:
@thomas-zahner fixed this. 🎉

PR #1500, opened 2024-09-15
Respect gitignore

thomas-zahner:
Closes https://github.com/lycheeverse/lychee/issues/1331

With this PR lychee can respect files ignored by git. jwalk is replaced with ignore.


thomas-zahner:
@mre I've got two questions:

1. The `test_readme_usage_up_to_date` test was annoying me a bit sometimes because all my editors trim empty lines when saving README.md. I've updated the test so that lines containing only whitespace are trimmed, so that empty lines are expected. Is this okay?
2. For now `--gitignore` is required so that ignored files are skipped. Would it be okay if we switched it around and update the flag to `--no-gitignore`? This would be a more sensible default in my opinion but also a breaking change.
3. Currently there is no way to include hidden files. (previously `skip_hidden(true)` now `hidden(true)`) Should I add an option (lychee-lib) and flag (lychee-bin) to include hidden files?

mre:
Very nice!

> 1. The `test_readme_usage_up_to_date` test was annoying me a bit sometimes because all my editors trim empty lines when saving README.md. I've updated the test so that lines containing only whitespace are trimmed, so that empty lines are expected. Is this okay?

Yes, same. Sounds very reasonable to me. Thanks!

> 2. For now `--gitignore` is required so that ignored files are skipped. Would it be okay if we switched it around and update the flag to `--no-gitignore`? This would be a more sensible default in my opinion but also a breaking change.

Absolutely. I'm all for it. Please go forward and change it if you like. Maybe we could use the `--include` prefix to align it with the other options. Maybe `--include-gitignore` or `--include-hidden`?

> 3. Currently there is no way to include hidden files. (hardcoded previously as `skip_hidden(true)` now `hidden(true)`) Should I add an option (lychee-lib) and flag (lychee-bin) to include hidden files?

Yes, that sounds sensible to me. 👍

thomas-zahner:
@thomas-zahner Thanks for the answers. The PR is now ready for review.

Just one last thing; clippy complains about too many booleans in the `Collector` struct. Should I extract the three `skip` booleans into a separate struct? How would you refactor that?

```
pub struct Collector {
    basic_auth_extractor: Option<BasicAuthExtractor>,
    skip_missing_inputs: bool,
    skip_ignored: bool,
    skip_hidden: bool,
    include_verbatim: bool,
    use_html5ever: bool,
    base: Option<Base>,
}
```

mre:
Awesome. I don't like that lint too much. Maybe we can ignore it for now.

mre:
Coincidentally, it's the most ignored lint: https://github.com/rust-lang/rust-clippy/issues/5418
(Except for `cognitive_complexity`, which got removed.)

thomas-zahner:
Yeah I also felt like this is very picky and currently there are [292 results](https://grep.app/search?q=%5C%5Ballow%5C%28.%2Aclippy%3A%3Astruct_excessive_bools%5Cb&regexp=true&case=true&filter[lang][0]=Rust) which is quite a lot in comparison to other lints. Began extracting the three bools but it didn't feel like a benefit, only like more code. So I also prefer disabling the lint here.

mre:
I don't know why the `publish-check` is failing. It seems like that's unrelated and not a blocker.

For consistency, I'd propose to rename the options:

* `--no-ignore` → `--include-ignored` or `--include-gitignore`
* `--hidden` → `--include-hidden`

Here's the other options so far:

```sh
❯❯❯ lychee --help | rg 'exclude|include'                                                                                                                                                         
      --include <INCLUDE>
          URLs to check (supports regex). Has preference over all excludes
      --exclude <EXCLUDE>
      --exclude-file <EXCLUDE_FILE>
          Deprecated; use `--exclude-path` instead
      --exclude-path <EXCLUDE_PATH>
  -E, --exclude-all-private
          Equivalent to `--exclude-private --exclude-link-local --exclude-loopback`
      --exclude-private
      --exclude-link-local
      --exclude-loopback
      --exclude-mail
          Exclude all mail addresses from checking (deprecated; excluded by default)
      --include-mail
      --include-fragments
      --include-verbatim
```

mre (approved):

thomas-zahner:
@mre Yes exactly, I used the same names as ripgrep does as I find them very good.

`--include-gitignore` too me sounds like I'd include the `.gitignore` files themself for checking. But that's not the case. As you have mentioned all flags with `include` or `exclude` have to do with individual links not files (with one exception). So I prefer the shorter and clearer flags that align with ripgrep.

mre:
Ah, that makes sense. Thanks for the clarification and the awesome PR.

PR #1510, opened 2024-10-01
Add release-plz and clean up workflows

mre:
It's no secret that we're lacking behind on releases; partially, because they are slightly tedious to create.

My hope is that this will be resolved by introducing `release-plz` [1], which is automates publication whenever we push to `master`.

Along the way, I aligned the naming of the other workflows.

[1]: https://github.com/MarcoIeni/release-plz


mre:
The failing `Docker / build (pull_request)` is the one I'm hoping to resolve with merging this... 😬

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,31 @@
...
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push:
+    branches:
+      - main
```
thomas-zahner, line 13: I don't think we have a `main` branch currently. Do you plan to rename the `master` branch?

thomas-zahner:
@mre Sounds like a great idea! How does it have an affect on the failing Docker image build?

mre:
I think it will publish release binaries under the new binary names, so that should automatically fix the Docker build. At least that's my current working theory. 🙃

thomas-zahner:
Ah okay sounds good to me.

PR #1511, opened 2024-10-03
markdown.rs: make first line a top-level heading

sebastiaanspeck:
Fixes [`MD041` - First line in a file should be a top-level heading](https://github.com/DavidAnson/markdownlint/blob/v0.35.0/doc/md041.md)


mre:
Thanks!

PR #1520, opened 2024-10-08
Exclude `rel=dns-prefetch` links

### LINKED ISSUE #1499: `dns-prefetch` links should be excluded
Links of type [dns-prefetch](https://developer.mozilla.org/en-US/docs/Web/Performance/dns-prefetch) should not be checked by lychee. This is because they are only used to perform DNS lookups and never to fetch any content, e.g.:

```
<link rel='dns-prefetch' href='https://fonts.googleapis.com/' />
```

I am aware of an existing issue #897 which covers preconnect-type links but dns-prefetch is a separate type not covered by the proposed fix.

wackget:
Resolves #1499

As discussed in #1187

### DIFF
--- a/lychee-lib/src/extract/html/html5ever.rs
+++ b/lychee-lib/src/extract/html/html5ever.rs
@@ -76,10 +76,10 @@ impl TokenSink for LinkExtractor {
                     }
                 }
 
-                // Check and exclude rel=preconnect. Other than prefetch and preload,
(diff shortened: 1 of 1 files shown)


mre:
Nice. Can you add a unit test similar to `preconnect`.

mre:
Thanks. I added tests in https://github.com/lycheeverse/lychee/pull/1522. We also forgot to add support for the other HTML extractor, html5gum.

Issue #1462, opened 2024-07-03
Mastodon link is interpreted as email address


almereyda:
When running Lychee on [degrowth.net/organisations/instituto-resiliencia/index.html](https://degrowth.net/organisations/instituto-resiliencia/index.html), it reports, that the email address cannot be reached:

> #### Errors in public/organisations/instituto-resiliencia/index.html
>
>    [ERR] [instituto_resiliencia@15-15-15.social](https://github.com/degrowth/idn-website/pull/instituto_resiliencia@15-15-15.social) | Failed: Unreachable mail address: [instituto_resiliencia@15-15-15.social](mailto:instituto_resiliencia@15-15-15.social): Invalid: Email doesn't exist or is syntactically incorrect
>
> https://github.com/degrowth/idn-website/pull/23

This is a special case, since the string that is recognised is only used as the label for a hyperlink to `https://15-15-15.social/@instituto_resiliencia`. The link clearly does not point at a `mailto:` URI.

This is because of a design choice made in how Mastodon addresses are displayed. The @ is shown directly before of the string that is recognised as an email address, but not part of the link anchor.

![grafik](https://github.com/lycheeverse/lychee/assets/1645308/e07970b6-9e10-4ac2-b75d-f6c90fa4c408)

    <p>Mastodon: @<a href=[https://15-15-15.social/@instituto_resiliencia](view-source:https://15-15-15.social/@instituto_resiliencia) target=_blank>instituto_resiliencia@15-15-15.social</a><br>Twitter: @<a href=[https://twitter.com/ins_resiliencia](view-source:https://twitter.com/ins_resiliencia) target=_blank>ins_resiliencia</a></p>

The HTML in action:

> <p>Mastodon:&nbsp;@<a href="https://15-15-15.social/@instituto_resiliencia" target="_blank">instituto_resiliencia@15-15-15.social</a><br>Twitter: @<a href="https://twitter.com/ins_resiliencia" target="_blank">ins_resiliencia</a></p>

It appears `check-if-email-exists`, of which an outdated fork exists at [github.com/lycheeverse/check-if-email-exists](https://github.com/lycheeverse/check-if-email-exists), which has fallen back considerably behind [github.com/reacherhq/check-if-email-exists](https://github.com/reacherhq/check-if-email-exists), does not recognise that the link points some place else. Eventually by grepping the whole file for a regular expression matching an email pattern, not considering the semantics of the HTML.

What would be the best way to proceed from here?

[...]

PR #1528, opened 2024-10-13
 Ignore textContent links in html nodes

markogalevski:
This fixes issue #1462 by remove plaintext uri parsing
in html5ever and pruning attribute-less URIs in html5gum


mre:
Looks good. Tested locally and it works as expected. Thanks for your contribution, @markogalevski!

Issue #1534, opened 2024-10-18
Problems with upper case umlauts in fragments 


autoantwort:
I have the following markdown code:
```markdown
[änderungen](#änderungen)  
[Änderungen](#Änderungen)

## Änderungen
```
Both fragments are valid as you can test here: https://gist.github.com/autoantwort/27be2ae83db76f9227a9f04f6549ba62

But I get
```
➜  network-documentation git:(master) ✗ lychee -b . test.md --include-fragments
  2/2 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                             Issues found in 1 input. Find details below.

[test.md]:
   [ERROR] file:///.../test.md#%C3%A4nderungen

🔍 2 Total (in 0s) ✅ 1 OK 🚫 1 Error
```

In general fragments are case insensitive.

Issue #1467, opened 2024-07-18
Support URL Encoded Anchors


Silvenga:
It appears that lychee assumes a different behavior from browsers regarding URL encoded anchors.

Given an anchor, such as `#tangent%3A-kustomize`, targeting the element `id="tangent%3A-kustomize"` lychee fails with `Failed: Cannot find fragment`. This appears to only occur with anchors that are URL encoded. Oddly, this only happens when processing local files, and not remote files:

```bash
.\lychee -v --include-fragments "dist\posts\helm-and-gitops\index.html"

✗ [ERR] file:///C:/Users/.../dist/posts/helm-and-gitops/index.html#tangent%3A-kustomize | Failed: Cannot find fragment
```

```
.\lychee -v --include-fragments https://silvenga.com/posts/helm-and-gitops

✔ [200] https://silvenga.com/#tangent%3A-kustomize
```

Although, I think that URL is incorrect - it should be `https://silvenga.com/posts/helm-and-gitops#tangent%3A-kustomize`.

Double checking with browsers, both Firefox and Chromium-based browsers appear to support URL encoded anchors.

andreySadomovskiy:
Have the same issue, but with `%e2%80%99` symbol.

mre:
If someone wants to tackle this, here's what to do.

## Bug Summary
Lychee fails to handle URL-encoded anchors in local files.

## Reproduction Steps
1. Create test.html:
```bash
cat << EOF > test.html
<html>
    <body>
        <a href="#tangent%3A-kustomize">test</a>
        <h1 id="tangent%3A-kustomize">Test Header</h1>
    </body>
</html>
EOF
```

2. Run Lychee:
```bash
lychee -v --include-fragments test.html
```

Expected: Link is valid
Actual: Error - "Cannot find fragment"

## Fix
Update lychee to properly decode URL-encoded anchors in local files.

PR #1535, opened 2024-10-18
Ignore casing when processing markdown fragments + check for percent encoded ancors

autoantwort:
Fixes https://github.com/lycheeverse/lychee/issues/1534 and https://github.com/lycheeverse/lychee/issues/1467

### DIFF
--- a/fixtures/fragments/file.html
+++ b/fixtures/fragments/file.html
@@ -5,8 +5,8 @@
     <title>For Testing Fragments</title>
   </head>
   <body>
-    <section id="in-the-beginning">
-      <p>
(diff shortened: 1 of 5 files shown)


mre:
The lint issue should be fixed in `master` now, so you can rebase to get rid of that. Not sure about the failing test. Is there still a problem with the implementation, or do you just have to update the total link count in the failing test?

autoantwort:
> or do you just have to update the total link count in the failing test?

Yeah that was it

mre (approved):
Looks good to me. Thanks for the improvement!

PR #1537, opened 2024-10-19
Add interesting projects who use lychee

thomas-zahner:
Today I encountered some interesting projects using lychee

### DIFF
--- a/README.md
+++ b/README.md
@@ -690,7 +690,6 @@ We collect a list of common workarounds for various websites in our [troubleshoo
 - https://github.com/papers-we-love/papers-we-love
 - https://github.com/pingcap/docs
 - https://github.com/microsoft/WhatTheHack
-- https://github.com/Azure/ResourceModules
 - https://github.com/nix-community/awesome-nix
 - https://github.com/balena-io/docs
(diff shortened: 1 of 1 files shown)


mre:
Awesome list!
Let's add Gradle as well: https://github.com/gradle/gradle/blob/master/.github/workflows/check-markdown-links.yml

thomas-zahner:
Ah didn't know that :+1:

thomas-zahner:
I've added the link just to the repo, not the full path where lychee is used for consistency. Would you prefer all links pointing to a lychee config or the workflow file so one can quickly see how lychee is used in that project?

Also the links are getting more numerous in the README (and maybe other places as well?). Maybe we could consider adding lychee's GitHub action to lychee? Maybe something equivalent to `lychee --exclude-path benches --exclude-path fixtures . --verbose`

mre (approved):

mre:
Great list, thanks!

I like the idea of using lychee-action and the proposed config. Let's do that to in a separate pull request. 

I think linking to the repos makes sense as the config files could be renamed. We could have a separate config showcase on the lychee website at some point.

PR #1548, opened 2024-10-27
Refactor cache handling test to make it more robust

mre:
This pull request includes a refactoring of the cache handling test to make it more robust. The cache handling test now ensures a clean state before running.


mre:
@dmathieu, this should make the cache check a bit more robust.

We're improving the test setup and cleanup process, as well as ensuring a clean state before running tests. The most important changes include adding a delay to ensure file operations complete and improving file handling for test consistency.

Here's hoping that this will deal with the flakiness.

PR #1559, opened 2024-11-07
Add support for custom file extensions in link checking.

### LINKED ISSUE #410: Don't require `<input>`; Find all supported files in cwd automatically
Lychee supports a lot of files, I'd rather not have to maintain my own glob list. Can it just find and check all the supported filetypes like Prettier does for example?

mre:
This adds support for overwriting extensions:

```
lychee . --extensions md,html,txt,json,yaml
```

The above would only check these extensions.

This was enabled by moving to `ignore` (#1500 by @thomas-zahner).

I'm not 100% convinced about the design yet. Feedback welcome!
Guess we should use whatever [`ignore::types::Types`](https://docs.rs/ignore/0.4.23/ignore/types/struct.Types.html) for file extension matching as it's well-maintained and more exhaustive. Switching this would remove some custom code we have in `FileType`.

Fixes #410

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -332,7 +332,7 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
         collector
     };
 
-    let requests = collector.collect_links_with_ext(inputs, opts.config.extensions.clone());
+    let requests = collector.collect_links_from_file_types(inputs, opts.config.extensions.clone());
 
     let cache = load_cache(&opts.config).unwrap_or_default();
     let cache = Arc::new(cache);
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -332,7 +332,7 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
         collector
     };
 
-    let requests = collector.collect_links(inputs);
+    let requests = collector.collect_links_with_ext(inputs, opts.config.extensions.clone());
 
     let cache = load_cache(&opts.config).unwrap_or_default();
     let cache = Arc::new(cache);
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -238,9 +238,6 @@ pub(crate) struct Config {
     /// want to provide a long list of inputs (e.g. file1.html, file2.md, etc.)
     #[arg(
         long,
-        value_parser = |s: &str| -> Result<FileExtensions, String> {
-            Ok(FileExtensions::from(s.split(',').map(String::from).collect::<Vec<_>>()))
-        },
         long_help = "Test the specified file extensions for URIs when checking files locally.
     
 Multiple extensions can be separated by commas. Note that if you want to check filetypes,
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -6,9 +6,9 @@ use clap::builder::PossibleValuesParser;
 use clap::{arg, builder::TypedValueParser, Parser};
 use const_format::{concatcp, formatcp};
 use lychee_lib::{
-    Base, BasicAuthSelector, FileType, Input, StatusCodeExcluder, StatusCodeSelector,
-    DEFAULT_MAX_REDIRECTS, DEFAULT_MAX_RETRIES, DEFAULT_RETRY_WAIT_TIME_SECS, DEFAULT_TIMEOUT_SECS,
-    DEFAULT_USER_AGENT,
+    Base, BasicAuthSelector, FileExtensions, FileType, Input, StatusCodeExcluder,
+    StatusCodeSelector, DEFAULT_MAX_REDIRECTS, DEFAULT_MAX_RETRIES, DEFAULT_RETRY_WAIT_TIME_SECS,
+    DEFAULT_TIMEOUT_SECS, DEFAULT_USER_AGENT,
 };
 use secrecy::{ExposeSecret, SecretString};
 use serde::Deserialize;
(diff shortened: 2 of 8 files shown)


lychee-bin/src/main.rs
```diff
@@ -314,7 +314,7 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
         collector
     };
 
-    let requests = collector.collect_links(inputs);
+    let requests = collector.collect_links_with_ext(inputs, opts.config.extensions.clone());
```
thomas-zahner, line 317: I'd actually prefer `collect_links` over `collect_links_with_ext` because the latter sounds like the link itself is checked for some extension.

lychee-bin/src/options.rs
```diff
@@ -218,6 +219,25 @@ pub(crate) struct Config {
     #[serde(default)]
     pub(crate) no_progress: bool,
 
+    /// A list of custom extensions for link checking
```
thomas-zahner, line 222:
```suggestion
    /// A list of file extensions. Files not matching the specified extensions are skipped.
```

lychee-lib/src/collector.rs
```diff
@@ -92,28 +92,39 @@ impl Collector {
...
     /// (This can be a directory or a base URL)
     ///
     /// # Errors
     ///
     /// Will return `Err` if links cannot be extracted from an input
-    pub fn collect_links(self, inputs: Vec<Input>) -> impl Stream<Item = Result<Request>> {
+    pub fn collect_links_with_ext(
+        self,
+        inputs: Vec<Input>,
+        extensions: Vec<String>,
```
thomas-zahner, line 111: Can we rename `extensions` to `file_extensions` or wrap `Vec<String>` into a newtype named `FileExtensions`? Or maybe we could use the type that ripgrep uses - the one that is created with `TypesBuilder::new()`?

thomas-zahner:
@mre looks good overall. Does this mean that non Markdown/HTML files are no longer checked when the new argument  isn't specified?

mre:
Yeah, good point. I think plaintext is missing to be compatible with what we had before. That's the only missing format, or? Was considering to extend it to other formats like JSON or YAML as well.

lychee-bin/src/main.rs
mre: Ah, good point. I can switch it around. So what should the method that takes a list of extensions be called?

mre: Made the change (finally)

lychee-lib/src/collector.rs
mre: Introduced the newtype

mre:
@thomas-zahner, I made all changes. Can you check again?

lychee-lib/src/types/file.rs
```diff
@@ -1,8 +1,113 @@
...
+    fn from_iter<T: IntoIterator<Item = String>>(iter: T) -> Self {
+        Self(iter.into_iter().collect())
+    }
+}
+
+impl Iterator for FileExtensions {
+    type Item = String;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.pop()
```
thomas-zahner, line 97: Is this the "normal" way to implement an `Iterator`? When iterating with `iter()` then `FileExtensions` is emptied, right? Shouldn't it stay the same so that I can iterate over it as many times as I want?

lychee-lib/src/types/file.rs
```diff
@@ -1,8 +1,113 @@
...
+        self.0.contains(&file_extension.into())
+    }
+
+    /// Build the current list of file extensions into a file type matcher.
+    ///
+    /// # Errors
+    ///
+    /// Fails if an extension is `all` or otherwise contains any character that
+    /// is not a Unicode letter or number.
+    pub fn all(&self) -> super::Result<Types> {
```
thomas-zahner, line 56: I don't find the function name very intuitive. Can we do it like this? `impl TryFrom<FileExtensions> for Types`

lychee-lib/src/types/file.rs
```diff
@@ -1,8 +1,113 @@
...
+    /// Check if the list of file extensions contains the given file extension
+    pub fn contains<T: Into<String>>(&self, file_extension: T) -> bool {
+        self.0.contains(&file_extension.into())
+    }
+
+    /// Build the current list of file extensions into a file type matcher.
+    ///
+    /// # Errors
+    ///
+    /// Fails if an extension is `all` or otherwise contains any character that
```
thomas-zahner, line 54: What does ```Fails if an extension is `all` ``` mean?

This makes me think about a possibly related question - is it possible to check files without any file extension?

thomas-zahner (approved):
I haven't tested it yet but it looks good, I like the new type :+1:

lychee-lib/src/types/file.rs
mre: Yeah, this is called a [consuming iterator](https://doc.rust-lang.org/book/ch13-02-iterators.html#methods-that-consume-the-iterator). It's fine I think. 😉

mre: The alternative would be to clone the inner iterator, but my current thinking is, that the caller should do that if needed to be more explicit.

lychee-lib/src/types/file.rs
mre: This is actually a limitaton of `TypesBuilder`. 😜
See https://docs.rs/ignore/latest/ignore/types/struct.TypesBuilder.html#method.add
Granted, the documentation could be a bit more clear.

lychee-lib/src/types/file.rs
mre: That's a good idea! Changing.

mre: Done.

lychee-lib/src/types/file.rs
mre: >  is it possible to check files without any file extension?

Yeah, by specifying them as inputs, like before.

So

```
lychee foo bar
```

or

```
lychee 'myglob*'
```

I don't know if that's enough. We could add a setting like `lychee --extensions empty` or `none`, but it might be confusing. For now, I think it's good enough to support the existing behavior, but excellent observation, thanks! 👍

PR #1560, opened 2024-11-07
Rename `fail_map` to `error_map` for consistency

### LINKED ISSUE #1446: Consider renaming `fail_map` to `error_map`
We no longer use the term `failure` in our system; we now refer to it as `error`. Therefore, it makes sense to rename `fail_map` to `error_map` for consistency. Similar to #1443, this will introduce a breaking change, but since we haven't reached a stable version yet, I want to address these naming inconsistencies beforehand.

mre:
Fixes #1446


thomas-zahner (approved):
Looks good to me. Even updated the screencast, I  would never have thought about :) There is nothing else I can think of as I didn't find any reference in the documentation

mre:
That's one of the advantages of SVG: it's just text, so all occurrences will pop up automatically when searching for `fail_map`. No need to remember! 😃

The other part is that maybe we should add a page on the JSON output format to the docs website in the future. 

Thanks for the review.

thomas-zahner:
Ah, I only just realised how the SVG is created. That's quite genius :exploding_head:

PR #1561, opened 2024-11-07
Add support for custom headers in input processing

### LINKED ISSUE #1441: custom Header not sent
Lychee ignores the `--header` argument:

**Steps to reproduce:**

start local http server:
```
nc -lvp1234
```

Run lychee (tested with v0.15.1) with `--header`:

```
lychee 'http://127.0.0.1:1234' --header 'foo=bar'
```

Lychee should send `foo: bar`, but doesn't:
```
Listening on 0.0.0.0 1234
Connection received on localhost 36584
GET / HTTP/1.1
accept: */*
accept-encoding: gzip
host: 127.0.0.1:1234
```

mre:
Previously, we didn't pass (auth) headers to the initial set of inputs -- only to the links which were discovered in the initial input set. That's inconsistent and led to some confusion.

This PR fixes that. There is probably a nicer design for input parsing; the constructor has quite a few "dummy" params now, which tend to be set to the same values. I guess we can refactor that to, say, introduce the notion of an `InputBuilder` if it becomes to tedious to set these values. I guess we should decouple input parsing for the different input types in general because header handling is only needed for web requests and not for file inputs.

For now, I just introduced a helper method `Input::from_value`. 

This fixes #1441.

### DIFF
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -1,5 +1,5 @@
 use crate::archive::Archive;
-use crate::parse::parse_base;
+use crate::parse::{parse_base, parse_headers};
 use crate::verbosity::Verbosity;
 use anyhow::{anyhow, Context, Error, Result};
 use clap::builder::PossibleValuesParser;
@@ -195,9 +195,18 @@ impl LycheeOptions {
         } else {
             Some(self.config.exclude_path.clone())
         };
+        let headers = parse_headers(&self.config.header)?;
         self.raw_inputs
             .iter()
-            .map(|s| Input::new(s, None, self.config.glob_ignore_case, excluded.clone()))
+            .map(|s| {
+                Input::new(
+                    s,
+                    None,
+                    self.config.glob_ignore_case,
+                    excluded.clone(),
+                    headers.clone(),
+                )
+            })
             .collect::<Result<_, _>>()
             .context("Cannot parse inputs from arguments")
     }
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -3,6 +3,7 @@ use crate::{utils, ErrorKind, Result};
 use async_stream::try_stream;
 use futures::stream::Stream;
 use glob::glob_with;
+use http::HeaderMap;
 use ignore::WalkBuilder;
 use reqwest::Url;
 use serde::{Deserialize, Serialize};
@@ -105,7 +106,7 @@ impl Display for InputSource {
 }
 
 /// Lychee Input with optional file hint for parsing
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Input {
     /// Origin of input
     pub source: InputSource,
@@ -113,6 +114,8 @@ pub struct Input {
     pub file_type_hint: Option<FileType>,
     /// Excluded paths that will be skipped when reading content
     pub excluded_paths: Option<Vec<PathBuf>>,
+    /// Custom headers to be used when fetching remote URLs
+    pub headers: reqwest::header::HeaderMap,
 }
 
 impl Input {
@@ -129,6 +132,7 @@ impl Input {
         file_type_hint: Option<FileType>,
         glob_ignore_case: bool,
         excluded_paths: Option<Vec<PathBuf>>,
+        headers: reqwest::header::HeaderMap,
     ) -> Result<Self> {
         let source = if value == STDIN {
             InputSource::Stdin
@@ -194,7 +198,18 @@ impl Input {
             source,
             file_type_hint,
             excluded_paths,
+            headers,
         })
+    }
+
+    /// Convenience constructor with sane defaults
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the input does not exist (i.e. invalid path)
+    /// and the input cannot be parsed as a URL.
+    pub fn from_value(value: &str) -> Result<Self> {
+        Self::new(value, None, false, None, HeaderMap::new())
     }
 
     /// Retrieve the contents from the input
@@ -424,6 +439,8 @@ fn is_excluded_path(excluded_paths: &[PathBuf], path: &PathBuf) -> bool {
 
 #[cfg(test)]
 mod tests {
+    use http::HeaderMap;
+
     use super::*;
 
     #[test]
@@ -434,14 +451,15 @@ mod tests {
         assert!(path.exists());
         assert!(path.is_relative());
 
-        let input = Input::new(test_file, None, false, None);
+        let input = Input::new(test_file, None, false, None, HeaderMap::new());
         assert!(input.is_ok());
         assert!(matches!(
             input,
             Ok(Input {
                 source: InputSource::FsPath(PathBuf { .. }),
                 file_type_hint: None,
-                excluded_paths: None
+                excluded_paths: None,
+                headers: _,
             })
         ));
     }
@@ -454,7 +472,7 @@ mod tests {
         assert!(!path.exists());
         assert!(path.is_relative());
 
-        let input = Input::new(test_file, None, false, None);
+        let input = Input::from_value(test_file);
         assert!(input.is_err());
         assert!(matches!(input, Err(ErrorKind::InvalidFile(PathBuf { .. }))));
     }
@@ -497,7 +515,7 @@ mod tests {
 
     #[test]
     fn test_url_without_scheme() {
-        let input = Input::new("example.com", None, false, None);
+        let input = Input::from_value("example.com");
         assert_eq!(
             input.unwrap().source.to_string(),
             String::from("http://example.com/")
@@ -508,7 +526,7 @@ mod tests {
     #[cfg(windows)]
     #[test]
     fn test_windows_style_filepath_not_existing() {
-        let input = Input::new("C:\\example\\project\\here", None, false, None);
+        let input = Input::from_value("C:\\example\\project\\here");
         assert!(input.is_err());
         let input = input.unwrap_err();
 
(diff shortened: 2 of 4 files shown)


lychee-bin/src/options.rs
```diff
@@ -195,9 +195,18 @@ impl LycheeOptions {
         } else {
             Some(self.config.exclude_path.clone())
         };
+        let headers = parse_headers(&self.config.header)?;
```
mre, line 198: This is not ideal because we parse the headers twice: here and in `lychee-bin/src/client`. We should probably move the parsing to the Clap config.

lychee-bin/src/options.rs
```diff
@@ -16,6 +20,101 @@ use std::path::Path;
...
+//     where
+//         S: Serializer,
+//     {
+//         match headers {
+//             Some(h) => http_serde::header_map::serialize(h, serializer),
+//             None => serializer.serialize_none(),
+//         }
+//     }
+
+//     pub(super) fn deserialize<'de, D>(deserializer: D) -> Result<Option<HeaderMap>, D::Error>
```
thomas-zahner, line 53: What's up with this commented code?

thomas-zahner (changes requested):
@mre I think this PR still needs some polishing to do. Also there are merge conflicts by now

lychee-lib/src/types/input.rs
```diff
@@ -194,9 +198,20 @@ impl Input {
...
         })
     }
 
+    /// Convenience constructor with sane defaults
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the input does not exist (i.e. invalid path)
+    /// and the input cannot be parsed as a URL.
+    pub fn from_value(value: &str) -> Result<Self> {
```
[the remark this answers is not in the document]
mre: You mean `TryFrom<&str>`?
Yeah, I thought about that. I do like that methods are more "discoverable" at times. E.g. I can autocomplete in my IDE. With `::try_from("foo")` it's not as easy. 
I can also see how `TryFrom` is a solid alternative, however.

lychee-bin/src/options.rs
```diff
@@ -172,6 +271,66 @@ macro_rules! fold_in {
     };
 }
 
+/// Parse a single header into a [`HeaderName`] and [`HeaderValue`]
+///
+/// Headers are expected to be in format "key=value".
```
thomas-zahner, line 276: `Headers are expected to be in format "key=value".`

Comment is outdated

lychee-lib/src/types/input.rs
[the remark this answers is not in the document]
thomas-zahner: Yeah, I meant `TryFrom<&str>`. Hmm, agree that IDE support is not as nice, but personally I'd still prefer it that way

lychee-bin/src/options.rs
```diff
@@ -172,6 +271,66 @@ macro_rules! fold_in {
...
+        value: &std::ffi::OsStr,
+    ) -> Result<Self::Value, clap::Error> {
+        let header_str = value.to_str().ok_or_else(|| {
+            clap::Error::raw(
+                clap::error::ErrorKind::InvalidValue,
+                "Header value contains invalid UTF-8",
+            )
+        })?;
+
+        let mut headers = HeaderMap::new();
```
thomas-zahner, line 313: I don't think I quite understand how it works. We seem to create a new `HeaderMap` for each CLI argument we encounter, right? We then insert a single value as long as it's valid. So how do we end up with a HeaderMap containing multiple values. How are they merged?

mre:
⚠️ We might also want to tackle this at the same time: https://github.com/lycheeverse/lychee/pull/1561

lychee-lib/src/types/input.rs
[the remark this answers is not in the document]
mre: Done!

lychee-bin/src/options.rs
mre: removed

lychee-bin/src/options.rs
mre: true. fixed

lychee-bin/src/options.rs
```diff
@@ -654,4 +832,26 @@ mod tests {
         );
         assert_eq!(cli.cache_exclude_status, StatusCodeExcluder::new());
     }
+
+    #[test]
+    fn test_parse_custom_headers() {
+        assert_eq!(
+            parse_header("accept=text/html").unwrap(),
```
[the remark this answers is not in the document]
mre: true!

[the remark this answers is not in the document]
mre: Updated. 😊

lychee-bin/src/options.rs
mre: Yeah, you're right. There was a failing test, which covered that issue; I just didn't get around to fixing it.
The issue is that AFAICT, merging `HeaderMap`s is not doable in `Clap`. I wish there was an `Args::flatten` or so, but there isn't.
So we need a type that can be merged later -- not a `HeaderMap` and also not a `HashMap`, which overwrites existing values because that's not supported by Clap.
Instead, a `Vec<(String, String)>` _does_ work, but it's pretty ugly. Oh well...

I added an extension trait to convert from `Vec<(String, String)>` to `HeaderMap`. That's at least... something. But I'm really not too happy with that part. I would assume there's an even better way to do it. Perhaps I could go with `Vec<HeaderNameValue>` or something, but I ran out of patience for now.

thomas-zahner: True there might be a better solution but it's definitely okay like this

lychee-bin/src/options.rs
```diff
@@ -172,6 +176,89 @@ macro_rules! fold_in {
...
+    type Parser = HeaderParser;
+    fn value_parser() -> Self::Parser {
+        HeaderParser
+    }
+}
+
+/// Extension trait for converting a Vec of header pairs to a HeaderMap
+pub(crate) trait HeaderMapExt {
+    /// Convert a collection of header key-value pairs to a HeaderMap
+    fn from_header_pairs(headers: &Vec<(String, String)>) -> Result<HeaderMap, Error>;
```
thomas-zahner, line 245: What exactly is the purpose of the trait? Do you plan to use it in other places as well? Personally, I would have just used `impl TryFrom<Vec<(String, String)>> for HeaderMap`.

mre: It has to be an extension trait because of [orphan rules](https://rust-exercises.com/100-exercises/04_traits/02_orphan_rule.html): we don't control neither `Vec` nor `HeaderMap`.

A workaround would be to define our own type, but I find extension traits easier to read.
I'm not planning to use in any other places for now.

thomas-zahner: Ah I see, makes sense :+1:

mre:
@thomas-zahner, this is ready for another review. 😊

lychee-bin/src/options.rs
```diff
@@ -201,6 +206,89 @@ macro_rules! fold_in {
...
+    ) -> Result<Self::Value, clap::Error> {
+        let header_str = value.to_str().ok_or_else(|| {
+            clap::Error::raw(
+                clap::error::ErrorKind::InvalidValue,
+                "Header value contains invalid UTF-8",
+            )
+        })?;
+
+        match parse_single_header(header_str) {
+            Ok((name, value)) => Ok((name.to_string(), value.to_str().unwrap().to_string())),
```
thomas-zahner, line 256: Is this `unwrap()` safe to do? Shouldn't we map it to some `Err` value?

lychee-bin/src/options.rs
```diff
@@ -450,10 +557,23 @@ Example: --fallback-extensions html,htm,php,asp,aspx,jsp,cgi"
...
+
+Some websites require custom headers to be passed in order to return valid responses. 
+You can specify custom headers in the format 'Name: Value'. For example, 'Accept: text/html'.
+This is the same format that other tools like curl or wget use. 
+Multiple headers can be specified by using the flag multiple times."
+    )]
     #[serde(default)]
-    pub(crate) header: Vec<String>,
+    #[serde(deserialize_with = "deserialize_headers")]
+    pub header: Vec<(String, String)>,
```
thomas-zahner, line 576: What's the reason for not using `HashMap<String, String>` directly on here? This way we wouldn't need to convert between `Vec` and `HashMap` as often. The problem of merging and duplicates would be solved by that, I think. Does clap not support that out of the box?

thomas-zahner (approved):
Looks really good :ok_hand:

lychee-bin/src/options.rs
mre: Exactly, clap doesn't support it AFAICT.
I did some research and couldn't find a way to make it work. The problem was that the HashMap didn't get merged together properly. I would love to be wrong here `¯\_(ツ)_/¯`

lychee-bin/src/options.rs
mre: True. Fixed.

mre: As for context, it should be safe, because we read the values from TOML and stdin. Both should provide valid UTF-8. At least cargo, which is also a CLI app also just supports valid UTF-8. On the other side, it can't hurt to return an error for invalid UTF-8, so let's do that.

lychee-bin/src/options.rs
mre: I looked into it again:
the problem is clap's `action = clap::ArgAction::Append`. 
If I try to use a `HashMap`, the problem is that clap is trying to collect `(String, String)` tuples into a `HashMap<String, String>`, but it doesn't know how to do that conversion automatically.

Here's a branch with the issue. Just run the tests to check: https://github.com/lycheeverse/lychee/tree/1441-always-send-headers-hashmap

The workaround would be to use a separate config argument just for the CLI header parameters, which would be a `Vec<(String, String)>`, but at this point, we might as well just always use a `Vec<(String, String)>` everywhere...

thomas-zahner: Ah I see. Thanks for digging into it and adding the comment :rocket:

PR #1565, opened 2024-11-09
Add Nix shell

thomas-zahner:
So I've installed Nix OS a week ago. Because of this I created this [Nix shell](https://nixos.wiki/wiki/Development_environment_with_nix-shell). This way it should be possible to reliably build and develop lychee on any Nix system. I've taken the [Custom Rust version example](https://nixos.wiki/wiki/Rust) from Nix's documentation and simply added pkg-config and openssl. I'm not proficient at Nix so I'm not sure if this follows best practices. But I know it works on my system and since it's Nix it should basically work on any system :)

@niklaskorz what do you think about this.


mre:
From my side, this looks good.

niklaskorz:
Sorry that we didn't get to go through it during RustLab anymore. A few possible improvements:

1. Instead of setting rust-overlay, nixpkgs and other parameters as variables using `let`, make shell.nix a function instead, tkaing rust-overlay and pkgs as parameters that default to the values currently set in the variables. This makes them easier to override.
    ```nix
    {
      rust-overlay ? import (builtins.fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz"),
      pkgs = import <nixpkgs> { overlays = [ rust-overlay ]; },
      rustVersion ? "latest", # using a specific version: "1.62.0"
      rust ? pkgs.rust-bin.stable.${rustVersion}.default.override {
        extensions = [
          "rust-src" # for rust-analyzer
          "rust-analyzer" # usable by IDEs like zed-editor
        ];
      },
    }:
    pkgs.mkShell {
      # ...
    }
    ```
2. `fetchTarball` should be combined with a pinned version and hash to make the result reproducible, e.g.
    ```nix
    rust-overlay ? import (fetchTarball {
      url = "https://github.com/oxalica/rust-overlay/archive/23c4b3ba5f806fcf25d5a3b6b54fa0d07854c032.tar.gz";
      sha256 = "1iy52ms10dcp90r046ida0albxy0qkfk5yhc2ijp4r02hyd63rfr";
    }),
    ```
3. You may want to consider using some kind of dependency management tool to pin nixpkgs to a specific commit and to make it easier to update rust-overlay without having to manually update the commit and hash. This can be achieved through [npins](https://github.com/andir/npins) or through the experimental [flakes](https://nix.dev/concepts/flakes.html) feature. For the beginning it's fine to skip this step though!

thomas-zahner:
@niklaskorz No problem, thank you very much for the detailed comment. I will try to apply your suggestions and merge it this weekend, as I will be attending a [local Nix conference](https://www.meetup.com/nix-zurich/)

DashieTM (approved):

thomas-zahner:
@niklaskorz @mre I'm now using flakes as they were also suggested by colleagues at the Nix meetup. The Rust version can be easily changed in the variable and it works flawlessly on my side

mre:
Looks good on my end, but I have no idea of Nix.

niklaskorz:
> For info, this config is only working for x86. We should be able to make it compatible with other platforms, for example by using [nix-systems](https://github.com/nix-systems/nix-systems). But for now this should suffice.

Supporting multiple systems without external dependencies isn't too hard either, see e.g. https://github.com/niklaskorz/rustlab2024-wgpu/blob/ch1/flake.nix

mre:
@thomas-zahner, I talked to @tirimia about Nix and that there's a PR for lychee. He took a look and created another PR. Maybe you can check it out and see which one we're going to merge in the end. 😉 https://github.com/lycheeverse/lychee/pull/1575

thomas-zahner:
@tirimia Thanks for you suggestions on [your PR](https://github.com/lycheeverse/lychee/pull/1575#issuecomment-2532476615). I agree that the .envrc could potentially be annoying/unwanted. So I've now removed it and also made the flake compatible with other platforms as suggested also by @niklaskorz.

I have also tried to add a build config but didn't succeed yet. If you want you could to add it via PR, that would be great. If you have any other suggestions on this PR, please let my know. Otherwise I'll merge these changes.

thomas-zahner:
I've now managed to provide a build config, inspired by [nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/tools/networking/lychee/default.nix). I've also added a short `Development` section to the README to document that `nix develop` and `nix build` can now be used on Nix systems.

@mre are you okay with the changes to the README? Wasn't sure if I should put the section in front.

README.md
```diff
@@ -34,6 +35,12 @@ Available as a command-line utility, a library and a [GitHub Action](https://git
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
+## Development
+
+After [installing Rust](https://www.rust-lang.org/tools/install) use [Cargo](https://doc.rust-lang.org/cargo/) for building and testing.
```
mre, line 40:
```suggestion
After [installing Rust](https://www.rust-lang.org/tools/install) use [cargo](https://doc.rust-lang.org/cargo/) for building and testing.
```

README.md
```diff
@@ -34,6 +35,12 @@ Available as a command-line utility, a library and a [GitHub Action](https://git
...
+On Linux the OpenSSL package [is required](https://github.com/seanmonstar/reqwest?tab=readme-ov-file#requirements) to compile `reqwest`, a dependendy of lychee.
```
mre, line 41:
```suggestion
On Linux the OpenSSL package [is required](https://github.com/seanmonstar/reqwest?tab=readme-ov-file#requirements) to compile `reqwest`, a dependency of lychee.
```

mre (commented):
Small typo, but apart from that looks good.

README.md
thomas-zahner: @EricCrosson Thanks, I agree

flake.nix
```diff
@@ -0,0 +1,61 @@
...
+
+      packages = forAllSystems (
+        system:
+        let
+          pkgs = nixpkgs.legacyPackages.${system};
+          code = import ./default.nix { inherit pkgs; };
+        in
+        {
+          default = code.app;
+        }
```
tirimia, line 58:
```suggestion
        in
        { 
          default = pkgs.lychee.overrideAttrs {
            src = ./.;
            version = self.rev or self.dirtyShortRev;
            cargoDeps = pkgs.rustPlatform.importCargoLock {
              lockFile = ./Cargo.lock;
            };
          };
        };  
```
This way you leverage all the goodies in the official nixpkgs recipe without duplication

tirimia: And you can also remove the `default.nix`

tirimia (changes requested):

README.md
mre: Oh, I didn't recognize. Thanks!

flake.nix
thomas-zahner: Ah I see, that makes sense :+1:

mre:
Thanks @niklaskorz, @tirimia, and @thomas-zahner for the discussion and for supporting each other. Happy to see this merged. ✨

Issue #1432, opened 2024-05-30
Missing 0.15.x releases on docker hub


FlorianLudwig:
On docker hub no tags for 0.15 releases can be found:

https://hub.docker.com/r/lycheeverse/lychee/tags?page=&page_size=&ordering=&name=0.15

thomas-zahner:
Thanks for reporting. Hmm, for some reason the release pipeline did not publish them..
We re-ran them manually so `0.15.0` and `0.15.1` [are now available](https://hub.docker.com/r/lycheeverse/lychee/tags?page=&page_size=&ordering=&name=0.15)

eread:
@thomas-zahner @mre It looks like this persists with the last couple of versions also. Only SHA tags, no version tags.

mre:
It feels like we broke it when switching to release-plz. [I tried to re-run the pipeline](https://github.com/lycheeverse/lychee/actions/runs/11799525793/job/32868236741), but it didn't create a new version tag. Not entirely sure how to fix it. Maybe someone has an idea.

eread:
> It feels like we broke it when switching to release-plz. [I tried to re-run the pipeline](https://github.com/lycheeverse/lychee/actions/runs/11799525793/job/32868236741), but it didn't create a new version tag. Not entirely sure how to fix it. Maybe someone has an idea.

@mre I noticed at the bottom of _Push Image (debian)_ https://github.com/lycheeverse/lychee/actions/runs/11708052907/job/32608866699 this warning:

> WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

Looking at the project docs at: https://github.com/docker/build-push-action, it looks like you might need to set `with.push = true` to have the push actually happen.

Because that isn't set anywhere in this project, I don't understand how any pushes are happening.

I wonder if anything has changed at: https://github.com/docker/build-push-action/releases? 🤔

mre:
Very interesting. I think it pushed in the past at least -- otherwise there would be no images. But yeah, we should set that. If you like, you could add it and we'll give it a shot?

eread:
> Very interesting. I think it pushed in the past at least -- otherwise there would be no images. But yeah, we should set that. If you like, you could add it and we'll give it a shot?

@mre I think I figured it out!

Looking more closely (not sure why I missed this before): https://github.com/lycheeverse/lychee/blame/master/.github/workflows/docker.yml, there's:

```yaml
[...]

PR #1568, opened 2024-11-18
Allow Docker pushes on release branches

eread:
As outlined in https://github.com/lycheeverse/lychee/issues/1432#issuecomment-2477830138, the current workflow won't push to the Docker registry when a release is cut because releases now include a pull request.

This pull request increases the number of conditions under which a Docker image can be pushed.

Closes https://github.com/lycheeverse/lychee/issues/1432.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -94,7 +94,9 @@ jobs:
           context: .
           file: ${{ env.DOCKERFILE }}
           platforms: ${{ env.DOCKER_PLATFORMS }}
-          push: ${{ github.event_name != 'pull_request' && github.actor != 'dependabot[bot]' }}
+          push: ${{ (github.event_name != 'pull_request' &&
+                     github.actor != 'dependabot[bot]') ||
+                     startsWith(github.ref, 'lychee-v') }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
@@ -104,7 +106,9 @@ jobs:
           context: .
           file: ${{ env.DOCKERFILE_ALPINE }}
           platforms: ${{ env.DOCKER_PLATFORMS }}
-          push: ${{ github.event_name != 'pull_request' && github.actor != 'dependabot[bot]' }}
+          push: ${{ (github.event_name != 'pull_request' &&
+                     github.actor != 'dependabot[bot]') ||
+                     startsWith(github.ref, 'lychee-v') }}
           tags: ${{ steps.meta-alpine.outputs.tags }}
           labels: ${{ steps.meta-alpine.outputs.labels }}
 


.github/workflows/docker.yml
```diff
@@ -104,7 +106,9 @@ jobs:
           context: .
           file: ${{ env.DOCKERFILE_ALPINE }}
           platforms: ${{ env.DOCKER_PLATFORMS }}
-          push: ${{ github.event_name != 'pull_request' && github.actor != 'dependabot[bot]' }}
+          push: ${{ (github.event_name != 'pull_request' &&
+                     github.actor != 'dependabot[bot]') ||
+                     startsWith(github.ref_name, 'release-plz') }}
```
mre, line 111: Looks good, but is the `ref_name` for `release-plz` documented somewhere? I was wondering if we should check for a tag name instead. 🤔

eread: @mre It's documented here: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context, but taking a closer look, it might not work for a pull request.

eread: @mre I've refactored this pull request. WDYT?

eread: @mre Did you want me to ask another maintainer to review?

mre: Oh, sorry for dropping the ball on this. 😅

.github/workflows/docker.yml
```diff
@@ -94,7 +94,9 @@ jobs:
           context: .
           file: ${{ env.DOCKERFILE }}
           platforms: ${{ env.DOCKER_PLATFORMS }}
-          push: ${{ github.event_name != 'pull_request' && github.actor != 'dependabot[bot]' }}
+          push: ${{ (github.event_name != 'pull_request' &&
+                     github.actor != 'dependabot[bot]') ||
+                     startsWith(github.ref, 'lychee-v') }}
```
mre, line 99: I think we always want to push -- even when we commit to `master`. Otherwise, the Docker image cannot be tested before release.

mre:
Thanks for your contribution (and your patience), @eread. 🥳 
Let's see if it works. 🤞

eread:
> Thanks for your contribution (and your patience), @eread. 🥳 Let's see if it works. 🤞

Thanks for the merge @mre! I didn't see of an easy way test it for sure. Hopefully a new release isn't too far away 😄

eread:
@mre @thomas-zahner We got a run of the new logic: https://github.com/lycheeverse/lychee/actions/runs/12389086296/job/34581417146

It looks like `push` still evaluated to false.

Also, the Docker command was run with `--tag lycheeverse/lychee:pr-1554 --tag lycheeverse/lychee:sha-007d93a`, rather than `--tag lycheeverse/lychee:0.18.0`. So still more to do there. 🤔

eread:
> @mre @thomas-zahner We got a run of the new logic: https://github.com/lycheeverse/lychee/actions/runs/12389086296/job/34581417146
> 
> It looks like `push` still evaluated to false.
> 
> Also, the Docker command was run with `--tag lycheeverse/lychee:pr-1554 --tag lycheeverse/lychee:sha-007d93a`, rather than `--tag lycheeverse/lychee:0.18.0`. So still more to do there. 🤔

@trask I don't suppose you know how to resolve this?

mre:
The tags are because the pipeline didn't run on a release. The version tag is missing.

So either we're not listening to the right event or the logs belong to a different pipeline run.

mre:
Yeah, so we don't trigger a pull request on releases. 

Typically this is done with 

```yaml
on:
  release:
    types: [published]
```

However, in our case we have a wait until the release binaries are created; otherwise the build would fail.

I don't know if there's a way to say that the workflow depends on another one. Alternatively, we could just have a "wait" step in the pipeline, which tries to curl the binary every 30 seconds or so until it succeeds.

mre:
We could trigger on release as shown above and use an action like https://github.com/marketplace/actions/wait-on to wait until the binaries were created.

eread:
> Yeah, so we don't trigger a pull request on releases.
> 
> Typically this is done with
> 
> ```yaml
> on:
>   release:
>     types: [published]
> ```
> 
> However, in our case we have a wait until the release binaries are created; otherwise the build would fail.
> 
> I don't know if there's a way to say that the workflow depends on another one. Alternatively, we could just have a "wait" step in the pipeline, which tries to curl the binary every 30 seconds or so until it succeeds.

@mre I must've misunderstood how these work: https://github.com/lycheeverse/lychee/pull/1598. I guess the release is created when this is merged, and so the release isn't in the context of the pull request anymore.

eread:
@mre It looks like you can make jobs depend on each other with `needs`: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

eread:
> We could trigger on release as shown above and use an action like https://github.com/marketplace/actions/wait-on to wait until the binaries were created.

@mre What about running when a tag is pushed: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore?

eread:
> > We could trigger on release as shown above and use an action like https://github.com/marketplace/actions/wait-on to wait until the binaries were created.
> 
> @mre What about running when a tag is pushed: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore?

@mre Or maybe we should remove this "escape early" configuration: https://github.com/lycheeverse/lychee/blob/fc006bddc5a63067895799b83cbb9a4cbcabca6a/.github/workflows/docker.yml#L31?

mre:
Hum, I'm not sure why that was there. [I removed it.](https://github.com/lycheeverse/lychee/commit/a94a71261ed2846f621e4f80414ec1c864c28e93). Thanks for the idea. We can give it another try on the next release.

mre:
> @mre It looks like you can make jobs depend on each other with needs: [docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds)

`needs` is just for expressing dependencies between jobs, not workflows. For that, you'd use [workflows](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run). In our case, we already do that. See [here](https://github.com/lycheeverse/lychee/blob/a94a71261ed2846f621e4f80414ec1c864c28e93/.github/workflows/docker.yml#L6).

PR #1573, opened 2024-11-26
Fix retries

trask:
I'm not seeing any retries.

I'm very new to rust, so probably missing how it's supposed to work, but this change made retries work for me.

### DIFF
--- a/lychee-lib/src/retry.rs
+++ b/lychee-lib/src/retry.rs
@@ -74,6 +74,9 @@ impl RetryExt for reqwest::Error {
         } else if let Some(status) = self.status() {
             status.should_retry()
         } else {
+            // We omit checking if error.is_status() since we check that already.
+            // However, if Response::error_for_status is used the status will still
+            // remain in the response object.
             false
         }
     }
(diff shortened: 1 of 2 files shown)


lychee-lib/src/retry.rs
```diff
@@ -71,10 +71,9 @@ impl RetryExt for reqwest::Error {
             } else {
                 false
             }
+        } else if let Some(status) = self.status() {
+            status.should_retry()
         } else {
-            // We omit checking if error.is_status() since we check that already.
```
mre, line 75: Can we keep this comment?

mre:
Thanks for the PR! Makes sense.
Can you add a test to check that it works?
Ideally, a unit test in the same file and an integration test in https://github.com/lycheeverse/lychee/blob/master/lychee-bin/tests/cli.rs. The integration test can be modeled after the existing tests. For example, you can use a mock http server as in this test: https://github.com/lycheeverse/lychee/blob/10056dc5e939be40a697bf30fecb9a9fdfec2d95/lychee-bin/tests/cli.rs#L547-L557

The unit test would be about creating a new status object and checking that it returns the correct result when calling `should_retry()` on it. For some reason, we don't have unit tests for that yet (hence why it's probably broken), but it's a great opportunity to change that. 😉

trask:
unit test and integration test added!

mre:
Brilliant! Thanks for the quick turnaround and for the clean implementation.

Issue #452, opened 2022-01-10
Support setting a base for absolute links


norswap:
By default, lychee doesn't seem to check absolute links (`/`-prefixed), except when setting the `--base` argument (this is weird by itself, since `--base`  should be for the resolution of absolute link as per documentation).

It would be great to:
- check absolute links by default, using the directory that lychee is run from as the root
- have a `--root`  argument that works like `--base` but for absolute links

mre:
Hi, we decided to not check absolute links by default, because it could generate a lot of false-positives when assuming the current directory as a root. For instance, many static site generators put files into a subfolder like `public` or `dist` and in this case, assuming using the directory that lychee is run from as the root would not work.

`--base` can be a base url (e.g. https://example.com) or a directory path (acting like the `--root` you proposed). Setting both would be confusing. When to use the url and when the path when encountering an absolute link?

norswap:
Hey Mathias! Thanks for the prompt answer!

There might a bit of a confusion, my suggestion is not about path vs URLs. Here's what I propose:
- both `--base` and `--root` can be set to either a path or an url
- when encoutering a relative path, prepend the value of `--base`
- when encountering an absolute path (starting with `/`, but not `http://`, `file://`, etc), prepend the value of `--root`
- only check absolute paths if the `--root` option is provided

What do you think?

dkarlovi:
I've created a similar issue for another project: https://github.com/wjdp/htmltest/issues/184

Basically, if you're generating for a specific subdirectory, it doesn't work as expected, for example:

```
# imagine a generator here which creates all URLs like <a href="/something">
$ ./my-generator --base=https://example.com --out=./public
$ lychee --offline --base ./public ./public
🔍 1464 Total ✅ 1269 OK 🚫 0 Errors 💤 195 Excluded
```

works as expected.

```
# imagine a generator here which creates all URLs like <a href="/some/path/something">
$ ./my-generator --base=https://example.com/some/path --out=./public
$ lychee --offline --base ./public ./public
🔍 1464 Total ✅ 16 OK 🚫 1253 Errors (HTTP:1253) 💤 195 Excluded
```

It's trying to find files at `./public/some/path` which doesn't exist, since the content of the `./public` folder will be deployed to that location as is.

If I hack it with:
```
$ mkdir public/some
$ ln -s ../../public public/some/path
$ ll -d public/some/path
[...]

Issue #668, opened 2022-06-30
`--offline` flag is misleading


polarathene:
```
--offline                Only check local files and block network requests
```

> `--offline` is misleading. I think a better name would be `--include-local` or maybe even remove it altogether and only check if `--base` is provided. - [Quoting @mre ](https://github.com/lycheeverse/lychee/issues/660#issuecomment-1164057080)

As a new user, I tried to use `lychee` for checking links from local files only. No external requests was expected, [I tried to troubleshoot why they were being made](https://github.com/lycheeverse/lychee/issues/660) and thought `--offline` would apply to my inputs (which were local files with incorrect path, that `lychee` tried to request externally).

I was mistaken and `--skip-missing` (_hide and avoid exiting due to network failures technically, but works for local files that don't exist too_) was the more appropriate option as it applied to inputs, whereas `--offline` does not (_only links parsed from inputs to check_).

---

I am not sure, but `--scheme 'file'` may provide equivalent functionality that `--offline` does? (_skipping other URIs rather than blocking?_)

Using `--scheme 'file'` against this example HTML content, `src="assets/img/bg-water.webp"` was checked against the local file, while querying the local webserver version at `localhost/index.html` the resource was skipped. So it seems to match `--offline`, if a little less obvious.

The opposite behaviour (_to check only network requests and ignore local file links_) could be `--scheme http https`, although that will exclude other URIs like `data:` too.

mre:
I remember now that we wanted to deprecate `--offline` in favor of https://github.com/lycheeverse/lychee/issues/452#issuecomment-1026277830.
This needs to happen before we hit 1.0.

mre:
In #1576, @trask implemented `--root-dir`, which was the missing piece to deprecate/remove `--offline` support.

polarathene:
When would `--include` variants make sense with that naming pattern?

Would another type `--only-remote` / `--only-local` make more sense? Or is `--include` just a pattern and a no-op/error in this case?

theletterf:
I really like this proposal. :)

mre:
@polarathene, sorry for the one-year delay. ^^

In my mind, `--exclude-local`/`--exclude-remote` fit better than `--include-local`/`--include-remove`, since we're literally excluding subsets of links from the checking phase. The default behavior of lychee is to check all links. It would be counterintuitive to subvert that mode of thinking by using the `--include` pattern to include subsets of links. Just my 2ct.

PR #1576, opened 2024-11-28
Introduce --root-dir

trask:
Related to https://github.com/lycheeverse/lychee/issues/452#issuecomment-1892863035

I tried a couple of more ambitious refactoring, but they kept spiraling out of (my) control, so tried to keep it minimal.


lychee-lib/src/utils/path.rs
```diff
@@ -42,14 +42,14 @@ fn dirname(src: &'_ Path) -> Option<&'_ Path> {
 /// Returns Ok(None) in case of an absolute local link without a `base_url`
 pub(crate) fn resolve(src: &Path, dst: &Path, base: &Option<Base>) -> Result<Option<PathBuf>> {
     let resolved = match dst {
-        relative if dst.is_relative() => {
+        relative if !dst.starts_with("/") => {
```
trask, line 45: this change is unrelated to the rest of the PR

I needed it because I'm running on Windows where a path only seems to be considered absolute if it contains a drive letter

trask:
Also, I don't understand the (remaining) CI failures if you could give some hint, thanks!

lychee-lib/src/utils/request.rs
```diff
@@ -125,24 +125,22 @@ fn truncate_source(source: &InputSource) -> InputSource {
 /// it will not be added to the `HashSet`.
 pub(crate) fn create(
     uris: Vec<RawUri>,
-    input_content: &InputContent,
+    source: &InputSource,
```
trask, line 128: this was just a simplification to make it easier to write tests, since only `input_content.source` is used by this function

lychee-lib/src/types/input.rs
```diff
@@ -140,11 +140,7 @@ impl Input {
                 Ok(url) if url.scheme() == "http" || url.scheme() == "https" => {
                     InputSource::RemoteUrl(Box::new(url))
                 }
-                Ok(_) => {
-                    // URL parsed successfully, but it's not http or https
-                    return Err(ErrorKind::InvalidFile(PathBuf::from(value)));
-                }
-                _ => {
+                Ok(_) | _ => {
```
trask, line 143: Another change only because I'm running on Windows.

If either of the Windows specific changes are problematic I can pull them out, I'll be running it via the docker image anyways.

trask: ...and this one did cause a test failure, so just reverted it

lychee-lib/src/lib.rs
```diff
@@ -71,6 +71,7 @@ pub mod remap;
 /// local IPs or e-mail addresses
 pub mod filter;
 
+/// Test utilities
```
trask, line 74: added this to make CI happy:

```
error: missing documentation for a module
  --> lychee-lib/src/lib.rs:76:1
   |
76 | pub mod test_utils;
   | ^^^^^^^^^^^^^^^^^^
   |
note: the lint level is defined here
  --> lychee-lib/src/lib.rs:47:9
   |
47 | #![deny(missing_docs)]
   |         ^^^^^^^^^^^^
```

lychee-lib/src/utils/request.rs
```diff
@@ -53,14 +53,14 @@ fn try_parse_into_uri(raw_uri: &RawUri, source: &InputSource, base: &Option<Base
     let uri = match Uri::try_from(raw_uri.clone()) {
         Ok(uri) => uri,
         Err(_) => match base {
+            Some(Base::RootPath(_)) | None => match source {
+                InputSource::FsPath(root) => create_uri_from_file_path(root, &text, base)?,
+                _ => return Err(ErrorKind::UnsupportedUriType(text)),
+            },
             Some(base_url) => match base_url.join(&text) {
                 Some(url) => Uri { url },
                 None => return Err(ErrorKind::InvalidBaseJoin(text.clone())),
```
mre, line 62: I read this like
- If `--root-path` is set, try to append the root to the file path to create the full URL (this is the new behavior).
- Otherwise, if base_url is set, join the base URL with the given text to create the full URL (the existing behavior).

I'm not sure, but I *think* we need to focus on the `text` rather than the `Base` here. Because if `text` is a relative URL, we check if we have a base URL, whereas if `text` is an absolute URL, we check if we have a root path.
So, base URL and root path are orthogonal. They could be set at the same time and serve a different purpose.

Which raises the question if we should make `RootPath` a variant of `Base` or if it's an entirely separate type.

Sorry if that description is a bit vague, it's been a while since I looked into that part and the explanation might be wrong.

lychee-lib/src/utils/path.rs
mre: That's surprising, because this is how I understood the documentation of `Path` in the standard library:
It should handle Windows correctly, see [`Path::is_absolute`](https://doc.rust-lang.org/std/path/struct.Path.html#method.is_absolute).

(Note that `Path::is_relative` is simply implemented as `!Path::is_absolute`.)

lychee-lib/src/types/base.rs
```diff
@@ -27,18 +29,28 @@ impl Base {
                 let full_path = path.join(link);
                 Url::from_file_path(full_path).ok()
             }
+            Self::RootPath(_path) => {
+                // this is unused currently because joining on RootPath is handled by create_uri_from_file_path
+                unreachable!()
+            }
```
mre, line 35: I wonder if we should make `RootPath` a separate type. It feels unrelated to `Base`; the `unreachable` code path could be seen as an indicator of that. 
Also see my other comment for an explanation.

mre:
Hey @trask,

this is a very solid attempt and a very long-standing issue of the project, so thanks a lot for looking into that!

I haven't checked all the edge-cases yet, but I think you're on the right track.
However, I wonder if it supports setting `base-url` and `root-path` at the same time? They serve a different purpose: `base-url` is used for relative URLs, while `--root-path` is used for absolute URLs. Setting both at the same time is expected to be a common scenario.
I've added an explanation to the section in the code that seemed the most relevant.

On another note, we could use the opportunity and deprecate `--offline` at the same time.  
See https://github.com/lycheeverse/lychee/issues/668 for more information.
In order to deprecate it, we should add a warning message when it's used and update the help message accordingly to
point people to the new `--root-path` option.
We can also add a note in the README about the deprecation and create a pull request to update the documentation
at https://github.com/lycheeverse/lycheeverse.github.io.

But we don't have to do that in one go. Let's take it step by step.

lychee-lib/src/utils/path.rs
trask: yeah, but I think(?) this condition should be about `abc` (relative) vs `/xyz` (absolute), e.g.

```
<html>
    <body>
        <a href="abc">home</a>
        <a href="/xyz">home</a>
    </body>
</html>
```

but `is_relative()` returns true for both `abc` and `/xyz` on Windows

I don't mind reverting it though, as I'll be running lychee via the docker image anyways

trask:
Thanks for the feedback! I've incorporated your suggestions.

lychee-lib/src/utils/path.rs
```diff
@@ -29,55 +24,35 @@ pub(crate) fn absolute_path(path: PathBuf) -> PathBuf {
...
-}
-
 /// Resolve `dst` that was linked to from within `src`
 ///
 /// Returns Ok(None) in case of an absolute local link without a `base_url`
-pub(crate) fn resolve(src: &Path, dst: &Path, base: &Option<Base>) -> Result<Option<PathBuf>> {
+pub(crate) fn resolve(
+    src: &Path,
+    dst: &PathBuf,
+    ignore_absolute_local_links: bool,
```
trask, line 33: this was previously only ever being called (outside of tests) with base `None`

lychee-lib/src/utils/path.rs
```diff
@@ -9,11 +9,6 @@ use std::path::{Path, PathBuf};
 static CURRENT_DIR: Lazy<PathBuf> =
     Lazy::new(|| env::current_dir().expect("cannot get current dir from environment"));
 
-/// Returns the base if it is a valid `PathBuf`
-fn get_base_dir(base: &Option<Base>) -> Option<PathBuf> {
-    base.as_ref().and_then(Base::dir)
-}
-
```
trask, line 16: unused after https://github.com/lycheeverse/lychee/pull/1576/files#r1864989573

lychee-lib/src/utils/path.rs
```diff
@@ -29,55 +24,33 @@ pub(crate) fn absolute_path(path: PathBuf) -> PathBuf {
...
 
-/// A cumbersome way to concatenate paths without checking their
-/// existence on disk. See <https://github.com/rust-lang/rust/issues/16507>
-fn join(base: PathBuf, dst: &Path) -> PathBuf {
-    let mut abs = base.into_os_string();
-    let target_str = dst.as_os_str();
-    abs.push(target_str);
-    PathBuf::from(abs)
-}
-
```
trask, line 80: unused, see https://github.com/lycheeverse/lychee/pull/1576/files#r1864989573

lychee-bin/tests/cli.rs
```diff
@@ -393,6 +393,42 @@ mod cli {
...
+            .arg("/resolve_paths")
+            .arg("--base")
+            .arg(&dir)
+            .arg(dir.join("resolve_paths").join("index.html"))
+            .env_clear()
+            .assert()
+            .success()
+            .stdout(contains("3 Total"))
+            .stdout(contains("3 OK"));
+    }
```
trask, line 430: copy of `test_resolve_paths` with splitting the base into two parts (base + root path)

lychee-lib/src/types/base.rs
```diff
@@ -30,15 +30,6 @@ impl Base {
...
 
-    /// Return the directory if the base is local
-    #[must_use]
-    pub(crate) fn dir(&self) -> Option<PathBuf> {
-        match self {
-            Self::Remote(_) => None,
-            Self::Local(d) => Some(d.clone()),
-        }
-    }
-
```
trask, line 41: unused, see https://github.com/lycheeverse/lychee/pull/1576/files#r1864989573

lychee-lib/src/utils/request.rs
```diff
@@ -278,10 +276,155 @@ mod tests {
...
+        let requests = create(uris, &source, &None, &base, &None);
+
+        assert_eq!(requests.len(), 1);
+        assert!(requests
+            .iter()
+            .any(|r| r.uri.url.as_str() == "https://example.com/path/page.html#fragment"));
+    }
+
+    #[test]
+    fn test_relative_url_resolution_from_root_path() {
```
trask, line 291: copied and modified the 5 tests above
* one set of copies named `*_from_root_path`
* one set of copies named `*_from_root_path_and_base_url`

examples/collect_links/collect_links.rs
```diff
@@ -21,7 +21,7 @@ async fn main() -> Result<()> {
         },
     ];
 
-    let links = Collector::new(None) // base
+    let links = Collector::new(None, None) // root_path and base
```
mre, line 24: I wonder if we should use `Collector::default` here, where `base` and `root_path` are set to `None`.

lychee-lib/src/utils/path.rs
mre: Aha! Yeah, that makes sense. 

Since we'll no longer use `is_relative` from the standard library, I suggest we encapsulate that behavior in our own function and adding a few unit tests here to check our expectations. 



```rust
use std::path::Path;

/// Check if a given path is relative
///
/// Note that we don't use [`Path::is_relative`] here, as it returns true
/// for both `abc` and `/xyz` on Windows and `/xyz` is expected to be a relative
/// path as per our link checking logic
fn is_relative_path(path: &Path) -> bool {
    !is_absolute_path(path)
}

/// Check if a given path is absolute
fn is_absolute_path(path: &Path) -> bool {
    path.starts_with("/")
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_relative_paths() {
        let paths = vec!["abc"];
        for path in paths {
            assert!(is_relative_path(Path::new(path)));
        }
    }
    
    #[test]
    fn test_absolute_paths() {
        let paths = vec!["/xyz", "/foo/bar"];
        for path in paths {
            assert!(is_absolute_path(Path::new(path)));
        }
    }

    #[test]
    #[cfg(windows)]
    fn test_windows_path_behavior() {
      todo!()
    }
}
```

lychee-lib/src/utils/request.rs
```diff
@@ -48,8 +49,20 @@ fn create_request(
...
+    base: &Option<Base>,
+) -> Result<Uri> {
+    let mut text = raw_uri.text.clone();
+    if text.starts_with('/') {
+        if let Some(path) = root_path {
+            if let Some(path_str) = path.to_str() {
+                text = format!("{path_str}{text}");
+            }
+        }
+    }
```
mre, line 65: Maybe we could put this logic into a separate function/method and add some tests for it. Not sure about the function name yet &mdash; `make_absolute`?

mre:
About the linter errors: you can run [clippy](https://github.com/rust-lang/rust-clippy) locally with `make lint`. It should give you the same output.

examples/collect_links/collect_links.rs
trask: sure, done

lychee-lib/src/utils/path.rs
trask: when I went to make this change I got confused and noticed I couldn't reproduce my problem anymore, so just reverted for the change in this PR, if I run into Windows issues in the future I'll send separate PR

lychee-lib/src/utils/request.rs
trask: makes sense, I added the function and tests, you probably won't like the name I gave it (sorry, I'm a verbose Java dev 😂), I'm happy to rename it, just let me know what you'd like

trask:
> About the linter errors: you can run [clippy](https://github.com/rust-lang/rust-clippy) locally with `make lint`. It should give you the same output.

yay it's green! 💚

lychee-lib/src/utils/request.rs
mre: oh, I like explicit names; no worries

mre:
Just tested it locally and it works as expected. Good work! 🥳 

@norswap, @dkarlovi, @vanbroup, @george-gca, can any of you test this version and provide some feedback? Does it cover your use-cases? 🙏 

You can install this lychee version with

```sh
cargo install --git https://github.com/trask/lychee --branch root-path
```

And then it should be the first thing in your `$PATH` or you use the absolute path.
E.g. for me

```
> which lychee
/Users/mendler/.cargo/bin/lychee
```

Then you can just test as normal.


Question to the group: shall we name the option `--root-dir` instead of `--root-path` to express that we always expect a directory here? A path could also refer to a file, and that's an invalid input.

Oh and @trask, let me know if you'd like to deprecate `--offline` in this pull request as mentioned [earlier](https://github.com/lycheeverse/lychee/pull/1576#issuecomment-2508997282) or if we should do that in a separate PR. Whatever you prefer.

trask:
> Question to the group: shall we name the option --root-dir instead of --root-path to express that we always expect a directory here? A path could also refer to a file, and that's an invalid input.

`--root-dir` makes sense to me

`--root-dir` and `--base-dir` are very close, do you still think it [makes sense to change](https://github.com/lycheeverse/lychee/issues/452#issuecomment-1328971849) `--base-dir` to `--base-url`?

> Oh and @trask, let me know if you'd like to deprecate `--offline` in this pull request as mentioned [earlier](https://github.com/lycheeverse/lychee/pull/1576#issuecomment-2508997282) or if we should do that in a separate PR. Whatever you prefer.

sure, I'd prefer to do this in a follow-up PR.

trask:
Renamed `--root-path` to `--root-dir`

mre:
@george-gca, any updates? Did you find the time to give it a try?

thomas-zahner:
Code-wise it looks good. But now I'm confused about its usage.

When I'm in the fixtures directory I do:

```
$ cargo run -- resolve_paths_from_root_dir/ --root-dir resolve_paths_from_root_dir/

[resolve_paths_from_root_dir/nested/index.html]:
   [ERROR] file:///home/thomas/Projects/lychee/fixtures/resolve_paths_from_root_dir/nested/resolve_paths_from_root_dir/nested/about/index.html#missing | Cannot find file
   [ERROR] file:///home/thomas/Projects/lychee/fixtures/resolve_paths_from_root_dir/nested/resolve_paths_from_root_dir/nested/about/index.html#fragment | Cannot find file
   [ERROR] file:///home/thomas/Projects/lychee/fixtures/resolve_paths_from_root_dir/nested/resolve_paths_from_root_dir/nested/another%20page | Cannot find file
   [ERROR] file:///home/thomas/Projects/lychee/fixtures/resolve_paths_from_root_dir/nested/resolve_paths_from_root_dir/nested | Cannot find file
   [ERROR] file:///home/thomas/Projects/lychee/fixtures/resolve_paths_from_root_dir/nested/resolve_paths_from_root_dir/nested/about | Cannot find file

🔍 7 Total (in 0s) ✅ 2 OK 🚫 5 Errors
```

But:

```
$ cargo run -- resolve_paths_from_root_dir/ --root-dir ..

🔍 7 Total (in 0s) ✅ 7 OK 🚫 0 Errors
```

Why isn't the first one working? Why is the second working?

trask:
> @trask Why isn't the first one working? Why is the second working?

relative `--root-dir` paths are definitely odd (and not covered currently by the `cli.rs` tests)

`--root-dir` is prepended to all absolute local links

and so when `--root-dir` is a relative path, it effectively turns absolute local links into relative links

e.g. running

> cargo run -- resolve_paths_from_root_dir/ --root-dir resolve_paths_from_root_dir/

turns the absolute local link `/nested` into a relative link `resolve_paths_from_root_dir/nested` which get resolved against the html file's directory and results in `file:///home/thomas/Projects/lychee/fixtures/resolve_paths_from_root_dir/nested/resolve_paths_from_root_dir/nested`

and running

> cargo run -- resolve_paths_from_root_dir/ --root-dir ..

turns the absolute local link `/nested` into a relative link `../nested` which get resolved against the html file's directory and results in `file:///home/thomas/Projects/lychee/fixtures/resolve_paths_from_root_dir/nested`

It probably makes more sense to resolve relative `--root-dir` into an absolute dir, I'll give this a try...

trask:
> An alternative would be to prevent relative links by throwing an error. It would be simpler to implement while avoiding ambiguity and unwanted side effects.
> 
> The help text would need to be updated to indicate that an absolute directory is expected.

done!

mre:
@thomas-zahner, what do you think? Can you try it again?

george-gca:
Wouldn't it be better to accept both absolute and relative path, and handle them internally? Maybe doing something like suggested in [these answers](https://stackoverflow.com/questions/30511331/getting-the-absolute-path-from-a-pathbuf) on stackoverflow? One uses PathBuf itself, the other would add a new dependency.

george-gca:
@mre would this be useful for [my case](https://github.com/lycheeverse/lychee/issues/452#issuecomment-1892595070)?

mre:
Ah, sorry. It's not applicable for your case because you have a more complex routing setup. `root-dir` is only if you want to prepend an absolute path before a relative link.

lychee-bin/src/main.rs
```diff
@@ -288,7 +288,13 @@ fn underlying_io_error_kind(error: &Error) -> Option<io::ErrorKind> {
 async fn run(opts: &LycheeOptions) -> Result<i32> {
     let inputs = opts.inputs()?;
 
-    let mut collector = Collector::new(opts.config.base.clone())
+    if let Some(root_dir) = &opts.config.root_dir {
+        if root_dir.is_relative() {
+            bail!("`--root_dir` must be an absolute path");
```
thomas-zahner, line 293: One last thing; I think we should move this to `Collector::new`. lychee can also be used as library through lychee-lib so that's another entrypoint where this invariant should be enforced

mre: Good idea.

lychee-lib/src/collector.rs
```diff
@@ -21,22 +23,47 @@ pub struct Collector {
...
+            skip_ignored: true,
+            root_dir: None,
+            base: None,
+        }
+    }
+}
+
 impl Collector {
     /// Create a new collector with an empty cache
-    #[must_use]
```
trask, line 29: clippy told me to remove this:

> warning: this function has a `#[must_use]` attribute with no message, but returns a type already marked as `#[must_use]`

lychee-bin/src/main.rs
trask: done

norswap:
> Just tested it locally and it works as expected. Good work! 🥳
> 
> @norswap, @dkarlovi, @vanbroup, @george-gca, can any of you test this version and provide some feedback? Does it cover your use-cases? 🙏

Unfortunately I haven't been at the company at which we were using this for more than one year now :') They've also moved the specs this was to be used on, so it's hard to say.

thomas-zahner:
@trask Awesome, thank you :+1:
From my side this looks good @mre

mre:
I changed my mind regarding relative path handling for `--root-dir`. Toying around with the feature and writing the [docs](https://lychee.cli.rs/recipes/root-dir/) for `--root-dir`, I found the lack of support for `.` to be a bit tedious. 

```sh
# ✅ Good
lychee --root-dir "$(pwd)/public"

# ❌ Bad: will fail
lychee --root-dir ./public/
```

Maybe we should implement what @george-gca suggested all along: we simply always canonicalize the `root-dir`:

```rust
use std::fs::canonicalize;
use std::path::PathBuf;

fn main() {
        let path = PathBuf::from(".");
        println!("{:?}", canonicalize(path));
}
```

@trask, @george-gca, @thomas-zahner, I've created a separate issue for that here: https://github.com/lycheeverse/lychee/issues/1606.

PR #1581, opened 2024-12-10
Rust 1.83.0

thomas-zahner:
Address clippy warnings / errors when using Rust 1.83

### DIFF
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -13,20 +13,12 @@ use crate::{
     Base, BasicAuthCredentials, ErrorKind, Request, Result, Uri,
 };
 
-/// Extract basic auth credentials for a given URL.
-fn extract_credentials(
-    extractor: &Option<BasicAuthExtractor>,
-    uri: &Uri,
-) -> Option<BasicAuthCredentials> {
-    extractor.as_ref().and_then(|ext| ext.matches(uri))
-}
-
 /// Create a request from a raw URI.
 fn create_request(
     raw_uri: &RawUri,
     source: &InputSource,
-    base: &Option<Base>,
-    extractor: &Option<BasicAuthExtractor>,
+    base: Option<&Base>,
+    extractor: Option<&BasicAuthExtractor>,
 ) -> Result<Request> {
     let uri = try_parse_into_uri(raw_uri, source, base)?;
     let source = truncate_source(source);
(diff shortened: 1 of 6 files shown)


lychee-lib/src/utils/request.rs
```diff
@@ -126,16 +126,15 @@ fn truncate_source(source: &InputSource) -> InputSource {
...
     input_content: &InputContent,
-    base: &Option<Base>,
-    extractor: &Option<BasicAuthExtractor>,
+    base: Option<&Base>,
+    extractor: Option<&BasicAuthExtractor>,
 ) -> HashSet<Request> {
-    let base = base
-        .clone()
-        .or_else(|| Base::from_source(&input_content.source));
+    let base_fallback = Base::from_source(&input_content.source);
```
thomas-zahner, line 132: @mre This is the only change I'm unsure about. Not really happy with this. Creating the `base_fallback` even when we don't need it is quite ugly. But I couldn't get it to work with inlining it or moving it into the `or_else` closure, because of lifetime issues. Do you have another idea?

thomas-zahner:
@mre One additional question - just stumbled upon the test `test_exclude_example_domains` which is excluded in CI explicitly and it fails consistently. Shouldn't we either fix this test or remove it? I can also create an issue for that.

I see that the test is older than the fixture file. So maybe the test started to fail because the fixture was changed without updating the test?

trask (commented):
several of these changes were made in #1576 (hoping we can merge that one first 🤞)

lychee-lib/src/utils/request.rs
mre: I agree. We can always disable the lint and add a comment.

mre:
> @mre One additional question - just stumbled upon the test test_exclude_example_domains which is excluded in CI explicitly and it fails consistently. Shouldn't we either fix this test or remove it? I can also create an issue for that.

Yes, we can remove it. If the fixture is easy to be updated, we can also do that, but let's not sweat it. 😉

thomas-zahner:
@trask oh didn't know. Thanks for the info

mre:
@thomas-zahner, #1576 got merged, so I guess you can rebase on top of `master`. 😃

thomas-zahner:
There are no more meaningful changes in this PR, all changes and clippy suggestions were already applied in #1576

PR #1614, opened 2025-01-18
Try the new GitHub Arm runners

mre:


mre:
Seems like there are no runners available. 😝

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-arm64
```
javierhonduco, line 22: Saw your post on Mastodon. I think you might have to drop the 64. Just arm. It's cleaner. (Jokes aside this should be a lint / warn)

mre: 🫡 you're a nice person, thanks!

mre: Oh, and I understood your reference. Quite funny. 😆

mre:
Getting a nice backtrace. 🤣 

```
thread 'rustc' panicked at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869/compiler/rustc_serialize/src/opaque.rs:269:45:
range start index 18230290903111630069 out of range for slice of length 39963722
```

javierhonduco:
https://github.com/rust-lang/rust/issues/135867 – very interesting!

javierhonduco:
In case it helps, after reading the comments seems like Ubuntu 22.04 doesn't have this issue. I've been running my Rust project's CI on this OS version without any issues

mre:
Cool, let's try!

mre:
Almost sounds like I forgot to enable a feature flag...

Issue #1619, opened 2025-01-30
feat: Sort (markdown) output by file path


mahlzahn:
Currently, each run of lychee sorts the files where broken links are found in another kind of random order. That makes it difficult to compare different outputs.

I suggest to let lychee by default sort the headings of the output.

mre:
Good idea. 

The content is in a data structure called `error_map`, which is not sorted (beause it's a `HashMap`).
We iterate over the `error_map` in multiple places, dependening on the output format:


https://github.com/lycheeverse/lychee/blob/3fb94e064cf618352bdcca10742bd21ebdb05a86/lychee-bin/src/formatters/stats/markdown.rs#L128-L136

https://github.com/lycheeverse/lychee/blob/3fb94e064cf618352bdcca10742bd21ebdb05a86/lychee-bin/src/formatters/stats/compact.rs#L40-L48

https://github.com/lycheeverse/lychee/blob/3fb94e064cf618352bdcca10742bd21ebdb05a86/lychee-bin/src/formatters/stats/detailed.rs#L52-L74


The proper way to change this would be to return the `error_map` sorted by key. We could add a helper method, e.g. `sorted_error_map` to the stats and use that in the above places instead of directly iterating over `error_map`.

I'd be thankful for a pull request if anyone wants to tackle this. Should be pretty straightforward.

mahlzahn:
This is implemented now and outputs are nicely sorted. Thanks!

PR #1622, opened 2025-02-05
Sort compact/detailed/markdown error output by file path

sudfud:
Implements the feature requested in #1619 

The helper function is named `sort_stat_map` as it's used for both errors and suggestions in the markdown module, and it could be used for other `ResponseStats` maps as well.

`sort_stat_map` returns a `Vec` of key-value pairs, it's only used for iterating through each pair so using a map isn't currently necessary.

The JSON and raw formatters are unchanged since they don't loop through the `ResponseStats` maps. 

This is my first pull request, so any feedback/suggestions are more than welcome and appreciated :alien:

### DIFF
--- a/lychee-bin/src/formatters/stats/mod.rs
+++ b/lychee-bin/src/formatters/stats/mod.rs
@@ -91,7 +91,7 @@ mod tests {
         // Add responses to stats
         // Responses are added to a HashMap, so the order is not preserved
         for source in &test_sources {
-            for response in test_response_urls.iter() {
+            for response in &test_response_urls {
                 test_stats.add(make_test_response(response, source.clone()));
             }
         }
--- a/lychee-bin/src/formatters/stats/mod.rs
+++ b/lychee-bin/src/formatters/stats/mod.rs
@@ -10,10 +10,27 @@ pub(crate) use json::Json;
 pub(crate) use markdown::Markdown;
 pub(crate) use raw::Raw;
 
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Display
+};
+
 use crate::stats::ResponseStats;
+use lychee_lib::InputSource;
 use anyhow::Result;
 
 pub(crate) trait StatsFormatter {
     /// Format the stats of all responses and write them to stdout
     fn format(&self, stats: ResponseStats) -> Result<Option<String>>;
+}
+
+// Convert error_map to a sorted Vec of key-value pairs
+fn sort_stat_map<T>(error_map: &HashMap<InputSource, HashSet<T>>) -> Vec<(&InputSource, &HashSet<T>)>
+where T: Display
+{
+    let mut errors: Vec<(&InputSource, &HashSet<T>)> = error_map.iter().collect();
+
+    errors.sort_by(|(source, _), (other_source, _)| source.cmp(other_source));
+
+    errors
 }
--- a/lychee-bin/src/formatters/stats/mod.rs
+++ b/lychee-bin/src/formatters/stats/mod.rs
@@ -12,25 +12,107 @@ pub(crate) use raw::Raw;
 
 use std::{
     collections::{HashMap, HashSet},
-    fmt::Display
+    fmt::Display,
 };
 
 use crate::stats::ResponseStats;
-use lychee_lib::InputSource;
 use anyhow::Result;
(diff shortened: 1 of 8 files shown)


lychee-bin/src/formatters/stats/mod.rs
```diff
@@ -10,10 +10,27 @@ pub(crate) use json::Json;
...
+// Convert error_map to a sorted Vec of key-value pairs
+fn sort_stat_map<T>(error_map: &HashMap<InputSource, HashSet<T>>) -> Vec<(&InputSource, &HashSet<T>)>
+where T: Display
+{
+    let mut errors: Vec<(&InputSource, &HashSet<T>)> = error_map.iter().collect();
+
+    errors.sort_by(|(source, _), (other_source, _)| source.cmp(other_source));
+
+    errors
+}
```
mre, line 36: I think we should also sort the values, while we're at it:

```rust
fn sort_stat_map<T>(error_map: &HashMap<InputSource, HashSet<T>>) -> Vec<(&InputSource, Vec<&T>)>
where 
    T: Display + Ord,
{
    let mut entries: Vec<_> = error_map
        .iter()
        .map(|(source, set)| {
            let mut sorted_values: Vec<&T> = set.iter().collect();
            sorted_values.sort();
            (source, sorted_values)
        })
        .collect();
    
    entries.sort_by_key(|(source, _)| *source);
    entries
}
```

Also, was there a reason why this isn't a method on `ResponseStats`?
https://github.com/lycheeverse/lychee/blob/master/lychee-bin/src/stats.rs#L53

I think it would be cool to say `&stats.sorted_error_map()` instead of `super::sort_stat_map(&stats.error_map)`, but there might be a reason why it's not a method.

mre:
Awesome! Much appreciated. 😎 

Here's what's left to do:

- [ ] Fix the formatting error (cargo fmt). See https://github.com/lycheeverse/lychee/actions/runs/13152082310/job/36701216646?pr=1622
- [ ] Perhaps make `sort_stat_map` a method?
- [ ] Add a unit test to make sure it works. E.g. you can write a test in the [`ResponseStats`](https://github.com/lycheeverse/lychee/blob/master/lychee-bin/src/stats.rs) file if you like to convert the `sort_stat_map` function to a method. 
- [ ] Add an integration test to [cli.rs](https://github.com/lycheeverse/lychee/blob/master/lychee-bin/tests/cli.rs) to check that the output is as expected on the command line.

Let me know if anything is unclear, or you need help.

lychee-bin/src/formatters/stats/mod.rs
sudfud: > Also, was there a reason why this isn't a method on ResponseStats?

I did originally plan to make it a method, but I ran into an issue in the markdown module. Specifically, I noticed that suggestions are iterated/formatted after the errors:

https://github.com/lycheeverse/lychee/blob/3fb94e064cf618352bdcca10742bd21ebdb05a86/lychee-bin/src/formatters/stats/markdown.rs#L103-L112

In order to keep the suggestion list consistent with the error list, I decided to make a function that could be used to sort the suggestion map as well.

I think this could also be done without sorting the suggestion map ( e.g. by looping through the sorted error keys and using them as suggestion keys ), but since it's my first PR I wanted to play it safe and keep code changes to a minimum. This is also why I opted to make a function that takes a `&HashMap` for input, so that the parameters for `write_stats_per_input` wouldn't need to be changed.

> I think we should also sort the values, while we're at it:

Sounds good 👍 I can start adding this and working on the other remaining tasks after I get home from work.

mre: Yup, sounds good. If you like, feel free to add that as a comment to `sort_stat_map`. 👍

sudfud:
Unit/integration tests for `sort_error_map` have been added.

I also added the [human-sort](https://crates.io/crates/human-sort) crate to lychee-bin's dependencies so that the items can be sorted in natural order. human-sort doesn't have any additional dependencies.

The `Ord` and `PartialOrd` implementations that I previously added to `InputSource` have been removed so human-sort doesn't need to be added to lychee-lib as well.

mre (approved):

mre:
@sufud, thanks for your work. This was a great addition.

PR #1629, opened 2025-02-10
renamed `base` to `base_url` (fixes #1607)

### LINKED ISSUE #1607: Rename `--base` to `--base-url`
`--base` is not explicit enough about the fact that it expects a URL as an argument.
I propose to rename it to `--base-url`. 
Related comment by @MichaIng: https://github.com/lycheeverse/lychee/issues/1574#issuecomment-2526378085

Also update the docs at https://github.com/lycheeverse/lycheeverse.github.io.

Arteiii:
Changed the configuration parameter `base` to `base_url` to improve clarity (fixes #1607)

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -177,6 +177,11 @@ fn load_config() -> Result<LycheeOptions> {
         warn!("WARNING: `--exclude-mail` is deprecated and will soon be removed; E-Mail is no longer checked by default. Use `--include-mail` to enable E-Mail checking.");
     }
 
+    // TODO: Remove this warning and the parameter with 1.0
+    if opts.config.base.is_some() {
+        warn!("WARNING: `--base` is deprecated and will soon be removed; use `--base-url` instead.");
+    }
+
     // Load excludes from file
     for path in &opts.config.exclude_file {
         let file = File::open(path)?;
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -179,7 +179,9 @@ fn load_config() -> Result<LycheeOptions> {
 
     // TODO: Remove this warning and the parameter with 1.0
     if opts.config.base.is_some() {
-        warn!("WARNING: `--base` is deprecated and will soon be removed; use `--base-url` instead.");
+        warn!(
+            "WARNING: `--base` is deprecated and will soon be removed; use `--base-url` instead."
+        );
     }
 
     // Load excludes from file
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -295,7 +295,18 @@ fn underlying_io_error_kind(error: &Error) -> Option<io::ErrorKind> {
 async fn run(opts: &LycheeOptions) -> Result<i32> {
     let inputs = opts.inputs()?;
 
-    let mut collector = Collector::new(opts.config.root_dir.clone(), opts.config.base_url.clone())?
+    // TODO: Remove this section after `--base` got removed with 1.0
+    let base = match (opts.config.base.clone(), opts.config.base_url.clone()) {
+        (None, None) => None,
+        (Some(base), None) => Some(base),
+        (None, Some(base_url)) => Some(base_url),
+        (Some(_base), Some(base_url)) => {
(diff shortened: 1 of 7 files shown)


lychee-bin/src/options.rs
```diff
@@ -453,7 +453,7 @@ separated list of accepted status codes. This example will accept 200, 201,
     /// e.g. <https://example.com> or `/path/to/public`
```
mre, line 453: We think we should also update this help message. `/path/to/public` is no longer recommended and will soon be removed. People should use `root_url` instead. Here's a suggestion:

```
/// Base URL used to resolve relative URLs during link checking
/// Example: "https://example.com"
```

lychee-bin/src/main.rs
```diff
@@ -177,6 +177,13 @@ fn load_config() -> Result<LycheeOptions> {
         warn!("WARNING: `--exclude-mail` is deprecated and will soon be removed; E-Mail is no longer checked by default. Use `--include-mail` to enable E-Mail checking.");
     }
 
+    // TODO: Remove this warning and the parameter with 1.0
```
mre, line 180: We can use the `deprecated` macro here: https://stackoverflow.com/a/37766834/270334

lychee.example.toml
```diff
@@ -75,7 +75,7 @@ header = ["name=value", "other=value"]
 remap = ["https://example.com http://example.invalid"]
 
 # Base URL or website root directory to check relative URLs.
-base = "https://example.com"
+base-url = "https://example.com"
```
mre, line 78:
```suggestion
base_url = "https://example.com"
```

lychee-bin/src/main.rs
Arteiii: Uuh ye I thought about this but it will only show errors on compile time. Right?

So if someone runs the compiled version there won't be any warning

mre: Oh, sorry, all good. You're right. Got confused there for a moment.

lychee-bin/src/main.rs
```diff
@@ -177,6 +177,13 @@ fn load_config() -> Result<LycheeOptions> {
...
+    if opts.config.base.is_some() {
+        warn!(
+            "WARNING: `--base` is deprecated and will soon be removed; use `--base-url` instead."
+        );
+    }
```
mre, line 185: Since we print a warning and continue, but we only use the new `base_url` param further down, wouldn't that lead to unexpected behavior if the user overlooks the message (which can easily go unnoticed in CI)? E.g. the user set `base`, but we use `base_url` only (without handling the case where `base` was set while `base_url` was not).

If so, I suggest checking to add some code which falls back on `base` if it's set (haven't tested the code):

```rust
// TODO: Remove this section after `--base` got removed with 1.0
let base = match (opts.config.base, opts.config.base_url) {
  (None, None) => None,
  (Some(base), None) => Some(base),
  (None, Some(base_url)) => Some(base_url),
  (Some(base), Some(base_url)) => {
    warn!("WARNING: Both, `--base` and `--base-url` are set. Using `base-url` and ignoring `--base` (as it's deprecated).";
    Some(base_url)
  },
}

// Later...
let mut collector = Collector::new(opts.config.root_dir.clone(), base.clone())?
```

mre:
Just one final remark; rest of the code looks fine.

lychee-bin/src/main.rs
Arteiii: Well, basically base still works the same at the moment, but I dont know what would happen if both were set...

Arteiii:
```
lychee . --base-url https://github.com/ --base https://github.com/
   [WARN ] WARNING: `--base` is deprecated and will soon be removed; use `--base-url` instead.
   [WARN ] WARNING: Both, `--base` and `--base-url` are set. Using `base-url` and ignoring `--base` (as it's deprecated).
```

Arteiii:
well I actually don't know why this test fails...

mre:
I think it's just a network blip. Let's retry!

mre:
Hum, no. The example.com website is up, so no clue. It's unrelated to your changes, but it would be nice to get that green tick. I can merge it if you like or we try again later.

Arteiii:
Haha idk, weird i can Check tomorrow

mre:
It was just a network issue. Now fixed. Thanks for the pull request! 😃

Issue #1636, opened 2025-02-17
Use `<URL>` in markdown output


Aariq:
Currently, GitHub actions has difficulty parsing markdown output (see https://github.com/lycheeverse/lychee-action/issues/271)
Rather than using `[URL](URL)` to format links in markdown output, consider using `<URL>`, which is also valid markdown when the link text is just the URL.  These are parsed correctly by GitHub actions output and are clickable links.  E.g.: https://github.com/Aariq/test-renv-cache/actions/runs/13377077841/job/37358471928

mre:
Thanks for creating the issue.

For anyone who wants to send in a PR, here's the part of the code that needs to be changed:

https://github.com/lycheeverse/lychee/blob/d6bbf85145940fde88e079fd304c9b1603ff4265/lychee-bin/src/formatters/stats/markdown.rs#L69-L74

Aariq:
Never done anything in Rust, but looks like you've got a clear CONTRIBUTING.md and I can see the fix, so I'll give it a shot.

nobkd:
Isn't it just basically the following?

```diff
 let mut formatted = format!( 
-    "* [{}] [{}]({})", 
+    "* [{}] <{}>", 
     response.status.code_as_string(), 
     response.uri, 
-    response.uri, 
 ); 
```

mre:
Yeah, basically. Maybe a bunch of tests will fail because of the new format, but that's about it.

PR #1638, opened 2025-02-18
format markdown URLs with <> instead of []() (#1636)

Aariq:
Changes formatting of URLs in markdown output so it works well with GitHub Actions output (#1636, https://github.com/lycheeverse/lychee-action/issues/271)

### DIFF
--- a/lychee-bin/src/formatters/stats/markdown.rs
+++ b/lychee-bin/src/formatters/stats/markdown.rs
@@ -67,9 +67,8 @@ fn stats_table(stats: &ResponseStats) -> String {
 /// Optional details get added if available.
 fn markdown_response(response: &ResponseBody) -> Result<String> {
     let mut formatted = format!(
-        "* [{}] [{}]({})",
+        "* [{}] <{}>",
         response.status.code_as_string(),
-        response.uri,
         response.uri,
     );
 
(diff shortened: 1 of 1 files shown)


lychee-bin/src/formatters/stats/markdown.rs
```diff
@@ -67,10 +67,9 @@ fn stats_table(stats: &ResponseStats) -> String {
 /// Optional details get added if available.
 fn markdown_response(response: &ResponseBody) -> Result<String> {
     let mut formatted = format!(
-        "* [{}] [{}]({})",
+        "* [{}] <{}>",
```
mre, line 70: Why the angular brackets? I thought we wanted to go for the "plain" format? E.g. `"{}"` instead of `<{}>`?

Aariq: angle brackets create [clickable links in markdown](https://www.markdownguide.org/basic-syntax/#urls-and-email-addresses) and they also still work in GitHub actions (tested with a [simple action](https://github.com/Aariq/test-renv-cache/actions/runs/13377077841/job/37358471928))

nobkd: Github may support automatically detecting links and making them clickable, but I think it's still better to explicitly state, that it's a link.

And I'm not sure, if stuff like https://example.com/myendingpunctuation. Would be correctly displayed

mre: Sounds convincing. 👍 Thanks for the explanation.

mre:
Thanks for the improvement @Aariq! 🥳

PR #1642, opened 2025-02-23
Release the macOS binaries also as `.tar.gz`

pepicrft:
I tried to install Lychee with [Ubi](https://github.com/houseabsolute/ubi) through [Mise](https://mise.jdx.dev), and it fails in macOS because `ubi expects the binaries to be in a compressed file, and that's not the case for the macOS binaries.

This PR fixes that by also including the a `.tar.gz` version for the macOS binary.

### DIFF
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -123,13 +123,23 @@ jobs:
           hdiutil create -fs HFS+ -srcfolder dmg -volname lychee lychee.dmg
           tar -czvf lychee.tar.gz -C dmg/ lychee
 
-      - name: Upload binary
+      - name: Upload .dmg
         uses: actions/upload-release-asset@v1
         env:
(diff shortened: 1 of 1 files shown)


mre:
Thanks! I don't know how to test this, but I guess we'll find out if it works with the next release.

pepicrft:
@mre when is the next release planned?

mre:
Probably not too far into the future. I was hoping to get https://github.com/lycheeverse/lychee/pull/1624 in, but it needs more work, so I'll probably aim for https://github.com/lycheeverse/lychee/pull/1559 and perhaps https://github.com/lycheeverse/lychee/pull/1648 and then we can cut a feature release, 0.19.

I don't know if https://github.com/lycheeverse/lychee/pull/1647 can be done by then as well, but we can always do another release once it's ready, so no rush. @devin-purple fyi. 😃

PR #1648, opened 2025-03-02
Refactor input dumping and path retrieval with extension filtering

### LINKED ISSUE #1767: Dumping behaves differently compared to link checking
According to documentation & help message:

```
--dump
    Don't perform any link checking. Instead, dump all the links extracted from inputs that would be checked

--dump-inputs
    Don't perform any link extraction and checking. Instead, dump all input sources from which links would be collected
```

However, in reality both flags behave differently compared to "normal" link checking without one of these flags. Namely, I've observed `.gitignore`d files to be always included with `--dump` and `--dump-inputs` whereas with "normal" link checking ignored files are skipped. This might have been introduced with 41c1b971c78e564d336abc5e960878548debfe94, but maybe dumping always behaved differently than from actual link checking.

@mre do you agree that this is a bug?

mre:
Refactor the `dump-inputs` functionality to print all files checked for links and implement file path retrieval with extension filtering in the `Input` module.

I added this because it was hard to troubleshoot the extension changes, but I figured it would also be helpful to people who want to see which files would be checked when they write `lychee .`.

As an added benefit, `dump-inputs` now behaves more like the `check` command and properly ignores hidden files etc.

This is a follow-up to #1559.
Closes #1767.

### DIFF
--- a/lychee-bin/src/commands/dump_inputs.rs
+++ b/lychee-bin/src/commands/dump_inputs.rs
@@ -31,7 +31,7 @@ pub(crate) async fn dump_inputs(
 
     // Collect all sources with deduplication
     let mut seen_sources = HashSet::new();
-    
+
     for input in inputs {
         let sources_stream = input.get_sources(
             valid_extensions.clone(),
--- a/lychee-bin/src/commands/dump_inputs.rs
+++ b/lychee-bin/src/commands/dump_inputs.rs
@@ -29,6 +29,9 @@ pub(crate) async fn dump_inputs(
     // Create the path filter once outside the loop for better performance
     let excluded_path_filter = lychee_lib::filter::PathExcludes::new(excluded_paths)?;
 
+    // Collect all sources with deduplication
+    let mut seen_sources = HashSet::new();
+    
     for input in inputs {
         let sources_stream = input.get_sources(
             valid_extensions.clone(),
@@ -40,7 +43,10 @@ pub(crate) async fn dump_inputs(
 
         while let Some(source_result) = sources_stream.next().await {
             let source = source_result?;
-            write_out(&mut writer, &source)?;
+            // Only print if we haven't seen this source before
+            if seen_sources.insert(source.clone()) {
+                write_out(&mut writer, &source)?;
+            }
         }
     }
 
--- a/lychee-bin/src/commands/dump_inputs.rs
+++ b/lychee-bin/src/commands/dump_inputs.rs
@@ -1,4 +1,5 @@
 use lychee_lib::{FileExtensions, Input, Result};
+use std::collections::HashSet;
 use std::fs;
 use std::io::{self, Write};
 use std::path::PathBuf;
@@ -12,7 +13,7 @@ use crate::ExitCode;
 /// by lychee, including file paths, URLs, and special sources like stdin.
 /// It respects file extension filtering and path exclusions.
 pub(crate) async fn dump_inputs(
-    inputs: Vec<Input>,
+    inputs: HashSet<Input>,
     output: Option<&PathBuf>,
     excluded_paths: &[String],
     valid_extensions: &FileExtensions,
--- a/lychee-bin/src/commands/dump_inputs.rs
+++ b/lychee-bin/src/commands/dump_inputs.rs
@@ -6,11 +6,11 @@ use tokio_stream::StreamExt;
 
 use crate::ExitCode;
 
-/// Dump all input sources to stdout without extracting any links and checking
-/// them.
+/// Print all input sources to stdout, without extracting or checking links.
 ///
-/// Uses the Input handler to properly handle different input sources and respect
-/// file extension filtering.
+/// This command outputs the resolved input sources that would be processed
+/// by lychee, including file paths, URLs, and special sources like stdin.
+/// It respects file extension filtering and path exclusions.
 pub(crate) async fn dump_inputs(
     inputs: Vec<Input>,
     output: Option<&PathBuf>,
@@ -24,14 +24,16 @@ pub(crate) async fn dump_inputs(
     }
 
     let mut writer = super::create_writer(output.cloned())?;
+    
+    // Create the path filter once outside the loop for better performance
+    let excluded_path_filter = lychee_lib::filter::PathExcludes::new(excluded_paths)?;
 
     for input in inputs {
-        let excluded_path_filter = lychee_lib::filter::PathExcludes::new(excluded_paths)?;
         let sources_stream = input.get_sources(
             valid_extensions.clone(),
             skip_hidden,
             skip_gitignored,
-            excluded_path_filter,
+            excluded_path_filter.clone(),
         );
         tokio::pin!(sources_stream);
 
--- a/lychee-bin/src/commands/dump_inputs.rs
+++ b/lychee-bin/src/commands/dump_inputs.rs
@@ -24,7 +24,7 @@ pub(crate) async fn dump_inputs(
     }
 
     let mut writer = super::create_writer(output.cloned())?;
-    
+
     // Create the path filter once outside the loop for better performance
     let excluded_path_filter = lychee_lib::filter::PathExcludes::new(excluded_paths)?;
 
@@ -33,7 +33,7 @@ pub(crate) async fn dump_inputs(
             valid_extensions.clone(),
             skip_hidden,
             skip_gitignored,
-            excluded_path_filter.clone(),
+            &excluded_path_filter,
         );
         tokio::pin!(sources_stream);
 
--- a/lychee-bin/src/commands/dump_inputs.rs
+++ b/lychee-bin/src/commands/dump_inputs.rs
@@ -14,7 +14,7 @@ use crate::ExitCode;
 pub(crate) async fn dump_inputs(
     inputs: Vec<Input>,
     output: Option<&PathBuf>,
-    excluded_paths: &[PathBuf],
+    excluded_paths: &[String],
     valid_extensions: &FileExtensions,
     skip_hidden: bool,
     skip_gitignored: bool,
@@ -26,22 +26,18 @@ pub(crate) async fn dump_inputs(
     let mut writer = super::create_writer(output.cloned())?;
 
     for input in inputs {
-        let paths_stream =
-            input.get_file_paths(valid_extensions.clone(), skip_hidden, skip_gitignored);
-        tokio::pin!(paths_stream);
+        let excluded_path_filter = lychee_lib::filter::PathExcludes::new(excluded_paths)?;
+        let sources_stream = input.get_sources(
+            valid_extensions.clone(),
+            skip_hidden,
+            skip_gitignored,
+            excluded_path_filter,
+        );
+        tokio::pin!(sources_stream);
 
-        while let Some(path_result) = paths_stream.next().await {
-            let path = path_result?;
-
-            // Skip excluded paths
-            if excluded_paths
-                .iter()
-                .any(|excluded| path.starts_with(excluded))
-            {
-                continue;
-            }
-
-            write_out(&mut writer, &path.to_string_lossy())?;
+        while let Some(source_result) = sources_stream.next().await {
+            let source = source_result?;
+            write_out(&mut writer, &source)?;
         }
     }
 
--- a/lychee-bin/src/commands/dump_inputs.rs
+++ b/lychee-bin/src/commands/dump_inputs.rs
@@ -1,0 +1,53 @@
+use lychee_lib::{FileExtensions, Input, Result};
+use std::fs;
+use std::io::{self, Write};
+use std::path::PathBuf;
+use tokio_stream::StreamExt;
+
+use crate::ExitCode;
+
+/// Dump all input sources to stdout without extracting any links and checking
+/// them.
+///
+/// Uses the Input handler to properly handle different input sources and respect
+/// file extension filtering.
+pub(crate) async fn dump_inputs(
+    inputs: Vec<Input>,
+    output: Option<&PathBuf>,
+    excluded_paths: &[PathBuf],
+    valid_extensions: &FileExtensions,
+    skip_hidden: bool,
+    skip_gitignored: bool,
+) -> Result<ExitCode> {
+    if let Some(out_file) = output {
+        fs::File::create(out_file)?;
+    }
+
+    let mut writer = super::create_writer(output.cloned())?;
+
+    for input in inputs {
+        let paths_stream =
+            input.get_file_paths(valid_extensions.clone(), skip_hidden, skip_gitignored);
+        tokio::pin!(paths_stream);
+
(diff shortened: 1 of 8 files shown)


lychee-lib/src/types/input.rs
```diff
@@ -333,40 +440,6 @@ impl Input {
         Ok(input_content)
     }
 
-    fn glob_contents(
```
mre, line 336: This method was no longer being used.

lychee-lib/src/types/input.rs
```diff
@@ -414,6 +487,20 @@ impl Input {
     }
 }
 
+/// Helper function to check if a file path matches any of the given extensions
+fn file_extensions_match(path: &Path, extensions: &FileExtensions) -> bool {
+    // If the path has no extension, check if we accept plaintext files
+    // Note: We treat files without extensions as plaintext
+    if path.extension().is_none() {
+        return extensions.contains("txt") || extensions.contains("");
+    }
```
mre, line 496: This is a bit of a workaround for files without extensions. It works and probably matches the user intent, so I'd argue it's fine.

thomas-zahner (changes requested):
Whew, that's quite a lot of changes. I didn't check every single detail nor did I test it locally. Are the existing tests covering the changes? I just see that you made a lot of changes and didn't add any tests but removed some.

lychee-lib/src/types/input.rs
```diff
@@ -193,50 +194,87 @@ impl Input {
...
+                            // Enable or disable standard filters based on skip_gitignored parameter.
+                            // This controls:
+                            // - Hidden files/directories (skip files starting with '.')
+                            // - Parent directory gitignore rules
+                            // - .ignore files
+                            // - .gitignore files
+                            // - Global git ignore files
+                            // - .git/info/exclude files
+                            // When skip_gitignored is true, these filters are enabled (files will be skipped).
+                            // When skip_gitignored is false, these filters are disabled (all files will be included).
```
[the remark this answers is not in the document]
mre: agreed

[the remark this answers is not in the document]
mre: shortened, thanks

lychee-lib/src/types/input.rs
```diff
@@ -251,29 +289,96 @@ impl Input {
...
-                            Ok(content) => yield content,
+                        // For individual files, yield if not excluded and matches extensions
+                        if !self.is_excluded_path(path) && file_extensions_match(path, &file_extensions) {
+                            yield path.clone();
                         }
                     }
                 },
+                InputSource::Stdin => {
+                    // Stdin is handled specially - we yield a marker path
+                    yield PathBuf::from("stdin");
```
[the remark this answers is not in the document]
mre: You're absolutely right. Turns out, we don't need that special handling at all. The only minor downside is that we print `stdin` and `string` when we call `--dump-inputs`, but that might actually be a feature as you mentioned because at least it's not misleading if there's a file that's called just that.

lychee-lib/src/types/input.rs
```diff
@@ -193,50 +194,87 @@ impl Input {
...
+                            }
+                        }
                     }
-                }
-                InputSource::FsPath(ref path) => {
+                },
+                InputSource::FsPath(path) => {
                     if path.is_dir() {
+                        // For directories, we walk through and yield matching files
+                        println!("Walking through directory: {}", path.display());
```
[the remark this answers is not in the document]
mre: done

lychee-lib/src/types/input.rs
```diff
@@ -193,50 +194,87 @@ impl Input {
...
-                    let content = Self::url_contents(url).await;
-                    match content {
-                        Err(_) if skip_missing => (),
-                        Err(e) => Err(e)?,
-                        Ok(content) => yield content,
-                    }
+            match &self.source {
+                InputSource::RemoteUrl(url) => {
+                    // For URLs, we just yield the URL string as a path
+                    yield PathBuf::from(url.to_string());
```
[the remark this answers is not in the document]
mre: cleaning up.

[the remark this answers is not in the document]
mre: Remember that this is used by dump-inputs first and foremost. So I don't know what's the best option here. On one side you're right that a URL is not a file path, on the other side it is an input that the user specified and it might help with troubleshooting. We could rename the method name to reflect that. 

But maybe it's not a big deal and I can just comment it out for now and explain the reasoning and we can change that in case someone complains. 😆

lychee-lib/src/types/input.rs
```diff
@@ -193,50 +194,87 @@ impl Input {
...
+    /// Get all file paths matching this input source.
     ///
-    /// If the input is a path, only search through files that match the given
-    /// file extensions.
+    /// This method returns a stream of paths that match the input source,
+    /// taking into account file extensions and exclusions.
+    ///
+    /// # Parameters
+    ///
+    /// * `file_extensions` - The file extensions to filter on
```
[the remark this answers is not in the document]
mre: True. Removed them.

lychee-lib/src/types/input.rs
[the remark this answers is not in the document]
mre: Added comment

lychee-lib/src/types/input.rs
```diff
@@ -251,29 +289,96 @@ impl Input {
...
+            // We can use `get_file_paths` to get the paths, which will handle
+            // filtering by file extensions and exclusions
+            let mut paths_stream = Box::pin(self.get_file_paths(file_extensions, skip_hidden, skip_gitignored));
+
+            while let Some(path_result) = paths_stream.next().await {
+                match path_result {
+                    Ok(path) => {
+                        // Handle special marker paths
+                        let path_str = path.to_string_lossy();
+                        if path_str == "stdin" || path_str == "string" {
```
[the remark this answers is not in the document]
mre: Cleaned up. 😄

lychee-lib/src/types/input.rs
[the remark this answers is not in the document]
mre: Well, this currently causes a test to fail:

```rust
    #[test]
    fn test_dump_inputs_url() -> Result<()> {
        let mut cmd = main_command();
        cmd.arg("--dump-inputs")
            .arg("https://example.com")
            .assert()
            .success()
            .stdout(contains("https://example.com"));

        Ok(())
    }
 ```

That means, if a user types:

```
lychee --dump-inputs https://example.com
```

Do we expect to print `https://example.com` or just return with an empty output?
We need to decide what is the path of least surprise here. I'm fine either way. Thoughts?

File: lychee-lib/src/types/input.rs [hunk 9]
```diff
@@ -208,50 +209,76 @@ impl Input {
...
+    pub fn get_file_paths(
+        &self,
+        file_extensions: FileExtensions,
         skip_hidden: bool,
         skip_gitignored: bool,
-        // If `Input` is a file path, try the given file extensions in order.
-        // Stop on the first match.
-        file_extensions: FileExtensions,
-    ) -> impl Stream<Item = Result<InputContent>> {
+    ) -> impl Stream<Item = Result<PathBuf>> + '_ {
```
thomas-zahner: Wouldn't it make sense to update the return type to something like `impl Stream<Item = Result<InputSource>>` and yield the input sources directly? Then we could also print `InputSource::Stdin` and the other cases with for example `<stdin>`. Alternatively, maybe we could introduce a new enum type?

mre: Huh, that's clever. 🤔

mre: I finally got around to refactoring this. Can you take a look?

mre:
From my side, this is ready to be merged. I changed `get_file_paths` to `get_input_sources`, which yielded a much cleaner design. Thanks for the tip, @thomas-zahner. 

The Docker image is currently failing because of https://github.com/lycheeverse/lychee/pull/1774 and I hope it will be fixed with the next lychee release (0.20.0).

lychee-bin/src/commands/dump_inputs.rs
```diff
@@ -0,0 +1,49 @@
+use lychee_lib::{FileExtensions, Input, Result};
+use std::fs;
+use std::io::{self, Write};
+use std::path::PathBuf;
+use tokio_stream::StreamExt;
+
+use crate::ExitCode;
+
+/// Dump all input sources to stdout without extracting any links and checking
+/// them.
```
thomas-zahner:
```suggestion
/// Print all input sources to stdout, without extracting or checking links.
```

lychee-bin/src/commands/dump_inputs.rs
```diff
@@ -0,0 +1,49 @@
...
+///
+/// Uses the Input handler to properly handle different input sources and respect
```
thomas-zahner: I feel like the term `Input handler` isn't very accurate. Not fully sure what you meant to say with this first part of the sentence

lychee-lib/src/types/input.rs
```diff
@@ -209,6 +210,106 @@ impl Input {
         }
     }
 
+    /// Get all input sources for content processing.
+    ///
+    /// This method returns a stream of input sources from the input source, taking into
```
thomas-zahner:
```suggestion
    /// This method returns a stream of input sources for the given input, taking into
```

lychee-bin/src/commands/dump_inputs.rs
```diff
@@ -0,0 +1,49 @@
...
+    skip_gitignored: bool,
+) -> Result<ExitCode> {
+    if let Some(out_file) = output {
+        fs::File::create(out_file)?;
+    }
+
+    let mut writer = super::create_writer(output.cloned())?;
+
+    for input in inputs {
+        let excluded_path_filter = lychee_lib::filter::PathExcludes::new(excluded_paths)?;
```
thomas-zahner: Is there a good reason for not moving this line before the loop? I don't think we need to create new `PathExcludes` for every `input` right? `get_sources` only makes use of `excluded_path_filter` as reference so let's create it once before the loop and only pass references.

lychee-lib/src/types/input.rs
```diff
@@ -307,30 +430,64 @@ impl Input {
     /// # Errors
     ///
     /// Returns an error if the globbing fails with the expanded pattern.
-    pub fn get_sources(self) -> impl Stream<Item = Result<String>> {
+    pub fn get_sources(
+        self,
+        file_extensions: FileExtensions,
+        skip_hidden: bool,
+        skip_gitignored: bool,
+        excluded_paths: PathExcludes,
```
thomas-zahner: Making this `&PathExcludes` should work

lychee-bin/tests/cli.rs
```diff
@@ -1838,23 +1837,94 @@ mod cli {
...
+    fn test_dump_inputs_with_extensions() -> Result<()> {
+        let mut cmd = main_command();
+        let test_dir = fixtures_path().join("dump_inputs");
+
+        cmd.arg("--dump-inputs")
+            .arg("--extensions")
+            .arg("md,txt")
+            .arg(test_dir)
+            .assert()
+            .success()
```
thomas-zahner, line 1944: Would prefer an equality comparison to multiple individual `contains`. This way the test will be much more accurate.
The actual output is:

```
fixtures/dump_inputs/some_file.txt
fixtures/dump_inputs/subfolder/file2.md
fixtures/dump_inputs/subfolder/test.html
fixtures/dump_inputs/markdown.md
```

Maybe the order is not guaranteed? In that case we should maybe sort the lines or assert the line count to 4?

lychee-bin/tests/cli.rs
```diff
@@ -1838,23 +1837,94 @@ mod cli {
...
+            .stdout(is_empty());
+
+        // Test with --hidden flag
+        main_command()
+            .arg("--dump-inputs")
+            .arg("--hidden")
+            .arg(test_dir)
+            .assert()
+            .success()
+            .stdout(contains(".hidden"));
```
thomas-zahner: Would also prefer an equality check here. Actual output is: `fixtures/hidden/.hidden/file.md`

lychee-bin/tests/cli.rs
```diff
@@ -1838,23 +1837,94 @@ mod cli {
...
+        let mut cmd = main_command();
+        let test_file = fixtures_path().join("TEST.md");
+
         cmd.arg("--dump-inputs")
-            .arg("fixtures")
+            .arg(&test_file)
             .assert()
             .success()
-            .stdout(contains("fixtures"));
+            .stdout(contains("TEST.md"));
```
thomas-zahner: Same here, actual output: `fixtures/TEST.md`

lychee-lib/src/types/input.rs
```diff
@@ -239,63 +341,84 @@ impl Input {
...
+                            InputSource::Stdin => {
+                                let content = Self::stdin_content(self.file_type_hint).await?;
+                                yield content;
+                            },
+                            InputSource::String(s) => {
+                                let content = Self::string_content(&s, self.file_type_hint);
+                                yield content;
+                            },
+                            InputSource::FsGlob { .. } => {
+                                // This shouldn't happen as `get_input_sources` expands the glob patterns
```
thomas-zahner: Would it make sense to add `unreachable!` to detect logical errors here? Or return an error?

thomas-zahner (approved):
Looks great :+1: and actually fixes https://github.com/lycheeverse/lychee/issues/1767

lychee-bin/src/commands/dump_inputs.rs
mre: Good idea. Changed.

lychee-bin/src/commands/dump_inputs.rs
mre: Correct, changed.

lychee-bin/src/commands/dump_inputs.rs
mre: Not really. I changed it. Moved it outside the loop. 👍

lychee-lib/src/types/input.rs
mre: Thanks, fixed

lychee-lib/src/types/input.rs
mre: It does. 👍

lychee-lib/src/types/input.rs
mre: Done. Great idea

lychee-bin/tests/cli.rs
mre: Good idea. Fixed.

lychee-bin/tests/cli.rs
mre: Yes.

mre:
Here I was, thinking that this would be a 10-minute change. So many edge-cases and caveats (and distractions, I guess)!
What a lift.

Anyway, this is finally ready now. As always, thanks for the thoughtful review @thomas-zahner.

PR #1668, opened 2025-03-28
Fix accept/exclude range syntax and docs

### LINKED ISSUE #1667: --cache-exclude-status claims to support half-open ranges, but doesn't
The documentation for the `--cache-exclude-status` option claims that it can accept the `500..` half-open range pattern (given in the list of examples), however these patterns are actually rejected by the current [implementation of `FromStr` for `AcceptRange`](https://github.com/lycheeverse/lychee/blob/1872b6eadf37e080a6e7b0819b8cb577fbe92b79/lychee-lib/src/types/accept/range.rs).

(Speaking of which, the regex used by this `FromStr` impl looks mildly suspicious: what does the trailing `+` repetition specifier on the HTTP status code parser at the end of the range actually mean?)

HadrienG2:
Fixes #1667 .

Along the way I also reworded the docs in a manner that felt easier to read for me, and made sure we have tests for off-by-one scenarios because these errors are so easy to make :) I updated the README's help output as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,16 +354,18 @@ Options:
       --cache-exclude-status <CACHE_EXCLUDE_STATUS>
           A list of status codes that will be ignored from the cache
 
-          The following accept range syntax is supported: [start]..[=]end|code. Some valid
+          The following exclude range syntax is supported: [start]..[[=]end]|code. Some valid
           examples are:
 
-          - 429
-          - 500..=599
-          - 500..
+          - 429 (excludes the 429 status code only)
+          - 500.. (excludes any status code >= 500)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -402,7 +404,7 @@ Options:
   -u, --user-agent <USER_AGENT>
           User agent
 
-          [default: lychee/x.y.z]
+          [default: lychee/0.18.1]
```
mre, line 407: Can we change that backt to x.z.y? We use it as a wildcard pattern to avoid that the tests fail when we bump the version without any other changes.

PR #1681, opened 2025-04-16
make clippy happy

JayJayArr:
resolves all clippy issues
sry for the messed up commit history


lychee-lib/Cargo.toml
```diff
@@ -30,7 +30,7 @@ ip_network = "0.4.1"
 linkify = "0.10.0"
 log = "0.4.27"
 octocrab = "0.44.0"
-once_cell = "1.21.3"
+# once_cell = "1.21.3"
```
thomas-zahner, line 33: We should get rid of this completely. But I'll sort it out myself.

PR #1687, opened 2025-05-02
Remove hardcoded rule for handling erroneous status codes differently

### LINKED ISSUE #1661: Question: Excluding error codes
Another question. Is there a way to skip certain error codes? I notice there is an include (`--accept`) but didn't see an exclude for codes.

thomas-zahner:
Closes #1661


thomas-zahner:
So I tracked down the code that is responsible for the current behaviour. Turns out it has nothing to do with clap, default values or range parsing. Instead it seems like a design decision made in the past. Also it's really not much code that is responsible for that.

Basically, if a HTTP code is not explicitly accepted (via `--accept` or its default value) we use [`error_for_status_ref`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.error_for_status_ref) to handle erroneous status codes in the code.

Removing this explicit handling would also imply that we remove the `Status::Redirected` and `Status::UnknownStatusCode` variants.

Also I wasn't sure at first if the new variant should be added to the `Status` or `ErrorKind` struct. But since I think it should behave like a "normal" error I've added it to `ErrorKind`.

This means it's not counted like unknowns as a new category:

```
🔍 1 Total (in 0s) ✅ 0 OK 🚫 0 Errors ❓ 1 Unknown
```

But instead like normal errors:

```bash
$ echo 'https://httpstat.us/200' | cargo run - --accept 201 --verbose

   [ERROR] https://httpstat.us/200 | Rejected status code: 200 OK (this depends on your "accept" configuration)

[stdin]:
   [ERROR] https://httpstat.us/200 | Rejected status code: 200 OK (this depends on your "accept" configuration)

```

@mre what are your thoughts on this. Is this approach okay with you? If so, we should of course update the documentation accordingly and mention the breaking changes in the release notes.

mre:
I agree.

The rule should be that `Status` is strictly set by the checking logic and should directly relate to a website's status (as in "status code"). It's nothing that lychee should temper with.
The error part is mostly for expressing how we *handle* the status that comes back from the website. So, yeah, if a user defines that they don't want to accept a certain status code as "valid", that should surface as an error.

thomas-zahner:
@mre The PR is ready for re

thomas-zahner:
@mre The change has more implications than I initially expected. What do you think about the PR?

lychee-lib/src/client.rs
```diff
@@ -237,7 +237,10 @@ pub struct ClientBuilder {
     /// Set of accepted return codes / status codes.
     ///
     /// Unmatched return codes/ status codes are deemed as errors.
-    accepted: Option<HashSet<StatusCode>>,
+    ///
+    /// TODO: accept all "valid" status codes by default. Maybe use `AcceptRange`?
```
thomas-zahner, line 241: Will have to resolve this

lychee-lib/src/retry.rs
```diff
@@ -105,6 +105,8 @@ impl RetryExt for ErrorKind {
         }) = self.github_error()
         {
             source.should_retry()
+        } else if let Self::RejectedStatusCode(StatusCode::TOO_MANY_REQUESTS) = self {
```
thomas-zahner, line 108: And not quite sure if this the optimal approach

lychee-lib/src/checker/website.rs
```diff
@@ -88,7 +88,7 @@ impl WebsiteChecker {
     /// Check a URI using [reqwest](https://github.com/seanmonstar/reqwest).
     async fn check_default(&self, request: Request) -> Status {
         match self.reqwest_client.execute(request).await {
-            Ok(ref response) => Status::new(response, self.accepted.clone()),
+            Ok(response) => Status::new(&response, &self.accepted),
```
mre, line 91: Nice one!

lychee-lib/src/client.rs
mre: Yeah, I believe [`AcceptRange`](https://docs.rs/lychee-lib/latest/lychee_lib/struct.AcceptRange.html) is the way to go. We should also link to the docs where helpful.

lychee-lib/src/client.rs
```diff
@@ -357,7 +360,7 @@ impl ClientBuilder {
             None => builder,
         }
         .build()
-        .map_err(ErrorKind::NetworkRequest)?;
+        .map_err(ErrorKind::BuildRequestClient)?;
```
mre, line 363: Yes! Better.

lychee-lib/src/retry.rs
mre: Yeah, not sure about this. Doesn't that mean: "if a user decided to reject a [429 Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) status code, you should still retry"? To me, that sounds like the user should be in control here; if they decided to consider 429s as hard errors, so be it. 😉 That should be the final answer, and we shouldn't retry. So I'd remove that condition and see where it breaks. (Unless I'm misunderstanding.)

lychee-lib/src/types/status.rs
```diff
@@ -240,12 +235,13 @@ impl Status {
...
             Status::Excluded => "EXCLUDED".to_string(),
             Status::Error(e) => match e {
-                ErrorKind::NetworkRequest(e)
-                | ErrorKind::ReadResponseBody(e)
-                | ErrorKind::BuildRequestClient(e) => match e.status() {
-                    Some(code) => code.as_str().to_string(),
-                    None => "ERROR".to_string(),
-                },
+                ErrorKind::RejectedStatusCode(code) => code.as_str().to_string(),
+                ErrorKind::ReadResponseBody(e) | ErrorKind::BuildRequestClient(e) => {
```
mre, line 239: Did you leave out the `ErrorKind::NetworkRequest(e)` case on purpose on this line? It was there before.

mre (changes requested):
Added some remarks. Overall, I like the approach and would go forward with this.

lychee-lib/src/types/status.rs
thomas-zahner: Yes. Initially, I even tried to remove `ErrorKind::NetworkRequest` but I realised that this doesn't really make sense. (though it would still be possible)

Until now it was used for:

1. Status codes that are considered erroneous by reqwest by using `response.error_for_status_ref()` . This method creates a `reqwest::Error` with the internal `Kind` of `Status(StatusCode)`.
2. Any other `reqwest::Error` that is not handled in a special way 

With my changes the first point is undone. So from now on only the second case applies. This means that when matching on our `ErrorKind::NetworkRequest(reqwest::Error)` the `reqwest::Error` will never be of `Kind` `Status(StatusCode)` which also means that `e.status()` will never be `Some(_)`.

lychee-lib/src/retry.rs
thomas-zahner: So actually I think we should leave it like this. Otherwise it means that lychee will never retry requests when a 429 occurs. The reason I noticed it in the first place was because without these lines there are test failures. If we don't keep these two lines that would mean that `--max-retries` would now only apply to the above and below cases. (octocrab and reqwest errors)

lychee-lib/src/types/error.rs
```diff
@@ -14,7 +15,8 @@ use crate::{Uri, basic_auth::BasicAuthExtractorError, utils};
 #[derive(Error, Debug)]
 #[non_exhaustive]
 pub enum ErrorKind {
-    /// Network error while handling request
+    /// Network error while handling request.
+    /// This does not include erroneous status codes, `RejectedStatusCode` will used in that case.
```
mre:
```suggestion
    /// This does not include erroneous status codes, `RejectedStatusCode` will be used in that case.
```

lychee-lib/src/types/status.rs
mre: Makes sense, thanks.

lychee-lib/src/retry.rs
mre: Sounds good. Can we add a comment to explain that behavior. I'm sure I'll forget why it's there otherwise. 😅

thomas-zahner:
Thanks for the test. For some reason I forgot about that, this makes a lot of sense!

PR #1691, opened 2025-05-09
Remove exclude mail

thomas-zahner:
Updated the documentation in https://github.com/lycheeverse/lycheeverse.github.io/pull/93 accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Needed SHELL since I'm using zsh
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
 IMAGE_NAME := "lycheeverse/lychee"
 
 .PHONY: help
@@ -36,7 +36,7 @@ run: ## Run project locally
 
 .PHONY: docs
 docs: ## Generate and show documentation
-	cargo doc --open 
+	cargo doc --open
 
 .PHONY: lint
 lint: ## Run linter
@@ -55,6 +55,6 @@ doc: ## Open documentation
 
 .PHONY: screencast
 screencast: ## Create a screencast for the docs
-	termsvg rec --command=assets/screencast.sh recording.asc 
+	termsvg rec --command=assets/screencast.sh recording.asc
 	termsvg export --minify recording.asc --output=assets/screencast.svg
 	rm recording.asc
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,5 +1,5 @@
 # Needed SHELL since I'm using zsh
-SHELL := /bin/bash
+SHELL := /usr/bin/env bash
```
mre, line 2: nice

lychee-bin/src/client.rs
```diff
@@ -34,25 +34,6 @@ pub(crate) fn create(cfg: &Config, cookie_jar: Option<&Arc<CookieStoreMutex>>) -
         .map(|value| StatusCode::from_u16(*value))
         .collect::<Result<HashSet<_>, _>>()?;
 
-    // `exclude_mail` will be removed in 1.0. Until then, we need to support it.
```
mre, line 37: Yeah, technically the comment says "removed in 1.0", but I think it's fine to do so earlier. We probably even talked about it at some point. 😆

mre (approved):
The change looks good, but it will break [a lot of code](https://github.com/search?q=--exclude-mail+lychee&type=code)... 💣 💥 
We have to do so at some point, but I wonder if you'd like to go through the list and pick the most prominent examples and we create an issue on their side to notify them about the changes?

eyestoreoye:
Impressive project structure! The architecture really demonstrates solid design principles.

lychee-bin/src/client.rs
thomas-zahner: True and yeah we did talk about it :smile:

thomas-zahner:
Created the following issues:

* https://github.com/rafaribe/home-ops/issues/3549
* https://github.com/andkret/Cookbook/issues/230
* https://github.com/dend/awesome-product-management/issues/825

mre:
Thanks!

> Do I understand correctly that depending on a fixed version of lychee-action means that lychee-bin / lychee-lib are fixed?

Yeah, because of https://github.com/lycheeverse/lychee-action/blob/82202e5e9c2f4ef1a55a3d02563e1cb6041e5332/action.yml#L30

thomas-zahner:
That makes sense :+1: I will merge then

PR #1711, opened 2025-05-22
Upgrade to 2024 edition

JayJayArr:
This PR upgrades all of lychee to the Rust 2024 edition, not sure if this is helpful or if lychee needs some steps before this can be done.
MSRV would have to be increased to 1.85.0.

### DIFF
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -1,14 +1,14 @@
 use crate::archive::Archive;
 use crate::parse::parse_base;
 use crate::verbosity::Verbosity;
-use anyhow::{anyhow, Context, Error, Result};
+use anyhow::{Context, Error, Result, anyhow};
 use clap::builder::PossibleValuesParser;
-use clap::{arg, builder::TypedValueParser, Parser};
+use clap::{Parser, arg, builder::TypedValueParser};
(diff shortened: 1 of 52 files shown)


lychee-bin/src/options.rs
```diff
@@ -165,7 +165,7 @@ impl OutputMode {
 
 // Macro for generating default functions to be used by serde
 macro_rules! default_function {
-    ( $( $name:ident : $T:ty = $e:expr; )* ) => {
+    ( $( $name:ident : $T:ty = $e:expr_2021; )* ) => {
```
mre, line 168: I think `expr` should still work. This is just a safeguard.
See https://doc.rust-lang.org/edition-guide/rust-2024/macro-fragment-specifiers.html#migration
Same for the other `expr_2021` items.

mre:
Good job! Just one remark.

lychee-bin/src/options.rs
JayJayArr: Thanks for bringing this up, might confuse otherwise!

mre:
Guess you have to rebase one more time; sorry.

mre (approved):
Good times. Appreciate it!

PR #1712, opened 2025-05-23
Fix `test_exclude_example_domains`

### LINKED ISSUE #1688: Fix `test_exclude_example_domains`
In [the past](https://github.com/lycheeverse/lychee/pull/1581#issuecomment-2531867201) I stumbled upon the test `test_exclude_example_domains`. And just now I did again because weirdly enough it is excluded in the Makefile explicitly unlike any other test. After a little investigation I realised that the test fails with certain feature flags. This was probably the reason the test was excluded because it seemed flaky.

To reproduce `cd` into lychee-bin and then:

```
$ cargo test -- test_exclude_example_domains

test cli::test_exclude_example_domains ... ok
```

```
$ cargo test --all-features -- test_exclude_example_domains

test cli::test_exclude_example_domains ... FAILED
```

thomas-zahner:
Closes #1688

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,8 +45,8 @@ lint: ## Run linter
 
 .PHONY: test
 test: ## Run tests
-	cargo nextest run --all-targets --all-features --filter-expr '!test(test_exclude_example_domains)'
-	cargo nextest run --filter-expr 'test(test_exclude_example_domains)'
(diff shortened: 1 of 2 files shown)


mre:
Makes sense! Thanks!

PR #1713, opened 2025-05-23
fix: only check the fragment when it's a file

kemingy:
- related to https://github.com/lycheeverse/lychee/issues/1709

What's in this PR:

- only check fragments for files
- better err type
- more info in err msg

BTW, I'm not sure if the `Client.check_fragment` is used by any caller. cc @mre


mre:
Thanks for the quick turnaround on this one! Can you post a comment in the linked issue?

kemingy:
> Thanks for the quick turnaround on this one! Can you post a comment in the linked issue?

Sure.

PR #1720, opened 2025-06-04
Archive in library

### LINKED ISSUE #1455: Suggestions with `lychee_lib`
Currently the `--archive` and `--suggest` functionality is only inside of `lychee_bin` and not `lychee_lib`. Would you agree that making it available through the library makes sense? This could be relevant in the future for the app I'm developing at work.

thomas-zahner:
Closes #1455

I've moved the core functionality to lychee-lib, added an example and updated the docs.
Did some debugging with the flaky tests and improved them accordingly.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -43,6 +43,7 @@
               pkgs.pkg-config
               pkgs.openssl
               rust
+              pkgs.cargo-nextest
             ];
           };
         }


examples/archive/Cargo.toml
mre, line 1: Thanks for adding a library example here

lychee-bin/src/formatters/stats/markdown.rs
```diff
@@ -154,12 +154,11 @@ impl StatsFormatter for Markdown {
 
 #[cfg(test)]
 mod tests {
-
     use http::StatusCode;
     use lychee_lib::{CacheStatus, InputSource, Response, ResponseBody, Status, Uri};
     use reqwest::Url;
 
-    use crate::archive::Suggestion;
+    use crate::formatters::stats::Suggestion;
```
mre: Should we move that to a separate module? Doesn't feel right if it's in stats. Probably we have to move the color codes then, but it's probably fine?

lychee-lib/src/archive/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if the `reqwest` client cannot be built, the request itself fails
+    /// or the API response cannot be parsed.
+    pub async fn get_snapshot(&self, url: &Url, timeout: Duration) -> Result<Option<Url>, Error> {
+        let function = match self {
+            Archive::WaybackMachine => wayback::get_wayback_link,
+        };
+
+        function(url, timeout).await
```
mre, line 36: Maybe we should rename that to `lookup_snapshot` or `get_archive_snapshot` or something?

lychee-lib/src/archive/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// Query the `Archive` to try and find the latest snapshot of the specified `url`.
+    /// Returns `None` if the specified `url` hasn't been archived in the past.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the `reqwest` client cannot be built, the request itself fails
+    /// or the API response cannot be parsed.
+    pub async fn get_snapshot(&self, url: &Url, timeout: Duration) -> Result<Option<Url>, Error> {
+        let function = match self {
+            Archive::WaybackMachine => wayback::get_wayback_link,
```
mre: Now the naming seems to be a bit unaligned. Perhaps we should now call it `get_archive_snapshot` or so?

lychee-lib/src/archive/wayback/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+use serde::de::Error as SerdeError;
+use serde::{Deserialize, Deserializer};
+
+use http::StatusCode;
+use reqwest::{Client, Error, Url};
+
+static WAYBACK_URL: LazyLock<Url> =
+    LazyLock::new(|| Url::parse("https://archive.org/wayback/available").unwrap());
+
+pub(crate) async fn get_wayback_link(url: &Url, timeout: Duration) -> Result<Option<Url>, Error> {
```
mre: I prefer the "snapshot" naming. It's a good idea to call it that and we should use that everywhere.

lychee-lib/Cargo.toml
```diff
@@ -53,6 +54,7 @@ secrecy = "0.10.3"
 serde = { version = "1.0.219", features = ["derive"] }
 serde_with = "3.12.0"
 shellexpand = "3.1.1"
+strum = { version = "0.27.1", features = ["derive"] }
```
mre, line 57: Did you check if we can remove any dependencies from the binary cargo toml?

lychee-lib/src/archive/wayback/mod.rs
thomas-zahner: Totally agree with the more consistent naming :+1:

lychee-lib/Cargo.toml
thomas-zahner: Yes, but unfortunately we can't. `strum` would now be used in both lychee-bin and lychee-lib. But that should be okay I guess. The reason is that I'd like to reuse `enum Archive` for lychee-bin and for the CLI argument with `clap`. Reimplementing the enum would be quite ugly.

lychee-bin/src/formatters/stats/markdown.rs
thomas-zahner: Ah, you're right. It doesn't fit into `stats` at all. I've now simply moved it up by one level. I think it should make sense there as it's pretty similar to `formatters::duration::Duration`.

mre:
Great. ✌️

PR #1725, opened 2025-06-11
Redirect reporting

thomas-zahner:
We want to report redirects at the end of the link check process, the same way we do for example with `--suggest`. The output format `markdown` for example should then also contain a section with all listed redirects.

**Summary bug**

There is a category `Redirected` in the summary. However, redirects were never counted. This is fixed with 6b28e2b.
This means `echo "https://http.codes/301 " | cargo run - --format detailed` yielded

```
📝 Summary
---------------------
🔍 Total............1
✅ Successful.......1
⏳ Timeouts.........0
🔀 Redirected.......0
👻 Excluded.........0
❓ Unknown..........0
🚫 Errors...........0
```

now it yields

```
📝 Summary
---------------------
🔍 Total............1
✅ Successful.......0
⏳ Timeouts.........0
🔀 Redirected.......1
👻 Excluded.........0
❓ Unknown..........0
🚫 Errors...........0
```

## Questions

1. I think it makes sense to show a `Redirects` section in the end just as we currently do with `Suggestions`. But I think we should show them by default without introducing a new feature/flag. Tying the feature to `--suggest` wouldn't make sense in my opinion. Do you agree?
2. As you can see in this PR I introduced a `redirect_map`. It is used to keep track of the redirects that lychee followed. This allows us to create the new summary section. (not yet implemented) The map is wrapped in an `Arc<Mutex<_>>`. Do you think this is okay? I probably could try to use channels instead. The reason for the `Arc<Mutex<_>>` is that `redirect::Policy::custom` takes `Fn(Attempt) -> Action + Send + Sync + 'static` as argument, so we can't mutate data in there. I also briefly considered using getting rid of `redirect::Policy::custom` and implement our own mechanism. But that turns out to be way more effort then necessary.

### DIFF
--- a/lychee-bin/src/stats.rs
+++ b/lychee-bin/src/stats.rs
@@ -72,7 +72,11 @@ impl ResponseStats {
             Status::Error(_) => self.errors += 1,
             Status::UnknownStatusCode(_) => self.unknown += 1,
             Status::Timeout(_) => self.timeouts += 1,
-            Status::Redirected(_) => self.redirects += 1,
+            Status::Redirected {
+                original: _,
+                resolved: _,
+                code: _,
+            } => self.redirects += 1,
             Status::Excluded => self.excludes += 1,
             Status::Unsupported(_) => self.unsupported += 1,
             Status::Cached(cache_status) => {
--- a/lychee-bin/src/stats.rs
+++ b/lychee-bin/src/stats.rs
@@ -89,6 +89,7 @@ impl ResponseStats {
 
     /// Add a response status to the appropriate map (success, fail, excluded)
     fn add_response_status(&mut self, response: Response) {
+        // TODO: redirect map?
         let status = response.status();
         let source = response.source().clone();
         let status_map_entry = match status {
--- a/lychee-lib/src/checker/website.rs
+++ b/lychee-lib/src/checker/website.rs
@@ -10,7 +10,11 @@ use async_trait::async_trait;
 use http::{Method, StatusCode};
 use octocrab::Octocrab;
 use reqwest::{Request, Response};
-use std::{collections::HashSet, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 #[derive(Debug, Clone)]
 pub(crate) struct WebsiteChecker {
@@ -50,6 +54,8 @@ pub(crate) struct WebsiteChecker {
 
     /// Utility for performing fragment checks in HTML files.
     fragment_checker: FragmentChecker,
+
+    redirect_map: Arc<Mutex<HashMap<url::Url, url::Url>>>,
 }
 
 impl WebsiteChecker {
@@ -57,6 +63,7 @@ impl WebsiteChecker {
     pub(crate) fn new(
         method: reqwest::Method,
         retry_wait_time: Duration,
+        redirect_map: Arc<Mutex<HashMap<url::Url, url::Url>>>,
         max_retries: u64,
         reqwest_client: reqwest::Client,
         accepted: HashSet<StatusCode>,
@@ -70,6 +77,7 @@ impl WebsiteChecker {
             reqwest_client,
             github_client,
             plugin_request_chain,
+            redirect_map,
             max_retries,
             retry_wait_time,
             accepted,
@@ -87,6 +95,10 @@ impl WebsiteChecker {
         let mut status = self.check_default(clone_unwrap(&request)).await;
         while retries < self.max_retries {
             if status.is_success() || !status.should_retry() {
+                if let Some(url) = self.redirect_map.lock().unwrap().get(dbg!(request.url())) {
+                    todo!("Redirected: {}", url)
+                }
+
                 return status;
             }
             retries += 1;
@@ -94,6 +106,7 @@ impl WebsiteChecker {
             wait_time = wait_time.saturating_mul(2);
             status = self.check_default(clone_unwrap(&request)).await;
         }
+
         status
     }
 
--- a/lychee-lib/src/checker/website.rs
+++ b/lychee-lib/src/checker/website.rs
@@ -116,8 +116,15 @@ impl WebsiteChecker {
                 if self.include_fragments && status.is_success() && method == Method::GET {
                     return self.check_html_fragment(status, response).await;
                 }
-                if let Some(redirection) = self.redirect_map.lock().unwrap().get(&url) {
-                    todo!("Redirected from {} to {}", url, redirection)
+
+                if let Some(code) = status.code() {
+                    if let Some(resolved) = self.redirect_map.lock().unwrap().get(&url).cloned() {
+                        return Status::Redirected {
+                            original: url,
+                            resolved,
+                            code,
+                        };
+                    }
                 }
 
                 status
--- a/lychee-lib/src/checker/website.rs
+++ b/lychee-lib/src/checker/website.rs
@@ -95,10 +95,6 @@ impl WebsiteChecker {
         let mut status = self.check_default(clone_unwrap(&request)).await;
         while retries < self.max_retries {
             if status.is_success() || !status.should_retry() {
-                if let Some(url) = self.redirect_map.lock().unwrap().get(dbg!(request.url())) {
-                    todo!("Redirected: {}", url)
-                }
-
                 return status;
             }
             retries += 1;
@@ -113,14 +109,18 @@ impl WebsiteChecker {
     /// Check a URI using [reqwest](https://github.com/seanmonstar/reqwest).
     async fn check_default(&self, request: Request) -> Status {
         let method = request.method().clone();
+        let url = request.url().clone();
         match self.reqwest_client.execute(request).await {
             Ok(response) => {
                 let status = Status::new(&response, &self.accepted);
                 if self.include_fragments && status.is_success() && method == Method::GET {
-                    self.check_html_fragment(status, response).await
-                } else {
-                    status
+                    return self.check_html_fragment(status, response).await;
                 }
+                if let Some(redirection) = self.redirect_map.lock().unwrap().get(&url) {
+                    todo!("Redirected from {} to {}", url, redirection)
+                }
+
+                status
             }
             Err(e) => e.into(),
         }
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -335,7 +335,7 @@ impl ClientBuilder {
         let max_redirects = self.max_redirects;
         let redirect_policy = redirect::Policy::custom(move |attempt| {
             if attempt.previous().len() > max_redirects {
-                attempt.error("too many redirects")
+                attempt.stop()
             } else {
                 debug!("Redirecting to {}", attempt.url());
                 attempt.follow()
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -855,45 +855,28 @@ mod tests {
             .mount(&mock_server)
             .await;
 
-        let client = ClientBuilder::builder()
+        let res = ClientBuilder::builder()
             .max_redirects(0_usize)
             .build()
             .client()
+            .unwrap()
+            .check(redirect_uri.clone())
+            .await
             .unwrap();
+        assert_eq!(
+            res.status(),
+            &Status::Redirected(StatusCode::PERMANENT_REDIRECT)
+        );
 
-        let res = client.check(redirect_uri.clone()).await.unwrap();
-        assert!(res.status().is_error());
-
-        let client = ClientBuilder::builder()
+        let res = ClientBuilder::builder()
             .max_redirects(1_usize)
             .build()
             .client()
+            .unwrap()
+            .check(redirect_uri)
+            .await
             .unwrap();
-
-        let res = client.check(redirect_uri).await.unwrap();
-        assert!(res.status().is_success());
-    }
-
-    #[tokio::test]
-    async fn test_limit_max_redirects() {
-        let mock_server = wiremock::MockServer::start().await;
-
-        // Set up permanent redirect loop
-        let template = wiremock::ResponseTemplate::new(StatusCode::PERMANENT_REDIRECT)
-            .insert_header("Location", mock_server.uri().as_str());
-        wiremock::Mock::given(wiremock::matchers::method("GET"))
-            .respond_with(template)
-            .mount(&mock_server)
-            .await;
-
-        let client = ClientBuilder::builder()
-            .max_redirects(0_usize)
-            .build()
-            .client()
-            .unwrap();
-
-        let res = client.check(mock_server.uri()).await.unwrap();
-        assert!(res.status().is_error());
+        assert_eq!(res.status(), &Status::Ok(StatusCode::OK));
     }
 
     #[tokio::test]
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -13,7 +13,11 @@
     clippy::default_trait_access,
     clippy::used_underscore_binding
 )]
-use std::{collections::HashSet, sync::Arc, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 use http::{
     StatusCode,
@@ -331,16 +335,7 @@ impl ClientBuilder {
             HeaderValue::from_static("chunked"),
         );
 
-        // Custom redirect policy to enable logging of redirects.
-        let max_redirects = self.max_redirects;
-        let redirect_policy = redirect::Policy::custom(move |attempt| {
-            if attempt.previous().len() > max_redirects {
-                attempt.stop()
-            } else {
-                debug!("Redirecting to {}", attempt.url());
-                attempt.follow()
-            }
-        });
+        let mut redirect_map = Arc::new(Mutex::new(HashMap::new()));
 
         let mut builder = reqwest::ClientBuilder::new()
             .gzip(true)
@@ -348,7 +343,7 @@ impl ClientBuilder {
             .danger_accept_invalid_certs(self.allow_insecure)
             .connect_timeout(Duration::from_secs(CONNECT_TIMEOUT))
             .tcp_keepalive(Duration::from_secs(TCP_KEEPALIVE))
-            .redirect(redirect_policy);
+            .redirect(redirect_policy(redirect_map.clone(), self.max_redirects));
 
         if let Some(cookie_jar) = self.cookie_jar {
             builder = builder.cookie_provider(cookie_jar);
@@ -392,6 +387,7 @@ impl ClientBuilder {
         let website_checker = WebsiteChecker::new(
             self.method,
             self.retry_wait_time,
+            redirect_map.clone(),
             self.max_retries,
             reqwest_client,
             self.accepted,
@@ -404,6 +400,7 @@ impl ClientBuilder {
         Ok(Client {
             remaps: self.remaps,
             filter,
+            redirect_map: redirect_map.clone(),
             email_checker: MailChecker::new(),
             website_checker,
             file_checker: FileChecker::new(
@@ -415,6 +412,30 @@ impl ClientBuilder {
     }
 }
 
+fn redirect_policy(
+    redirect_map: Arc<Mutex<HashMap<url::Url, url::Url>>>,
+    max_redirects: usize,
+) -> redirect::Policy {
+    let redirect_policy = redirect::Policy::custom(move |attempt| {
+        if let Some(first) = attempt.previous().first() {
+            redirect_map
+                .lock()
+                .unwrap()
+                .insert(first.clone(), attempt.url().clone());
+        }
+
+        dbg!(&redirect_map); // todo
+
+        if attempt.previous().len() > max_redirects {
+            attempt.stop()
+        } else {
+            debug!("Redirecting to {}", attempt.url());
+            attempt.follow()
+        }
+    });
+    redirect_policy
+}
+
 /// Handles incoming requests and returns responses.
 ///
 /// See [`ClientBuilder`] which contains sane defaults for all configuration
@@ -424,7 +445,7 @@ pub struct Client {
     /// Optional remapping rules for URIs matching pattern.
     remaps: Option<Remaps>,
 
-    /// Rules to decided whether each link should be checked or ignored.
+    /// Rules to decide whether a given link should be checked or ignored.
     filter: Filter,
 
     /// A checker for website URLs.
@@ -435,6 +456,8 @@ pub struct Client {
 
     /// A checker for email URLs.
     email_checker: MailChecker,
+
+    redirect_map: Arc<Mutex<HashMap<url::Url, url::Url>>>,
 }
 
 impl Client {
@@ -461,15 +484,6 @@ impl Client {
             source,
             ..
         } = request.try_into()?;
-
-        // Allow filtering based on element and attribute
-        // if !self.filter.is_allowed(uri) {
-        //     return Ok(Response::new(
-        //         uri.clone(),
-        //         Status::Excluded,
-        //         source,
-        //     ));
-        // }
 
         self.remap(uri)?;
 
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -335,7 +335,7 @@ impl ClientBuilder {
             HeaderValue::from_static("chunked"),
         );
 
-        let mut redirect_map = Arc::new(Mutex::new(HashMap::new()));
+        let redirect_map = Arc::new(Mutex::new(HashMap::new()));
 
         let mut builder = reqwest::ClientBuilder::new()
             .gzip(true)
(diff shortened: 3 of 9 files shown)


mre:
Glad you're tackling this.

> 1. I think it makes sense to show a Redirects section in the end just as we currently do with Suggestions. But I think we should show them by default without introducing a new feature/flag. Tying the feature to --suggest wouldn't make sense in my opinion. Do you agree?

Yeah, that's fine. Although the naming of the flag is a bit confusing then. Maybe `--show-suggestions` to make it more explicit? I know that it's another breaking change, but it might be worth it in the long run to change the naming now.

> 2. As you can see in this PR I introduced a redirect_map. It is used to keep track of the redirects that lychee followed. This allows us to create the new summary section. (not yet implemented) The map is wrapped in an Arc<Mutex<_>>. Do you think this is okay? I probably could try to use channels instead. The reason for the `Arc<Mutex<_>>` is that `redirect::Policy::custom` takes `Fn(Attempt) -> Action + Send + Sync + 'static` as argument, so we can't mutate data in there. I also briefly considered using getting rid of `redirect::Policy::custom` and implement our own mechanism. But that turns out to be way more effort then necessary.

Looking at planned features (like per-host rate-limiting and recursion), I believe we'll end up adding more metadata to requests and responses in the future anyway.

The question is if we should start with this PR already and instead of a redirect-map, we add a `redirects` field, which is a `Vec<Url>` or so. It could serve as a sort of "audit log" for a request, and my hope is that it fits in nicely with the request chain. There might be constraints I'm not aware of yet, though.

thomas-zahner:
Sure, no problem. Thanks for the idea in your second point, I'll take a look at it.

> Maybe --show-suggestions to make it more explicit?

So you think we should rename the feature flag even though we don't change its functionality?

mre:
I think we should probably separate redirect tracking from the existing `--suggest` flag, though I'm finding it tricky to land on a definitive answer here. The core issue seems to be that `--suggest` was designed for "show alternative URLs you should use instead" but redirect tracking feels different - it's more about "what actually happened during requests."

I initially thought redirects could be considered suggestions since you might want to update links to point directly to the final destination, but the reality is more nuanced. Many redirects are intentional and shouldn't be "fixed" - URL shorteners, CDN routing, domain migrations with proper redirects, etc. Treating every redirect as a "suggestion to fix" might create noise and dilute the actionable nature of actual suggestions. If we at some point introduce an `--apply` flag, we can no longer separate between fixing broken links and fixing redirects, which are might not be an issue.

The use cases feel somewhat different too. Suggestions tend to answer "what's broken and needs fixing" while redirects answer "what path did this request take." Broken links generally force action - your site is broken until you fix them. Redirects are more about inefficiency - your site works fine, you're just making extra HTTP requests. Perhaps some users are also just curious about the redirect chain (similar to `traceroute`), but don't care about archive links?

I'm leaning toward keeping `--suggest` focused on actionable fixes for broken links and adding another flag, maybe `--redirects`, for redirect tracking. This would give users clearer expectations about what each flag does and let people choose exactly the information they need. But I can see arguments for keeping them together too - it's not as clear-cut as I would like it to be.

thomas-zahner:
Yes I agree with that. That was basically my initial proposal, maybe you misunderstood me.

> Tying the feature to --suggest wouldn't make sense in my opinion. Do you agree?

`--suggest` can also be slow or unreliable which is another reason not to tie the features together into a single CLI flag.

> But I think we should show them by default without introducing a new feature/flag.

What do you think about this? I'm also okay with your proposal of a new flag called `--redirects`. But maybe we want to always show the redirects, only depending on the verbosity, output format and mode.

mre:
> What do you think about this? I'm also okay with your proposal of a new flag called `--redirects`. But maybe we want to always show the redirects, only depending on the verbosity, output format and mode.

Yes! Good idea to show them by default. Or alternative with `--verbose`? But by default is fine of course.

mre:
> Yes I agree with that. That was basically my initial proposal, maybe you misunderstood me.

Yes, probably. Sorry about that.

thomas-zahner:
> The question is if we should start with this PR already and instead of a redirect-map, we add a redirects field, which is a Vec<Url> or so. It could serve as a sort of "audit log" for a request, and my hope is that it fits in nicely with the request chain. There might be constraints I'm not aware of yet, though.

Yes, it probably makes sense to do that at one point. However, this will not resolve the `Arc<Mutex<_>>` situation these are two separate concerns. The problem is that we build a single reqwest `Client` (which makes sense) for all requests and pass a single `redirect::Policy` argument for handling redirects. Given this situation we need to somehow keep track of the redirects, which I did with `Arc<Mutex<_>>` for now. I could now try to use another approach, such as channels. The only other alternative I see, is to not use reqwest's redirection handling and implement our own mechanism. (your suggestion might also assume this) But looking at the code I think we should try to use reqwest's battle tested redirection handling as it would be too much (unnecessary) effort to implement our own mechanisms.

lychee-lib/src/client.rs
```diff
@@ -415,6 +412,28 @@ impl ClientBuilder {
...
+                .lock()
+                .unwrap()
+                .insert(first.clone(), attempt.url().clone());
+        }
+
+        if attempt.previous().len() > max_redirects {
+            attempt.stop()
+        } else {
+            debug!("Redirecting to {}", attempt.url());
+            attempt.follow()
```
mre: Do we want to update the redirect map in that case? I don't know, but I in general it would be nice to have the full "chain" of redirects for reporting.

lychee-lib/src/client.rs
```diff
@@ -855,45 +867,28 @@ mod tests {
...
             .await;
 
-        let client = ClientBuilder::builder()
+        let res = ClientBuilder::builder()
             .max_redirects(0_usize)
             .build()
             .client()
+            .unwrap()
+            .check(redirect_uri.clone())
+            .await
```
mre: How about we create the client as before and check the `redirect_uri` in a separate step?

mre:
Makes sense, yeah; `reqwest::redirect::Policy` is pretty nice.

The draft looks good.
Perhaps we would like to introduce an actual type for tracking the redirects? How about this?

```rust
use std::{collections::HashMap, sync::{Arc, Mutex}};
use url::Url;

#[derive(Debug, Clone)]
pub struct RedirectTracker(Arc<Mutex<HashMap<Url, Url>>>);

impl RedirectTracker {
    pub fn new() -> Self {
        Self(Arc::new(Mutex::new(HashMap::new())))
    }

    pub fn record_redirect(&self, original: Url, resolved: Url) {
        if let Ok(mut map) = self.0.lock() {
            map.insert(original, resolved);
        }
    }

    pub fn get_resolved(&self, original: &Url) -> Option<Url> {
        self.0.lock().ok()?.get(original).cloned()
    }

    pub fn all_redirects(&self) -> HashMap<Url, Url> {
        self.0.lock().unwrap_or_else(|_| {
            // Handle poisoned mutex gracefully
            HashMap::new()
        }).clone()
    }

    pub fn clear(&self) {
        if let Ok(mut map) = self.0.lock() {
            map.clear();
        }
    }
}

impl Default for RedirectTracker {
    fn default() -> Self {
        Self::new()
    }
}
```

I also thought of a simple type alias, but I believe this is a bit more declarative and encapsulates the actual tracking.
However, I don't want us to lose the ability to refactor the code later, so feel free to keep it simple.

lychee-lib/src/client.rs
thomas-zahner: Ah yes, recording the whole redirection chain probably make sense. This would mean something like this `RedirectTracker(Arc<Mutex<Vec<Vec<Url>>>>)`, right?

thomas-zahner:
@mre Thanks, I like the `RedirectTracker` type :+1:

lychee-lib/src/client.rs
mre: Correct.

thomas-zahner:
**FYI**

In 75135375d0d274ab4eecc6138dc850567617bf62 I removed `Status::Redirected`. It was not used anyways and conceptually it should not exist. `Status` represents the final state after having checked a link. So after following redirects the status should resolve normally.

When the redirect limit is reached the resulting status should be treated just like any normal status code. Depending on the accept config this will be either `Error(ErrorKind::RejectedStatusCode(3XX))` or `Ok(3XX)` when `--accept` is configerd as such.

This is the best and simplest approach in my opinion because of the configurability with `--accept`. I've also removed `ErrorKind::TooManyRedirects` because it's not used. It was only ever produced by converting from a `reqwest::Error` when `e.is_redirect`. But this only happens when something really is wrong with HTTP redirection (e.g. with the headers),
so converting that into `ErrorKind::NetworkRequest(e)` makes more sense. If we didn't overwrite the redirection policy and used the default policy than it migh have made sense to keep it. But this is no longer the case and the behaviour is covered with the new test `test_max_redirects`.

mre:
Fully agree. Great explanation. 👌

lychee-lib/src/types/redirect_tracker.rs
```diff
@@ -0,0 +1,67 @@
...
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+use url::Url;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Default)]
+/// A list of URLs that were followed through HTTP redirects,
+/// starting from the original URL and ending at the final destination.
+/// Each entry in the chain represents a step in the redirect sequence.
+pub struct RedirectChain(Vec<Url>);
```
mre: Now that I think about it, the naming could be a bit ambiguous. We already use `Chain` for requests and it's the chain-of-command pattern. Should we just call this `Redirects` maybe?

lychee-lib/src/types/redirect_tracker.rs
```diff
@@ -0,0 +1,67 @@
...
+
+impl From<Vec<Url>> for RedirectChain {
+    fn from(value: Vec<Url>) -> Self {
+        Self(value)
+    }
+}
+
+impl RedirectChain {
+    /// Count how many times a redirect was followed.
+    /// This is the lenght of the chain minus one
```
mre:
```suggestion
    /// This is the length of the chain minus one
```

lychee-lib/src/types/redirect_tracker.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub(crate) fn redirect_count(&self) -> usize {
+        self.0.len().saturating_sub(1)
+    }
+}
+
+#[derive(Debug, Clone)]
+/// Keep track of HTTP redirections for reporting
+pub(crate) struct RedirectTracker(Arc<Mutex<HashMap<Url, RedirectChain>>>);
```
mre: How about `RedirectHistory` or `RedirectCache` instead?

lychee-lib/src/types/redirect_tracker.rs
```diff
@@ -0,0 +1,67 @@
...
+        Self(Arc::new(Mutex::new(HashMap::new())))
+    }
+
+    pub(crate) fn record_redirect(&self, redirect_chain: &[Url]) {
+        if let Ok(mut map) = self.0.lock() {
+            if let Some(first) = redirect_chain.first() {
+                map.insert(first.clone(), RedirectChain(redirect_chain.to_vec()));
+            }
+        }
+    }
```
mre: I think you could flatten that a bit.
I also got a bit confused by the `if let Some(first) = redirect_chain.first()` thing, so I tried to add a doc-comment to explain what's going on. Please correct me if I'm wrong here.

```rust
/// Records a redirect chain, using the original URL as the key.
/// 
/// The first URL in the chain is treated as the original request URL,
/// and the entire chain (including the original) is stored as the value.
/// This allows later lookup of redirect paths by the initial URL.
pub(crate) fn record_redirect(&self, redirect_chain: &[Url]) {
   let mut map = self.0.lock().ok()?;
   let first = redirect_chain.first()?;
   map.insert(first.clone(), RedirectChain(redirect_chain.to_vec()));
}
```

lychee-lib/src/types/redirect_tracker.rs
```diff
@@ -0,0 +1,67 @@
...
+            Status::Ok(code) => {
+                if let Some(chain) = self.get_resolved(url) {
+                    Status::Redirected(code, chain)
+                } else {
+                    Status::Ok(code)
+                }
+            }
+            s => s,
+        }
+    }
```
mre: And early return would help make that easier to read, in my opinion:

```rust
pub(crate) fn handle_redirected(&self, url: &Url, status: Status) -> Status {
    let Status::Ok(code) = status else {
        return status;
    };
    
    match self.get_resolved(url) {
        Some(chain) => Status::Redirected(code, chain),
        None => Status::Ok(code),
    }
}
```

mre (commented):
Looks good already! I know this isn't fully done yet, but I still added a few remarks I had while reading the code, just so that I don't forget.

thomas-zahner:
@mre Thanks for taking a look already! I've just now finished some final changes and the PR is now in a state I'm happy with, apart from this one other comment. Will reply to your comments later.

lychee-lib/src/test_utils.rs
```diff
@@ -34,6 +34,38 @@ where
         .unwrap()
 }
 
+/// Set up a mock server which has two routes: `/ok` and `/redirect`.
+/// Calling `/redirect` returns a HTTP Location header redirecting to `/ok`
+pub async fn redirecting_mock_server<F, R>(f: F)
```
thomas-zahner: @mre I've extracted this new function because I wanted to reuse it in another test. But I realised that I need it once in lychee-lib and once in lychee-bin. I've now temporarily just duplicated the code. What would be the best solution here? The reason I can't import it in lychee-bin is because its gated behind `cfg(test)` which is then unavailable in the tests in lychee-bin. Would it make sense to create a new module/crate with test utils which we could then use in both?

lychee-lib/src/types/redirect_tracker.rs
thomas-zahner: Thanks for the doc comment, that's exactly my thoughts. It would be possible to not store the first element as `Redirects` value in the `HashMap` and I have considered that. However, I think this is a much simpler design and it works well.

Using `?` here doesn't work, as I don't return an `Option` or `Result`. Would you return a `Result`? For now I've combined the nested `if let` into a single one.

lychee-lib/src/types/redirect_tracker.rs
thomas-zahner: Ah right. I've now used `.map` which I even prefer a bit more

lychee-lib/src/test_utils.rs
mre: Yes, that's the way to go. 👍

thomas-zahner:
> In https://github.com/lycheeverse/lychee/commit/75135375d0d274ab4eecc6138dc850567617bf62 I removed Status::Redirected. It was not used anyways and conceptually it should not exist.

I've reconsidered this statement and realised that this is not the path we should follow. I've reverted mentioned commit, so that `Status::Redirected` still exists. It is true, that previously `Status::Redirected` was useless because a link check could never resolve to that status.

# Summary

## Updated default behaviour

```
echo 'http://google.com/jobs' | lychee -

🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors
```

Now we list the `Redirects` category separately. It's no longer `OK`, however lychee still exits with `0`.
This makes it much easier for users to realise that their links might potentially be outdated.

```
echo 'http://google.com/jobs' | cargo run -

🔍 1 Total (in 0s) ✅ 0 OK 🚫 0 Errors 🔀 1 Redirects
```

## Max redirects

The following lead to a hard error.

```
echo 'http://google.com/jobs' | lychee - --max-redirects 0

[stdin]:
   [ERROR] http://google.com/jobs | Too many redirects: error following redirect for url (http://google.com/jobs)

```

Now it does no longer result in a "hard" error, but the link is just no longer followed so that the result will normally be a 3XX status code.

```
echo 'http://google.com/jobs' | cargo run - --max-redirects 0

[stdin]:
     [301] http://google.com/jobs | Rejected status code (this depends on your "accept" configuration): Moved Permanently

```

## Detailed output and JSON

Previously it was impossible to discern redirects.

```
echo 'http://google.com/jobs' | lychee - --format json --verbose

     [200] http://google.com/jobs

{
  "total": 1,
  "successful": 1,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 0,
  "excludes": 0,
  "errors": 0,
  "cached": 0,
  "success_map": {
    "stdin": [
      {
        "url": "http://google.com/jobs",
        "status": {
          "text": "200 OK",
          "code": 200
        }
      }
    ]
  },
  "error_map": {},
  "suggestion_map": {},
  "excluded_map": {},
  "duration_secs": 0,
  "detailed_stats": true
}

```

Now it is and users get new useful information.

```
echo 'http://google.com/jobs' | cargo run - --format json --verbose

     [200] http://google.com/jobs | Redirect: Followed 2 redirects resulting in OK.

{
  "total": 1,
  "successful": 0,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 1,
  "excludes": 0,
  "errors": 0,
  "cached": 0,
  "success_map": {},
  "error_map": {},
  "suggestion_map": {},
  "redirect_map": {
    "stdin": [
      {
        "url": "http://google.com/jobs",
        "status": {
          "text": "Redirect",
          "code": 200,
          "redirects": [
            "http://google.com/jobs",
            "https://www.google.com/jobs",
            "https://jobs.google.com/about/"
          ]
        }
      }
    ]
  },
  "excluded_map": {},
  "duration_secs": 0,
  "detailed_stats": true
}
```

mre:
> `[301] http://google.com/jobs | Rejected status code (this depends on your "accept" configuration): Moved Permanently`

This error message sounds a bit misleading to me given that changing the `accept` configuration would not improve the situation. Maybe we should change it to "Too many redirects"?

lychee-bin/src/formatters/response/plain.rs
```diff
@@ -69,12 +70,12 @@ mod plain_tests {
     fn test_format_response_with_redirect_status() {
         let formatter = PlainFormatter;
         let body = mock_response_body(
-            Status::Redirected(StatusCode::MOVED_PERMANENTLY),
+            Status::Redirected(StatusCode::MOVED_PERMANENTLY, Redirects::default()),
```
mre: We can introduce Redirects::zero() or Redirects::none() to make that clearer. Of course we could put that into a separate impl block with `#[cfg(test)]`.

lychee-bin/src/stats.rs
```diff
@@ -37,13 +37,15 @@ pub(crate) struct ResponseStats {
...
+    /// Store successful responses (if `detailed_stats` is enabled)
     pub(crate) success_map: HashMap<InputSource, HashSet<ResponseBody>>,
-    /// Map to store failed responses (if `detailed_stats` is enabled)
+    /// Store failed responses (if `detailed_stats` is enabled)
     pub(crate) error_map: HashMap<InputSource, HashSet<ResponseBody>>,
     /// Replacement suggestions for failed responses (if `--suggest` is enabled)
     pub(crate) suggestion_map: HashMap<InputSource, HashSet<Suggestion>>,
-    /// Map to store excluded responses (if `detailed_stats` is enabled)
+    /// Store redirected responses with their redirection list (if `detailed_stats` is enabled)
+    pub(crate) redirect_map: HashMap<InputSource, HashSet<ResponseBody>>,
```
mre, line 47: Do we really store the redirection list here as the comment suggests?

lychee-lib/src/checker/website.rs
```diff
@@ -187,20 +193,34 @@ impl WebsiteChecker {
...
-                if self
-                    .check_website_inner(&uri.to_https()?, &default_chain)
+        let status = self.check_website_inner(uri, &default_chain).await;
+        let status = self
+            .handle_insecure_url(uri, &default_chain, status)
+            .await?;
+        Ok(self.redirect_history.handle_redirected(&uri.url, status))
+    }
+
+    async fn handle_insecure_url(
```
mre, line 205: I believe we should add comments for new functions, even internal ones.

lychee-lib/src/client.rs
```diff
@@ -434,6 +429,19 @@ impl ClientBuilder {
     }
 }
 
+fn redirect_policy(redirect_history: RedirectHistory, max_redirects: usize) -> redirect::Policy {
```
mre, line 434: We should add a doc comment here to describe the logic.

lychee-lib/src/client.rs
```diff
@@ -481,15 +489,6 @@ impl Client {
             ..
         } = request.try_into()?;
 
-        // Allow filtering based on element and attribute
```
mre, line 484: This was dead code but can we replace it with a TODO? I think we still want to do that. We could even create an issue.

lychee-lib/src/types/redirect_history.rs
```diff
@@ -0,0 +1,68 @@
+use crate::Status;
+use serde::Serialize;
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
+use url::Url;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Default, Serialize)]
+/// A list of URLs that were followed through HTTP redirects,
```
mre, line 9: Recently started to put the doc comments before the derives. Looks a bit cleaner im my opinion.

test-utils/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "test-utils"
+authors = ["Thomas Zahner <thomas.zahner@protonmail.ch>", "Matthias Endler <matthias@endler.dev>"]
+edition = "2024"
+description = "Common test utilities for the lychee project, intended for internal use only"
```
mre, line 5: We could add `publish = false` to prevent accidental publication

thomas-zahner:
> > `[301] http://google.com/jobs | Rejected status code (this depends on your "accept" configuration): Moved Permanently`
> 
> This error message sounds a bit misleading to me given that changing the `accept` configuration would not improve the situation. Maybe we should change it to "Too many redirects"?

@mre That was the behaviour before this PR, it always resulted in `Too many redirects`. Now as we simply stop following further redirects when the limit is reached. Providing `--accept 301` actually does make the error go away as indicated by the error message. So I think the error message is perfectly accurate.

> What's the benefit of switching to macros? Or was it necessary because of the changes?

The macros `mock_server` and `redirecting_mock_server` previously were two duplicated functions in the tests of lychee-lib and lychee-bin. Now we share the single definition. `mock_server` was a macro already. I then ported all functions from lychee-lib/src/test_utils.rs to the new test-utils crate to make them reusable across both crates. I chose to make the functions macros because otherwise we have dependency cycles which is rightfully prevented by the compiler. For example `get_mock_client_response` uses `ClientBuilder`. This means test-utils depends on lychee-lib if it weren't a macro. For testing we add test-utils as dev-dependency to lychee-lib. This creates an unresolvable cycle. I've now update the doc comment to explain it.

lychee-bin/src/stats.rs
thomas-zahner: Yes, see my previous comment section `Detailed output and JSON`. It's the new `redirect_map` field. I've added it for a more detailed JSON output. Maybe `with their redirection list` was a bit misleading? I've removed it now.

lychee-bin/src/formatters/response/task.rs
```diff
@@ -60,12 +60,12 @@ mod task_tests {
...
         let formatter = TaskFormatter;
         let body = mock_response_body(
-            Status::Redirected(StatusCode::MOVED_PERMANENTLY),
+            Status::Redirected(StatusCode::MOVED_PERMANENTLY, Redirects::none()),
             "https://example.com/redirect",
         );
         assert_eq!(
             formatter.format_response(&body),
-            "- [ ] [301] https://example.com/redirect | Redirect (301 Moved Permanently): Moved Permanently"
+            "- [ ] [301] https://example.com/redirect | Redirect: Followed 0 redirects resulting in Moved Permanently."
```
mre: Idea: maybe we can change the status message a bit. 

```suggestion
            "- [ ] [301] https://example.com/redirect | Redirect: Followed 0 redirects resulting in: Moved Permanently."
```

or

```rust
Redirect: Followed 0 redirects resulting with final status: Moved Permanently.
```

lychee-lib/src/types/redirect_history.rs
thomas-zahner: I agree. Fixed for the whole project with bfd231c47a883fb1f4a4d0ed415525d132a04318

lychee-lib/src/client.rs
thomas-zahner: Yeah, creating an issue for that definitely makes sense. Could you do that? Because I'm not quite sure what the comment and disabled code tried to convey. To me, the comment does not seem to relate to the code below.

Is the comment requesting a feature like: Allow filtering based on _HTML_ elements and _HTML_ attributes? If so, the comment is in the wrong place.

lychee-bin/src/formatters/response/task.rs
thomas-zahner: It's now `Redirect: Followed 0 redirects resolving to the final status of: Moved Permanently`. WDYT? `resulting with final status` sounded grammatically incorrect to me.

mre: yes, sounds good

Issue #1297, opened 2023-11-14
Security: don't print provided HTTP header value as part of help info


sanmai-NL:
Printing the actual provided HTTP request header is not needed to help the user. But it does add an information leak in logs. I think that a custom header in most applications of Lychee in practice, would contain sensitive values.

## Actual behavior

```console
$ read gitlabtoken
read> glpat-SECRET
$ lychee --header "PRIVATE-TOKEN=$gitlabtoken" https://gitlab.com/mysecretproject
Error: Header value must be of the form key=value, got PRIVATE-TOKEN: glpat-SECRET
```

## Desired behavior

```console
$ read gitlabtoken
read> glpat-SECRET
$ lychee --header "PRIVATE-TOKEN=$gitlabtoken" https://gitlab.com/mysecretproject
Error: Header value must be of the form key=value.
```

sanmai-NL:
Note that in the previous example, the secret token is provided on the command line, rather than read as a secret. This also a security antipattern, since it leaks secrets to the shell logs. An additional enhancement would be to allow header specification in a file (not a configuration file, useful to be shareable and with non-sensitive content).

sanmai-NL:
Given #1298, this issue could be quick-fixed by keeping the current functionality intact but warning in the docs and UI that this parameter should never be used for secrets and for values that are host specific.

Alternatively, if this feature scope seems too limiting, the feature could be deprecated in favor of the solution for #1298.

mre:
I like it.

* [x] Change error message in https://github.com/lycheeverse/lychee/blob/b882a14161e186d1f40af4f8ae5c40dc04103fbb/lychee-bin/src/parse.rs#L11

We should also add https://github.com/lycheeverse/lychee/issues/1298, but we can discuss that over there.

mre:
This got fixed by @thomas-zahner in #1728. It will be released with the next version (v0.19.1 or v0.20.0).

PR #1728, opened 2025-06-12
Don't echo potentially sensitive header data

thomas-zahner:
Closes https://github.com/lycheeverse/lychee/issues/1297


lychee-bin/src/options.rs
```diff
@@ -212,19 +212,23 @@ macro_rules! fold_in {
...
 fn parse_single_header(header: &str) -> Result<(HeaderName, HeaderValue)> {
     let parts: Vec<&str> = header.splitn(2, ':').collect();
     match parts.as_slice() {
         [name, value] => {
             let name = HeaderName::from_bytes(name.trim().as_bytes())
-                .map_err(|e| anyhow!("Invalid header name '{}': {}", name.trim(), e))?;
+                .map_err(|e| anyhow!("Unable to read header name: {}", e))?;
             let value = HeaderValue::from_str(value.trim())
-                .map_err(|e| anyhow!("Invalid header value '{}': {}", value.trim(), e))?;
+                .map_err(|e| anyhow!("Unable to read header value: {}", e))?;
```
mre: Maybe we should log the header name here as well, which would help with troubleshooting? That should be safe and we know that we were able to parse it.

thomas-zahner: You are right :+1: changed it accordingly

mre (approved):

mre:
Great fix. 👍

PR #1730, opened 2025-06-16
fix: skip the fragment check if the uri doesn't contain fragment

kemingy:
As discussed in:
- https://github.com/lycheeverse/lychee/issues/1709

Otherwise, this may cause "Too many open files" because the `self.check_fragment` needs to open the file to build the `FragmentInput`.


mre:
Great progress. Thanks!

Issue #1723, opened 2025-06-11
Improve or add more information when a "Network error" error is reported


afalhambra-hivemq:
Currently, it's quite difficult to debug and troubleshoot why an error reported by Lychee is happening.
At the moment for some links we only get something like:

```
Network error: error sending request for url (https://www.foobar.com/) Maybe a certificate error?
```

Even though, we have followed the documentation regarding this [issue](https://lychee.cli.rs/troubleshooting/network-errors/), and increased the log level, but still it is pretty difficult to know what went wrong and how to fix it.

Would be nice to just simply show any further information about why any error of this kind has occurred or the underlying error.

mre:
The relevant code where we extract additional information from the underlying error is [here](https://github.com/lycheeverse/lychee/blob/31b2525a8d0422030161f01d92c85be94bacc492/lychee-lib/src/utils/reqwest.rs#L19). As you can see, there's not much going on except us trying to trim the first part, which is "error trying to connect:". If that string doesn't exist in the `reqwest` error, we simply use the string that gets returned. So we don't strip any information. (Also see [this quick test code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4a07ebf5ce586bb9cb1a2e81773f1ffa)).

In summary, that means that there's no additional information being returned from [`reqwest`](https://github.com/seanmonstar/reqwest), so I guess that would be an issue on their end? Unless someone finds a bug in that code.

afalhambra-hivemq:
Thanks @mre for the quick reply. I see.
If not further information can be logged, then it is what it is... 
All of a sudden we're getting quite a lot of these errors in our GitHub workflows, to our own websites... so everything looks fine using cURL command in our terminals.
As mentioned we updated the ca-certificates in the Ubuntu GitHub runner, used the flag `--insecure` and increased the verbosity but nothing found. 
So I'm just considering to use the Lychee Docker image instead, to see if the problem is coming from the environment.

afalhambra-hivemq:
Is there any further recommendation about how to tackle these errors? They are completely random failures from run to run, but mostly all of them related to our own hosted links.

mre:
If it helps, you can share a set of failing links with me via email or paste them here. I'll take a look then.

afalhambra-hivemq:
That would be really awesome. Thanks a lot. 

Here is a sample report we get:

[link-check-results1.md](https://github.com/user-attachments/files/20692026/link-check-results1.md)

[...]

PR #1731, opened 2025-06-16
Refactor error reporting

mre:
Related to https://github.com/lycheeverse/lychee/issues/1723

The goal is to get as much information from the error as possible. 
This started out as an experiment, but now I think it would be useful to provide better error messages.

The downside is that we'd have to pay close attention to changes in upstream error messages (in hyper/reqwest). I feel like that is a sensible tradeoff to make, given that the error messages rarely change, and we merely check for specific, relatively "robust" keywords.

### DIFF
--- a/lychee-lib/src/utils/reqwest.rs
+++ b/lychee-lib/src/utils/reqwest.rs
@@ -59,7 +59,6 @@ impl ErrorRules {
         }
 
         self.fallback
-            .clone()
             .unwrap_or_else(|| format!("Unhandled error: {error_msg}"))
     }
 }


lychee-lib/src/types/error.rs
```diff
@@ -182,33 +188,152 @@ impl ErrorKind {
...
-            }
-            _ => self.source().map(ToString::to_string),
+                        if let octocrab::Error::GitHub { source, .. } = &**e {
+                            Some(source.message.clone())
+                        } else {
+                            // Fall back to generic error analysis
+                            Some(e.to_string())
+                        }
+                    }
+            ErrorKind::InvalidFilePath(_uri) => None,
```
thomas-zahner: Is there a specific reason for returning `None` instead of a message? If not, we should probably add a message.

lychee-lib/src/types/error.rs
```diff
@@ -182,33 +188,152 @@ impl ErrorKind {
...
+            ErrorKind::Utf8(_utf8_error) => Some(
+                "Invalid UTF-8 sequence found. File contains non-UTF-8 characters".to_string()
+            ),
+            ErrorKind::BuildGithubClient(error) => Some(format!(
+                "Failed to create GitHub client: {error}. Check token and network connectivity",
+            )),
+            ErrorKind::InvalidGithubUrl(url) => Some(format!(
+                "Invalid GitHub URL format: '{url}'. Check URL syntax",
+            )),
+            ErrorKind::InvalidFragment(_uri) => None,
```
thomas-zahner: What's the reason here?

lychee-lib/src/types/error.rs
```diff
@@ -182,33 +188,152 @@ impl ErrorKind {
...
+            ErrorKind::BasicAuthExtractorError(basic_auth_extractor_error) => Some(format!(
+                "Basic authentication error: {basic_auth_extractor_error}. Check credentials format",
+            )),
+            ErrorKind::Cookies(reason) => Some(format!(
+                "Cookie handling error: {reason}. Check cookie file format",
+            )),
+            ErrorKind::StatusCodeSelectorError(status_code_selector_error) => Some(format!(
+                "Status code selector error: {status_code_selector_error}. Check accept configuration",
+            )),
+            ErrorKind::InvalidIndexFile(_path) => None,
```
thomas-zahner: And what's the reason here?

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,310 @@
...
-        return after.trim().to_string();
+/// This traverses the error chain to extract specific failure details and provides
+/// user-friendly explanations with actionable suggestions when possible.
+///
+/// The advantage of this approach is that we can be way more specific about the
+/// errors which can occur, rather than just returning a generic error message.
+/// The downside is that we have to maintain this code as reqwest and hyper
+/// evolve. However, this is a trade-off we are willing to make for better user
+/// experience.
+#[allow(clippy::too_many_lines)]
```
thomas-zahner: It would be lovely to get rid of this line by extracting a few new functions from this function.

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,310 @@
...
+    if let Some(status) = error.status() {
+        let reason = status.canonical_reason().unwrap_or("Unknown");
+        return format!(
+            "HTTP {}: {} - check URL and server status",
+            status.as_u16(),
+            reason
+        );
+    }
+
+    // Traverse error chain for detailed analysis
```
thomas-zahner: A new method `traverse_error_chain` here

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,310 @@
...
+                        }
+                        _ => format!(
+                            "I/O error ({kind_name}). Check network connectivity and server status",
+                        ),
+                    }
+                }
+            };
+        }
+
+        // Check for hyper-specific errors
```
thomas-zahner, line 124: a new method `handle_hyper_error`

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,310 @@
...
+            }
+            if hyper_msg.contains("invalid content-length") {
+                return "Invalid Content-Length header from server".to_string();
+            }
+
+            return format!("HTTP protocol error: {hyper_error}");
+        }
+
+        // Check for URL parsing errors
+        if let Some(url_error) = err.downcast_ref::<url::ParseError>() {
```
thomas-zahner, line 130: a new method `handle_url_parse_error`

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,310 @@
...
+                }
+                url::ParseError::InvalidPort => "Invalid URL: invalid port number".to_string(),
+                url::ParseError::RelativeUrlWithoutBase => {
+                    "Invalid URL: relative URL without base".to_string()
+                }
+                _ => format!("Invalid URL format: {url_error}"),
+            };
+        }
+
+        // Check for generic error types by examining their string representation
```
thomas-zahner, line 134: a new method `handle_generic_error_by_message`

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,310 @@
...
+        // Protocol-specific errors
+        if error_msg.contains("protocol") && error_msg.contains("not supported") {
+            return "Protocol not supported. Check URL scheme (http/https)".to_string();
+        }
+
+        source = err.source();
+    }
+
+    // Fallback to basic reqwest error categorization
+    if error.is_connect() {
```
thomas-zahner, line 388: a new method `handle_fallback_reqwest_error`

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,310 @@
...
+///
+/// The advantage of this approach is that we can be way more specific about the
+/// errors which can occur, rather than just returning a generic error message.
+/// The downside is that we have to maintain this code as reqwest and hyper
+/// evolve. However, this is a trade-off we are willing to make for better user
+/// experience.
+#[allow(clippy::too_many_lines)]
+pub(crate) fn analyze_error_chain(error: &reqwest::Error) -> String {
+    // First check reqwest's built-in categorization
+    if error.is_timeout() {
```
thomas-zahner, line 93: One method for these `error.is_timeout`, `error.is_redirect` and `error.status` called something like `handle_common_reqwest_error`

mre:
@thomas-zahner Thanks for the excellent feedback! I've addressed all your comments:

  Changes Made:

  1. Added helpful error messages for the three None cases:
    - InvalidFilePath: Now returns "File not found at path: '{uri}'. Check if file exists
   and path is correct"
    - InvalidFragment: Now returns "Fragment not found in document: '{uri}'. Check if 
  fragment exists or page structure"
    - InvalidIndexFile: Now returns "Index file not found in directory: '{}'. Check if 
  index.html or other index files exist"

  2. Refactored `analyze_error_chain` to eliminate `clippy::too_many_lines`:
    - First, I broke the 287-line function into focused helper functions
    - Further evolved it into a clean builder pattern using ErrorRules
    - All manual loops eliminated in favor of declarative rule definitions

  Reasoning:

  The original approach with individual if contains() chains was repetitive and hard to maintain. I've implemented an ErrorRules builder that encapsulates the pattern matching logic:

```rust
  ErrorRules::new()
      .rule(&["expired", "NotValidAtThisTime"], "SSL certificate expired. Site needs to 
  renew certificate")
      .rule(&["hostname", "NotValidForName"], "SSL certificate hostname mismatch. Check 
  URL spelling")
      .fallback("SSL certificate error. Check certificate validity")
      .match_error(error_msg)
```

This approach provides better maintainability: Rules are declarative, and adding new patterns ist just adding new `.rule()` calls. Also, each rule set can have custom fallback behavior.

I hope the code now more readable. What do you think?

lychee-lib/src/types/error.rs
mre: Changed and added helpful error message.

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,421 @@
...
+    }
+
+    /// Set a fallback message if no rules match
+    fn fallback(mut self, message: impl Into<String>) -> Self {
+        self.fallback = Some(message.into());
+        self
+    }
+
+    /// Match against the error message and return the appropriate response
+    fn match_error(&self, error_msg: &str) -> String {
```
thomas-zahner: One tiny improvement we could do.
 
```suggestion
    fn match_error(self, error_msg: &str) -> String {
```

lychee-lib/src/utils/reqwest.rs
```diff
@@ -1,39 +1,421 @@
...
+        for rule in &self.rules {
+            if rule.matches(error_msg) {
+                return rule.message().to_string();
+            }
+        }
+
+        self.fallback
+            .clone()
+            .unwrap_or_else(|| format!("Unhandled error: {error_msg}"))
```
thomas-zahner: Then cloning should no longer be necessary, right?

```suggestion
        self.fallback
            .unwrap_or_else(|| format!("Unhandled error: {error_msg}"))
```

mre:
Thanks for the thorough review, @thomas-zahner

PR #1735, opened 2025-06-16
Add xml schema found in xsd files to list of exclusions

tooomm:
- Adding xml schema found in XSD files, see e.g. https://www.w3schools.com/xml/schema_intro.asp
- Escape `.`'s in URLs
- Reorder one entry for better readability

Linking https://github.com/lycheeverse/lychee-action/issues/297

<br>

~~Should the dots in the regex be escaped to match explicitly? @mre~~ Done, https://github.com/lycheeverse/lychee/pull/1735/commits/93090c30b9cfa76257702fd182f8e8943eac26ac

### DIFF
--- a/lychee-lib/src/filter/mod.rs
+++ b/lychee-lib/src/filter/mod.rs
@@ -41,15 +41,15 @@ static UNSUPPORTED_DOMAINS: LazyLock<HashSet<&'static str>> = LazyLock::new(|| {
 
 /// Pre-defined exclusions for known false-positives
 const FALSE_POSITIVE_PAT: &[&str] = &[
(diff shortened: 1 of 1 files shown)


thomas-zahner:
Thanks for the PR. You are right, we should escape all `.` in URLs. A false positive is quite unlikely but technically it's still incorrect to not escape them. Can you escape all URLs in the `FALSE_POSITIVE_PAT` slice?

In our documentation it is also explained like that: https://lychee.cli.rs/recipes/excluding-links/#advanced-examples

mre:
Looks good, thanks @tooomm and thanks for the review @thomas-zahner.

PR #1738, opened 2025-06-19
Fix docker tagging for release versions

### LINKED ISSUE #1628: Publish Docker tags with semver
For example, alpine tags are published in the format: `sha-7c4b132-alpine`.

This makes it impossible to make it work with [renovate](https://docs.renovatebot.com/modules/versioning/docker/).

> Are commit hashes supported?
> No, Renovate ignores Docker image tags that look like a Git commit hash.

Why is it not published in semver format (like on github releases)? For example: `0.18.0-alpine`.

Thanks.

trask:
I hope will resolve #1628

This is the relevant part of the workflow to understand what's going on:

https://github.com/lycheeverse/lychee/blob/3592972d6462ceb6081c3945cc43121bc3ce2039/.github/workflows/docker.yml#L46-L48

This is configuring `docker/metadata-action`.

And [`type=semver`](https://github.com/docker/metadata-action#typesemver)

> Will be used on a [push tag event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) and requires a valid [semver](https://semver.org/) Git tag

Even if docker/metadata-action is being used on branch push, I think that [`${{ github.ref_name }}`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) is the branch name when the workflow is triggered by a branch push (and the tag name when the workflow is triggered by a tag push).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,16 +1,11 @@
 name: Docker
 on:
   workflow_dispatch:
-  workflow_run:
-    workflows: 
-      - "Release Binary"
-    types:
-      - completed
-    branches:
-      - master
   push:
     branches:
       - master
+    tags:
+      - 'lychee-v*'
   pull_request:
     branches:
       - master


mre:
Thanks, this should ensure that the Docker build runs after `release-plz` has finished (which is necessary because it depends on the release binaries).

Looks good to me.

eread:
> Thanks, this should ensure that the Docker build runs after `release-plz` has finished (which is necessary because it depends on the release binaries).
> 
> Looks good to me.

@trask @mre @thomas-zahner If I'm not mistaken, this is the Docker workflow run that happened when the `lychee-v0.20.1` Git tag was pushed: https://github.com/lycheeverse/lychee/actions/runs/17205362504.

On the plus side, the Docker command was run with `push`, so it would've created new Docker images if it had succeeded.

On the down side:

- The command didn't have a Docker image tag for `0.20.1`, only for the SHA: https://github.com/lycheeverse/lychee/actions/runs/17205362504/job/48804656733#step:9:218
- It looks like `wget` couldn't find the correct binary yet: https://github.com/lycheeverse/lychee/actions/runs/17205362504/job/48804656733#step:9:427. I guess that means having a tag doesn't mean having a binary yet?

This PR removed `workflow_run`, which means it no longer waits for a successful run of "Release Binary" workflow? Should it? Would that work given Release Binary seems to be failing a lot: https://github.com/lycheeverse/lychee/actions/workflows/release-binary.yml.

mre:
Yeah, you're right, just because we have a tag doesn't mean with have the release binaries. That has been haunting us for quite a while. Thats' also why `workflow_run` is not a solution. Perhaps we can do a super simple thing and wait until the binaries get built with a bash script?

```yaml
- name: Login to DockerHub
        uses: docker/login-action@v3
        with:
          username: ${{ secrets.DOCKER_USERNAME }}
          password: ${{ secrets.DOCKER_PASSWORD }}

      - name: Wait for binary release
        if: ${{ (github.event_name != 'pull_request' &&
                 github.actor != 'dependabot[bot]') ||
                 startsWith(github.ref, 'lychee-v') }}
        run: |
          LYCHEE_VERSION="${{ github.ref == 'refs/heads/master' && 'nightly' || 'latest' }}"
          
          BASE_URL=$(case "$LYCHEE_VERSION" in
            "latest" | "nightly") echo "https://github.com/lycheeverse/lychee/releases/latest/download";;
            *) echo "https://github.com/lycheeverse/lychee/releases/download/$LYCHEE_VERSION";;
          esac)
          
          X86_URL="${BASE_URL}/lychee-x86_64-unknown-linux-gnu.tar.gz"
          ARM_URL="${BASE_URL}/lychee-aarch64-unknown-linux-gnu.tar.gz"
          
          echo "Waiting for both binaries to be available..."
          
          while true; do
            if curl -sSf --head "$X86_URL" >/dev/null 2>&1 && curl -sSf --head "$ARM_URL" >/dev/null 2>&1; then
              echo "Both binaries ready!"
              break
            fi
            sleep 20
          done

      - name: Push Image (debian)
        uses: docker/build-push-action@v6
        # ... rest of your existing config
```

I haven't tested this, but that's my best guess. Either way, I think we should move this discussion into a follow-up issue. @eread, are you up for that?

eread:
> I haven't tested this, but that's my best guess. Either way, I think we should move this discussion into a follow-up issue. @eread, are you up for that?

Not sure about that solution. I guess if it had a maximum number of iterations (rather than an infinite number), it might work ok. It's a shame that `wget` and `curl` both seem to regard a `404` as a fatal error that can't be retried.

On the first issue (no release tag), I've raised: https://github.com/lycheeverse/lychee/issues/1833. Given the actions seem to be able to be run manually (so someone can manually intervene when they fail to find a release binary and retry later), I thought the issue of no release tag was more important.

Issue #1495, opened 2024-09-05
No results for staging site. X-Robots-Tag: none a problem?


jannisborgers:
I found [this thread](#68) after I tried lychee on multiple of my sites when deploying them to a staging environment and always had problems of no results. 

Basic auth is a must in these situations, so I was happy that lychee supports it. 

As an additional complication, the CMS we’re using sends an `X-Robots-Tag: none` response header in staging environments, as it is deliberately not supposed to be indexed. Is that something that lychee supports, or does it ignore that header? From the messages in the above thread, I could not find out if robots.txt is ignored at the moment, or not.

Right now, I get the following response:

```
🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors
```

The format I’m using is: 
```
lychee --basic-auth 'user:password https://subdomain.domain.tld' https://subdomain.domain.tld
```

There are about 500 links on that page, I verified with curl that the basic auth is working correctly. It returns the HTML response.

mre:
lychee ignores all response headers and robots.txt because we're not indexing the page. The problem must be elsewhere. 

Can you save the html into a file and use that as the input?

```
lychee -vvv foo.html
```

If that works, it's because the website doesn't serve the HTML to lychee. In that case you can try curl as a user agent. 
https://lychee.cli.rs/troubleshooting/network-errors/#try-a-different-user-agent

If that doesn't work, lychee might have issues passing the URLs from the document. In that case, could you post an expert of the HTML file?

jannisborgers:
Hi @mre — thanks for the quick response!

### Locally
I tried it on the index.html file locally, with the exact command you provided, and it yields:

```
🔍 529 Total (in 1s) ✅ 6 OK 🚫 518 Errors 💤 4 Excluded
```

The errors come from the static html usign the password-protected URLs, so this is the result I was expecting from the local version.

### cURL user-agent
I tried using lychee with the cURL user-agent like described, but it still yields:

```
🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors
```

But using curl like this works and returns the HTML correctly:
```
curl user:password https://subdomain.domain.tld
```

So the way I see it, cURL itself is working, but lychee isn’t, even when using cURL as a user agent.

### Basic auth problem?
I suspect that the basic auth of lychee is the source of the problem. I tested on other staging sites that had basic auth and they all came back empty-handed. The output is the same as when I omit `--basic-auth` on sites with basic auth:

```
[...]

PR #1748, opened 2025-06-27
Fix basic auth

thomas-zahner:
I've captured the bug by updating the test in 997f1f7cd1c454fbfbc3f8e2f44ff50ad98b3f9a

Closes https://github.com/lycheeverse/lychee/issues/1495

### DIFF
--- a/lychee-lib/src/collector.rs
+++ b/lychee-lib/src/collector.rs
@@ -1,5 +1,6 @@
 use crate::ErrorKind;
 use crate::InputSource;
+use crate::types::url_extractor::UrlExtractor;
 use crate::{
     Base, Input, Request, Result, basic_auth::BasicAuthExtractor, extract::Extractor,
     types::FileExtensions, types::uri::raw::RawUri, utils::request,
@@ -9,7 +10,9 @@ use futures::{
     StreamExt,
     stream::{self, Stream},
 };
+use http::HeaderMap;
 use par_stream::ParStreamExt;
+use reqwest::Client;
 use std::path::PathBuf;
 
 /// Collector keeps the state of link collection
@@ -25,6 +28,7 @@ pub struct Collector {
     use_html5ever: bool,
     root_dir: Option<PathBuf>,
     base: Option<Base>,
+    headers: HeaderMap,
 }
 
 impl Default for Collector {
@@ -36,6 +40,7 @@ impl Default for Collector {
             use_html5ever: false,
             skip_hidden: true,
             skip_ignored: true,
+            headers: HeaderMap::new(),
             root_dir: None,
             base: None,
         }
@@ -61,6 +66,7 @@ impl Collector {
             use_html5ever: false,
             skip_hidden: true,
             skip_ignored: true,
+            headers: HeaderMap::new(),
             root_dir,
             base,
         })
@@ -84,6 +90,13 @@ impl Collector {
     #[must_use]
     pub const fn skip_ignored(mut self, yes: bool) -> Self {
         self.skip_ignored = yes;
+        self
+    }
+
+    /// Skip files that are ignored
+    #[must_use]
+    pub fn headers(mut self, headers: HeaderMap) -> Self {
+        self.headers = headers;
         self
     }
 
@@ -148,10 +161,22 @@ impl Collector {
                 let default_base = global_base.clone();
                 let extensions = extensions.clone();
                 let basic_auth_extractor = self.basic_auth_extractor.clone();
+                let headers = self.headers.clone();
                 async move {
                     let base = match &input.source {
                         InputSource::RemoteUrl(url) => Base::try_from(url.as_str()).ok(),
                         _ => default_base,
+                    };
+
+                    let client = Client::builder()
+                        .build()
+                        .map_err(ErrorKind::BuildRequestClient)
+                        .unwrap(); // TODO
+
+                    let extractor = UrlExtractor {
+                        basic_auth_extractor,
+                        headers,
+                        client,
                     };
 
                     input
@@ -160,7 +185,7 @@ impl Collector {
                             skip_hidden,
                             skip_ignored,
                             extensions,
-                            basic_auth_extractor,
+                            extractor,
                         )
                         .map(move |content| (content, base.clone()))
                 }
@@ -191,7 +216,7 @@ impl Collector {
 mod tests {
     use std::{collections::HashSet, convert::TryFrom, fs::File, io::Write};
 
-    use http::{HeaderMap, StatusCode};
+    use http::StatusCode;
     use reqwest::Url;
 
     use super::*;
@@ -241,7 +266,13 @@ mod tests {
         let _file = File::create(&file_path).unwrap();
         let input = Input::new(&file_path.as_path().display().to_string(), None, true, None)?;
         let contents: Vec<_> = input
-            .get_contents(true, true, true, FileType::default_extensions(), None)
+            .get_contents(
+                true,
+                true,
+                true,
+                FileType::default_extensions(),
+                UrlExtractor::default(),
+            )
             .collect::<Vec<_>>()
             .await;
 
@@ -254,7 +285,13 @@ mod tests {
     async fn test_url_without_extension_is_html() -> Result<()> {
         let input = Input::new("https://example.com/", None, true, None)?;
         let contents: Vec<_> = input
-            .get_contents(true, true, true, FileType::default_extensions(), None)
+            .get_contents(
+                true,
+                true,
+                true,
+                FileType::default_extensions(),
+                UrlExtractor::default(),
+            )
             .collect::<Vec<_>>()
             .await;
 
--- a/lychee-lib/src/collector.rs
+++ b/lychee-lib/src/collector.rs
@@ -105,7 +105,7 @@ impl Collector {
         self
     }
 
-    /// Set client to use when resolving input URLs
+    /// Set client to use for checking input URLs
     #[must_use]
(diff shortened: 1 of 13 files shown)


lychee-lib/src/collector.rs
```diff
@@ -38,6 +43,8 @@ impl Default for Collector {
             skip_ignored: true,
             root_dir: None,
             base: None,
+            headers: HeaderMap::new(),
+            client: Client::new(),
```
thomas-zahner, line 51: Do you think it's okay to use `Client::new()` here? Error handling might be a bit annoying in this case.

> This method panics if a TLS backend cannot be initialized, or the resolver cannot load the system configuration.
Use Client::builder() if you wish to handle the failure as an Error instead of panicking.

thomas-zahner:
Pipeline fails because of Rust 1.88 was released yesterday with new clippy suggestions. See: https://github.com/lycheeverse/lychee/pull/1749

mre:
https://github.com/lycheeverse/lychee/pull/1749 is merged so you can rebase 👍

lychee-lib/src/collector.rs
mre: Right, that's a bit tricky. The collector is public, so if someone were to build a library around lychee, they could not gracefully handle that case on their end because the code would panic. 😕
But since `default()` returns `Self` and not `Result<Self>` we also can't return the error of `Client::build` either.

Then again, all of this is unlikely, because library authors would probably not use `Collector::default`, but `Collector::new`.
I think it's an edge-case, though, so it's probably fine. Would add a doc-comment to warn people about that behavior and point them to `Collector::new` is they wanted to handle that error case.

lychee-lib/src/collector.rs
```diff
@@ -87,6 +98,20 @@ impl Collector {
...
     }
 
+    /// Set headers to use when resolving input URLs
+    #[must_use]
+    pub fn headers(mut self, headers: HeaderMap) -> Self {
+        self.headers = headers;
+        self
+    }
+
+    /// Set client to use when resolving input URLs
```
mre: Isn't it about checking the URLs?

```suggestion
    /// Set client to use for checking input URLs
```

lychee-lib/src/types/basic_auth/credentials.rs
```diff
@@ -77,8 +76,8 @@ impl BasicAuthCredentials {
 }
 
 #[async_trait]
-impl Handler<Request, Status> for Option<BasicAuthCredentials> {
-    async fn handle(&mut self, mut request: Request) -> ChainResult<Request, Status> {
+impl<Response> Handler<Request, Response> for Option<BasicAuthCredentials> {
+    async fn handle(&mut self, mut request: Request) -> ChainResult<Request, Response> {
```
mre, line 87: Switching from `Status` to `Response` here reads so much more naturally.

lychee-lib/src/types/resolver.rs
```diff
@@ -0,0 +1,77 @@
...
+        } else {
+            FileType::from(url.as_str())
+        };
+
+        let credentials = request::extract_credentials(
+            self.basic_auth_extractor.as_ref(),
+            &Uri { url: url.clone() },
+        );
+
+        let chain: RequestChain = Chain::new(vec![Box::new(credentials), Box::new(self.clone())]);
```
mre: Thanks for moving more parts to `RequestChain`.

lychee-lib/src/types/resolver.rs
```diff
@@ -0,0 +1,77 @@
...
+    pub basic_auth_extractor: Option<BasicAuthExtractor>,
+    pub headers: HeaderMap,
+    pub client: reqwest::Client,
+}
+
+type RequestChain = Chain<Request, Result<String>>;
+
+impl UrlContentResolver {
+    /// Fetch remote content by URL.
+    /// This method is not intended to check if a URL is functional but
```
mre: Nitpick

```suggestion
    /// Fetch remote content by URL.
    ///
    /// This method is not intended to check if a URL is functional but
```

lychee-lib/src/types/resolver.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// to get a URL's content and process the content.
+    pub async fn url_contents(&self, url: Url) -> Result<InputContent> {
+        // Assume HTML for default paths
+        let file_type = if url.path().is_empty() || url.path() == "/" {
+            FileType::Html
+        } else {
+            FileType::from(url.as_str())
+        };
```
mre: Old code, but nowadays, I'd write it like this:

```rust
let file_type = match url.path() {
  _ if path.is_empty() || path == "/" => FileType::Html,
  _ => FileType::from(url.as_str())
};
```

I'm not sure if it's idiomatic, but to me it reduces clutter somehow. 😆 (Not saying we have to change this, though.)

lychee-lib/src/types/resolver.rs
```diff
@@ -0,0 +1,77 @@
...
+        let request = self
+            .client
+            .request(reqwest::Method::GET, url.clone())
+            .build()
+            .map_err(ErrorKind::BuildRequestClient)?;
+
+        let content = match chain.traverse(request).await {
+            ChainResult::Next(_) => unreachable!(
+                "ChainResult::Done is unconditionally returned from the last chain element"
+            ),
```
mre: That feels like a bit of a wart in our chain implementation.
I wonder how we could avoid that. 🤔 
Should `traverse` return the final `Result` instead? It's probably not possible, but I haven't thought much about it.

lychee-lib/src/types/resolver.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        Ok(input_content)
+    }
+}
+
+#[async_trait]
+impl Handler<Request, Result<String>> for UrlContentResolver {
+    async fn handle(&mut self, mut request: Request) -> ChainResult<Request, Result<String>> {
+        request.headers_mut().extend(self.headers.clone());
+        ChainResult::Done(execute_request(&self.client, request).await)
```
mre: Wouldn't it be possible to return the modified request instead and handle the execution in a separate handler? Might be cleaner.

lychee-lib/src/types/resolver.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+async fn execute_request(client: &Client, request: Request) -> Result<String> {
```
mre: Maybe a bit more explicit

```suggestion
async fn get_request_body_text(client: &Client, request: Request) -> Result<String> {
```

lychee-lib/src/collector.rs
thomas-zahner: Yes that might be better terminology

lychee-lib/src/types/basic_auth/credentials.rs
thomas-zahner: Yeah agree previously it was too specific/strict. Now for example I can use use `Result<String>` as in `UrlContentResolver`

lychee-lib/src/types/resolver.rs
thomas-zahner: Agree that it's a tiny bit better, as we reduce the duplication of `url.path()`

lychee-lib/src/collector.rs
thomas-zahner: Agree, it would be nice to do `impl Default for Result<Collector>`, but the orphan rule prevents that. :disappointed: I've now added a doc comment.

lychee-lib/src/types/resolver.rs
thomas-zahner: Yes, I'm also not really proud of the `unreachable!` here. We cover it with the updated test though so I could still sleep at night if we merge it like this. But it's basically like Rust's `Option` type with `unwrap` where we are sure that there is `Some` value. I can try to introduce some additional types so that we could prove this with the type system, but not sure if this is worth the effort.

lychee-lib/src/types/resolver.rs
thomas-zahner: Yes, can do that

lychee-lib/src/collector.rs
mre: Orphan rules and the fact that the `Default` is designed for infallible instantiation. It makes sense because getting a default value should be an unambiguous operation. In our case, it's a bit of an edge-case and I think the improved ergonomics outweigh the downsides.

lychee-lib/src/types/resolver.rs
mre: Makes me wonder if we should even use a request chain here. After all, we just call `extract_credentials` and then `client.request`, right? The chain would be helpful for inversion of control like in situations where we want to return the chain and someone else "merges" it with other handlers and then calls the entire thing.

However, in this function here, the execution is very imperative as in "do this, then do that." 🤔 So maybe we just get rid of the request chain here? Or was there a specific reason for using it? The alternative would be to make the entire resolver a handler in the request chain. Then it would make sense.

Feedback welcome!

thomas-zahner: Actually I totally agree. I just had this vague idea in mind of using the request chain and it turned out to be possible to use it. However, it also felt quite a bit like overengineering. So I'll simplify that.

lychee-lib/src/types/resolver.rs
thomas-zahner: Resolved, as I removed usage of the `Chain` completely.

thomas-zahner:
@mre Thanks, the code is now quite a bit simpler. What do you think?

mre:
Very nice!

PR #1749, opened 2025-06-27
Clippy 1.88

thomas-zahner:
Pipeline fails because of Rust 1.88 was released yesterday with new clippy suggestions.

The most interesting part of this PR is that I removed `#[ignore]` from two tests as I noticed that they are actually passing. I assume it's because the issues were addressed in the meantime. Or is there a reason we should still ignore them?

### DIFF
--- a/lychee-lib/src/extract/html/html5gum.rs
+++ b/lychee-lib/src/extract/html/html5gum.rs
@@ -464,12 +464,7 @@ mod tests {
         assert_eq!(uris, expected);
     }
 
-    // TODO: This test is currently failing because we don't handle nested
-    // verbatim elements of the same type correctly. The first closing tag will
-    // lift the verbatim flag. This is a known issue and could be handled by
-    // keeping a stack of verbatim flags.
     #[test]
-    #[ignore]
     fn test_include_verbatim_nested_identical() {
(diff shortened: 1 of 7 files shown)


lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -464,12 +464,7 @@ mod tests {
         assert_eq!(uris, expected);
     }
 
-    // TODO: This test is currently failing because we don't handle nested
-    // verbatim elements of the same type correctly. The first closing tag will
-    // lift the verbatim flag. This is a known issue and could be handled by
-    // keeping a stack of verbatim flags.
     #[test]
-    #[ignore]
```
mre, line 472: Ah nice, so this no longer failing? Thanks for removing the `#[ignore]`.

lychee-lib/src/extract/markdown.rs
```diff
@@ -314,7 +314,6 @@ or inline like `https://bar.org` for instance.
     }
 
     #[test]
-    #[ignore]
```
mre, line 317: thx

mre:
Very good. Thanks for updating!

thomas-zahner:
Sure no problem :+1:

Issue #1751, opened 2025-06-29
Incorrect fragment check for `index.html` in a sub directory


ocavue:
Steps to reproduce:

```bash
$ git clone https://github.com/issueset/lychee-fragments-issue

$ cd lychee-fragments-issue

$ lychee --version
lychee 0.19.1

$ lychee --include-fragments --fallback-extensions html --no-progress --offline -v './**/*.html'
     [200] file:///private/tmp/lychee-fragments-issue/foo#invalid-id # ⬅️ This is unexpected
     [200] file:///private/tmp/lychee-fragments-issue/foo/index.html#valid-id
     [200] file:///private/tmp/lychee-fragments-issue/foo#valid-id
     [200] file:///private/tmp/lychee-fragments-issue/bar#valid-id
   [ERROR] file:///private/tmp/lychee-fragments-issue/bar#invalid-id | Cannot find fragment
     [200] file:///private/tmp/lychee-fragments-issue/bar.html#valid-id
   [ERROR] file:///private/tmp/lychee-fragments-issue/bar.html#invalid-id | Cannot find fragment
   [ERROR] file:///private/tmp/lychee-fragments-issue/foo/index.html#invalid-id | Cannot find fragment

Issues found in 1 input. Find details below.

[index.html]:
   [ERROR] file:///private/tmp/lychee-fragments-issue/bar#invalid-id | Cannot find fragment
   [ERROR] file:///private/tmp/lychee-fragments-issue/bar.html#invalid-id | Cannot find fragment
   [ERROR] file:///private/tmp/lychee-fragments-issue/foo/index.html#invalid-id | Cannot find fragment

🔍 8 Total (in 0s) ✅ 5 OK 🚫 3 Errors
```

This reproduction repo contains a file `foo/index.html`, and there isn't a fragment `invalid-id` in this file. However, `lychee` still pass the check for `<a href="foo#invalid-id">2</a>`

## Expected behavior 

All four links with incorrect `#invalid-id` are found

## Actual behavior  

Only three of them are found

ocavue:
The following line seems off:

https://github.com/lycheeverse/lychee/pull/1713/files#diff-7b6a0f8237f5bd7a9eb7637faf689671a8fe2505dcb764889d4155ffcdefbd89R69

If I have the following document content:

```
Skip the fragment check for directories like: [empty](empty_dir/#this_fragment_does_not_exist).
```

I would expect this to generate an error instead of passing silently.

mre:
[...]

PR #1752, opened 2025-06-29
fix: resolve index file inside a directory

ocavue:
Closes https://github.com/lycheeverse/lychee/issues/1751 

This PR addresses issue #1751 by adding an "index" file check logic into Lychee. When the URL points to a local directory, it attempts to locate the index file within that directory and verify its fragments.

### DIFF
--- a/lychee-lib/src/checker/file.rs
+++ b/lychee-lib/src/checker/file.rs
@@ -105,63 +105,111 @@ impl FileChecker {
     ///
     /// Returns a `Status` indicating the result of the check.
     async fn check_path(&self, path: &Path, uri: &Uri) -> Status {
-        if path.exists() {
-            return self.check_existing_path(path, uri).await;
+        let file_path = self.resolve_file_path(path);
+        let has_fragment = uri.url.fragment().is_some_and(|x| !x.is_empty());
+
+        // If file_path exists, check this file
+        if file_path.is_some() {
+            return self.check_file(&file_path.unwrap(), uri).await;
         }
-
-        self.check_with_fallback_extensions(path, uri).await
-    }
-
-    /// Checks an existing path, optionally verifying fragments for HTML files.
-    ///
-    /// # Arguments
-    ///
-    /// * `path` - The path to check.
-    /// * `uri` - The original URI, used for error reporting.
-    ///
-    /// # Returns
-    ///
-    /// Returns a `Status` indicating the result of the check.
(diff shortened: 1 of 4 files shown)


lychee-lib/src/checker/file.rs
```diff
@@ -105,63 +105,111 @@ impl FileChecker {
...
+
+        // If file_path exists, check this file
+        if file_path.is_some() {
+            return self.check_file(&file_path.unwrap(), uri).await;
+        }
+        // If path is a directory, and we cannot find an index file inside it,
+        // and we don't have a fragment, just return success. This is for
+        // backward compatibility.
+        else if path.is_dir() && !has_fragment {
+            return Status::Ok(StatusCode::OK);
```
ocavue, line 119: I added this logic to ensure that the following test can pass:

https://github.com/lycheeverse/lychee/blob/8f2f746bf932dd0d5b0de3a4364b0e01e20617c8/fixtures/fragments/file1.md?plain=1#L68-L70

While I personally prefer this test to fail, I may not be aware of all the context surrounding it.

kemingy: I just found that I forgot to add the fragment to this test case. So it doesn't test the new code in #1713.

Instead, it has been treated as a success by lychee for a long time.

I'm not sure whether it's a common case for some scenarios or not.

ocavue: @kemingy Hey 👋


If I take out the quoted code in this comment, in other words, treat `[empty](empty_dir/#no_exist_fragment)` as an error, will this cause the "Too many open files" error to reappear? I'm not entirely clear on how checking link fragments relates to the "Too many open files" error.

kemingy: No. I think it's safe to treat the empty_dir#fragment as an error.

ocavue: Nice! I'll keep this PR as it is to limit its scope. I'll open another PR to make `empty_dir#fragment` an error.

mre:
Fantastic. That should work!

mre:
> I'll open another PR to make empty_dir#fragment an error.

Yes, much appreciated.

ocavue:
> > I'll open another PR to make empty_dir#fragment an error.
> 
> Yes, much appreciated.

https://github.com/lycheeverse/lychee/pull/1756

Issue #1660, opened 2025-03-02
Reference style links are checked or reported multiple times


theofidry:
Taken from this job: https://github.com/box-project/box/actions/runs/13615816913/job/38058706641

And an excerpt in case the job output is removed later:

```
✔ [200] https://secure.php.net/manual/en/phar.setsignaturealgorithm.php
✔ [200] https://secure.php.net/manual/en/book.zlib.php
✔ [200] https://symfony.com/slack-invite
✔ [200] https://symfony.com/slack-invite
## Summary

| Status        | Count |
|---------------|-------|
| 🔍 Total      | 644   |
| ✅ Successful | 642   |
| ⏳ Timeouts   | 0     |
| 🔀 Redirected | 0     |
| 👻 Excluded   | 0     |
| ❓ Unknown    | 0     |
| 🚫 Errors     | 2     |

## Errors per input

### Errors in doc/docker.md

* [404] [https://docs.docker.com/samples/library/php/](https://docs.docker.com/samples/library/php/) | Failed: Network error: Not Found

### Errors in ./doc/docker.md

* [404] [https://docs.docker.com/samples/library/php/](https://docs.docker.com/samples/library/php/) | Failed: Network error: Not Found
*
```

You can see the 404 appears twice. The underlying code was something like this:

```md
blablabla [`docker-php-ext-install`][docker-php-ext-install] blablabla [`docker-php-ext-install`][docker-php-ext-install]

[docker-php-ext-install]: https://deadlink.com
```

In other words I am using a reference-style link in my document and it looks like it is either checked or reported for every usage, rather than once.

thomas-zahner:
The implementation should probably make use of some sort of HashSet that keeps track of all paths that previously were [canonicalized](https://doc.rust-lang.org/std/fs/fn.canonicalize.html). This should make it possible to remove duplicates from the file checking process, so that a file that was provided via `README.md` and via `./docs/../README.md` is checked only once.

tooomm:
Is this fully addressed with [#1781](https://github.com/lycheeverse/lychee/pull/1781)?

mre:
Yeah, should be. Thanks for the heads-up. Closing.

thomas-zahner:
@tooomm Thanks for the contribution. Yes, I think the problem reported in this issue is solved. I also mentioned [canonicalizing](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) file inputs.

[...]

PR #1761, opened 2025-07-04
Fix duplicate file paths in error reporting

tosky1125:
## Description

This PR fixes the issue where the same file is reported multiple times in error output when accessed through different path representations (e.g., `doc/docker.md` vs `./doc/docker.md`).

Fixes #[issue-number]

## Problem

When `lychee` discovers files through different methods (CLI arguments vs document links), the same file can be represented with different path formats:
- `doc/docker.md` (from glob pattern)
- `./doc/docker.md` (from document link)

This causes duplicate entries in the error report, making it appear as if the same error occurred in multiple files.

## Solution

Implemented path normalization in the `Display` trait for `InputSource::FsPath` to ensure consistent path representation across all platforms:

- Removes redundant `./` prefixes (including multiple occurrences like `./././`)
- Filters out intermediate `.` components (e.g., `doc/./file.md` → `doc/file.md`)
- Converts Windows backslashes to forward slashes for cross-platform consistency
- Preserves important path characteristics (absolute paths, parent directory references, trailing slashes)
- Handles edge cases like empty paths and root directories

## Changes

- Modified `Display` implementation for `InputSource` in `lychee-lib/src/types/input.rs`
- Added comprehensive test coverage for various path formats and edge cases

## Testing

Added extensive tests covering:
- [x] Basic path normalization (`./doc/file.md` → `doc/file.md`)
- [x] Multiple current directory markers (`./././doc/file.md` → `doc/file.md`)
- [x] Windows path normalization (`doc\file.md` → `doc/file.md`)
- [x] Cross-platform consistency
- [x] Edge cases (empty paths, trailing slashes, absolute paths)
- [x] All existing tests pass

## Notes

- The normalization only affects display output, not the actual file processing logic
- Maintains backward compatibility with existing behavior
- No performance impact as normalization only occurs during error reporting

### DIFF
--- a/lychee-lib/src/types/input.rs
+++ b/lychee-lib/src/types/input.rs
@@ -90,13 +90,64 @@ impl Serialize for InputSource {
 
 impl Display for InputSource {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str(match self {
-            Self::RemoteUrl(url) => url.as_str(),
-            Self::FsGlob { pattern, .. } => pattern,
-            Self::FsPath(path) => path.to_str().unwrap_or_default(),
-            Self::Stdin => "stdin",
-            Self::String(s) => s,
-        })
+        match self {
+            Self::RemoteUrl(url) => f.write_str(url.as_str()),
+            Self::FsGlob { pattern, .. } => f.write_str(pattern),
+            Self::FsPath(path) => {
+                use std::path::Component;
+
+                let mut result = String::new();
+                let mut components = path.components().peekable();
+
+                while let Some(component) = components.next() {
+                    match component {
+                        Component::Prefix(prefix) => {
+                            // Windows drive letters (C:)
+                            result.push_str(&prefix.as_os_str().to_string_lossy());
+                        }
+                        Component::RootDir => {
+                            // Root directory - just add single slash
+                            result.push('/');
+                        }
+                        Component::CurDir => {
+                            // Skip "." components
+                            continue;
+                        }
+                        Component::ParentDir => {
+                            // Keep ".." components
+                            if !result.is_empty() && !result.ends_with('/') {
+                                result.push('/');
+                            }
+                            result.push_str("..");
+                        }
+                        Component::Normal(name) => {
+                            // Add separator if needed (but not after root)
+                            if !result.is_empty() && !result.ends_with('/') {
+                                result.push('/');
+                            }
+                            result.push_str(&name.to_string_lossy());
+                        }
+                    }
+                }
+
+                // Handle empty result (e.g., from just ".")
+                if result.is_empty() {
+                    result.push('.');
+                }
+
+                // Preserve trailing slash if original path had it
+                let original_str = path.to_string_lossy();
+                if (original_str.ends_with('/') || original_str.ends_with('\\'))
+                    && !result.ends_with('/')
+                    && result != "." {
+                    result.push('/');
+                }
+
+                f.write_str(&result)
+            },
+            Self::Stdin => f.write_str("stdin"),
+            Self::String(s) => f.write_str(s),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


tosky1125:
The test failures appear to be unrelated to this PR's changes. The failing tests are:
- `test_formatted_file_output`
- `test_exclude_multiple_urls`

Both are failing due to `https://creativecommons.org/publicdomain/zero/1.0/` returning 403 Forbidden, which seems to be an external service issue or rate limiting in the CI environment.

The changes in this PR only affect path display normalization and all path-related tests pass successfully.

mre (commented):
Interesting idea. 

- The issue name is not filled out in the pull request description. 
- There are a few comments in the PR code, which I would clean up. Will highlight them. 
- Needs an Integration test in cli.rs
- Does Windows path conversion in multiple places. I'm also not sure if there aren't any edge cases with that. Maybe there's a library for it. Also wondering if we even need to change the Windows paths because I think it's good to know the exact path that was checked, even if that means the output is different on different platforms. 
- I think the path cleanup should be in a separate method (or even module). Display should not have a lot of business logic, but just call into other methods if needed.

lychee-lib/src/types/input.rs
```diff
@@ -90,13 +90,57 @@ impl Serialize for InputSource {
...
 impl Display for InputSource {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str(match self {
-            Self::RemoteUrl(url) => url.as_str(),
-            Self::FsGlob { pattern, .. } => pattern,
-            Self::FsPath(path) => path.to_str().unwrap_or_default(),
-            Self::Stdin => "stdin",
-            Self::String(s) => s,
-        })
+        use std::path::{Component, PathBuf}; // Move to top of function
```
mre: Guess this can be cleaned up?

lychee-lib/src/types/input.rs
```diff
@@ -90,13 +90,57 @@ impl Serialize for InputSource {
...
+
+        match self {
+            Self::RemoteUrl(url) => f.write_str(url.as_str()),
+            Self::FsGlob { pattern, .. } => f.write_str(pattern),
+            Self::FsPath(path) => {
+                // Use to_string_lossy() for safety with non-UTF8 paths
+                let original = path.to_string_lossy();
```
mre: You're only using `original` later. I would move it further down.

lychee-lib/src/types/input.rs
```diff
@@ -90,13 +90,57 @@ impl Serialize for InputSource {
...
+
+                let mut result = normalized.to_string_lossy().into_owned();
+
+                // Handle empty path
+                if result.is_empty() {
+                    result = ".".to_string();
+                }
+
+                // Preserve trailing slash if present in original
+                // Fix collapsible_if warning
```
mre: Redundant comment

lychee-lib/src/types/input.rs
```diff
@@ -90,13 +90,57 @@ impl Serialize for InputSource {
...
+                    return f.write_str(&original);
+                }
+
+                // For relative paths, do component-based normalization
+                let normalized: PathBuf = path
+                    .components()
+                    .filter(|comp| !matches!(comp, Component::CurDir))
+                    .collect();
+
+                let mut result = normalized.to_string_lossy().into_owned();
```
mre: Looks weird. I wonder if we can avoid the string conversion or at least the `into_owned`.

lychee-lib/src/types/input.rs
```diff
@@ -90,13 +90,57 @@ impl Serialize for InputSource {
...
+                    && !result.ends_with('/')
+                    && result != "."
+                {
+                    result.push('/');
+                }
+
+                // Ensure forward slashes on Windows
+                #[cfg(windows)]
+                {
+                    result = result.replace('\\', "/");
```
mre: This is the same code as above. I also don't want to litter the code with windows checks. I would see if there's a method in the standard library or a crate in the Rust, which handles the path conversion. My hunch is that it's not that easy.

lychee-lib/src/types/input.rs
```diff
@@ -496,6 +540,124 @@ mod tests {
...
+    fn test_normalized_path_display() {
+        // Test that paths with and without "./" prefix are displayed identically
+        let path1 = InputSource::FsPath(PathBuf::from("doc/docker.md"));
+        let path2 = InputSource::FsPath(PathBuf::from("./doc/docker.md"));
+        assert_eq!(path1.to_string(), path2.to_string());
+        assert_eq!(path2.to_string(), "doc/docker.md");
+    }
+
+    #[test]
+    fn test_path_display_comprehensive() {
```
mre: We should use the rstest crate for this.

lychee-lib/src/types/input.rs
```diff
@@ -496,6 +540,124 @@ mod tests {
...
+            ("doc\\sub/file.md", "doc/sub/file.md"),
+            (".\\doc/file.md", "doc/file.md"),
+            // Windows absolute paths
+            ("C:\\Users\\file.md", "C:/Users/file.md"),
+            (
+                "C:\\Users\\Documents\\file.md",
+                "C:/Users/Documents/file.md",
+            ),
+            // UNC paths
+            ("\\\\server\\share\\file.md", "//server/share/file.md"),
```
mre: This is strange. I was not expecting that it changes the `\\\\` to `//`. Could potentially be confused with a different path, but I don't have an example right now. Not sure.

mre: If anything, why not just a single slash in the beginning? Not sure how UNC paths work.

lychee-lib/src/types/input.rs
```diff
@@ -496,6 +540,124 @@ mod tests {
...
+        ];
+
+        for (input, expected) in windows_cases {
+            let source = InputSource::FsPath(PathBuf::from(input));
+            assert_eq!(
+                source.to_string(),
+                expected,
+                "Failed for Windows input: '{}'",
+                input
```
mre: This input variable could be inlined into the format string.

tosky1125:
@mre Thank you for the thorough review! I've addressed most of your feedback:

### Changes made:
1.  **Extracted normalization logic**: Moved path normalization from `Display` impl to a dedicated `InputSource::normalize_path_display()` method for better separation of concerns.

2. **Added integration test**: Added `test_normalized_path_display` in `cli.rs` to verify that paths like `./test.md` are displayed consistently as `test.md`.

3. **Removed Windows path conversion**: As you suggested, I've removed the backslash-to-forward-slash conversion to preserve platform-specific paths. Now only `.` normalization is performed.

4. **Cleaned up comments**: Removed redundant comments and moved variable declarations closer to their usage.

5. **Simplified implementation**: The current implementation only removes leading `./` for consistency, avoiding complex path manipulations.

Additionally, I've confirmed that the CI failures are unrelated to this PR:

The failing tests expect:
- `successful: 10`
- `errors: 0`

But actually get:
- `successful: 7` 
- `errors: 1` (403 from creativecommons.org)

Our path normalization only affects relative paths (removes `./` prefix) and preserves absolute paths unchanged. The JSON serialization works correctly with our changes.

This is purely an external service issue affecting all PRs.

lychee-lib/src/types/input.rs
```diff
@@ -88,15 +88,50 @@ impl Serialize for InputSource {
...
+            ".".to_string()
+        } else {
+            cleaned.to_string_lossy().into_owned()
+        };
+
+        // Preserve trailing slash for "./"
+        if path_str == "./" {
+            "./".to_string()
+        } else if has_trailing_slash && result != "." && !result.ends_with('/') {
+            format!("{result}/") // 수정된 부분
```
mre: Comment can be removed

tosky1125: removed 👍

thomas-zahner:
Thank you for the PR. This is an interesting idea. However, this does not really fix https://github.com/lycheeverse/lychee/issues/1660 as to my understanding. This now "normalises" output files to some specific format. But the affected files are still checked multiple times and also displayed multiple times. To really address the issue we would need to filter out such duplicates before checking them.

lychee-bin/tests/cli.rs
```diff
@@ -2222,4 +2222,24 @@ mod cli {
...
+
+        let mut cmd = main_command();
+        cmd.arg("test.md").arg("./test.md").current_dir(dir.path());
+
+        let output = cmd.output()?;
+        let stdout = String::from_utf8_lossy(&output.stdout);
+
+        // Both should be displayed as "test.md" (not "./test.md")
+        assert!(stdout.contains("[test.md]:"));
+        assert!(!stdout.contains("[./test.md]:"));
```
thomas-zahner, line 2241: To really fix https://github.com/lycheeverse/lychee/issues/1660 as to my understanding the assertions must check that `test.md` (or `./test.md`) occurs exactly once in `stdout`. In the current state of the PR this assertion will fail because `test.md` occurs twice.

mre:
> this does not really fix https://github.com/lycheeverse/lychee/issues/1660 

That is correct. @tosky1125, can you remove the `Fixes #1660` from the description? I think the PR is still valuable to be merged as is because it cleans up the output a bit. @thomas-zahner, do you agree?

lychee-bin/600c0d6d-eacd-4fba-bef0-bd79a4da72c2.json
```diff
@@ -0,0 +1,27 @@
+{
```
thomas-zahner, line 1: What's the purpose of all these JSON files? They don't seem to be used in the test, right?

lychee-lib/src/types/input.rs
```diff
@@ -88,15 +88,50 @@ impl Serialize for InputSource {
...
 impl Display for InputSource {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str(match self {
-            Self::RemoteUrl(url) => url.as_str(),
-            Self::FsGlob { pattern, .. } => pattern,
-            Self::FsPath(path) => path.to_str().unwrap_or_default(),
-            Self::Stdin => "stdin",
-            Self::String(s) => s,
-        })
+        match self {
```
thomas-zahner, line 128: I'd much prefer the previous position of `match` as this didn't duplicate `f.write_str` five times

thomas-zahner (commented):
After giving it some thought, I came to the conclusion that I would prefer not to merge this. As mentioned earlier addressing the root cause to eliminate duplicates would be great, but this PR does not solve that. 

This PR applies some cosmetic changes to the output. However, I actually prefer the current output because it contains more information.

Given the following example `cargo run *.md ./*.md --dump-inputs` we currently get

```
CONTRIBUTING.md
README.md
./CONTRIBUTING.md
./README.md
```

with this PR I would expect the following (noticed that it's not yet the case because this PR only affects normal link checking as of now, not dumping [which would be another reason for not merging the PR because I expect consistent behaviour as a user across all commands])

```
CONTRIBUTING.md
README.md
CONTRIBUTING.md
README.md
```

This removes some useful information in my opinion. I no longer understand as easily why my files were included twice for link checking.

mre:
I went ahead and closed this as it doesn't resolve an immediate problem. Thanks anyway for the idea and the PR, @tosky1125.

lychee-bin/600c0d6d-eacd-4fba-bef0-bd79a4da72c2.json
thomas-zahner: Realised that these JSON files were likely created with the `test_json_output` macro after test failures. This was because the files were not cleaned up after a failing assertion. I've fixed that behaviour with https://github.com/lycheeverse/lychee/pull/1771/commits/ec4bedb3d4d21ecc0a2db9b6f94423d4f43ca8ba

thomas-zahner:
Yes, thank you nonetheless for the idea and your efforts @tosky1125.

PR #1764, opened 2025-07-09
docs: add `fallback_extensions` to example config file

kemingy:
This is very common for documentation tools like vitepress, astro, etc.


mre:
Good stuff!

Issue #1620, opened 2025-01-30
Not excluding .png


sekyondaMeta:
Running the workflow below does not seem to exclude the .png file. I have tried differernt permutations and combinations of adding the regex but nothing seems to work. Any ideas would be greatly appreciated.

Workflow:
`name: link check on PR

on:
  pull_request:
    branches: [main]

jobs:
  linkChecker:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 1

      - name: Get Changed Files
        id: changed-files
        uses: tj-actions/changed-files@v41

      - name: Check for Skip Label
        id: skip-label
        uses: actions/github-script@v6
        with:
          script: |
            const labels = await github.rest.issues.listLabelsOnIssue({
              owner: context.repo.owner,
              repo: context.repo.repo,
              issue_number: context.issue.number
            });
            return labels.data.some(label => label.name === 'skip-link-check');
          
      - name: Check Links
        if: steps.skip-label.outputs.result == 'false'
        uses: lycheeverse/lychee-action@v1
        with:
          args: --exclude '\.(pdf|zip|png|jpg)$' --accept=200,403,429 --base . --verbose --no-progress ${{ steps.changed-files.outputs.all_changed_files }}
          token: ${{ secrets.CUSTOM_TOKEN }}
          fail: true
          
      - name: Skip Message
        if: steps.skip-label.outputs.result == 'true'
        run: echo "Link check was skipped due to the presence of the 'skip-link-check' label."
        
      - name: Suggestions
        if: failure()
        run: |
          echo -e "\nPlease review the links reported in the Check links step above."
          echo -e "If a link is valid but fails due to a CAPTCHA challenge, IP blocking, login requirements, etc., consider adding such links to .lycheeignore file to bypass future checks.\n"
          exit 1`

error message:

`Run lycheeverse/lychee-action@v1
Run # Cleanup artifacts from previous run in case it crashed
lychee
Run /home/runner/work/_actions/lycheeverse/lychee-action/v1/entrypoint.sh
Error: Cannot read input content from file recipes_source/customaddandround.png`

`Caused by:
    stream did not contain valid UTF-8`

mre:
The key issue is this:

```
lychee --exclude '\.(pdf|zip|png|jpg)$' --accept=200,403,429 --base . --verbose --no-progress .
```

[...]

PR #1766, opened 2025-07-11
Regular expressions for `exclude_path`

### LINKED ISSUE #1608: Does `exclude_path` support regular expressions?
The docs (https://lychee.cli.rs/recipes/excluding-paths) seem to indicate that it does

> Regular expressions are also supported.
> ...
> exclude_path = ["node_modules", "^./dir/", ".*/dev/.*"]

but I can't get it to work, and I don't see any tests in the repo for it, thanks.

thomas-zahner:
Closes #1608


lychee-lib/src/filter/regex_filter.rs
```diff
@@ -0,0 +1,46 @@
...
+        Self {
+            regex: RegexSet::empty(),
+        }
+    }
+}
+
+impl PartialEq for RegexFilter {
+    fn eq(&self, other: &Self) -> bool {
+        // Workaround, see https://github.com/rust-lang/regex/issues/364
+        self.regex.patterns() == other.regex.patterns()
```
thomas-zahner: That's the only thing in the PR I'm not really happy about. `Input` has the field `excluded_paths: PathExcludes` and `Input` derives `PartialEq` so `PathExcludes` which is a type alias for `RegexFilter` also has to implement `PartialEq` which is not really supported by the `regex` crate. Maybe it could be possible to remove the filed from `Input`. But not sure if this is worth the effort.

mre:
Just a note that we should also change back the docs as part of this work: https://github.com/lycheeverse/lycheeverse.github.io/pull/92 😉

lychee-lib/src/filter/regex_filter.rs
mre: Yeah, probably `excluded_paths` in `Input` is wrong.

The `Input` should not know about `PathExcludes`. I think it was just tempting to add that, but maybe it's easier to remove it than we think? Might be worth a try.

lychee.example.toml
```diff
@@ -104,10 +104,10 @@ include_verbatim = false
 # Ignore case of paths when matching glob patterns.
 glob_ignore_case = false
 
-# Exclude URLs and mail addresses from checking (supports regex).
+# Exclude URLs and mail addresses from checking. The values are treated as regular expressions
```
mre, line 107: Should we add an example for a useful regular expression here?

lychee.example.toml
```diff
@@ -104,10 +104,10 @@ include_verbatim = false
...
 exclude = ['^https://www\.linkedin\.com', '^https://web\.archive\.org/web/']
 
-# Exclude these filesystem paths from getting checked.
+# Exclude paths from getting checked. The values are treated as regular expressions
```
mre, line 115: same here

lychee.example.toml
thomas-zahner: Agree :+1:

lychee-lib/src/filter/regex_filter.rs
thomas-zahner: Turned out to be easier than expected. `excluded_paths` is now simply part of the `Collector` instead of `Input`.

thomas-zahner:
I realised that this actually fixes two issues we tracked separately: https://github.com/lycheeverse/lychee/issues/1608 and https://github.com/lycheeverse/lychee/issues/1620

mre:
This looks done, right? We could merge and then create a new release as it fixes an important bug.

thomas-zahner:
Yes it's done. I will update the docs accordingly.

(update: see https://github.com/lycheeverse/lycheeverse.github.io/pull/99)

PR #1775, opened 2025-07-26
fix comment for `ErrorKind::InvalidFragment`

katrinafyi:
the comment doesn't make sense and it is identical to InvalidFilePath which is right above it, so I reason this was a copy/paste mistake.


mre:
Thanks for fixing. 😃

PR #1776, opened 2025-07-27
tests: refactor `test_fragments` to clarify expected successes/fails

katrinafyi:
this PR improves `test_fragments` by separating the URLs into separate
lists for expected failures and expected successes. each of these are
searched in stdout and/or stderr to check that the URL has the expected
outcome. this makes the expected test behaviour much more obvious, where
it was previously only implied by the "x total" and "x OK" strings.

this PR makes no changes to actual behaviour. hopefully, it will make it
easier to test future changes to the fragment behaviour.

previously, the test only checked that stderr contains a list of URLs.
because `--verbose` is specified, stderr would contain both successful and
failing URLs, so this didn't really test much. it only tested that the
URLs were detected by lychee.

also, search terms are now suffixed with whitespace to prevent `file.html`
from matching `file.html#something`. the test should consider these as
different links, but the previous code had the potential to match
`file.html#something` when looking for `file.html`.

in making this PR, we find that a fragment link to an empty file is
always treated as a valid link because it is detected as a plaintext
file. maybe there should be a warning when a URL points to a plaintext
file and a fragment check is requested
([relevant code](https://github.com/lycheeverse/lychee/blob/ea415c8db4597c383f88b5fc9894fd86a7245494/lychee-lib/src/utils/fragment_checker.rs#L133)).

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1877,56 +1877,73 @@ mod cli {
         let mut cmd = main_command();
         let input = fixtures_path().join("fragments");
 
-        cmd.arg("--verbose")
+        let mut result = cmd
             .arg("--include-fragments")
+            .arg("--verbose")
             .arg(input)
             .assert()
-            .failure()
-            .stderr(contains("fixtures/fragments/file1.md#fragment-1"))
-            .stderr(contains("fixtures/fragments/file1.md#fragment-2"))
-            .stderr(contains("fixtures/fragments/file1.md#code-heading"))
-            .stderr(contains("fixtures/fragments/file2.md#custom-id"))
-            .stderr(contains("fixtures/fragments/file1.md#missing-fragment"))
-            .stderr(contains("fixtures/fragments/file2.md#fragment-1"))
-            .stderr(contains("fixtures/fragments/file1.md#kebab-case-fragment"))
-            .stderr(contains(
-                "fixtures/fragments/file1.md#lets-wear-a-hat-%C3%AAtre",
-            ))
-            .stderr(contains("fixtures/fragments/file2.md#missing-fragment"))
-            .stderr(contains("fixtures/fragments/empty_file#fragment"))
-            .stderr(contains("fixtures/fragments/file.html#a-word"))
-            .stderr(contains("fixtures/fragments/file.html#in-the-beginning"))
-            .stderr(contains("fixtures/fragments/file.html#in-the-end"))
-            .stderr(contains(
-                "fixtures/fragments/file1.md#kebab-case-fragment-1",
-            ))
-            .stderr(contains("fixtures/fragments/file.html#top"))
-            .stderr(contains("fixtures/fragments/file2.md#top"))
-            .stderr(contains(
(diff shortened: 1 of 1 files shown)


mre:
Yeah, we've certainly outgrown our old fragment testing setup, so I'm thankful for the refactor.

> in making this PR, we find that a fragment link to an empty file is
always treated as a valid link because it is detected as a plaintext
file. maybe there should be a warning when a URL points to a plaintext
file and a fragment check is requested

Makes sense, yes. Would you like to add the warning as part of this PR or a separate one? I'm fine with either option.

katrinafyi:
I've made the change but I'm looking at [other discussions](https://www.github.com/lycheeverse/lychee/issues/1709) now and the warning could be hit often, for instance if a user has lots of local binary files. This seems more noisy than helpful. That said, people shouldn't be attaching fragments to binary file links anyway so maybe it's okay.

mre:
Maybe instead of `warning` we could set it to `info`. I think it's fine to only print it in verbose mode if it's expected to occur often.

katrinafyi:
It is done :)

lychee-bin/tests/cli.rs
```diff
@@ -1877,60 +1877,92 @@ mod cli {
...
-            .stderr(contains(
-                "https://raw.githubusercontent.com/lycheeverse/lychee/master/fixtures/fragments/zero.bin#fragment",
-            ))
-            .stdout(contains("42 Total"))
-            .stdout(contains("31 OK"))
+            .failure();
+
+        let expected_successes = vec![
+            "fixtures/fragments/empty_dir",
+            "fixtures/fragments/empty_file#fragment", // XXX: is this a bug? a fragment in an empty file is being treated as valid
```
mre, line 1889: You're right to question this. I think we should treat fragment links to empty files as invalid.

Our job is to identify links that won't provide a meaningful user experience. A fragment reference in an empty file is functionally broken: it promises to take the user to specific content that doesn't exist. While the file itself may be reachable, the fragment reference is semantically meaningless, making this a legitimate case to flag as invalid.

I'd be willing to accept a pull request, which changes this, and I would welcome a PR for that.

mre:
Thanks for the nice change, @katrinafyi.

lychee-bin/tests/cli.rs
katrinafyi: I do agree in principle, but I couldn't think of a nice way to implement it. Fragments in empty file are treated as existing because empty files are detected as plaintext. But I think it would be too heavy-handed to reject all fragments on plaintext files, especially since plaintext is the fallback file type for unknown files. 

I think the info message is a okay for now, and maybe someone more experienced can look at it :) Maybe the first step would be to differentiate plaintext files and unknown file types, so you could handle them differently.

mre: Oh, that's a really good idea. I guess in the long run, we should consider deeper file inspection anyway. This gives us more flexibility and takes away some of the guesswork.

thomas-zahner:
Thanks @katrinafyi this test also used to bother me a bit as it was so imprecise

Issue #542, opened 2022-03-07
Allow index files when checking a folder


dkarlovi:
_(note: this was previously discussed in https://github.com/lycheeverse/lychee/issues/452#issuecomment-1051025563)_

When deploying HTML files to some common platforms like Netlify or Gitlab Pages (or even on a typical Apache HTTP environment), you can rely on the platform to use an "index file", namely, for an URL like `https://example.com/foo`, it will serve `./public/foo/index.html` if such a file exists.

This means that, the heuristic should look like this:
1. is the "index file" feature enabled?
2. if so, does the URL naturally resolve to a folder?
3. if so, does the folder contain a file called `index.html` (or `index.htm`, maybe allow passing a list when enabling the feature)
4. if so, the URL is actually valid

The feature could be enabled as `lychee --allow-index-file index.html` (where the name is used in 3)

mre:
Is this for checking files offline or online?

I'm guessing Apache would forward the request to the correct `index.html`/`index.html` file automatically with a redirect.

For offline checks, this could indeed be helpful. At the moment we only check if the directory exists where the `index.html` has to be in. The idea was that the web server would "do the right thing" if the folder existed. For example, if no `index.html` existed, it might show a directory listing instead if that is configured.
So in this case, even if the no `index.html` existed, this could still be a valid request. We'd only know by looking at the server config.

I like that your idea doesn't require knowledge about the server config and instead the user would be responsible for making the decision.

dkarlovi:
> Is this for checking files offline or online?

Only offline, I think?

> I like that your idea doesn't require knowledge about the server config and instead the user would be responsible for making the decision.

Indeed, I actually have a use case where this should *not* work, as in, even if the `index.html` file exists, pointing to its parent repo should be treated as a failure (because the deployment target does *not* support the index file feature).

lebensterben:
there're too much work for this feature request.

for netlify, you need to parse per-directory configuration and check "redirects". See https://docs.netlify.com/configure-builds/file-based-configuration/

for apache, you need DocumentRoot.

for gitlab we need to check "pages" job in .gitlab-ci.yml. specifically we need pages.artifects.paths to determine where index file is stored in a repository.

dkarlovi:
@lebensterben I think there's a misunderstanding: you don't need to do any of that. You **let the user decide** if index files are supported or not for their use case with a runtime flag. If they say the index files are supported, lychee checks for them. If the user says they're not supported, lychee doesn't check for them.

[...]

Issue #269, opened 2021-06-24
Re-define `lychee` as both a link checker and a link linter


lebensterben:
Currently, `lychee` is just a link ___checker___, which conceptually has the following components:
1. It parses a config file or CLI argument, and turns it into an internal configuration
2. An __overseer__  is created given the configurations. It serves several purposes:
   - Detection: It detects/scrape HTML links and EMAIL addresses from the input file(s)
   - Filtering: It checks whether the link should be checked or skipped
   - Dispatcher: It sends URLs to __workers__ who verify whether a given URL is valid by checking whether it's accessible.
   - Logging: It keeps an record of URL it encounters and the status of each URL. (Valid, Ignored, Timeout, etc)
3. The result is returned either as CLI output or to a log file.

---

We can extend `lychee` so that it's also a link ___linter___.

Note that the __overseer__ hands the job to its __workers__, and __workers__ are only responsible for validating the URLs. The __linting__ facility should also be done by __overseer__.

Conceptually, __linting__ could happen before and/or after the __overseer__ dispatches the job. For example:
- Before: If the user wants to deny a certain pattern in URL, the __overseer__ don't need to send URLs that match the pattern to the __workers__, instead it directly logs it.
- After: If the user wants to avoid the use of absolute/relative links, the __overseer__ first verify whether a link is valid, and if it is, then it check whether it's in the desired style before logging.
- Around (before and after): If the user wants to use HTTPS links whenever possible, the __overseer__ first sends an HTTP link to __workers__, and if it's valid the __overseer__ resends it HTTPS counterpart the __workers__, and logs the result accordingly.

untitaker:
This is labelled as `design-feedback` because it already goes into the nitty-gritty of which code unit does what, but from what i can tell what is actually proposed here is to expand lychee's feature scope such that it can be used to enforce policies on HTML, such as the already mentioned "need to use HTTPs links everywhere" and "links can't match this pattern".

You can already enforce some basic pattern-policies today: Use `--dump` and a wrapper shell script. What can't be done is the HTTPs enforcement in the way you imagine it.

Is the list of policies in your OP exhaustive?

mre:
@lebensterben any thoughts?

untitaker:
prior art: https://github.com/wjdp/htmltest

mre:
Link validation is a whole other use-case with a lot of design decisions to consider along the way. We have to be careful to keep the scope manageable. I guess we can commit to the following:
[...]

PR #1777, opened 2025-07-27
add `--index-files` flag, and turn off index file checking by default

katrinafyi:
this pr introduces the `--index-files` argument. this argument lets the user specify a list of index files names to require for local directory links. if specified, at least one index file must exist in order for a directory link to be valid.

this pr also makes the default behaviour more consistent. after this pr, index files are never used by default and directories are valid as long as they exist. this is a small difference from the current behaviour which implicitly searches `index.{ext}` (implemented in https://github.com/lycheeverse/lychee/pull/1752) - this causes the +1 error count in `test_fragments`.

this pr also makes some clarifying points about the interactions of `--index-files`, `--fallback-extensions`, and fragment checking.
- for a given link, only one of index files or fallback extensions will take effect. this is determined by whether the original path exists or not. if the path exists and is a directory, index files are applied. if the path does not exist, fallback extensions are applied.
- when a link resolves to a directory (e.g., in the default behaviour), the directory is treated as a file with no fragments  for the purpose of fragment checking, 

some more details about the changes:
- add `checker::file::tests` module with some specific unit tests for the local file checker
- adds `InvalidIndexFile` to ErrorKind to display a more helpful "cannot find index file in directory", rather than "cannot find file"
- adds `fixture_uri` method to test_utils to construct a URI relative to the fixtures dir. importantly, this supports building a URI with a fragment.
- restructured the file.rs code to separate methods for applying index files and fallback extensions. these are changed to return a Result so they can report specific errors, rather than just option.
- the code logic handling fragment checking directories is pushed into the `check_fragment` method. previously, it was an ad-hoc `if` statement higher up the call stack.
- the index files/fallback extension behaviour is determined by a match statement in `check_path`. hopefully, this makes it more obvious what cases will lead to particular code paths.

this pr can be tested with:
```
cargo run -- fixtures/fragments --include-fragments
```
note that `fixtures/fragments/sub_dir#a-link-inside-index-html-inside-sub-dir` now fails with Cannot find fragment, because index.html is no longer searched by default.

the full diff of `cargo run -- fixtures/fragments --include-fragments` between previous and this pr is below. also note that the `empty_dir` cases with a fragment have the more specific "cannot find fragment" message, because the directory does exist.
```diff
--- a   2025-07-27 22:35:20.601240030 +1000
+++ b   2025-07-27 22:35:50.691518775 +1000
@@ -1,10 +1,11 @@
 Issues found in 2 inputs. Find details below.
 
 [fixtures/fragments/file1.md]:
-[ERROR] file:///home/x/progs/lychee/fixtures/fragments/empty_dir#non-existing-fragment-3 | Cannot find file
-[ERROR] file:///home/x/progs/lychee/fixtures/fragments/empty_dir#non-existing-fragment-4 | Cannot find file
+[ERROR] file:///home/x/progs/lychee/fixtures/fragments/empty_dir#non-existing-fragment-3 | Cannot find fragment
+[ERROR] file:///home/x/progs/lychee/fixtures/fragments/empty_dir#non-existing-fragment-4 | Cannot find fragment
 [ERROR] file:///home/x/progs/lychee/fixtures/fragments/file1.md#missing-fragment | Cannot find fragment
 [ERROR] file:///home/x/progs/lychee/fixtures/fragments/file2.md#missing-fragment | Cannot find fragment
+[ERROR] file:///home/x/progs/lychee/fixtures/fragments/sub_dir#a-link-inside-index-html-inside-sub-dir | Cannot find fragment
 [ERROR] file:///home/x/progs/lychee/fixtures/fragments/sub_dir#non-existing-fragment-1 | Cannot find fragment
 [ERROR] file:///home/x/progs/lychee/fixtures/fragments/sub_dir#non-existing-fragment-2 | Cannot find fragment
 [ERROR] file:///home/x/progs/lychee/fixtures/fragments/sub_dir_non_existing_1 | Cannot find file
@@ -15,4 +16,4 @@
 [ERROR] file:///home/x/progs/lychee/fixtures/fragments/file.html#in-THE-begiNNing | Cannot find fragment
 [ERROR] file:///home/x/progs/lychee/fixtures/fragments/file.html#in-the-end | Cannot find fragment
 
-🔍 42 Total (in 0s) ✅ 31 OK 🚫 11 Errors
+🔍 42 Total (in 0s) ✅ 30 OK 🚫 12 Errors
```

then, try
```
cargo run -- fixtures/fragments --index-files index.html --include-fragments    
```
and observe the output contains
```
[fixtures/fragments/file1.md]:
   [ERROR] file:///home/x/progs/lychee/fixtures/fragments/empty_dir | Cannot find index file within directory
   [ERROR] file:///home/x/progs/lychee/fixtures/fragments/empty_dir#non-existing-fragment-3 | Cannot find index file within directory
   [ERROR] file:///home/x/progs/lychee/fixtures/fragments/empty_dir#non-existing-fragment-4 | Cannot find index file within directory
```

of course, you should also try to run the new test suite. i hope that the test code is readable and is an understandable description of the behaviour.
```
cargo nextest run -p lychee-lib --lib checker::file::tests
```

i would appreciate any comments! especially proofreading and feedback on all the documentation and messages.

this pr properly implements https://github.com/lycheeverse/lychee/issues/542. also related to https://github.com/lycheeverse/lychee/issues/1765


README.md
```diff
@@ -452,12 +452,34 @@ Options:
...
+
+      --index-files <INDEX_FILES>
+          When checking locally, resolves directory links to an index file within the
+          directory. The argument is a comma-separated list of index file names to search
+          for. Index files are attempted in the order given, and at least one must exist
+          in order for a directory link to be considered valid.
+
+          The special name `.` can be used to resolve directory links to the directory
+          itself. When `.` is specified, a directory link will be accepted as long as the
+          directory exists. This is the default behavior.
```
mre: This is pretty clever

README.md
mre, line 1: In one of your commits you mentioned that fallback extensions and index files are essentially the same thing, e.g. they both modify the set of accepted paths.

Have you changed your mind on this?
Just wondering, because I would not mind merging both features if that makes the API cleaner. We can still make breaking changes at this point. But your current implementation looks fine to me anyway.

katrinafyi: I have changed my mind. That comment was from before I had the idea to determine index vs fallback based on the original path's existence, and I was struggling to make them interact nicely when both were applied.

Now, I think that keeping the options separate is a more obvious API. It also lets us report different kinds of error to help the user.

README.md
katrinafyi: I wonder if it's too clever. For instance, I have to apply this default value at lots of different levels - CLI parsing and ClientBuilder both set it. 

Is this okay? Or maybe the API should take Options? 

Edit: In fact, I just found a bug to do with setting the default value 😖

katrinafyi:
Test fails at https://example.net and seems unrelated/flaky. It does seem to be broken when visiting in Firefox, but it works in Chrome and curl. Weird.

README.md
thomas-zahner: It's an interesting idea. But I think it's overly complicated. Also if I didn't misunderstand there does not seem to be any use-case for this ever. If a user ever provides `.` as a value (e.g. `--index-files index.html,index.md,.`) it should be exactly equivalent to not providing the whole flag at all.

So the only reason for `.`'s existence is for internal representation which is much cleaner represented with an empty list and probably a new condition withing `apply_index_files`. (or `Option` if really necessary)

lychee-lib/src/checker/file.rs
```diff
@@ -96,132 +108,155 @@ impl FileChecker {
...
-    /// * `uri` - The original URI, used for error reporting.
+    /// * `uri` - The original URI, used for checking and error reporting.
     ///
     /// # Returns
     ///
     /// Returns a `Status` indicating the result of the check.
     async fn check_path(&self, path: &Path, uri: &Uri) -> Status {
-        let file_path = self.resolve_file_path(path);
-        let has_fragment = uri.url.fragment().is_some_and(|x| !x.is_empty());
+        let path = match path.metadata() {
```
thomas-zahner, line 128: This match is a great solution, better than before :+1:
I just think that the comments inside the match are not necessary, they just repeat the code.

lychee-lib/src/checker/file.rs
```diff
@@ -238,3 +273,196 @@ impl FileChecker {
...
+        assert_filecheck!(
+            &checker,
+            "filechecker/dir_with_extension",
+            Status::Error(InvalidFilePath(_))
+        );
+    }
+
+    #[tokio::test]
+    async fn test_empty_index_list_corner() {
+        // empty index_files list will reject all directory links
```
thomas-zahner, line 426: So this is the case we could redefine. "no specified index files == skip check" If you find this weird then the alternative would be to wrap it with `Option` and make `None` skip the check

lychee-lib/src/checker/file.rs
```diff
@@ -96,132 +108,155 @@ impl FileChecker {
...
+                };
 
-        for ext in &extensions_to_try {
-            let index_path = dir_path.join(format!("index.{ext}"));
-            if index_path.is_file() {
-                return Some(index_path);
-            }
-        }
-        None
+                let path = dir_path.join(filename);
```
thomas-zahner, line 224: Should be able to replace `join` with [`with_file_name`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_file_name). Should resolve the comment in `test_index_file_traversal_corner`

README.md
katrinafyi: `--index-files index.html,index.md,.` is different to omitting the flag because it will use an index file if it exists before falling back to the directory. this could be useful if somebody has a setup like this.

 file names _after_ a `.`, however, will have no effect

lychee-lib/src/checker/file.rs
katrinafyi: i don't think the comments hurt. also, i think some cases like the `Err(e)` and the last `Ok(_)` deserve some explanation. these cases are less obvious because their condition implicitly includes the negation of the previous `if` conditions.

but if you direct me to remove them, i will do so. just let me know which you want to remove.

lychee-lib/src/checker/file.rs
katrinafyi: the docs for with_file_name point to [set_file_name](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.set_file_name) which says
> The argument is not sanitized, so can include separators. This behavior may be changed to a panic in the future.

so, at this point in time, it will probably not affect the behaviour. but i'll make the change anyway.

katrinafyi: i tried to make the change and it is a bit troublesome. using `.with_file_name` strips the last path component, effectively making it relative to the parent directory, rather than a file within the directory as we would like.

to avoid this, you can use 
```
dir_path.join("nothing").with_file_name(filename)
```
but this is really just identical to `.join`.

so i don't think it's worth making the change here. if you want, a separate PR could add validation to the CLI arg to ensure no slashes are present.

thomas-zahner: Ah I see. Yes we should leave it in that case.

README.md
katrinafyi: The field has been changed to an option :) See https://github.com/lycheeverse/lychee/commit/f6b678eb49a1f1bd28ec5c50f2e4eb91360cd4a7 for more commentary.

thomas-zahner: Thanks for the clarification and for the new commit. I'm starting to think that we should now either:

Rename the flag from `--index-files` to something like `--remap-directory`. (maybe you have a another idea) This captures the current behaviour better IMO, as this is not necessarily about index files. (though this might be the most common use-case) Especially since we can provide arbitrary values such as `.`, `..` and `some/subdirectory/random.file`. Or only allow values of simple file names that don't affect the directory in which the remapped files lie. But we pretty much already agreed that the latter case is not a priority. So my preference is renaming the flag. The flag description and documentation of course can still explain the common "index" use-case.

katrinafyi: I mean, I can't imagine a use for this flag which _doesn't_ include index files in some way. `.` is the _only_ directory name which will be accepted, and this is to support the case for _optional_ index files. If you want, I can add this use case to the help and test cases to make it more visible. If you pass `..` or some other path to a directory, those entries will never be considered a valid index file. This is recorded by one of the corner test cases. 

And we've established that path traversal into nested directories is accidental rather than intentional, so I wouldn't want to generalise the option name to include this unsupported use. In the other comment thread, I talked about eventually restricting this with validation.

So, my recommendation is to keep the current name. It's clear to users who will be looking for this feature, and I don't think anyone will be surprised by its name or functionality.

But again, if you want to overrule me, let me know your desired name and I will make it so.

katrinafyi: The other problem is remap is already a defined term in the Lycheeverse. remap-directories has the connotation that this is a directory-specific version of remap, but its functionality is not nearly that powerful. It can't rewrite directories in the middle of paths, for instance. Nor can it change its behaviour based on the URL being checked.

I'd rather have a name which is very slightly too narrow than one which is too broad. (And I don't want to think of another name rn :))

dkarlovi:
@katrinafyi  how is this addressing #1765, should that be addressed here? Assuming you want to explicitly say index files are **not** supported and want to make linking to a dir an error?

@thomas-zahner noted:
> But we could introduce a new feature flag to make lychee reject any directories that do not contain an index.html file.

Since the default is "link to a directory is valid", this feature is saying "link to a directory with `<index file>` is valid", we seem to need a feature flag to say "link to a directory is not valid" which can also be read as "index files **not** supported", maybe `--no-index-files` or similar?

katrinafyi:
@dkarlovi That issue has had a few back and forths.

- It was originally called "Links to an empty directory should fail when running offline" and in the text, I wrote about requiring a index.html. That would be implemented by `--index-files index.html`. This was my intention when filing the issue.
- The title was changed to "Links to directories should fail when running offline" without qualification on the directories. With this PR, that could be done by using `--index-files ''`. This is a bit obscure but it will work. Idk if this would be a common enough case to warrant mentioning in the help.

dkarlovi:
> With this PR, that could be done by using --index-files ''. This is a bit obscure but it will work. Idk if this would be a common enough case to warrant mentioning in the help.

Ah, OK. :+1:  I'd definitely make at least a mention about it in the docs because it's definitely not obvious.

The use case we have for this is: when you're working on Gitlab, you can directly browse your static site in the MR artifacts, but it doesn't support index files, linking to a dir is a mistake which should get detected. When you deploy the site to Gitlab pages, it **does** support index pages so linking to an index file directly is a mistake which should be detected.

katrinafyi:
Gitab lets you browse an artifact in a MR? That's really cool! For your Gitlab pages, this PR won't be able to flag links that unnecessarily point to index.html - but maybe you have another approach for that already.

I'll make the changes to the docs :)

dkarlovi:
> Gitab lets you browse an artifact in a MR? That's really cool!

Yes! It's a REALLY cool feature because you can browse the site directly from the MR without deploying it anywhere which automatically uses the access control for the project.

> this PR won't be able to flag links that unnecessarily point to index.html - but maybe you have another approach for that already.

Not really, do you have any ideas?

thomas-zahner:
@katrinafyi Ah sorry for the confusion caused by renaming the issue. `--index-files ''` to reject all directories makes a lot of sense and is quite obvious behaviour in my opinion :+1: 

> this PR won't be able to flag links that unnecessarily point to index.html - but maybe you have another approach for that already.

True, that's something quite different that doesn't have much to do with this PR. @dkarlovi I would probably do it as follows:

```
unnecessary_links=$(lychee --dump --exclude '.*' --include '^file://.*/index\.html(#.*)?$' YOUR_DIRECTORY)
count=$(echo $unnecessary_links | wc -l)
(( $count > 0 )) && echo "Found $count links unnecessary links" && exit 1
```

lychee-lib/src/checker/file.rs
```diff
@@ -238,3 +287,195 @@ impl FileChecker {
...
+        );
+        assert_filecheck!(
+            &checker_no_indexes,
+            "filechecker/empty_dir",
+            Status::Error(InvalidIndexFile(_))
+        );
+    }
+
+    #[tokio::test]
+    async fn test_index_list_of_directories_corner() {
```
thomas-zahner, line 441: I thought about these tests for a moment now (test_index_list_of_directories_corner and  test_index_file_traversal_corner). I think we should commit to either being quite restrictive or keep the option more flexible. I'd like to suggest that we keep it flexible and remove the special handling for `.` and allow any directories instead, so that these test cases here don't result in `Status::Error(InvalidIndexFile(_))`. This should also simplify the documentation & help message somewhat.

lychee-lib/src/checker/file.rs
```diff
@@ -238,3 +287,195 @@ impl FileChecker {
...
+            "filechecker/empty_dir",
+            Status::Error(InvalidIndexFile(_))
+        );
+    }
+
+    #[tokio::test]
+    async fn test_index_file_traversal_corner() {
+        // index file names can contain path fragments and they will be traversed.
+        // being able to do this is not necessarily a good thing, and maybe it should
+        // be changed. this test just records the current behaviour.
```
thomas-zahner: Then you can update this comment that this is normal & expected behaviour.

lychee-lib/src/checker/file.rs
```diff
@@ -96,132 +112,165 @@ impl FileChecker {
...
+            .find_map(|filename| {
+                // for some special index file names, we accept directories as well
+                // as files.
+                let exists = match filename.as_str() {
+                    "." => Path::exists,
+                    _ => Path::is_file,
+                };
+
+                let path = dir_path.join(filename);
+                exists(&path).then_some(path)
```
thomas-zahner, line 225: And this would be simplified to `Path::exists(&path).then_some(path)`

README.md
thomas-zahner: Okay let's keep the flag name then. Ah true, in the current state a parameter like `..` doesn't make sense because directories are not accepted, except for `.`. But I actually think it's not necessary then to handle `.` differently than all other directories

katrinafyi:
@dkarlovi I would've suggested a shell pipeline in the same spirit as @ thomas-zahner's suggestion :) With this external approach, you'd have to be careful if you have multiple index file names.

Speculating about a future Lychee implementation of that check, it's definitely possible by building on this PR. The logic would go like:
- if we're checking a local link and its filename is within index_files, then
- apply index file resolution on the parent directory and consider the found file, then
- if the found file is equal to our original link, flag this to the user.

So it's all technically _possible_, but I think the bigger obstacle is that this is more of a lint rather than a validity check - both the directory link and index.html are usable links. Then, to show this information to the user, it could be reported as an error using the existing infrastructure, or it could be reported as a new kind of "lint" check. I don't know if these changes would be accepted into the project.

Maybe in future we will get some kind of plugin architecture for Lychee so you could add custom checks.

lychee-lib/src/checker/file.rs
thomas-zahner: You can keep them if you prefer. It's just that if I were to refactor the code in the future for some reason I personally would not keep such comments, as I think they don't add any value. There's no new insight about "why" we do this and "what" it does is obvious from the code. To me this is like a form of code duplication so I would remove all comments inside this match.

dkarlovi:
> is is more of a lint rather than a validity check 

This is a great point. :+1:

> it could be reported as a new kind of "lint" check. I don't know if these changes would be accepted into the project.

I do remember #269 and https://github.com/lycheeverse/lychee/discussions/880 where this exact topic was previously discussed: it does seem there's some interest for link linting once you go beyond validity, I'll link to your comment from the discussion.

Won't spam your PR any more.

lychee-lib/src/checker/file.rs
katrinafyi: Do you really want to commit to such unjustified generality? For... slightly shorter docs and code? I struggle to imagine a use case. To be clear, you're asking for an expansion of scope in two directions:
- Accepting any directory. This alone would make the `--index-files` name wildly inaccurate. It also raises the question of: Do you recurse and apply index file resolution to _that_ directory? Why, or why not? Why should we even have to think about this?
- Enshrining the ability to traverse arbitrary paths. Why? Not that Lychee is security-critical, but unexpected path traversal is a recurring source of vulnerabilities. I don't see why this should be a desired feature. I've repeated that I would be happy to see validation that restricts accepted values. 

There is also one technical problem with this... `--index files ''` is parsed by clap as `Some([""])`. We want this argument to reject all directory links and at the moment this works because the `''` name is excluded because it resolves to a directory. This is not insurmountable, but it is currently load bearing in this way :) Maybe I should change this.

Btw I'm sorry, this is getting really long but unfortunately I have opinions :pensive: 

update: i've added a test case to guard against a regression of `--index-files ''` https://github.com/lycheeverse/lychee/commit/dcb1a1dbea412a08044d714dff95f1f7b17ac2e9

edit: after reading about [Directory index](https://httpd.apache.org/docs/current/mod/mod_dir.html#directoryindex), I'd be okay with allowing arbitrary path traversal. but I don't think this needs to be encouraged by a docs mention or more general flag name. I still think it should be files-only aside from `.`.

I also point out that if you allow path traversal, we should also allow absolute links. this would be hideously complicated because it needs to go through the base-url and/or rootdir machinery, and so it's possible for a directory link to get resolved into a remote URL link.

edit: endorsing path traversal is done. i decided we don't care about base-url/root-dir and it will always be literal files on disk.

lychee-bin/tests/cli.rs
```diff
@@ -2288,4 +2288,98 @@ mod cli {
...
+        main_command()
+            .arg(&input)
+            .arg("--index-files")
+            .arg("index.html,.")
+            .assert()
+            .success()
+            .stdout(contains("4 OK"));
+
+        // checking fragments will accept the index_dir#fragment link,
+        // but reject empty_dir#fragment because empty_dir dens't have
```
```suggestion
        // but reject empty_dir#fragment because empty_dir doesn’t have
```

lychee-lib/src/checker/file.rs
thomas-zahner: > update: i've added a test case to guard against a regression of --index-files '' https://github.com/lycheeverse/lychee/commit/dcb1a1dbea412a08044d714dff95f1f7b17ac2e9

Thanks. The tests are great and keep the PR manageable :+1: 

> endorsing path traversal is done. i decided we don't care about base-url/root-dir and it will always be literal files on disk.

Sounds sensible to me.

> This alone would make the `--index-files` name wildly inaccurate.

True, I initially wanted to keep the option more general as discussed previously with renaming the flag. But if we stick to this use-case - as we probably should - and keep the flag name I'm okay with this.

thomas-zahner (commented):
@katrinafyi I think the PR is pretty much ready to be merged. Can you resolve the conflicts?

katrinafyi:
Has rust got a new version that's learned more compiler warnings?

Do you want to fix those in a separate PR or should I just add it here? I'm not usually a rust user, so I wouldn't be super sure of the changes.

Edit: or the third option is to rollback the CI version for the time being.

mre:
Good point. Opened https://github.com/lycheeverse/lychee/pull/1785 for that.

thomas-zahner:
@katrinafyi Yes, Clippy updates often add new lints leading to new warnings which we reject in CI. It might make sense to switch to a fixed Rust+Clippy version for CI/CD in the future. The pipeline should now pass if you rebase on top of master

mre:
From my side, this looks good! Thanks!
@thomas-zahner, feel free to merge if you're fine with it.

thomas-zahner:
@katrinafyi Thanks a lot for the PR and for bearing with my discussions :+1:

mre:
Yes, thanks @katrinafyi, you do a lot of great work recently. We all highly appreciate it.

dkarlovi:
@katrinafyi thank you from my side too, as you can probably tell I'm pretty eager to have this feature available!  :clap:

katrinafyi:
Thank you for your kind words :) I feel like it's unwarranted, because I'm really just fixing things for my own use case. I'm glad it's appreciated, though. I hope I haven't been too overbearing in your project.

ofek:
I can't get this to work when defined via the config file. Is that not supported?

katrinafyi:
@ofek Sorry, it should've been supported, but there was a bug that silently ignored certain config.toml fields and this was one of the affected fields.

The bug has been fixed but it's not released yet (https://github.com/lycheeverse/lychee/pull/1843). It is in the nightly version, so you can try that if you need.

ofek:
Great, thanks! Just to confirm, this is indeed the proper option to choose when generators like MkDocs are configured to use [directory URLs](https://www.mkdocs.org/user-guide/configuration/#use_directory_urls) right? For example I'm doing:

```
lychee site --index-files index.html
```

katrinafyi:
Yes, that use looks right. I don't know this specific generator, but other options which might be useful are --root-dir (for root-relative links) and --fallback-extensions (for resolving filename to filename.html).

PR #1774, opened 2025-07-23
Add support for 64-bit ARM binary

eread:
Add support for 64-bit ARM binary and use in Alpine Linux-based Docker images.

Fixes: https://github.com/lycheeverse/lychee/issues/1745.

PR #1778, opened 2025-07-28
Fix possible outputs of 'arch' command in Alpine Linux-based Dockerfile

eread:
A follow up to: https://github.com/lycheeverse/lychee/pull/1774.

That pull request seemed to correctly provide a new `lychee-aarch64-unknown-linux-musl.tar.gz` build (see nightly: https://github.com/lycheeverse/lychee/releases/tag/nightly), but incorrectly changed what the expected output from Alpine Linux's `arch` command was.

This pull request restores the original parsing of the output, but retains the new format of what is set in `$ARCH` for use later. It might be possible to further refine the Dockerfile, but the intention here is to make as few changes as possible each time.

### DIFF
--- a/Dockerfile-CI.alpine.Dockerfile
+++ b/Dockerfile-CI.alpine.Dockerfile
@@ -5,8 +5,8 @@ ARG LYCHEE_VERSION="latest"
 
 RUN apk add --no-cache ca-certificates jq wget \
     && ARCH=$(case $(arch) in \
-        "amd64") echo "x86_64";; \
-        "arm64") echo "aarch64";; \
+        "x86_64") echo "x86_64";; \
+        "aarch64") echo "aarch64";; \
         *) echo "Unsupported architecture" && exit 1;; \
         esac) \
     && BASE_URL=$(case $LYCHEE_VERSION in \


eread:
@mre @willnode Could you review?

The way to know this is it the correct fix is (assuming Docker is set up locally):

1. Run `docker run --rm -ti --platform "linux/arm64" alpine:latest sh`.
1. In the running container, run `arch`. You'll see `aarch64` is returned.
1. Exit that running container.
1. Run `docker run --rm -ti --platform "linux/amd64" alpine:latest sh`.
1. In the running container, run `arch`. You'll see `x86_64` is returned.

With this change, I can also locally build a Docker image with the Alpine Linux Docker file for the `linux/amd64` platform by running (assuming `buildx` is available):

```shell
docker buildx build --platform "linux/amd64" --file Dockerfile-CI.alpine.Dockerfile .
```

Building for `linux/arm64` platform locally doesn't work yet, but I think that's unrelated to the changes here and in https://github.com/lycheeverse/lychee/pull/1774. This is the command for that (assuming `buildx` is available):

```shell
docker buildx build --platform "linux/arm64" --file Dockerfile-CI.alpine.Dockerfile .
```

If we run `wget` with `-v` (verbose) instead of `-q` (quiet) within the Dockerfile, it seems like we're trying to pull both of these:

- `https://github.com/lycheeverse/lychee/releases/latest/download/lychee-aarch64-unknown-linux-musl.tar.gz` (which is found)
- `https://github.com/lycheeverse/lychee/releases/download/lychee-v0.19.1/lychee-aarch64-unknown-linux-musl.tar.gz` (which is not found, and you can imagine that's expected)

It is likely that when the next release happens, this problem will go away. I haven't really looked into if this should be the expected behavior (wanting both "latest" and "latest released").

mre:
I didn't test it locally, but given that the setup is broken in master right now and that this PR reverts to the previous behavior, I'd say it's fine to merge.

eread:
> I didn't test it locally, but given that the setup is broken in master right now and that this PR reverts to the previous behavior, I'd say it's fine to merge.

Thanks @mre. Do you think a new release is far away?

mre:
I hope to get a release out after the next batch of fixes/features which are still open. Would like to at least update the dependencies before. https://github.com/lycheeverse/lychee/pull/1779
Help wanted.

eread:
@mre @thomas-zahner I managed to figure out the problem I was having locally (https://github.com/lycheeverse/lychee/pull/1778#issuecomment-3124851141).

When `wget` tries to access `https://github.com/lycheeverse/lychee/releases/latest/download/`, it's redirected not to the `nightly` (which now has the additional build) but rather to `0.19.1`, which doesn't have the additional build yet.

If I update the Docker file to access the nightly when `LYCHEE_VERSION` is set to either `nightly` or `latest` (which is the default if `LYCHEE_VERSION` is unset), it works!

I've raised a pull request (https://github.com/lycheeverse/lychee/pull/1802) in case you want that behavior for the project.

PR #1786, opened 2025-08-08
Add snap config and GitHub Action

steinbro:
Adds a package format requested but not yet fulfilled from #59. Also adds a workflow to automatically build and publish to the [snap store](https://snapcraft.io/hey-mail). The latter requires credentials to be configured in CI variables to publish from the appropriate account.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -1,4 +1,4 @@
-name: Build and publish snap
+name: Snap
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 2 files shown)


mre:
Thanks for that awesome PR @steinbro!

Added the credentials. It's still failing because it can't find them. Not sure what's up. Do I need to merge before it can find the secrets? Probably not. It might be because it's on your fork. I checked the instructions over on `snapcore/action-publish` and I can't see any missing steps. 🤔 

In general, we should only run this workflow on new releases. We should add that condition to the pipeline (at least as a comment for now). Can you do that? You can check the other workflows for reference.

steinbro:
Snaps have a concept of "channels" for distributing e.g. nightly vs. stable builds. I've tweaked the workflow to publish to the `edge` channel for all commits to master, and to `stable` for all commits with a version tag. It should now also automatically assign a version number from the git tag as well.

I'm not sure what's going on with the build secrets, but I've also now set the job to not run at all on pull requests.

mre:
Very nice. The PR looks good to me, so I merged it. If it doesn't work, we can always tweak it later. Thanks for the contribution!

Issue #1718, opened 2025-05-27
[UX] uclear how to use --base-url, --root-dir and --remap together to get a built folder checked


dkarlovi:
Ref #1606 

Let's use this for an example:

```
mkdir -p public/assets public/download public/test1 public/test2
touch public/assets/app.css public/assets/app.js public/download/something.json public/test1/index.html
```

and the `public/index.html` file:
```html
<html>
<head>
    <link rel="stylesheet" href="/sub/dir/another/assets/app.css">
    <link rel="stylesheet" href="https://example.com/sub/dir/another/assets/app.css">
</head>
<body>

<dl>
    <dt>index</dt>
    <dd><a href="/sub/dir/another">/sub/dir/another</a></dd>
    <dd><a href="/sub/dir/another/">/sub/dir/another/</a></dd>

    <dt>file</dt>
    <dd><a href="/sub/dir/another/download/something.json">/sub/dir/another/download/something.json</a></dd>

    <dt>homepage</dt>
    <dd><a href="/sub/dir/another/test1/">/sub/dir/another/test1/</a></dd>
    <dd><a href="/sub/dir/another/test2/">/sub/dir/another/test2/</a></dd>
</dl>

</body>
</html>
```

When I run Lychee like so

```
lychee --base-url https://example.com/sub/dir/another --root-dir `pwd`/public public/index.html --offline
```

I expect these things to happen:
1. it checks everything (currently it excludes everything)
2. it finds all the links are valid, **except** `test2` because we didn't create an `index.html` file in it (I'm assuming I need to remap this?)
3. it treats URLs prefixed with `base-url` as offline in this case (because it's equivalent to the others)

Am I missing something or should this already work as is? I'm using just-built master.

katrinafyi:
Hi! For the past week I've been interrogating lychee for this use case. I also think it would be really nice to have some documentation for this use case!

In the meantime, I can offer some observations from what I've managed to piece together. Currently,
- `base-url` is used as part of relative _and_ absolute links, and
- `root-dir` is used in absolute links only.

Specifically, when encountering a relative or absolute link, Lychee constructs the target URL in this way:
```python
<a href="relative/1"> --> {parent(base-url)}/relative/1
<a href="/absolute/1"> --> {origin(base-url)}{root-dir}/absolute/1
```
where `origin` gets the origin of its argument and `parent` gets the parent of the URL if it doesn't end in `/` or returns the original URL if it does end in `/`.

[...]

PR #1787, opened 2025-08-08
clarify `--base-url` and `--root-dir` and their interactions

katrinafyi:
this more clearly documents the behaviour of these flags. this makes no changes to behavior. as always, this documents the usage as i observe right now and does not presume that this behaviour is good or correct.

i also hope to clarify some more subtle points:
- the base-url is uniformly applied to every local file being checked, regardless of the filesystem path of each local file. this means that a relative link `a/b` resolves to exactly the same URL no matter which local file it appears in.
- base-url is affected by trailing slashes.
- define behavior when root-dir and base-url are both specified. namely, root-dir becomes a subpath of the base-url's domain.

honestly, the more i think about it, the more i think this behavior is wrong or has diverged from its original intention. for instance, let's look at [the docs recipe about `--root-dir`][1]:

[1]: https://lychee.cli.rs/recipes/root-dir/#using-both-together

> Sometimes you need both:
>
> ```bash
> lychee \
>   --root-dir "$(pwd)/public" \
>   --base-url https://example.com/ \
>   "public/**/*.html"
> ```
>
> This tells lychee:
>
> 1. Look for `/`-prefixed files in `./public/`
> 2. Resolve relative links against https://example.com/

with the above points in mind, this example is nonsensical. firstly, the root-dir will make absolute links become something like `https://example.com/path/to/pwd/public/absolute-link` which is unlikely to exist on the remote host. this is in contradiction with point (1) in the quote. secondly, /every/ relative link is resolved against exactly `https://example.com/`, without considering its position relative to root-dir. this seems wrong too.

but in any case, this PR clarifies the behavior as it is right now. this PR is mostly just my findings from https://github.com/lycheeverse/lychee/issues/1718#issuecomment-3142606156

### DIFF
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -667,8 +667,9 @@ root directory to the requested absolute link. For example, with a root-dir of
 `/root/dir`, a link to `/page.html` would be resolved to `/root/dir/page.html`.
 
 This option can be specified alongside `--base-url`. If both are given, an
-absolute link is resolved as: the domain name of the base URL, then the root
-dir path, then the absolute link's path."
+absolute link is resolved by constructing a URL from three parts: the domain
+name specified in `--base-url`, followed by the `--root-dir` directory path,
+followed by the absolute link's own path."
     )]
     #[serde(default)]
     pub(crate) root_dir: Option<PathBuf>,
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -628,15 +628,48 @@ separated list of accepted status codes. This example will accept 200, 201,
     #[serde(skip)]
     pub(crate) base: Option<Base>,
 
-    /// Base URL used to resolve relative URLs during link checking
+    /// Base URL used to resolve relative URLs in local files.
     /// Example: <https://example.com>
-    #[arg(short, long, value_parser= parse_base)]
+    #[arg(
+        short,
+        long,
+        value_parser = parse_base,
+        long_help = "Base URL to use when resolving relative URLs in local files. If specified,
+relative links in local files are interpreted as being relative to the given
+base URL.
+
+For example, given a base URL of `https://example.com/dir/page`, the link `a`
+would resolve to `https://example.com/dir/a` and the link `/b` would resolve
+to `https://example.com/b`. This behavior is not affected by the filesystem
+path of the file containing these links.
+
+Note that relative URLs without a leading slash become siblings of the base
+URL. If, instead, the base URL ended in a slash, the link would become a child
+of the base URL. For example, a base URL of `https://example.com/dir/page/` and
+a link of `a` would resolve to `https://example.com/dir/page/a`.
+
+Basically, the base URL option resolves links as if the local files were hosted
+at the given base URL address."
+    )]
     #[serde(default)]
     pub(crate) base_url: Option<Base>,
 
-    /// Root path to use when checking absolute local links,
-    /// must be an absolute path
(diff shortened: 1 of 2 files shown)


lychee-bin/src/options.rs
```diff
@@ -628,15 +628,48 @@ separated list of accepted status codes. This example will accept 200, 201,
...
+Basically, the base URL option resolves links as if the local files were hosted
+at the given base URL address."
+    )]
     #[serde(default)]
     pub(crate) base_url: Option<Base>,
 
-    /// Root path to use when checking absolute local links,
-    /// must be an absolute path
-    #[arg(long)]
+    /// Root path to use when checking absolute links in local files.
```
mre: Should we use the opportunity to make this change as well?

```suggestion
    /// Root directory to use when checking absolute links in local files.
```

(Same for `long_help`.)

lychee-bin/src/options.rs
```diff
@@ -628,15 +628,48 @@ separated list of accepted status codes. This example will accept 200, 201,
...
+required if absolute links appear in local files, otherwise those links will be
+flagged as errors. This must be an absolute path (i.e., one beginning with `/`).
+
+If specified, absolute links in local files are resolved by prefixing the given
+root directory to the requested absolute link. For example, with a root-dir of
+`/root/dir`, a link to `/page.html` would be resolved to `/root/dir/page.html`.
+
+This option can be specified alongside `--base-url`. If both are given, an
+absolute link is resolved as: the domain name of the base URL, then the root
+dir path, then the absolute link's path."
```
mre: That confused me a bit, mostly because the two parameters get confusing when used in combination. Is there an easier way to explain that? I can't come up with a good alternative right now, but maybe you have an idea. I liked your examples in the help message, so maybe another example would help or an ordered list of steps?

mre:
It is amazing that we start to properly document these options.
Added some comments, but your description is already much better than anything I ever came up with.

lychee-bin/src/options.rs
dkarlovi: > because the two parameters get confusing when used in combination

You can say that again #1718 :smile:

katrinafyi: > Is there an easier way to explain that? 

I'm not sure. as some background, in #1718, I wrote
```
<a href="relative/1"> --> {parent(base-url)}/relative/1
<a href="/absolute/1"> --> {origin(base-url)}{root-dir}/absolute/1
```
which I think is pretty clear. what you see in this PR is an attempt to translate that into prose.

it does lose some formality, but i worry that if we make it too explicit, it starts becoming more of a formal specification. this is not bad (indeed, it's good to be precise about these things) but I think help text isn't the right place for such formality. I want to make sure it doesn't get too overwhelming.

but anyway, maybe we can make it clearer with some phrasing changes:
```suggestion
absolute link is resolved to: the
domain name of --base-url, joined
with --root-dir, then joined with the
absolute link's path."
```
what're your thoughts?

mre: Yes, I like that much better. What do you think of this version?

```suggestion
absolute links are resolved by constructing a URL from three parts:
the domain name specified in `--base-url`, followed by the `--root-dir`
directory path, followed by the absolute link's own path."
```

mre:
Not sure why we kept this open for so long, sorry. It was ready to be merged earlier. 😅 Anyhow, it's merged now.

katrinafyi:
No worries, thanks! I think i just forgot to reply that I'd fixed the review comments :)

PR #1792, opened 2025-08-10
README: Cleanup input example

tooomm:
`--verbose` command in one of the examples on input options looks like a leftover


mre (approved):
Thanks!

PR #1793, opened 2025-08-10
Fix missing identifier in snap build action; add snap install to readme

steinbro:
Quick follow-up to #1786 since I saw the workflow isn't passing yet.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -13,6 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: snapcore/action-build@v1
+        id: build
       - uses: snapcore/action-publish@v1
         env:
           SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
--- a/README.md
+++ b/README.md
@@ -55,6 +55,12 @@ pacman -S lychee
 zypper in lychee
 ```
 
+### Ubuntu
+
+```sh
+snap install lychee
+```
+
 ### macOS
 
 Via [Homebrew](https://brew.sh):


mre (approved):
Thanks. I don't know how it would make the build work, but it's certainly great to update the docs.

steinbro:
The reason this fixed the error from the earlier PR is because the final publishing step uses the value of `steps.build.outputs.snap`, but there wasn't any step named `build`. You're now getting a [new error](https://github.com/lycheeverse/lychee/actions/runs/16866982712/job/47775226425) coming from the snap store:
```
Store operation failed:
- resource-forbidden: Cannot upload new revisions for name=lychee
```
Maybe you need to register the name under your account before you can upload packages?

mre:
Aha! That might be the case. I'll try that if I get the chance.

mre:
It would not let me register:

```
> snapcraft register lychee

We always want to ensure that users get the software they expect
for a particular name.

If needed, we will rename snaps to ensure that a particular name
reflects the software most widely expected by our community.

For example, most people would expect 'thunderbird' to be published by
Mozilla. They would also expect to be able to get other snaps of
Thunderbird as '$username-thunderbird'.

Would you say that MOST users will expect 'lychee' to come from
you, and be the software you intend to publish there? [y/N]: y
Registering lychee.
The name 'lychee' is already taken.
```

Can't find a package with that name through the snap search. Not sure what's going on. I reached out to the snapcraft team.

mre:
Update: it worked now! 🥳 

The fine folks at snapcraft already replied and we have the name now. The [upload worked](https://github.com/lycheeverse/lychee/actions/runs/16891823886/job/47855589990). Can you try to install it on your end?

steinbro:
It does install, though I forgot to enable privileges to let the app access files and the network. Fixed in #1796.

Also, only the unstable build was pushed; you may want to manually trigger a build for the latest stable release (you should be able to run the workflow against an arbitrary branch/tag from the [Actions panel](https://github.com/lycheeverse/lychee/actions/workflows/snap.yml)). Until there's a build on the stable channel, users will see the following when installing:
```
$ snap install lychee
error: snap "lychee" is not available on stable but is available to install on
       the following channels:

       edge       snap install --edge lychee

       Please be mindful pre-release channels may include features not
       completely tested or implemented. Get more information with 'snap info
       lychee'.
```

mre:
Right, thanks! Merged your PR. I just wanted to wait for the stable release until we ran a test. I can't test it myself right now, but if you let me know that it works, I can then go and push the stable release.

steinbro:
It works :)
```
$ snap install --edge lychee
lychee (edge) 0.19.1.dev from Matthias Endler (m-r-e) installed
$ lychee lychee/README.md 
   [WARN ] Error creating request: InvalidPathToUri("/CONTRIBUTING.md")
  111/111 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                        
🔍 111 Total (in 1s) ✅ 111 OK 🚫 0 Errors
```

mre:
Very nice! Thanks for the feedback!

PR #1799, opened 2025-08-14
Improve wikilink parsing and checking

JayJayArr:
- [x] Extract the correct filename from a Wikilink by removing potholes and fragments
- [x] Create a new Checker for wikilinks, make it clear to the users that a base-url is required
  - [x] Check Obsidian style wikilinks without headers
  - [x] strip fragments & potholes from wikilinks
  - [x] traverse the wiki-directory and check if any of the Filenames exist with any of the specified `--fallback-extensions`

### DIFF
--- a/lychee-lib/src/checker/wikilink/index.rs
+++ b/lychee-lib/src/checker/wikilink/index.rs
@@ -44,8 +44,10 @@ impl WikilinkIndex {
             .filter_map(Result::ok)
         {
             if let Some(filename) = entry.path().file_name() {
-                let mut lock = self.filenames.lock().unwrap();
-                lock.insert(filename.to_ascii_lowercase(), entry.path().to_path_buf());
+                self.filenames
+                    .lock()
+                    .unwrap()
+                    .insert(filename.to_os_string(), entry.path().to_path_buf());
             }
         }
     }
@@ -53,16 +55,10 @@ impl WikilinkIndex {
     /// Checks the index for a filename. Returning the absolute path if the name is found,
     /// otherwise returning None
     pub(crate) fn contains_path(&self, path: &Path) -> Option<PathBuf> {
-        match path.file_name() {
-            None => None,
-            Some(filename) => {
-                let filename_lock = self.filenames.lock().unwrap();
-                if filename_lock.contains_key(&filename.to_ascii_lowercase()) {
-                    filename_lock.get(&filename.to_ascii_lowercase()).cloned()
-                } else {
-                    None
-                }
-            }
-        }
+        self.filenames
+            .lock()
+            .unwrap()
+            .get(path.file_name()?)
+            .cloned()
     }
 }


JayJayArr:
@jrfnl I am having trouble understanding the MediaWiki links with `()` and `:`. 
Are there really markdown files named `Help:Links` or where do they point to?
Is `[[Extension:DynamicPageList (Wikimedia)]]` different to `[[Extension:DynamicPageList]]`

Could you please clarify this?

mre:
Guess you can rebase on top of `master` now to fix CI.

lychee-bin/tests/cli.rs
```diff
@@ -2941,4 +2941,58 @@ mod cli {
             .success()
             .stdout(contains("https://example.org")); // Should extract the link as plaintext
     }
```
mre, line 3049: Missing empty line

lychee-bin/tests/cli.rs
```diff
@@ -2941,4 +2941,58 @@ mod cli {
...
+            .arg(fixtures_path())
+            .assert()
+            .success();
+    }
+
+    #[test]
+    fn test_wikilink_fixture_with_fragments_obsidian_style_fixtures_excluded() {
+        let input = fixtures_path().join("wiki/obsidian-style-plus-headers.md");
+
+        //fragments should resolve all headers
```
mre: It's not a big deal, but I see that some comments are missing a space between `//` and the first word. Maybe we can be consistent here. 😉

lychee-lib/src/checker/file.rs
```diff
@@ -73,6 +81,10 @@ impl FileChecker {
     ///
     /// Returns a `Status` indicating the result of the check.
     pub(crate) async fn check(&self, uri: &Uri) -> Status {
+        //only populate the wikilink filenames if it is enabled
```
mre:
```suggestion
        //only populate the wikilink filenames if the feature is enabled
```

or maybe "if support is enabled"

lychee-lib/src/checker/file.rs
```diff
@@ -246,7 +262,7 @@ impl FileChecker {
     }
 
     /// Checks a resolved file, optionally verifying fragments for HTML files.
-    ///
+    ///u
```
mre: typo


```suggestion
    ///
```

mre: Are you a VIM user? 😉

lychee-lib/src/checker/file.rs
```diff
@@ -304,6 +320,25 @@ impl FileChecker {
             }
         }
     }
+
+    // Initializes the Index of the wikilink checker
```
mre:
```suggestion
    // Initializes the index of the wikilink checker
```

lychee-lib/src/checker/file.rs
```diff
@@ -304,6 +320,25 @@ impl FileChecker {
...
+    fn setup_wikilinks(&self) {
```
mre: I was wondering if we should explicitly call this `setup_wikilinks_index` or so to make the meaning clear.

lychee-lib/src/checker/file.rs
```diff
@@ -304,6 +320,25 @@ impl FileChecker {
...
+        self.wikilink_checker.index_files();
+    }
```
mre: Missing empty line

lychee-lib/src/checker/file.rs
```diff
@@ -304,6 +320,25 @@ impl FileChecker {
...
+    // Tries to resolve a link by looking up the filename in the wikilink index
+    // The
```
mre:
```suggestion

```

lychee-lib/src/checker/file.rs
```diff
@@ -304,6 +320,25 @@ impl FileChecker {
...
+    fn apply_wikilink_check(&self, path: &Path, uri: &Uri) -> Result<PathBuf, ErrorKind> {
+        let mut path_buf = path.to_path_buf();
+        for ext in &self.fallback_extensions {
+            path_buf.set_extension(ext);
+            match self.wikilink_checker.check(&path_buf, uri) {
+                Err(_) => {}
```
mre: We could add a `trace!` log here. Might be helpful for troubleshooting later on. For example, we could tell people to set `RUST_LOG=trace lychee` to send us more details. 

```suggestion
                Err(_) => { trace!("Tried to find wikilink at {path_buf}") }
```

lychee-lib/src/checker/file.rs
```diff
@@ -363,7 +398,7 @@ mod tests {
     #[tokio::test]
     async fn test_default() {
         // default behaviour accepts dir links as long as the directory exists.
-        let checker = FileChecker::new(None, vec![], None, true);
+        let checker = FileChecker::new(None, vec![], None, true, false);
```
mre: Eventually we'll have to create a proper `checker::Config` or `FileCheckerOptions` struct which holds all of these settings. 🙈 It's hard to tell what the params stand for by now. Oh well, maybe in a follow-up PR. 😆

lychee-lib/src/extract/markdown.rs
```diff
@@ -92,11 +92,29 @@ pub(crate) fn extract_markdown(
...
-                        Some(vec![RawUri {
-                            text: dest_url.to_string(),
-                            element: Some("a".to_string()),
-                            attribute: Some("href".to_string()),
-                        }])
+
+                        //Strip potholes (|) from wikilinks
+                        let mut stripped_dest_url = if has_pothole {
+                            pulldown_cmark::CowStr::Borrowed(&dest_url[0..dest_url.find('|').unwrap_or(dest_url.len())])
+                        }else {
```
mre:
```suggestion
                        } else {
```

`cargo fmt` should catch this I guess

lychee-lib/src/extract/markdown.rs
```diff
@@ -92,11 +92,29 @@ pub(crate) fn extract_markdown(
...
                         }
-                        Some(vec![RawUri {
-                            text: dest_url.to_string(),
-                            element: Some("a".to_string()),
-                            attribute: Some("href".to_string()),
-                        }])
+
+                        //Strip potholes (|) from wikilinks
+                        let mut stripped_dest_url = if has_pothole {
+                            pulldown_cmark::CowStr::Borrowed(&dest_url[0..dest_url.find('|').unwrap_or(dest_url.len())])
```
mre: I *think* this works as expected, but I'm not sure.

For `"|foo|"`, this would produce `""`.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=510ecf9db2234372843723643edc882b

Maybe you turn it into a function and add some unit tests. `rstest` is pretty nice for that. We use it in some other places for the tests already.

lychee-lib/src/extract/markdown.rs
```diff
@@ -92,11 +92,29 @@ pub(crate) fn extract_markdown(
...
+                        }else {
+                            dest_url.clone()
+                        };
+
+                        //Strip fragments (#) from wikilinks, according to the obsidian spec
+                        //fragments come before potholes
+                        if stripped_dest_url.contains('#') {
+                            stripped_dest_url = pulldown_cmark::CowStr::Borrowed(&dest_url[0..dest_url.find('#').unwrap_or(dest_url.len())]);
+                        }
```
mre: Same here, could be a function.

lychee-lib/src/utils/wikilink_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+use log::info;
+use std::collections::HashMap;
+use std::ffi::OsString;
+use std::path::Path;
+use std::sync::Mutex;
+use std::{path::PathBuf, sync::Arc};
+use walkdir::WalkDir;
+
+#[derive(Clone, Debug, Default)]
+// Indexes a given directory for filenames and the corresponding path
```
mre: This doc-string could be a bit clearer. I was not sure what you meant by corresponding path.

And while we're at it, maybe we can change the order to align with the rest:

```rust
/// Doc-string here
#[derive(Clone, Debug, Default)]
pub(crate) struct WikilinkChecker {
```

(Note that I used `///` here to make it show up in rust-doc.)

lychee-lib/src/utils/wikilink_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl WikilinkChecker {
+    pub(crate) fn new(base: Option<Base>) -> Self {
+        Self {
+            filenames: Arc::new(Mutex::new(HashMap::new())),
+            basedir: base,
+        }
+    }
+
+    pub(crate) fn index_files(&self) {
```
mre: Could use a doc-comment

lychee-lib/src/utils/wikilink_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+    filenames: Arc<Mutex<HashMap<OsString, PathBuf>>>,
+    basedir: Option<Base>,
+}
+
+impl WikilinkChecker {
+    pub(crate) fn new(base: Option<Base>) -> Self {
+        Self {
+            filenames: Arc::new(Mutex::new(HashMap::new())),
+            basedir: base,
+        }
```
mre: If you like the syntax, you can also write


```suggestion
        Self {
            basedir: base,
            ..default::Default()
        }
```

but that's a matter of preference. Just wasn't sure if you knew about it.

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::path::Path;
+use std::sync::Mutex;
+use std::{path::PathBuf, sync::Arc};
+use walkdir::WalkDir;
+
+#[derive(Clone, Debug, Default)]
+// Indexes a given directory for filenames and the corresponding path
+pub(crate) struct WikilinkChecker {
+    filenames: Arc<Mutex<HashMap<OsString, PathBuf>>>,
+    basedir: Option<Base>,
```
mre: Is the basedir really optional? Further down, you abort file indexing if it's not set.

lychee-lib/src/utils/wikilink_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+            }
+            Some(ref basetype) => match basetype {
+                Base::Local(localbasename) => {
+                    //Start file indexing only if the Base is valid and local
+                    info!(
+                        "Starting file indexing for wikilinks in {}",
+                        localbasename.display()
+                    );
+
+                    let mut filenameslock = self.filenames.lock().unwrap();
```
mre: No harm in returning a `Result` to get rid of the `unwrap`.
Also, I would rename the variables to use underscores more, like `filenames_lock` (or just `lock` in this case).

lychee-lib/src/utils/wikilink_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+                        .filter_map(std::result::Result::ok)
+                    {
+                        if let Some(filename) = entry.path().file_name() {
+                            filenameslock
+                                .insert(filename.to_ascii_lowercase(), entry.path().to_path_buf());
+                        }
+                    }
+                }
+                // A remote base is of no use for the wikilink checker, silently skip over it
+                Base::Remote(_remotebasename) => {}
```
mre: Maybe we could print an info message here

lychee-lib/src/utils/wikilink_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+            },
+        }
+    }
+
+    pub(crate) fn check(&self, path: &Path, uri: &Uri) -> Result<PathBuf, ErrorKind> {
```
mre: Could use a doc-string

And how about changing it to `contains_path(&self, path: &Path) -> bool`?
I think it should only take a `path`, because it only uses the `uri` for error creation, and that could happen on the call-site. As it is now, this method is only concerned about checking the index and not about error handling.

mre (commented):
Good work! I've added a bunch of comments, but don't worry.
Most of them are about very minor paper cuts. Most of it can be fixed pretty easily by clicking on the "accept suggestion" button, I hope. 😉

lychee-lib/src/checker/file.rs
JayJayArr: Guilty as charged :laughing:

lychee-lib/src/utils/wikilink_checker.rs
JayJayArr: Thanks for pointing that out! Learned something new here

lychee-lib/src/checker/file.rs
JayJayArr: Sounds like a good idea, its getting kinda confusing :sweat_smile:

lychee-lib/src/extract/markdown.rs
JayJayArr: Thanks for the Feedback, I am a little unsure about this though.
`"|foo|"` is not a valid wikilink. We check further down if the string is left empty, am I missing something here?

lychee-lib/src/utils/wikilink_checker.rs
JayJayArr: Sounds good, I'll opt for an Option here to be able to return the resolved Pathbuf

lychee-lib/src/checker/wikilink/index.rs
JayJayArr: Well, the argument `--base-url` is returned as an `Option` from clap but it is actually required here. I'll have a look if it can be made mandatory when specifying `--include-wikilinks`

lychee-lib/src/utils/wikilink_checker.rs
JayJayArr: I got it working by adding a new ErrorKind, but please let me know if you have another idea that you would prefer.

lychee-lib/src/utils/wikilink_checker.rs
mre: Yeah that's fine as well.

lychee-lib/src/extract/markdown.rs
mre: We could make that an invariant enforced on creation. For example, you could have a `clean_wikilink` or `remove_potholes` method, which returns a Result and throws an error if the final string is empty. You could also add a `WikiLink` type which has a `TryFrom<&str> for WikiLink`. That's a powerful pattern we use a lot.

lychee-lib/src/checker/wikilink/index.rs
mre: Then the base-dir should be checked on object creation. That means the WikiLinkChecker can only be created with a valid base-dir.

JayJayArr: Good point, implemented it as an Option, thanks for the patience!

mre:
@JayJayArr, any chance you could fix the merge conflicts? I would love to get this great work merged. 😊

JayJayArr:
There we go. Lets call it okay for now :sweat_smile:

lychee-bin/src/options.rs
```diff
@@ -851,7 +851,8 @@ and existing cookies will be updated."
     pub(crate) cookie_jar: Option<PathBuf>,
 
     #[allow(clippy::doc_markdown)]
-    /// Check WikiLinks in Markdown files
+    /// Check WikiLinks in Markdown files, this requires specifying --base-url
+    #[clap(requires = "base_url")]
```
mre, line 895: Nice! I had no clue that this exists.

lychee-lib/src/checker/file.rs
```diff
@@ -32,8 +33,12 @@ pub(crate) struct FileChecker {
     index_files: Option<Vec<String>>,
     /// Whether to check for the existence of fragments (e.g., `#section-id`) in HTML files.
     include_fragments: bool,
+    /// Whether to check for the existence of files linked to by Wikilinks
+    include_wikilinks: bool,
     /// Utility for performing fragment checks in HTML files.
     fragment_checker: FragmentChecker,
+    /// Utility for checking wikilinks, indexes files in a given directory
```
mre: A nitpick, but why is "wikilinks" written in lowercase here? Should it be uppercase?

### 
```suggestion
    /// Utility for checking Wikilinks, indexes files in a given directory
```

Same for the other occurrences. Sometimes it's uppercase and sometimes lowercase. We should decide on one way to spell it. Looks like the official spelling is "WikiLinks."

lychee-lib/src/checker/file.rs
```diff
@@ -73,6 +81,13 @@ impl FileChecker {
     ///
     /// Returns a `Status` indicating the result of the check.
     pub(crate) async fn check(&self, uri: &Uri) -> Status {
+        // only populate the wikilink filenames if the feature is enabled
+        if self.include_wikilinks {
+            match self.setup_wikilinks() {
+                Ok(()) => (),
+                Err(e) => return Status::Error(e),
+            }
+        }
```
mre: Here's an alternative way to write this:

```suggestion
        if self.include_wikilinks {
            if let Err(e) = self.setup_wikilinks() {
                return Status::Error(e);
            }
        }
```

lychee-lib/src/checker/file.rs
```diff
@@ -134,8 +149,12 @@ impl FileChecker {
     ) -> Result<Cow<'a, Path>, ErrorKind> {
         let path = match path.metadata() {
             // for non-existing paths, attempt fallback extensions
+            // if fallback extensions don't help, try wikilinks
             Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
-                self.apply_fallback_extensions(path, uri).map(Cow::Owned)
+                match self.apply_fallback_extensions(path, uri).map(Cow::Owned) {
+                    Ok(val) => Ok(val),
+                    Err(_) => self.apply_wikilink_check(path, uri).map(Cow::Owned),
```
mre: The `Ok(val) => Ok(val)` looked a bit noisy. What do you think about this?

```suggestion
            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
                self.apply_fallback_extensions(path, uri)
                    .or_else(|_| self.apply_wikilink_check(path, uri))
                    .map(Cow::Owned)
            }
```

lychee-lib/src/checker/file.rs
```diff
@@ -313,6 +332,34 @@ impl FileChecker {
...
+
+    // Initializes the index of the wikilink checker
+    fn setup_wikilinks(&self) -> Result<(), ErrorKind> {
+        match &self.wikilink_checker {
+            Some(checker) => checker.setup_wikilinks_index(),
+            None => Err(ErrorKind::WikilinkCheckerInit(
+                "Initialization failed, no checker instantiated".to_string(),
+            )),
+        }
+    }
```
mre: You can use a let-else

```rust
/// Initializes the index of the wikilink checker
fn setup_wikilinks(&self) -> Result<(), ErrorKind> {
    let Some(checker) = &self.wikilink_checker else {
        return Err(ErrorKind::WikilinkCheckerInit(
            "Initialization failed, no checker instantiated".to_string(),
        ));
    };
    checker.setup_wikilinks_index()
}
```

I also changed the comment into a doc-comment, which is helpful even for non-public methods.

mre: Thinking a little more about it, I'm not sure if it needs to return an error at all. Maybe we can just silently ignore the error or return a bool?

lychee-lib/src/checker/file.rs
```diff
@@ -313,6 +332,34 @@ impl FileChecker {
...
+                    None => {
+                        trace!("Tried to find wikilink {} at {}", uri, path_buf.display());
+                    }
+                    Some(resolved_path) => return Ok(resolved_path),
+                }
+            }
+        }
+
+        Err(ErrorKind::InvalidFilePath(uri.clone()))
+    }
```
mre: Sorry, I still have a bunch of issues with that code. 😅 

- I don't love the method name because the method does more than just applying the check. It looks like it resolves the path.
- The `if let Some(checker)` is inside the loop but the checker doesn't change
- The match on `contains_path` with `None => trace!()` is awkward
- You're iterating extensions but modifying the same path_buf

Here's an attempt to fix the above:

```rust
/// Resolves a wikilink reference to its actual file path.
///
/// Iterates through configured fallback extensions (e.g., `.md`, `.txt`) and checks
/// the wikilink index for a matching file. Returns the first match found.
///
/// # Errors
///
/// Returns `ErrorKind::InvalidFilePath` if no matching file is found in the index.
fn resolve_wikilink(&self, path: &Path, uri: &Uri) -> Result<PathBuf, ErrorKind> {
    let Some(checker) = &self.wikilink_checker else {
        return Err(ErrorKind::InvalidFilePath(uri.clone()));
    };

    for ext in &self.fallback_extensions {
        let mut candidate = path.to_path_buf();
        candidate.set_extension(ext);
        
        if let Some(resolved) = checker.contains_path(&candidate) {
            return Ok(resolved);
        }
        trace!("Wikilink not found: {} at {}", uri, candidate.display());
    }
    
    Err(ErrorKind::InvalidFilePath(uri.clone()))
}
```

There are a few more issues. For example, if the wikilink_checker is None, we should return a different type of error. Or ideally, we should not even be able to call that method on it.

If wikilink_checker can be None, you're doing defensive checks everywhere instead of encoding the constraint in the type system. I'm thinking that what we really want is a `WikiLinkResolver`. Then this will become nicer. 😉 

Here's what I mean:

```rust
struct WikilinkResolver {
    checker: WikilinkChecker,
    fallback_extensions: Vec<String>,
}

impl WikilinkResolver {
    /// Resolves a wikilink by searching the index with fallback extensions.
    fn resolve(&self, path: &Path, uri: &Uri) -> Result<PathBuf, ErrorKind> {
        for ext in &self.fallback_extensions {
            let mut candidate = path.to_path_buf();
            candidate.set_extension(ext);
            
            if let Some(resolved) = self.checker.contains_path(&candidate) {
                return Ok(resolved);
            }
            trace!("Wikilink not found: {} at {}", uri, candidate.display());
        }
        
        Err(ErrorKind::WikilinkNotFound(uri.clone()))
    }
}
```

Then we don't have to deal with the `None` check.

lychee-lib/src/checker/wikilink/index.rs
mre, line 1: At the moment, the wikilink functionality is scattered across multiple places and would be much cleaner as its own module.

Just one idea:

```
lychee-lib/src/checker/
├── mod.rs
├── file.rs
└── wikilink/
    ├── mod.rs
    ├── resolver.rs    # WikilinkResolver - the main resolution logic
    └── index.rs       # WikilinkIndex - the filename indexing data structure
```

This has the added advantage that we also have all the tests in one place and can write a proper module documentation.

We don't have to do this right away. It's okay if we keep it as is and only fix the other comments. But feel free to move stuff around as well. 😄

lychee-lib/src/checker/file.rs
JayJayArr: The two cases where this could fail is either because of 
a. a poisoned `Mutex` or 
b. the user given `base-url` is not local. 

I just realized, that the `FileChecker` itself is never called if the `base-url` is `Remote` so it would be a viable option to return a bool. 
Just out of curiosity: Do you see any downsides in returning an Error except cluttering up the `ErrorKinds`?

lychee-lib/src/checker/wikilink/index.rs
JayJayArr: I REALLY appreciate this remark! :heart:  
Right now I am still a little afraid to screw around too much in other peoples projects :sweat_smile: 
But I will have another look into it when I have the time to attack the `WikilinkResolver`

JayJayArr:
Apparently archive.org has some problems right now which breaks CI

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,89 @@
...
+use std::collections::HashMap;
+use std::ffi::OsString;
+use std::path::Path;
+use std::sync::Mutex;
+use std::{path::PathBuf, sync::Arc};
+use walkdir::WalkDir;
+
+/// Indexes a given directory mapping filenames to their corresponding path.
+///
+/// The `WikilinkIndex` Recursively checks all subdirectories of the given
```
mre:
```suggestion
/// The `WikilinkIndex` recursively checks all subdirectories of the given
```

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,89 @@
...
+            basedir: base,
+            ..Default::default()
+        };
+        index.start_indexing();
+        index
+    }
+
+    /// Populates the index of the `WikilinkIndex` on startup
+    ///
+    /// Recursively walks the base directory mapping each filename to an absolute filepath.
```
mre:
```suggestion
    /// Recursively walks the base directory, mapping each filename to an absolute filepath.
```

lychee-lib/src/checker/wikilink/index.rs
mre: Sorry, I meant


```suggestion
    basedir: Base,
```

and the constructor would take a `base: Base`.

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Populates the index of the `WikilinkIndex` on startup
+    ///
+    /// Recursively walks the base directory, mapping each filename to an absolute filepath.
+    /// The Index stays empty if no base directory is supplied or if the base directory is remote
+    pub(crate) fn start_indexing(&self) {
+        match &self.basedir {
+            None => {
+                // The Empty Index returns no results in this case
+                trace!("File indexing for Wikilinks aborted as no base directory is specified");
+            }
```
mre: This check will go away once we change `basedir: Option<Base>,` to `basedir: Base,` in the struct. 😊

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,89 @@
...
+            Some(base_type) => match base_type {
+                Base::Local(local_base_name) => {
+                    // Start file indexing only if the Base is valid and local
+                    info!(
+                        "Starting file indexing for wikilinks in {}",
+                        local_base_name.display()
+                    );
+
+                    let mut lock = self.filenames.lock().unwrap();
```
mre: It looks like you don't need to hold the lock for so long. You can move that directly above this line: 

```rust
lock.insert(filename.to_ascii_lowercase(), entry.path().to_path_buf());
```

(In general, try to hold the lock as briefly as possible to avoid lock-contention.)

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,89 @@
...
+                    for entry in WalkDir::new::<PathBuf>(local_base_name.into())
+                        // actively ignore symlinks
+                        .follow_links(false)
+                        .into_iter()
+                        .filter_map(std::result::Result::ok)
```
mre:
```suggestion
                        .filter_map(Result::ok)
```

mre: Have you tried that? I could be wrong here, depending on what `Result` points to in that context.

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,89 @@
...
+                        if let Some(filename) = entry.path().file_name() {
+                            lock.insert(filename.to_ascii_lowercase(), entry.path().to_path_buf());
+                        }
+                    }
+                }
+
+                // A remote base is of no use for the wikilink checker, return an error to the user
+                Base::Remote(remote_base_name) => {
+                    warn!("Error using remote base url for checking wililinks: {remote_base_name}");
+                }
```
mre: I think we should check that in `new` and return an error.

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,89 @@
...
+            },
+        }
+    }
+    /// Checks the index for a filename. Returning the absolute path if the name is found,
```
mre, line 55:
```suggestion
    }
    
    /// Checks the index for a filename. Returning the absolute path if the name is found,
```

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,90 @@
...
+            None => None,
+            Some(filename) => {
+                let filename_lock = self.filenames.lock().unwrap();
+                if filename_lock.contains_key(&filename.to_ascii_lowercase()) {
+                    Some(
+                        filename_lock.get(&filename.to_ascii_lowercase()).expect(
+                            "Could not retrieve inserted Path for discovered Wikilink-Path",
+                        ),
+                    )
+                    .cloned()
```
mre: That's a bit clunky. `expect` would panic here. I believe we should convert that to an `Option`:


```suggestion
                        filename_lock.get(&filename.to_ascii_lowercase()).ok(),
                    )
                    .cloned()
```

lychee-lib/src/checker/wikilink/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! `WikiLink` Module
+//!
+//! This module contains a Indexer and a Resolver for `WikiLinks`
```
mre, line 3: Can you elaborate a bit on what Indexer and Resolver means in this context? What are their responsibilities? How do they differ?

lychee-lib/src/checker/wikilink/resolver.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Clone, Debug)]
+pub(crate) struct WikilinkResolver {
+    checker: WikilinkIndex,
+    fallback_extensions: Vec<String>,
+}
+
+/// Tries to resolve a `WikiLink` by searching for the filename in the `WikilinkIndex`
+/// Returns the path of the found file if found, otherwise an Error
+impl WikilinkResolver {
+    pub(crate) fn new(basedir: Option<Base>, fallback_extensions: Vec<String>) -> Self {
```
mre: This also won't be optional anymore, once you refactor the `WikilinkIndex`. E.g.:


```suggestion
    pub(crate) fn new(basedir: Base, fallback_extensions: Vec<String>) -> Self {
```

lychee-lib/src/checker/wikilink/resolver.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// Resolves a wikilink by searching the index with fallback extensions.
+    pub(crate) fn resolve(&self, path: &Path, uri: &Uri) -> Result<PathBuf, ErrorKind> {
+        for ext in &self.fallback_extensions {
+            let mut candidate = path.to_path_buf();
+            candidate.set_extension(ext);
+
+            if let Some(resolved) = self.checker.contains_path(&candidate) {
+                return Ok(resolved);
+            }
+            trace!("Wikilink not found: {} at {}", uri, candidate.display());
```
mre:
```suggestion
            trace!("Wikilink {uri} not found at {candiate}", uri, candidate=candidate.display());
```

lychee-lib/src/extract/markdown.rs
```diff
@@ -311,6 +323,26 @@ pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
     out
 }
 
+fn clean_wikilink(input: &str, has_pothole: bool) -> Result<CowStr<'_>, ErrorKind> {
```
mre: Can we also move this method into our new wikilink module? Then we have everything in one place. You might want to even call it `wikilink::Wikilink` or so?

lychee-lib/src/extract/markdown.rs
```diff
@@ -311,6 +323,26 @@ pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
...
+    // Strip potholes (|) from wikilinks
+    let mut stripped_input = if has_pothole {
+        pulldown_cmark::CowStr::Borrowed(&input[0..input.find('|').unwrap_or(input.len())])
+    } else {
+        pulldown_cmark::CowStr::Borrowed(input)
+    };
+
+    // Strip fragments (#) from wikilinks, according to the obsidian spec
```
mre: Shall we link the Obsidian Spec here?

lychee-lib/src/extract/markdown.rs
```diff
@@ -311,6 +323,26 @@ pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
...
+    // fragments always come before potholes
+    if stripped_input.contains('#') {
```
mre: We should make `#` a constant. We use it here and in the line below. I actually thought we already defined that constant somewhere? If not, we should do that. 😅

mre:
Sorry, I've added a few more comments. But we're getting super close to finishing this work now. Hang in there. 🤩

lychee-lib/src/extract/markdown.rs
JayJayArr: I could find some references to `#`, mostly used for file comments, created a new one

lychee-lib/src/types/error.rs
```diff
@@ -178,6 +178,14 @@ pub enum ErrorKind {
         /// The reason the command failed
         reason: String,
     },
+
+    /// The extracted `WikiLink` could not be found by searching the directory
+    #[error("Could not find Wikilink by searching th given base directory")]
+    WikilinkNotFound(Uri),
```
thomas-zahner: Remove the `trace!` message in the resolver and instead make this error variant equally informative and useful.

```suggestion
    #[error("Wikilink {0} not found at {1}")]
    WikilinkNotFound(Uri, PathBuf),
```

Maybe `PathBuf` needs to be handled slightly differently if it doesn't implement `Display`.

lychee-lib/src/checker/wikilink/resolver.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Resolves a wikilink by searching the index with fallback extensions.
+    pub(crate) fn resolve(&self, path: &Path, uri: &Uri) -> Result<PathBuf, ErrorKind> {
+        for ext in &self.fallback_extensions {
+            let mut candidate = path.to_path_buf();
+            candidate.set_extension(ext);
+
+            if let Some(resolved) = self.checker.contains_path(&candidate) {
+                return Ok(resolved);
+            }
+            trace!(
```
thomas-zahner: Remove this and update `WikilinkNotFound` as mentioned in the other comment.

lychee-lib/src/checker/wikilink/index.rs
```diff
@@ -0,0 +1,80 @@
...
+            filenames: Arc::new(Mutex::new(HashMap::new())),
+        };
+        index.start_indexing();
+        index
+    }
+
+    /// Populates the index of the `WikilinkIndex` on startup
+    ///
+    /// Recursively walks the local base directory, mapping each filename to an absolute filepath.
+    /// The Index stays empty if no base directory is supplied or if the base directory is remote
```
thomas-zahner: > The Index stays empty if no base directory is supplied

How is that possible? Was it an `Option<Base>` in the past and you forgot to update the comment?

lychee-lib/src/extract/markdown.rs
```diff
@@ -97,8 +99,18 @@ pub(crate) fn extract_markdown(
...
+                        if let Ok(wikilink) = wikilink(&dest_url, has_pothole) {
+                            Some(vec![RawUri {
+                                text: wikilink.to_string(),
+                                element: Some("a".to_string()),
+                                attribute: Some("wikilink".to_string()),
+                                // wiki links start with `[[`, so offset the span by `2`
+                                span: span_provider.span(span.start + 2)
+                            }])
+                        } else {
+                            warn!("WARNING: The wikilink destination url {dest_url} could not be cleaned by removing potholes and fragments");
```
thomas-zahner: Warning is already doing formatting and prefixing `WARN`

```suggestion
                            warn!("The wikilink destination url {dest_url} could not be cleaned by removing potholes and fragments");
```

lychee-lib/src/types/error.rs
```diff
@@ -178,6 +178,14 @@ pub enum ErrorKind {
...
+
+    /// Error on creation of the `WikilinkResolver`
+    #[error("Failed to initialize Wikilink Checker")]
```
thomas-zahner: Also display the inner value here
```suggestion
    #[error("Failed to initialize wikilink checker: {0}")]
```

lychee-lib/src/checker/wikilink/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// In Markdown Links both '#' and '|' act as modifiers
+/// '#' links to a specific Header in a file
+/// '|' is used to modify the link name, a so called "pothole"
+const MARKDOWN_FRAGMENT_MARKER: char = '#';
+const MARKDOWN_POTHOLE_MARKER: char = '|';
+
+/// Clean a `WikiLink` by removing potholes and fragments from a `&str`
+pub(crate) fn wikilink(input: &str, has_pothole: bool) -> Result<CowStr<'_>, ErrorKind> {
+    // Strip potholes (|) from wikilinks
```
thomas-zahner: This confused me to think that the marker is the "pothole" but its the text afterwards.

```suggestion
    // Strip pothole marker (|) and pothole (text after marker) from wikilinks
```

lychee-lib/src/checker/wikilink/index.rs
thomas-zahner: Never mind, this is resolved with my commit.

JayJayArr:
Thanks for taking a look at the error handling here @thomas-zahner ! 
Added some more testcases for valid and invalid `WikiLinks`, if you still have any specific testcase in mind that we should check here please let me know. Added a check for unsupported characters as well :sweat_smile:

lychee-lib/src/types/error.rs
```diff
@@ -466,6 +487,9 @@ impl Hash for ErrorKind {
             Self::Cookies(e) => e.hash(state),
             Self::StatusCodeSelectorError(e) => e.to_string().hash(state),
             Self::PreprocessorError { command, reason } => (command, reason).hash(state),
+            Self::WikilinkNotFound(u, _p) => u.hash(state),
```
thomas-zahner: We should hash both parts, e.g. like with `PreprocessorError`

lychee-lib/src/types/error.rs
```diff
@@ -466,6 +487,9 @@ impl Hash for ErrorKind {
...
+            Self::WikilinkInvalidBase(e) => e.hash(state),
+            Self::WikilinkUnsupportedCharacter(_c, u) => u.hash(state),
```
thomas-zahner: Same here, but should be removed as mentioned in other comment

lychee-lib/src/checker/wikilink/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+use pulldown_cmark::CowStr;
+
+/// In Markdown Links both '#' and '|' act as modifiers
+/// '#' links to a specific Header in a file
+/// '|' is used to modify the link name, a so called "pothole"
+const MARKDOWN_FRAGMENT_MARKER: char = '#';
+const MARKDOWN_POTHOLE_MARKER: char = '|';
+/// A Link containing one of the following characters may not work as a link:
+/// `<https://help.obsidian.md/links#Supported+formats+for+internal+links>`
+const MARKDOWN_OBSIDIAN_PROBLEMATIC_CHARACTERS: [char; 4] = ['^', '%', '[', ']'];
```
thomas-zahner: Thanks for including the link, I see where you get it from. But this is only a  recommendation by the obsidian docs: `A string which contains the following characters MAY not work as a link`. It then goes on to link to [stack overflow](https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names) where the main answers don't even seem to mention these characters. So to me it sounds like Obsidian itself allows these characters, at least in certain cases, and also we want support for Wikilinks in general and not adhere to a specific implementation. (Obsidian in this case)

So I would really prefer to remove `WikilinkUnsupportedCharacter` and this very specific check. The links may still work even with those characters and if they do, then lychee should also accept them. Also it's really the parser's responsibility to deal with that IMO.

JayJayArr:
Added a Test from the `cli` and a Unit Test for the `WikilinkResolver` for this case. You did not miss it, it was indeed missing, sorry for the gap there!

thomas-zahner:
@JayJayArr Thank you very much for the changes and the work, this really is a great addition. I've just removed a `dbg!` expression you probably forgot to remove and simplified indexing with 0c3dcb53295fbc255d7511d78724e055437de2b8.

jrfnl:
@JayJayArr and everyone who reviewed and contributed to the discussion: thank you for working on this! I look forward to seeing this in the next release. 💞

JayJayArr:
Thanks to both of you @mre & @thomas-zahner for your support on this!

katrinafyi:
I was trying out this feature and I wrote these notes about things that I thought were unexpected:

- when a link is not found, the error message is long because it repeats the message twice. it also repeats the base directory twice and makes the invalid path `/home/x/progs/lychee/fixtures/home/x/progs/lychee/fixtures/Doesn't exist.md` - i don't know whether this is intentional.
   ```
      [ERROR] file:///home/x/progs/lychee/fixtures/Doesn't%20exist.md |
      Wikilink file:///home/x/progs/lychee/fixtures/Doesn't%20exist.md not found at /home/x/progs/lychee/fixtures/home/x/progs/lychee/fixtures/Doesn't exist.md:
      WikiLink file:///home/x/progs/lychee/fixtures/Doesn't%20exist.md could not be found at /home/x/progs/lychee/fixtures/home/x/progs/lychee/fixtures/Doesn't exist.md
   ```
- wikilinks [always ignore fragments](https://github.com/lycheeverse/lychee/blob/24becd700b4b2a496a901b11670a403c3cb5651f/lychee-lib/src/checker/wikilink/mod.rs#L32)? i can't see anything in the [obsidian help page](https://help.obsidian.md/links#Link+to+a+heading+in+a+note) which says they should be disregarded. [the unit tests](https://github.com/lycheeverse/lychee/blob/24becd700b4b2a496a901b11670a403c3cb5651f/lychee-bin/tests/cli.rs#L3084-L3117) also seem to say that they should be resolved (unless "resolve" is meant as "ignore").  i note that if fragments were to be processed, the current markdown extractor would have to be changed. atm, it makes anchors using `#kebab-case` which is different to what obsidian uses in wikilinks.

- wikilinks, if enabled, are tried on all file:// links, including absolute links. for instance, a link like `[a](file:///tmp/Usage.md)` could succeed when /tmp/Usage.md doesn't exist and a Usage.md exists inside the base folder. maybe it should be limited to relative links which only have filenames. 

- the wikilink feature does [its own directory walk](https://github.com/lycheeverse/lychee/blob/24becd700b4b2a496a901b11670a403c3cb5651f/lychee-lib/src/checker/wikilink/index.rs#L40). this is different from other directory walks in lychee and won't respect options like gitignore or hidden files. actively ignoring symlinks is also weird.

- to specify the indexing folder, i think `--root-dir` would've made more sense, or it could even be based off the inputs provided through command line. but it does work with `--base-url` so it's nbd.

Aside from the first point, I think these points are kinda hard to fix or don't necessarily need fixing. Just things to think about.

PR #1816, opened 2025-08-23
Always check files explicitly if specified by user or matched by user glob

mre:
Ensure that files specified via glob patterns are always checked, regardless of their extensions. This change addresses the issue where no links were found due to configuration errors. Additional tests have been added to verify glob pattern handling and extension checks.

Fixes lycheeverse/lychee-action#305

### DIFF
--- a/fixtures/example_dir/example.html
+++ b/fixtures/example_dir/example.html
@@ -1,0 +1,1 @@
+<a href="https://example.com/example_dir/html">This is an HTML file to test glob patterns in lychee.</a>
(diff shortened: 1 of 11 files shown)


mre:
This should be ready for review now. It's just two lines of actual logic changes followed by a bunch of tests to make sure I didn't break more. Once this is merged, I plan to do a hotfix release, 0.20.1 to resolve the downstream problems with lychee-action.

This now also covers individual files, which will always be checked if provided. The extension does not matter in that case.
@tooomm fyi

mre:
I will go ahead and merge this since this is a problem for some lychee right now.

thomas-zahner:
@mre Thanks for the fix, looks all good to me :+1:

PR #1827, opened 2025-08-29
Refactor Input Handling

mre:
I wasn't too happy with our input handling module anymore, so I tried to split up the functionality up into multiple smaller structs.

One benefit is that I got rid of the `unreachable!` branch for glob patterns, because that case can no longer happen: I introduced a new `ResolvedInput` type instead, which is the representation of plain file paths and raw inputs (strings, URLs, and `stdin`) where globs got resolved to paths as well.

The final result is an incremental improvement in my opinion. We could still do a lot more, however I wanted to keep the logic identical to before.

## Details

- Introduced `InputResolver` to centralize input source resolution and content extraction.
- Moved content-related logic from `Input` to `InputResolver`.
- Created new modules for input handling: `content`, `core`, `resolver`, and `source`.
- Updated `Collector` to utilize `InputResolver` for obtaining input sources and contents.
- Improved the docs

### DIFF
--- a/lychee-lib/src/types/input/input.rs
+++ b/lychee-lib/src/types/input/input.rs
@@ -152,110 +152,6 @@ impl Input {
         }
     }
 
-    /// Get all input sources for content processing.
-    ///
-    /// This method returns a stream of input sources for the given input, taking into
-    /// account the matching file extensions and respecting exclusions.
-    ///
-    /// This can be used for retrieving all inputs which lychee would check for
-    /// links.
-    ///
-    /// # Returns
-    ///
-    /// Returns a stream of `Result<InputSource>` for all matching input sources.
-    ///
-    /// # Errors
-    ///
-    /// Will return errors for file system operations or glob pattern issues
-    pub fn get_input_sources<'a>(
-        &'a self,
-        file_extensions: FileExtensions,
-        skip_hidden: bool,
-        skip_gitignored: bool,
-        excluded_paths: &'a PathExcludes,
-    ) -> impl Stream<Item = Result<InputSource>> + 'a {
-        try_stream! {
-            match &self.source {
-                InputSource::RemoteUrl(url) => {
-                    // Yield the remote URL as an input source
-                    yield InputSource::RemoteUrl(url.clone());
-                },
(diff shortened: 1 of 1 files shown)


lychee-lib/src/types/input/input.rs
```diff
@@ -442,7 +370,21 @@ impl Input {
...
+    /// - The glob pattern is invalid or expansion encounters I/O errors
+    /// - Directory traversal fails, including:
+    ///   - Permission denied when accessing directories or files
+    ///   - I/O errors while reading directory contents
+    ///   - Filesystem errors (disk errors, network filesystem issues, etc.)
+    ///   - Invalid file paths or symbolic link resolution failures
+    /// - `.gitignore` processing fails:
+    ///   - Reading or parsing `.gitignore` files
+    ///   - Processing ignore patterns for parent directories
+    ///   - Evaluating ignore rules against file paths
```
thomas-zahner: These cases sound pretty identical to me. Maybe a single line like:

```
Errors when reading or evaluating `.gitignore` or `.ignore` files
```

lychee-lib/src/types/input/input.rs
```diff
@@ -353,68 +285,64 @@ impl Input {
...
+                }
                 InputSource::String(ref s) => {
                     yield Self::string_content(s, self.file_type_hint);
                     return;
-                },
-                InputSource::FsGlob { .. } | InputSource::FsPath(..) => {
-                    // Handle FsPath and FsGlob sources
-                    // We can use `get_input_sources` to get the input sources, which will handle
-                    // filtering by file extensions and exclusions
-                    let mut sources_stream = Box::pin(self.get_input_sources(file_extensions, skip_hidden, skip_gitignored, &excluded_paths));
```
thomas-zahner, line 369: We remove the single usage of `Input::get_input_sources`. Is there a point in still keeping this method? Do library users ever need this?

mre: Right, I've cleaned that up. Thanks for the hint.

lychee-lib/src/types/input/input.rs
mre: Summarized. 😊

Issue #1635, opened 2025-02-17
problems with lychee pre-commit installation


dpinol:
1) With the following pre-commit configuration as per the README.md, running "pre-commit run --all lychee" fails with error "Executable `lychee` not found". I had to `cd` into the pre-commit environment and manually run "make install".
```
  - repo: https://github.com/lycheeverse/lychee.git
    rev: v0.15.1
    hooks:
      - id: lychee
        # Optionally include additional CLI arguments
        args: ["--no-progress", "--exclude", "file://"]
```

2) Above v0.15.1, the rev pre-commit field must have the syntax "lychee-v0.1X.X". Is this intended? All pre-commit hooks I've  seen have revisions with  format "v9.9.9"

thanks

mre:
It looks like there are two distinct issues being reported:

1. Pre-commit hook installation failure
The error "Executable `lychee` not found" suggests the pre-commit hook isn't properly installing the lychee binary. Having to manually `cd` into the pre-commit environment and run "make install" shouldn't be necessary. This likely indicates an issue with the hook's installation script.

Have you tried any of these steps?
- Clearing your pre-commit cache: `pre-commit clean`
- Updating pre-commit: `pre-commit autoupdate`

2. Version tag format change
The prefix was automatically added when we switched to release-plz. I'm not too happy with that change, as it caused many problems. At the moment, I wouldn't want to switch back to non-prefixed releases as I know it will break a lot of things again. Sorry about that.

Does that mean you changed the above version string from `rev: v0.15.1` to  `rev: lychee-v0.15.1`?
If not, you can try adapting your pre-commit config like this to see if it works out of the box:

```yaml
- repo: https://github.com/lycheeverse/lychee.git
  rev: lychee-v0.18.1  # If using newer versions
  hooks:
    - id: lychee
      args: ["--no-progress", "--exclude", "file://"]
```

(You probably have to re-install the hook.)

Would you be able to provide:
1. Your operating system details
2. The output of `pre-commit --version`

This would help us identify if this is a broader issue that needs to be addressed in the documentation or codebase.

lyndonlim27:
> Also, I wonder if someone else got it to work? If there's anyone reading this who uses the pre-commit configuration, please add a comment.

Hi I am facing the same issue as well.

<img width="702" alt="Image" src="https://github.com/user-attachments/assets/cd5dfe66-068b-4fc3-9104-5f508d263261" />

pre-commit configuration:

```bash
repos:
[...]

PR #1829, opened 2025-08-29
Fix pre-commit hooks

mre:
So I was thinking about the pre-commit hooks again. At the moment, the installation fails because lychee is a workspace and AFAICT, there is no way to say to the pre-commit project to install the Rust *binary* in the project (correct me if I'm wrong). 

The workaround is to provide a script which installs lychee. This has the advantage that we can use `cargo binstall`, if available`, to speed up the installation process. If it is not available, we fall back to a normal installation. The installation should happen automatically if users use the default `lychee` hook.

I also considered using our pre-built lychee binaries instead, but this would require us to check the platform we're on before fetching the correct binary. That would be additional work, but may worth it? 🤔 I could change that part, but I would like to hear if this is a thing that is typically done. I haven't checked if there's an example for that binary installation process anywhere. Either way, we could also do that in a separate PR. This PR should at least unblock most users for now (?).

Feedback welcome! 

## Implementation Details

- Add auto-installing lychee hook using script that tries cargo-binstall then cargo install
- Add lychee-system hook for users with existing installations
- Update docker hook to use current version (0.20.1)
- Add `PRE_COMMIT.md` documentation which explains hook options and version format
- Fix version tag format usage (lychee-v* instead of v*)

Closes: #1635

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,18 +1,30 @@
 - id: lychee
   name: lychee
-  description: "Run 'lychee' for fast, async, stream-based link checking"
+  description: "Run 'lychee' for fast, async, stream-based link checking (auto-installs)"
+  entry: scripts/lychee_pre_commit.sh
+  language: script
+  args: ["--no-progress"]
+  types: [text]
+  pass_filenames: true
+  require_serial: true
+  minimum_pre_commit_version: "2.9.2"
+- id: lychee-system
+  name: lychee-system
+  description: "Run 'lychee' using system installation (requires: cargo install lychee)"
   entry: lychee
   language: system
   args: ["--no-progress"]
   types: [text]
   pass_filenames: true
   require_serial: true
+  minimum_pre_commit_version: "2.9.2"
 - id: lychee-docker
-  name: lychee
-  description: "Run 'lychee' docker image for fast, async, stream-based link checking"
-  entry: lycheeverse/lychee:0.14
+  name: lychee-docker  
+  description: "Run 'lychee' docker image for fast, async, stream-based link checking (auto-installs)"
+  entry: lycheeverse/lychee:0.20.1
   language: docker_image
   args: ["--no-progress"]
   types: [text]
   pass_filenames: true
   require_serial: true
+  minimum_pre_commit_version: "2.9.2"
--- a/PRE_COMMIT.md
+++ b/PRE_COMMIT.md
@@ -1,0 +1,78 @@
+# Lychee Pre-commit Hooks
+
+This repository provides three pre-commit hook options for lychee link checking:
+
+## Quick Start
+
+Add this to your `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+  - repo: https://github.com/lycheeverse/lychee
+    rev: lychee-v0.20.1  # Use latest lychee-v* tag
+    hooks:
+      - id: lychee  # Auto-installs lychee
+```
+
+## Hook Options
+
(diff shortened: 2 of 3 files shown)


mre:
@dpinol, @lyndonlim27, I'm sorry that it took so long, but there's finally some progress on the pre-commit front. If you have some time, could you try it?

Here's how you can test the pre-commit hook from the PR branch:

```yaml
  repos:
      rev: fix-pre-commit-hook
      hooks:
        - id: lychee  # Auto-installs lychee
          args: ["--no-progress"]
```

## Quick Test Steps

1. Add the config above to .pre-commit-config.yaml
2. Run: pre-commit install
3. Run: pre-commit run lychee --all-files

I'd be thankful to everyone who gives feedback! 😊

mre:
I will go ahead and merge this to get some more eyeballs on the issue. The current version is broken anyway, so I don't think this will ruin anybody's workflow. Let me know if it does.

dpinol:
@mre Hi, it now works perfectly. Thanks!
 I just needed to change the revision
```
    rev: lychee-v0.20.1
    hooks:
      - id: lychee  # Auto-installs lychee
```

mre:
Thanks for the feedback.

Borda:
I tried to copy this https://github.com/lycheeverse/lychee/pull/1829#issuecomment-3244464383 to my local config and am still getting `Executable `lychee` not found` 🤔

mre:
Sorry to hear that. I don't use pre-commit myself, so I'm really at a loss here. But if you like, perhaps answering the following questions can help narrow down the problem:

- What operating system and architecture are you on?
- What shell do you use? 
- What version of `pre-commit` are you running? `pre-commit --version`
- Can you paste the **exact hook configuration** from your `.pre-commit-config.yaml`?
- How was `lychee` installed (outside pre-commit), if at all? I guess it should install it automatically, but just to be sure.
- Does `lychee` run successfully **outside pre-commit**?
- What is the output of the following commands?
   - `which lychee` (or `where lychee` on Windows)
   - `echo $PATH`
- Have you tried clearing pre-commit’s cache with `pre-commit clean`?

In general, I would suggest creating a new issue based on that information to get more eyeballs on the problem.

Borda:
> * What operating system and architecture are you on?

MacOS

> * What shell do you use?

zsh

> * What version of `pre-commit` are you running? `pre-commit --version`

pre-commit 4.5.1

> * Can you paste the **exact hook configuration** from your `.pre-commit-config.yaml`?

Exactly this one https://github.com/lycheeverse/lychee/pull/1829#issuecomment-3244464383

> * How was `lychee` installed (outside pre-commit), if at all? I guess it should install it automatically, but just to be sure.

Exactly this one https://github.com/lycheeverse/lychee/pull/1829#issuecomment-3244464383

> * Does `lychee` run successfully **outside pre-commit**?

Did not try, as it shall be irelevant, as pre-commit hooks runs in own env, not using local execution... 

> * What is the output of the following commands?
>   * `which lychee` (or `where lychee` on Windows)
>   * `echo $PATH`

dotn have local version installed

> * Have you tried clearing pre-commit’s cache with `pre-commit clean`?

Yes, I did

mre:
Okay, I see, that all seems very normal.

My current theory is that it works for users who already have lychee (or `~/.cargo/bin`) on PATH; the auto-install hook installs the binary but never adds Cargo's bin directory to PATH, so execution fails in clean pre-commit environments.

I took a stab at a fix in https://github.com/lycheeverse/lychee/pull/2002.

PR #1830, opened 2025-08-29
 chore(file checker): separate resolving and checking for easier testing

katrinafyi:
this does a simple refactor to break up the `check_path` function.   previously, this had the responsibility of resolving index   files/fallback extensions _and_ checking fragments. this was done in one   step which made it hard to test - in order to test the resolving, we had  to test it indirectly by checking fragments in certain files.
  
  this introduces a new `resolve_local_path` function which only applies  the index file + extensions, so we can use this in unit tests and make  assertions about the resolved path.
  
  this should have no changes to any behaviour.

~please note perf bug.~ also, idk if I go overboard with the macros ;-;

### DIFF
--- a/lychee-lib/src/checker/file.rs
+++ b/lychee-lib/src/checker/file.rs
@@ -156,12 +156,10 @@ impl FileChecker {
         //
         // (currently, this case is only reachable if `.` is in the index_files list.)
         match path {
-            Ok(dir_path) if dir_path.is_dir() => {
-                match self.apply_fallback_extensions(&dir_path, uri) {
-                    Ok(fallback_path) => Ok(Cow::Owned(fallback_path)),
-                    Err(_) => Ok(dir_path),
-                }
-            }
+            Ok(dir_path) if dir_path.is_dir() => self
+                .apply_fallback_extensions(&dir_path, uri)
+                .map(Cow::Owned)
+                .or(Ok(dir_path)),
             Ok(path) => Ok(path),
             Err(err) => Err(err),
         }


lychee-lib/src/checker/file.rs
```diff
@@ -110,35 +113,34 @@ impl FileChecker {
...
             }
 
             // other IO errors are unexpected and should fail the check
             Err(e) => Err(ErrorKind::ReadFileInput(e, path.to_path_buf())),
 
             // existing directories are resolved via index files
-            Ok(meta) if meta.is_dir() => self.apply_index_files(path).map(Cow::Owned),
+            Ok(meta) if meta.is_dir() => self.apply_index_files(path),
 
             // otherwise, path is an existing file - just return the path
```
mre: Have you tried to return a `Path` (with a lifetime)? This would technically be a breaking change, but that should be fine if we can avoid the copy.

katrinafyi: Return a &Path with lifetime from resolve_local_path? I haven't tried, but my intuition was it wouldn't work because we can't return a reference to locally owned data. The owned data would get deallocated upon leaving the function.

Previously, Cow was used to hold either a owned or borrowed value, but I got rid of Cow because it was somewhat inconvenient to work with. I could add it back if you think that's a reasonable solution.

mre: Silly me, you're right.

I played around with it and created https://github.com/rina-forks/lychee/pull/1. Can you take a look? I can see advantages and disadvantages on both sides, so I'm okay with keeping your version as well. 😉 If you decide to merge my PR, this PR should automatically get updated. 🪄

katrinafyi: It is merged, thanks for your help!

mre:
Yay, that's a solid improvement. Thanks @katrinafyi. 🙏

Issue #1579, opened 2024-12-06
Treats markdown link text as URL


dbr:
If I have a markdown where the link text is a broken link, but the actual link target is a valid URL,

```
$ cat test.md 
[https://dummyexample.gov/notexist (archive.org link)](https://web.archive.org/web/20241129184733/http://example.com)
```

I would expect it only checks the web.archive.org link, but it checks both the link-text and the URL

```
$ lychee -v test.md 
   [ERROR] https://dummyexample.gov/notexist
     [200] https://web.archive.org/web/20241129184733/http://example.com

Issues found in 1 input. Find details below.

[test.md]:
   [ERROR] https://dummyexample.gov/notexist

🔍 2 Total (in 2s) ✅ 1 OK 🚫 1 Error
```

This seems to relate to https://github.com/lycheeverse/lychee/issues/949#issuecomment-2400908728 except this is a slightly different case (as a markdown parser might helpfully convert `http://example.com` to `<a href="example.com">http://example.com</a>` - but it would surely be nonsensical to ever do the same for `dummyexample.gov` in the initial example - it would be a link inside a link)

mre:
Yeah, we started treating plaintext links in HTML files as false-positives in https://github.com/lycheeverse/lychee/pull/1528. Something similar could work for Markdown files.

@markogalevski fyi

thomas-zahner:
@mre I guess this is related to #1399? Should the text of  a Markdown link only be checked with the `--include-verbatim` option as mentioned there?

mre:
Yes, that sounds like the way to go! Thanks for recognizing the connection.

thomas-zahner:
@mre So to recap; "verbatim" for Markdown and HTML seem to have different meanings.

With HTML we only detect links as such if we can be sure that they are supposed to be valid links, meaning we ignore links form "plain text" that could be links.

```html
<p>This is considered verbatim and not found unless the flag is specified: https://example.com</p>
```

With Markdown we are less strict:

```markdown
This is not considered verbatim, meaning it is found even without the flag: https://example.com

only things like this are considered verbatim: `https://example.com`
```

I think we should give verbatim the same meaning for Markdown files as with HTML files. Do you agree? If so, I will try to change this behaviour and update the documentation accordingly, as the current documentation is not very clear about the meaning of verbatim. This will of course be a breaking change. It will prevent false positives like in this issue at the cost of detecting fewer links as long as `--include-verbatim` is not specified.

[...]

Issue #1399, opened 2024-04-04
a.textContent is checked even if the text is shortened for the link


jpoehnelt:
Is there a way to skip the check of `textContent` for the `a` tag?

```html
<a href="https://example.com/foo/1231241243124124124">https://example.com/foo/...</a>
```

_Originally posted by @MichaIng in https://github.com/lycheeverse/lychee/issues/259#issuecomment-865908305_

---

While checking our forum for outdated links, I recognised that also shortened URL text within anchor tags is checked. phpBB shortens long URLs. so that such an element is produced:
```
<a href="https://example.com/loong/looong/url/that/is/shorteneed/index.html" class="postlink">https://example.com/loong/looong/url/th ... index.html</a>
```
`lychee` now checks the URL itself but as well the text, where it identifies `https://example.com/loong/looong/url/th` as URL, which of course fails.

So once a basic functionality exists to exclude URLs/text based on HTML tags, the text content of tags/elements which have an explicit src/href/... attribute could be excluded by default, as AFAIK such shortening is quite common.

mre:
Just tested this and you're right. This is a bug, and we should exclude `textContent` unless `--include-verbatim` is enabled.

mre:
Related: https://github.com/lycheeverse/lychee/issues/1462#issuecomment-2380225133

thomas-zahner:
Did some investigation. This is indeed related to #1462 and according to my testing the bug reported in this issue was fixed with the relating PR #1528.

#1579 however is still not fixed, probably because the PR only affected HTML but not Markdown. This also means that "verbatim" currently has a different meaning for Markdown and HTML. I will try to fix #1579 and I will also try to make sure that the meaning of "verbatim" is consistent and better documented.

thomas-zahner:
Actually this is a duplicate of https://github.com/lycheeverse/lychee/issues/1462
So ignore my previous above comment. I might have been confused because I was piping the contents directly to lychee, while not specifying the default extension. (see note below)
While the issue was initially reported because of issues with mail addresses @mre realised that the verbatim behaviour was the root cause, as outlined in [this comment](https://github.com/lycheeverse/lychee/issues/1462#issuecomment-2380225133).

So this issues has actually been fixed for quite a wile now. Using master & 0.21.0 both seems to work.

## Reproduce

```
echo '<a href="https://example.com/real">https://example.com/fake</a>' > test.html

$ lychee -v --dump test.html
https://example.com/real (test.html) [excluded]

[...]

PR #1831, opened 2025-08-29
Ignore Links in Markdown Link Text

mre:
At the moment, links inside Markdown link texts get parsed and handled.

For example:

```md
[https://dummyexample.gov/notexist (archive.org link)](https://web.archive.org/web/20241129184733/http://example.com)
``` 

This would find https://dummyexample.gov/notexist.

This PR changes that and ignores those links unless the user explicitly enabled the check by setting `--include-verbatim`.

This is similar to the handling of HTML files. See https://github.com/lycheeverse/lychee/issues/1399.

Wikilinks behaved slightly differently and would be extracted twice when `--include-verbatim` was enabled - once from the WikiLink tag itself (which contains the destination URL) and once from the text content within the WikiLink. This caused duplicate URL entries in the output. That's why I introduced a separate flag for that.

Fixes: https://github.com/lycheeverse/lychee/issues/1579

## Details

- Respect `include_verbatim` flag when processing link text
- Fix WikiLink double extraction by using separate tracking
- Add tests for link text extraction behavior

### DIFF
--- a/lychee-lib/src/extract/markdown.rs
+++ b/lychee-lib/src/extract/markdown.rs
@@ -23,6 +23,7 @@ pub(crate) fn extract_markdown(
     // which is why we keep track of entries and exits while traversing the input.
     let mut inside_code_block = false;
     let mut inside_link_block = false;
+    let mut inside_wikilink_block = false;
 
     let parser = TextMergeStream::new(Parser::new_ext(input, md_extensions()));
     parser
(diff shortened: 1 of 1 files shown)


mre:
@dbr, it's been a very long time. So sorry for the delay. I'm not sure if you're still interested in the fix.

dbr:
Great, thanks for this!

PR #1832, opened 2025-08-29
Optimize InputSource handling with Cow and add benchmarks

mre:
Out of curiosity, I finally tackled this TODO from the code:

```rust
// TODO: consider using Cow (to avoid one .clone() for String types)
```

I took a look and it turns out it doesn't touch as many places as I thought it would, so I went with it.

The result is lesser allocations for places where we handle raw input strings.

I did add a micro-benchmark for input content creation and it turns out to be decently fast. 

On my machine:

  - Extract from large docs: 28% improvement (21.4ms vs 29.3ms baseline)
  - InputContent::from_string: 28% improvement (5.3µs vs 7.4µs baseline)
  - InputContent::from_owned_string: 9% improvement (14.2µs vs 15.7µs baseline)
  - from_string with owned baseline: 14% improvement (8.1µs vs 9.4µs baseline)

Not earth-shaking, but we're already operating at a decent level of performance by now, so I take it.

The question is: what do y'all think? Does it make the code too complex or is the complexity manageable? Feedback very welcome!

### DIFF
--- a/lychee-lib/src/types/input/content.rs
+++ b/lychee-lib/src/types/input/content.rs
@@ -6,6 +6,7 @@
 use super::source::InputSource;
 use crate::ErrorKind;
 use crate::types::FileType;
+use std::borrow::Cow;
 use std::fs;
 use std::path::PathBuf;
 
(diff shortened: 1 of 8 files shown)


lychee-lib/src/types/input/content.rs
```diff
@@ -24,9 +25,28 @@ impl InputContent {
...
-            source: InputSource::String(s.to_owned()),
+            source: InputSource::String(Cow::Owned(s.to_owned())),
+            file_type,
+            content: s.to_owned(),
+        }
+    }
+
+    #[must_use]
+    /// Create an instance of `InputContent` from an owned String
+    pub fn from_owned_string(s: String, file_type: FileType) -> Self {
```
thomas-zahner: Does it make sense to keep both methods `from_owned_string` and `from_static_str`?
Wouldn't it be possible to generalise it somehow like:

```
pub fn from_str<S: Into<Cow<'static, str>>>(s: S, file_type: FileType) -> Self
```

thomas-zahner (approved):
I think this is an improvement as it doesn't increase complexity very much :+1:

lychee-lib/src/types/input/content.rs
mre: Yeah, that would be possible. I'm still not sure which variant I prefer. The generic version is a bit harder to read. 😅

thomas-zahner: Well if it is possible, I personally would change it. But I wasn't quite sure whether you intentionally have two different methods, as you use `Cow::Owned` and `Cow::Borrowed` respectively.

mre: Yes, done

katrinafyi:
Should this todo also be changed?

https://github.com/lycheeverse/lychee/blob/d6b16bc457c4125eb3d102b6b319fb8f11802122/lychee-lib/src/utils/request.rs#L115-L120

Issue #1833, opened 2025-09-01
Tag Docker image with release version when release tag is pushed


eread:
At the moment, a new Docker image is pushed when a release Git tag is pushed. For example:

- Build command for `v0.20.1` (Debian): https://github.com/lycheeverse/lychee/actions/runs/17205362504/job/48804656733#step:9:218
- Build command for `v0.20.0` (Debian): https://github.com/lycheeverse/lychee/actions/runs/17148270680/job/48652300555#step:9:218

In both cases, there was a Docker image tag for the SHA corresponding to the commit of the Git tag, but no Docker image tag for the release. So:

- For `v0.20.1`, there was `--tag lycheeverse/lychee:sha-0a96dc2` (for https://github.com/lycheeverse/lychee/commit/0a96dc2fdc99faf5e1cd6ccc429b27da32f25985) but no `--tag lycheeverse/lychee:v0.20.1` or `--tag lycheeverse/lychee:0.20.1`.
- For `v0.20.0`, there was `--tag lycheeverse/lychee:sha-e510602` (for https://github.com/lycheeverse/lychee/commit/e5106023109a36848b8d4b387188ad033829d6bd) but no `--tag lycheeverse/lychee:v0.20.0` or `--tag lycheeverse/lychee:0.20.0`.

The Docker image tagging seems to be controlled by using the `docker/metadata-action` GitHub Action: https://github.com/docker/metadata-action. So some configuration change might be required to have a Docker image tagged with the release version number in addition to the SHA of the Git tag.

eread:
I wonder if the warnings we get for the **Docker (meta)** outputs is telling us why there are no release tags for the specific version:

- For Debian-based v0.20.1: https://github.com/lycheeverse/lychee/actions/runs/17205362504/job/48804656733#step:3:54
- For Debian-based v0.20.0: https://github.com/lycheeverse/lychee/actions/runs/17148270680/job/48652300555#step:3:54

I'm not sure that configuration at: https://github.com/lycheeverse/lychee/blob/be26d02c80825bb485ec504f39787a0df6310a34/.github/workflows/docker.yml#L41-L43 with `prefix=lychee-` is performing how intended. Is it meant to be stripping `lychee-` from the tag? 🤔

aravindan888:
can i work on this @eread

eread:
> can i work on this [@eread](https://github.com/eread)

[...]

PR #1834, opened 2025-09-01
Fix locating of version when tagging Docker images on Git tag push

eread:
To contribute to https://github.com/lycheeverse/lychee/issues/1833

Fix locating of version when tagging Docker images on Git tag push and also remove tagging major version only, because it would resolve to 'version 0' at the moment.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,9 +38,8 @@ jobs:
             type=schedule
             type=ref,event=branch
             type=ref,event=pr
-            type=semver,pattern={{version}},value=${{ github.ref_name }},prefix=lychee-
-            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},prefix=lychee-
-            type=semver,pattern={{major}},value=${{ github.ref_name }},prefix=lychee-
+            type=semver,pattern=lychee-{{version}},value=${{ github.ref_name }}
+            type=semver,pattern=lychee-{{major}}.{{minor}},value=${{ github.ref_name }}
             type=sha
             # set latest tag for default branch
             type=raw,value=latest,enable={{is_default_branch}}
@@ -60,9 +59,8 @@ jobs:
             type=schedule
             type=ref,event=branch
             type=ref,event=pr
-            type=semver,pattern={{version}},value=${{ github.ref_name }},prefix=lychee-
-            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},prefix=lychee-
-            type=semver,pattern={{major}},value=${{ github.ref_name }},prefix=lychee-
+            type=semver,pattern=lychee-{{version}},value=${{ github.ref_name }}
+            type=semver,pattern=lychee-{{major}}.{{minor}},value=${{ github.ref_name }}
             type=sha
             # set latest tag for default branch
             type=raw,value=latest,enable={{is_default_branch}}


.github/workflows/docker.yml
```diff
@@ -38,9 +38,8 @@ jobs:
             type=schedule
             type=ref,event=branch
             type=ref,event=pr
-            type=semver,pattern={{version}},value=${{ github.ref_name }},prefix=lychee-
-            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},prefix=lychee-
-            type=semver,pattern={{major}},value=${{ github.ref_name }},prefix=lychee-
```
eread, line 43: If I understand correctly, this will avoid tagging a Docker image **only** with the major version number, which is `0` at the moment.

The other lines will tag:

- with `major`.`minor`.`patch` (for example, `0.20.0` and `0.20.1`)
- with `major`.`minor` (for example, `0.20`)

.github/workflows/docker.yml
```diff
@@ -38,9 +38,8 @@ jobs:
...
+            type=semver,pattern=lychee-{{version}}
+            type=semver,pattern=lychee-{{major}}.{{minor}}
```
eread: If I understand correctly, `pattern` helps the action locate the version string itself to use to tag the Docker image. I think the fact that `pattern` is just currently the whole tag is why there are build warnings: https://github.com/lycheeverse/lychee/actions/runs/17205362504/job/48804656733#step:3:54.

I think `prefix` would **add** `lychee-` to the version that's found for tagging the Docker image - perhaps that's wanted so that it matches the tag. The last version tagged with a version number was simply tagged `0.15.0` (https://hub.docker.com/layers/lycheeverse/lychee/0.15.0/images/sha256-f71cb370f67fe81eccab4a8b0ff0539c2859f6fd1c80087078922eba07f9dace) and `0.15.0-alpine` (https://hub.docker.com/layers/lycheeverse/lychee/0.15.0-alpine/images/sha256-18cf46741b0a92244c55670895cf7893c75f19f097f469c1274bf6b53c197ece).

eread:
@mre Could you review this one against the documentation for `semver` type: https://github.com/marketplace/actions/docker-metadata-action#typesemver?

I'm imaging that if, for example, this change was in place when the `lychee-v0.20.1` tag was pushed:

1. The `type=semver` lines would be triggered, because `type=semver` is used on push tag events.
1. The `pattern=lychee-` settings would help pattern match just the version string, and would return proper semver.
1. By having this action return something valid twice, there would be two additional `--tag` options added to the Docker image build command in a later action.
 
If correct, there would've been, in the end:

 - `lycheeverse/lychee:0.20.1`
 - `lycheeverse/lychee:0.20`
 - `lycheeverse/lychee:0.20.1-alpine`
 - `lycheeverse/lychee:0.20-alpine` 

Does that make sense? Have I missed anything? Is that the tagging scheme you want, or do you want something different?

.github/workflows/docker.yml
mre: Yeah, that's better

mre:
Didn't take a look yet, but just posting a comment that this would affect the docs at https://github.com/lycheeverse/lycheeverse.github.io as well. Maybe some other places, too? Can't think of any right now, but if someone depends on the Docker image tagging, then now is the time speak up. 😆

eread:
> Didn't take a look yet, but just posting a comment that this would affect the docs at https://github.com/lycheeverse/lycheeverse.github.io as well. Maybe some other places, too? Can't think of any right now, but if someone depends on the Docker image tagging, then now is the time speak up. 😆

@mre Would this look dramatically different to what the project produced before the move to `release-plz` https://hub.docker.com/r/lycheeverse/lychee/tags?name=0.1?

eread:
@mre Based on the comment from @aravindan888 here: https://github.com/lycheeverse/lychee/issues/1833#issuecomment-3241130206, I've added back `value=${{ github.ref_name }}`. I interpret `value` as being a default value when the tag doesn't match anything. Not sure if we'll ever hit the logic for this, but it does make this a smaller change.

CC @trask 

WDYT @aravindan888?

eread:
@mre Did you need me to do anything else on this one? Or is it just left open for comment?

mre:
Sorry, there's one comment here, which should be addressed before merging. 
https://github.com/lycheeverse/lychee/issues/1833#issuecomment-3241130206
Can you take a look?

eread:
> Sorry, there's one comment here, which should be addressed before merging. [#1833 (comment)](https://github.com/lycheeverse/lychee/issues/1833#issuecomment-3241130206) Can you take a look?

Thanks for taking a look @mre. I think I addressed that, and mentioned in this comment: https://github.com/lycheeverse/lychee/pull/1834#issuecomment-3243524167. So I've put back `value=${{ github.ref_name }}` compared to an earlier version.

eread:
@mre I've also added another commit: https://github.com/lycheeverse/lychee/pull/1834/commits/88c48a289bf5753f2f064fe36f695ab57017440c to fix an error with a job (https://github.com/lycheeverse/lychee/actions/runs/18024185042/job/51287994384) I got when I rebased this PR.

Not sure what changed, but explicitly specifying `rustfmt` is documented: https://github.com/dtolnay/rust-toolchain?tab=readme-ov-file#inputs.

eread:
> @mre I've also added another commit: [88c48a2](https://github.com/lycheeverse/lychee/commit/88c48a289bf5753f2f064fe36f695ab57017440c) to fix an error with a job (https://github.com/lycheeverse/lychee/actions/runs/18024185042/job/51287994384) I got when I rebased this PR.
> 
> Not sure what changed, but explicitly specifying `rustfmt` is documented: https://github.com/dtolnay/rust-toolchain?tab=readme-ov-file#inputs.

@mre I've now removed that commit because it's been fixed upstream.

mre:
Looks fine to me. I don't yet know what implications it will have downstream, but we'll see about that. But from my side, this is worth merging to try and fix the Docker tags. Thanks for the work and for your patience, @eread.

eread:
> Looks fine to me. I don't yet know what implications it will have downstream, but we'll see about that. But from my side, this is worth merging to try and fix the Docker tags. Thanks for the work and for your patience, @eread.

Thanks for the merge @mre.

PR #1835, opened 2025-09-01
Refactor progress bar

rodic:
Refactor of progress‑bar

    - Introduced `LycheeProgressBar` – a thin wrapper around the `indicatif` library that centralises all progress‑bar logic (configuration, styling, ticking, length handling, and finishing).
    - Replaced scattered direct usages of `ProgressBar` with the new wrapper, improving readability and making future UI changes easier to manage.
    - Added a dedicated configuration struct (`LycheeProgressBarConfig`) with sensible defaults (template, increment step, tick interval, progress characters).
    - Tests


lychee-bin/src/commands/check.rs
```diff
@@ -173,34 +172,35 @@ async fn suggest_archived_links(
...
 where
     S: futures::Stream<Item = Result<Request>>,
 {
     tokio::pin!(requests);
+    let mut request_count = 0;
     while let Some(request) = requests.next().await {
         let request = request?;
-        if let Some(pb) = &bar {
-            pb.inc_length(1);
-            pb.set_message(request.to_string());
```
rodic, line 190: i don't think we need to set msg here?

thomas-zahner: We should keep it IMO because showing the extracted links and not only the link after having checked it, makes sense in my opinion.

lychee-bin/src/commands/check.rs
```diff
@@ -173,34 +172,35 @@ async fn suggest_archived_links(
...
-        }
+        request_count += 1;
         send_req
             .send(Ok(request))
             .await
             .expect("Cannot send request");
     }
+    if let Some(pb) = &bar {
+        pb.set_length(request_count);
+    }
```
thomas-zahner, line 192: The original version of this section made more sense.  Extracting the links from the inputs (and awaiting the requests, before executing the requests) and displaying the result after having executed the request are two independent highly concurrent processes. Because of that the progress bar size grows, even as we are already performing requests. So setting the progress bar length only after all requests were created makes no sense.

Run `cargo run fixtures/` with your PR version and you will see, how the progress bar remains at `0` during the whole link check process, displaying something like `2114/0`.

lychee-bin/src/lychee_progress_bar.rs
```diff
@@ -0,0 +1,96 @@
...
+        bar.set_message(initial_message);
+        bar.enable_steady_tick(config.tick_interval);
+
+        LycheeProgressBar { bar, config }
+    }
+
+    pub(crate) fn update(&self, message: Option<String>) {
+        self.bar.inc(self.config.increment);
+        if let Some(msg) = message {
+            self.bar.set_message(msg.clone());
```
thomas-zahner, line 49: I don't think we need to `clone` here

thomas-zahner (changes requested):
Thanks for the PR. Extracting the progress bar into a new struct and file does make sense and improves readability and maintainability :+1:

rodic:
@thomas-zahner thanks a lot for reviewing this. I'm going to address the comments as soon as I can. I'm a bit busy these days, so sorry for the late response.

lychee-bin/src/lychee_progress_bar.rs
```diff
@@ -0,0 +1,96 @@
...
+            template: "{spinner:.162} {pos}/{len:.238} {bar:.162/238} {wide_msg}",
+            increment: 1,
+            tick_interval: Duration::from_millis(500),
+            progress_chars: "━ ━",
+        }
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct LycheeProgressBar {
```
mre, line 24: Can we call it `ProgressBar` instead? I think the `Lychee` is sort of implied. 😉 Same for the filename.

mre:
@rodic any updates?

rodic:
I'm sorry @mre i'll try to get it done this week.

mre:
Ah no worries and no rush. Just wanted to avoid that this awesome work goes stale.

rodic:
i apologize again for being late, i didn't forget about it :)

thomas-zahner:
@rodic I'm closing this in favour of  #1914

There I've squashed this PR into one commit and dug a big deeper with refactoring. Thanks for the PR!

Issue #662, opened 2022-06-24
Option to provide a list of inputs from a file or stdin


polarathene:
While it's likely a niche case, if providing many explicit inputs to `lychee`, as file paths and URLs can be lengthy, I am a bit concerned about exceeding the character limit (at least I think there is one for some shells with command args).

It may be worthwhile having an option to provide inputs via a line delimited list instead? (`--inputs-from` or similar?)

---

My use-case was with a generated list of files to check (output from `git diff-tree`), which was simple enough to get, but slightly problematic providing as a list of args to `lychee` via a shell this way (especially with edge cases like file paths with spaces). I ended up using `xargs` for the first time to handle this, but found it required `-d '\n'` to work correctly.

I'm not super great with shell commands/scripts, and I struggled a bit with other approaches before finding the simple solution with `xargs`. Being able to read a list of lines from a file as inputs, via a supported option for `lychee` would prevent others like myself from stumbling about in bash/shell :sweat_smile: 

---

It's totally fine if you'd rather hold off on such an option until there's enough demand, or someone were to open a PR for it :+1: 

For anyone else that lands here, you can use whatever command to get your list of inputs (1 per line), and provide them to `lychee` with `xargs` like this: `your-command | xargs -d '\n' lychee --`, that simple :smile: (quotes around `\n` is important)

<details>
<summary>More info</summary>

`your-command` examples:
- `echo -e 'path/to/file\npath/to/another file.md'` (more useful in scripts or testing)
- `cat path/to/input-list-file` (if you stored your input list to a file)
- `git diff-tree --no-commit-id --name-only --diff-filter 'AM' -r --merge-base master remote/branch-name` (generated list output from a binary or script of your own)

These work with `xargs`, but prior to getting a solution with `xargs` I had tried approaches with `echo`/`printf` and such but I had problems especially when testing with file paths that had spaces:
- `xargs -d '\n' lychee -- < <(echo -e 'path/to/file\npath/to/another file.md')`
- `xargs -d '\n' lychee -- <<< $(echo -e 'path/to/file\npath/to/another file.md')`

[...]

PR #1836, opened 2025-09-03
Implement `--files-from` Option

mre:
This allows providing a list of inputs from a file or `stdin`.

## How it works

1. Create a `inputs.txt` file with all the inputs (files) you'd like to check for broken links.
2. Pass that to `lychee` with
   ```sh
   lychee --files-from inputs.txt
   ```

lychee will then use the files from `inputs.txt` as inputs. You can still provide custom inputs as well with

```sh
lychee --files-from inputs.txt another_input.md
```

The inputs file can contain comments, such as

```text
# Ignore this line
some_input.txt
# Ignore this line as well
```

Closes: https://github.com/lycheeverse/lychee/issues/662

### DIFF
--- a/lychee-bin/src/files_from.rs
+++ b/lychee-bin/src/files_from.rs
@@ -35,7 +35,8 @@ impl FilesFrom {
         lines
             .into_iter()
             .filter(|line| {
-                !line.trim().is_empty() && !line.trim_start().starts_with(COMMENT_MARKER)
+                let line = line.trim();
+                !line.is_empty() && !line.starts_with(COMMENT_MARKER)
             })
             .collect()
     }
--- a/lychee-bin/src/files_from.rs
+++ b/lychee-bin/src/files_from.rs
@@ -1,0 +1,130 @@
+//! File list reading functionality for --files-from option
+//!
+//! This module provides the `FilesFrom` struct which handles reading input file
+//! lists from any reader, with support for comments and empty line filtering.
+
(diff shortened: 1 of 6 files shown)


lychee-bin/src/files_from.rs
```diff
@@ -0,0 +1,130 @@
...
+        let inputs = Self::filter_lines(lines);
+        Ok(FilesFrom { inputs })
+    }
+
+    /// Filter out comments and empty lines from input
+    fn filter_lines(lines: Vec<String>) -> Vec<String> {
+        lines
+            .into_iter()
+            .filter(|line| {
+                !line.trim().is_empty() && !line.trim_start().starts_with(COMMENT_MARKER)
```
thomas-zahner:
```suggestion
                let line = line.trim();
                !line.is_empty() && !line.starts_with(COMMENT_MARKER)
```

thomas-zahner (approved):
LGTM :+1:

lychee-bin/src/files_from.rs
mre: Good idea!

thomas-zahner:
@mre Shouldn't it now also be possible to now specify `files_from` in the config file? It doesn't seem to work.

thomas-zahner:
Never mind, investigated and solved in 07213f8730f87cf96692f4a43306028792302664. Will soon create a PR

Issue #972, opened 2023-03-01
Absolute local paths on windows not recognized


Dobatymo:
Sorry if I am using it wrong or it's duplicate report.

Running lychee on a local absolute path on Windows gives the following error:
```
>lychee.exe "C:\asd"
Error: Network error

Caused by:
    0: builder error for url (c:\asd): URL scheme is not allowed
    1: URL scheme is not allowed
```
Running it from that directory with `lychee.exe .` works!

lebensterben:
this indeed is a bug.

Windows path is technically URL and the lychee tries to parse the argument as URL before parse it as a file path. And in the case the drive is mistaken as a scheme.

A simple workaround is to change the said order. But it won't work because someone may name a drive as "http". so "http:\\foo.bar\xyz" becomes both a legit Windows file path and a legit URL.

The true solution is:

Totally ban positional arguments. If you want to supply a file path, use --file, if you want to supply a glob, use --glob, etc.

Dobatymo:
Windows drive names can only be letters `a-z`, so `http:\` would not be a valid Windows path. URLs would have to start with a double slash, which Windows paths don't (ignoring network drives like `\\server\path\` for now...

But yeah, I agree that paths/URLs are a mess and it's best to explicitly distinguish them!

lebensterben:
in my made-up example `http:\foo.bar\xyz`:

- http is the drive name
- \foo.bar\xyz is an absolute path

> URLs would have to start with a schema and double slash

False. It doesn't have to contain double slash. Examples 

- `mailto:foo@bar.com`
- `foo:bar`

Dobatymo:
Yes sorry, I was incorrect about the double slashes. However Windows doesn't really have drive names, only drive letters afaik.

lebensterben:
okay after some research it turns out you're right.

on windows the drive letter must be one of A-Z.

(for example see https://learn.microsoft.com/en-us/windows-server/storage/disk-management/change-a-drive-letter)

So we can indeed guess whether a positional argument is a absolute path with drive on Windows.

mre:
Can we collect a few test cases and the expected outcome here?

lebensterben:
we really should not abuse positional argument even though we can guess whether it's a file path or a URL.

I think in most cases people invoke lychee against specific files or a glob pattern, so the practical choice is to allow path and glob as the positional argument.

[...]

Issue #1595, opened 2024-12-18
Do not assume input to be a URL if the local path doesn't exist


mre:
## Context

At the moment, lychee uses the `http://` scheme as a fallback for local input paths that don't exist:

```sh
> lychee foo
Error: Network error

Caused by:
    0: error sending request for url (http://foo/)
    1: client error (Connect)
    2: dns error: failed to lookup address information: nodename nor servname provided, or not known
    3: failed to lookup address information: nodename nor servname provided, or not known
```

The idea was to model the behavior after curl, which does the same:
https://github.com/curl/curl/blob/70ac27604a2abfa809a7b2736506af0da8c3c8a9/lib/urlapi.c#L1104-L1124

## Issue

Getting the scheme guessing correct is [quite tricky](https://github.com/curl/curl/issues/4856).

Furthermore, it can be misleading if the user assumes a local path exists, but doesn't. In this case, we make an unexpected network request.
This can even go unnoticed in CI/CD in case there happens to be a URL with the same name.

For example, assume we want to check a ZIP archive (which isn't supported, but might be in the future).
Furthermore, assume the file doesn't exist.
If we run:

```sh
lychee --dump-inputs url.zip
http://url.zip/
```

This would assume `http://url.zip/` instead! 💥 
(Yes, [`.zip` is a TLD](https://en.wikipedia.org/wiki/.zip_(top-level_domain)).)

## Proposal

I propose to remove the fallback to `http`. This is a breaking change, which we should make before 1.0.

## Necessary changes

Remove the condition here and return an error instead:
https://github.com/lycheeverse/lychee/blob/1780aa0daa70462e296f9811e06178f13fd945db/lychee-lib/src/types/input.rs#L180-L187 

Add a test to prove correct behavior.

Pull requests greatly appreciated.

mre:
Tackling this in #1837. Feedback welcome! Also, if someone could try it, I'd be super thankful.

```sh
cargo install --git https://github.com/lycheeverse/lychee/ --branch absolute-local-windows-paths
```

thomas-zahner:
Resolved via #1837

PR #1837, opened 2025-09-03
Fix Windows Absolute Path Parsing and Remove HTTP Assumption

mre:
This commit fixes issue #972 where Windows absolute paths like C:\path were incorrectly parsed as URLs with scheme `C:`.

I also took that as an opportunity to finaly remove the assumption that non-existent inputs get parsed as HTTP URLs. So an input `foo` would previously have been automatically converted to `http://foo/`, which is a bold and mostly incorrect presumption.  You can read more about it in the issue [here](https://github.com/lycheeverse/lychee/issues/1595).

Key changes in this PR:
- Added WindowsPath newtype with proper detection using pattern matching
- Moved Windows path logic to separate submodule for better organization
- Removed automatic HTTP assumption (foo -> http://foo/)
- Added InvalidInput error type with helpful error messages
- Updated all tests to reflect new behavior

Fixes https://github.com/lycheeverse/lychee/issues/972 and https://github.com/lycheeverse/lychee/issues/1595


katrinafyi:
I have quick comments. The first, and easiest, is that lowercase drive letters should be permitted. The second is that I think that the Unix and Windows logic should be unified so Windows also gains helpful hints like the full URL suggestion. The third is that (imo) it would be preferable to avoid custom WindowsPath parsing and logic. Windows paths are notoriously complicated.  Is there a reason why using PathBuf on Windows is not sufficient? Is there a need to recognise windows paths on Unix platforms? 

It would also be nice if there was a CI job to run windows tests.

mre:
Thanks for the feedback; I'll look into it.

mre:
Sorry for dropping the ball on this. Just a reminder what needs to be done so that I can quickly jump back in when I find the time:

- rebase
- support lowercase Windows paths
- Try to unify Unix and Windows logic
- Try to use PathBuf for Windows instead of WindowsPath

lychee-lib/src/types/input/input.rs
```diff
@@ -120,20 +120,26 @@ impl Input {
                     return;
                 }
                 InputSource::FsPath(ref path) => {
```
katrinafyi, line 122: if you tack a predicate onto the case, it can avoid indenting the entire match arm.
```suggestion
                InputSource::FsPath(ref path) if !skip_missing => {
```

lychee-lib/src/types/error.rs
```diff
@@ -116,6 +116,10 @@ pub enum ErrorKind {
     #[error("Invalid file path: {0}")]
     InvalidFile(PathBuf),
 
+    /// The given input is neither a valid file path nor a valid URL
+    #[error("{0}")]
+    InvalidInput(String),
```
katrinafyi, line 115: the doc comment here already prescribes the function of this error, so the error should be made less flexible. i.e., the message `"Input '{input}' not found as file and not a valid URL. Use full URL (e.g., https://example.com) or check file path."` should be moved into `[#error(...)]` and the error's data should be `input` rather than any error message.

or, the doc comment can be made more general.

lychee-lib/src/types/input/source.rs
```diff
@@ -52,21 +54,65 @@ impl InputSource {
...
+    /// - The input looks like a domain or word without a URL scheme (e.g., `example.com`)
+    /// - The input is a URL with an unsupported scheme (only `http://` and `https://` are supported)
+    /// - The glob pattern syntax is invalid
     pub fn new(input: &str, glob_ignore_case: bool) -> Result<Self, ErrorKind> {
         if input == Self::STDIN {
             return Ok(InputSource::Stdin);
         }
 
-        // We use [`reqwest::Url::parse`] because it catches some other edge cases that [`http::Request:builder`] does not
+        // Detect drive-letter paths with `Path::is_absolute()` This handles
```
katrinafyi: I'm really sorry to do this to you but.. why is this so complicated? I was hoping that removing the HTTP assumption would make it shorter, not longer.

I think that having a heuristic for things which are unambiguous "enough" to bypass skip-missing is weird and unexpected. The --skip-missing help would have to be changed to reflect this, but it's really hard to explain to the user (what counts as ambiguous? it's completely arbitrary).

It could be simplified a lot by changing --skip-missing slightly:
- If a command-line input is not-URL not-glob and doesn't exist, it will always be reported as an error. This avoids needing to decide if it's unambiguous enough.
- skip-missing only applies to "discovered" inputs, such as permission denied while traversing directories or cannot read file.

With this change, the code would look [something like this](https://github.com/rina-forks/lychee/blob/isthisenough/lychee-lib/src/types/input/source.rs#L65-L99).

.github/workflows/ci.yml
```diff
@@ -36,6 +36,18 @@ jobs:
       - name: Test
         run: make test
 
+  test-windows:
+    runs-on: windows-latest
```
katrinafyi: hooray!

lychee-lib/src/types/input/source.rs
mre: No worries. Thanks for the feedback. Your solution looks very clean. I will look into integrating it into my PR.

lychee-lib/src/types/input/input.rs
mre: Nice one.

katrinafyi (commented):
To get this merged, it might be easier to enable windows CI and tests in a follow-up PR. Removing http assumption by itself is an important fix that should not be stuck behind windows. I think we need a proper strategy for writing cross platform tests.

mre:
I took another stab at it today and addressed the outstanding review feedback. Moved out the Windows CI stuff to a different branch. @katrinafyi, can you take another look?

lychee-bin/tests/cli.rs
```diff
@@ -691,13 +691,13 @@ mod cli {
...
-    fn test_missing_file_ok_if_skip_missing() {
+    fn test_fails_if_input_file_missing_even_with_skip_missing() {
         let filename = format!("non-existing-file-{}", uuid::Uuid::new_v4());
         cargo_bin_cmd!()
             .arg(&filename)
             .arg("--skip-missing")
             .assert()
-            .success();
+            .failure();
     }
```
mre, line 702: Think we established that we only ignore missing files if the get "discovered" during the file scanning process and that we error out on non-existent input files. Correct me if I'm wrong here.

lychee-lib/src/types/error.rs
```diff
@@ -112,6 +112,12 @@ pub enum ErrorKind {
     #[error("Invalid file path: {0}")]
     InvalidFile(PathBuf),
```
katrinafyi: I believe this error isn't used anymore and could be removed?

lychee-lib/src/types/input/source.rs
```diff
@@ -87,45 +88,17 @@ impl InputSource {
...
-        // On Windows, a filepath can never be mistaken for a
-        // URL, because Windows filepaths use `\` and URLs use
-        // `/`
-        #[cfg(windows)]
+        // Assume all non-URL non-glob things are intended to be file paths.
+        // The error message should be general enough to suggest solutions to the user.
+        //
+        // For missing files, we defer validation to the stream processing level where
+        // --skip-missing can take effect. Domain-like inputs still get immediate
+        // InvalidInput errors for better user experience.
```
katrinafyi: I think this comment is wrong after the skip missing changes. It could just be deleted.

```suggestion
```

lychee-lib/src/types/input/source.rs
```diff
@@ -65,13 +65,14 @@ impl InputSource {
...
-            };
+            // Weed out Windows drive letters which will be incorrectly accepted as URLs.
+            match url.scheme() {
+                scheme if scheme.len() > 1 => return Ok(InputSource::RemoteUrl(Box::new(url))),
+                _ => {
+                    // Single character scheme (likely Windows drive letter)
+                    // Continue to file path handling
+                }
+            }
         }
```
katrinafyi: In https://github.com/lycheeverse/lychee/pull/2005, I added `parse_url_or_path` for handling this because it's used in the `--base-url` argument parsing too.

(you'll have to change the line above this block too - github won't let me include it in the suggestion)

```suggestion
        let input = match utils::url::parse_url_or_path(input) {
            Ok(url) => return Ok(InputSource::RemoteUrl(Box::new(url))),
            Err(input) => input,
        };
```

katrinafyi (approved):
Looks good, a few small comments.

lychee-lib/src/types/input/source.rs
mre: Ah, very nice; changed.

lychee-lib/src/types/input/source.rs
mre: True. Removed in the latest commit.

lychee-lib/src/types/error.rs
mre: That's correct. Thanks!

mre:
Nice! I've made those changes. It's much cleaner now. Thanks for the quick turnaround. 👍

katrinafyi (approved):
looks good on my end!

mre:
Thanks for reviewing @katrinafyi. Went ahead and merged this. @thomas-zahner fyi.

lychee-bin/tests/cli.rs
thomas-zahner: What does "getting discovered during the file scanning process" mean exactly? 😅
Or do you have a comment to point me to where you established this?

PR #1839, opened 2025-09-05
fix: add support for name attributes in HTML fragment extraction

### LINKED ISSUE #1838: Lychee cannot find fragment
Lychee 20.1 fails to find fragments

Example:

```
lychee-x86_64-windows.exe --include-fragments https://tinylog.org/javadoc/org/tinylog/ThreadContext.html
```

Output:

```
  20/20 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links

[https://tinylog.org/javadoc/org/tinylog/ThreadContext.html]:
   [ERROR] https://tinylog.org/javadoc/org/tinylog/ThreadContext.html#clear-- | Cannot find fragment
   [ERROR] https://tinylog.org/javadoc/org/tinylog/ThreadContext.html#method.detail | Cannot find fragment
   [ERROR] https://tinylog.org/javadoc/org/tinylog/ThreadContext.html#method.summary | Cannot find fragment
   [ERROR] https://tinylog.org/javadoc/org/tinylog/ThreadContext.html#skip.navbar.bottom | Cannot find fragment
   [ERROR] https://tinylog.org/javadoc/org/tinylog/ThreadContext.html#skip.navbar.top | Cannot find fragment

🔍 20 Total (in 0s) ✅ 13 OK 🚫 5 Errors ⛔ 2 Unsupported
```

If I open the links in any browser, it jumps to the fragment,

mre:
Fixes fragment checking for JavaDoc-generated HTML which uses <a name="anchor"> instead of id attributes for anchors.

This resolves a regression where lychee v0.20.1 was failing to find fragments that worked in v0.18.1, particularly for JavaDoc URLs like:
- https://example.com/javadoc/Class.html#method--
- https://example.com/javadoc/Class.html#skip.navbar.top

The fix maintains backward compatibility by checking both 'id' and 'name' attributes when extracting fragments from HTML documents.

Resolves #1838

### DIFF
--- a/lychee-lib/src/extract/html/html5gum.rs
+++ b/lychee-lib/src/extract/html/html5gum.rs
@@ -235,6 +235,12 @@ impl LinkExtractor {
             self.fragments.insert(id.to_string());
         }
 
+        // Also check for 'name' attributes for backward compatibility with older HTML
+        // standards and JavaDoc-generated HTML which uses <a name="anchor"> instead of id
+        if let Some(name) = self.current_attributes.get("name") {
+            self.fragments.insert(name.to_string());
+        }
+
         self.current_attributes.clear();
     }
 }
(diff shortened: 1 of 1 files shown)


lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -235,6 +235,12 @@ impl LinkExtractor {
             self.fragments.insert(id.to_string());
         }
 
+        // Also check for 'name' attributes for backward compatibility with older HTML
+        // standards and JavaDoc-generated HTML which uses <a name="anchor"> instead of id
```
thomas-zahner: Maybe you could add: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#name or https://stackoverflow.com/a/484781

So apparently in HTML 4.01it both could be used. Theoretically, it's no longer valid HTML 5, but it's still used by some tools/sites because of historical reasons.

thomas-zahner (approved):
Thanks for the fix :+1:

lychee-lib/src/extract/html/html5gum.rs
mre: Ah, neat. I've extended the documentation accordingly.

PR #1842, opened 2025-09-05
Add --default-extension argument to force file format

### LINKED ISSUE #1665: Add argument to force file format
In https://github.com/lycheeverse/lychee/pull/673, @HU90m suggested:

> My preference would be a [`ripgrep`](https://github.com/BurntSushi/ripgrep)/[`fd`](https://github.com/sharkdp/fd) style type option, specifically a `--default-type` option which allows the user to override the behaviour for a file without an extension, e.g. `--default-type html`.

We should introduce that `--default-type` argument.

mre:
## Summary
Fixes #1665 by implementing a `--default-extension` option that allows users to specify a default file extension for files without extensions.

## Problem
Users needed a way to manually specify the file format for files without clear extensions. This is common with:
- README files without extensions  
- Configuration files
- Scripts without standard extensions
- Files with ambiguous extensions

## Solution
Added `--default-extension` CLI option that:
- Accepts any valid file extension (e.g., `md`, `html`, `txt`)
- Uses the extension to determine the file type for processing
- Gracefully falls back to plaintext for unknown extensions
- Follows the same pattern as the existing `--extensions` option

## Examples
```bash
# Treat README as markdown
lychee --default-extension md README

# Treat index as HTML
lychee --default-extension html index  

# Treat CHANGELOG as plaintext
lychee --default-extension txt CHANGELOG

# Works with any supported extension
lychee --default-extension htm file_without_ext
lychee --default-extension markdown docs
```

## Implementation Details
- **CLI Option**: Added `--default-extension` argument to `Config` struct
- **File Type Processing**: Enhanced `FileType::from_extension()` to be public
- **Input Processing**: Updated `inputs()` method to convert extension to `FileType` hint
- **Error Handling**: Invalid extensions gracefully fall back to default plaintext behavior
- **Testing**: Comprehensive unit and integration tests

## Design Decision: Extensions vs Types
Initially considered using abstract types (`--default-type markdown`), but switched to extensions (`--default-extension md`) because:
- **More intuitive**: Users work with familiar file extensions
- **More flexible**: Easy to support any extension lychee recognizes
- **Consistent**: Matches the existing `--extensions` option pattern
- **Extensible**: Could easily support multiple extensions in the future

## Test Plan
- [x] Unit tests for `FileType::from_extension()` with valid/invalid extensions
- [x] Integration tests for CLI option with different file types
- [x] Error handling test for unknown extensions  
- [x] Manual testing with various file scenarios
- [x] All existing tests still pass

## Verification
**Before:**
Files without extensions were always treated as plaintext, potentially missing links in markdown/HTML content.

**After:** 
```bash
$ lychee --default-extension md README
# Now extracts markdown links from README file

$ lychee --default-extension html index  
# Now extracts HTML links from index file
```

The feature integrates seamlessly with existing functionality while providing users the control they need over file type detection.


lychee-lib/src/types/file.rs
```diff
@@ -281,4 +282,24 @@ mod tests {
...
+            FileType::from_extension("markdown"),
+            Some(FileType::Markdown)
+        );
+        assert_eq!(FileType::from_extension("md"), Some(FileType::Markdown));
+        assert_eq!(FileType::from_extension("MD"), Some(FileType::Markdown));
+        assert_eq!(FileType::from_extension("txt"), Some(FileType::Plaintext));
+        assert_eq!(FileType::from_extension("TXT"), Some(FileType::Plaintext));
+
+        // Invalid extension
+        assert_eq!(FileType::from_extension("invalid"), None);
```
thomas-zahner:
```suggestion
        assert_eq!(FileType::from_extension("unknown"), None);
```

An "invalid" file extension doesn't exist as to my understanding, right? Just extensions we don't know or handle.

lychee-lib/src/types/file.rs
```diff
@@ -171,7 +171,8 @@ impl FileType {
     }
 
     /// Get the [`FileType`] from an extension string
-    fn from_extension(ext: &str) -> Option<Self> {
```
thomas-zahner, line 174: Would prefer `ext` to be named `extension`. We could then rename `from_extension` to just `from`. Wouldn't it also make sense to instead do `impl From<&str> for FileType`?

README.md
```diff
@@ -371,6 +371,11 @@ Options:
 
           [default: md,mkd,mdx,mdown,mdwn,mkdn,mkdown,markdown,html,htm,txt]
 
+      --default-extension <EXTENSION>
+          Default file extension to treat files without extensions as having.
```
thomas-zahner, line 386: I'm really confused by this sentence 😅

```suggestion
          This is the default file extension that is applied to files without an extension.
```

lychee-bin/tests/cli.rs
```diff
@@ -2845,4 +2845,61 @@ mod cli {
...
+            .arg("--dump")
+            .arg(html_file_without_ext.path())
+            .assert()
+            .success()
+            .stdout(contains("https://html-example.com"));
+
+        Ok(())
+    }
+
+    /// Test that invalid --default-extension values are handled gracefully
```
thomas-zahner: Same here. I don't think there is a concept of "invalid" extensions. Just unknown extensions. So I would replace all `invalid` occurrences here with `unknown`.

lychee-lib/src/types/file.rs
mre: Yes, that's better.

README.md
mre: Probably just me editing the wording over and over again until I broke the grammar.

lychee-lib/src/types/file.rs
mre: Good idea. Done.

mre: I'm not 100% sure about `impl From<&str> for FileType`. We would lose some semantic meaning. 🤔

lychee-bin/tests/cli.rs
mre: done

lychee-lib/src/types/file.rs
thomas-zahner: Sure, it's fine if we keep this

README.md
thomas-zahner: @mre I noticed now that you merged your version. Did you forget about changing it? I still think this original version is quite confusing :smile:

mre: Oh, that was by mistake. I thought I updated the text but apparently not.

Issue #1768, opened 2025-07-16
Keep example config file and documentation up to date


katexochen:
This can currently only be configured via flag.

thomas-zahner:
I think there is no special distinction between CLI flags and the config files. We simply merge them at some point. Clap just renames `cache_exclude_status` to `cache-exclude-status` for CLI args. So can you try to make use add `cache_exclude_status` in the config file? Something like `cache_exclude_status = ["123", "456"]` should work.

So if it works this is a documentation issue really.

katexochen:
> Something like cache_exclude_status = ["123", "456"] should work.

yes, that works, thanks!

> So if it works this is a documentation issue really.

Agree. It would be nice to have the example config file up to date (and maybe include the defaults?).

I've updated the issue title to match that.

thomas-zahner:
Thanks, I wanted to propose exactly the same. So to properly resolve this issue we should do the following things:

- [x] Write a test that validates that every single argument is included in lychee.example.toml
- [x] Find a way to include the lychee.example.toml from the main branch into the [documentation](https://github.com/lycheeverse/lycheeverse.github.io/blob/90253d1c03713172585a27dd5fb09dfca787b062/src/content/docs/usage/config.md?plain=1#L28)

This way we get full documentation coverage 🙂 (similar to how we test the contents the README.md with test_readme_usage_up_to_date)

kemingy:
I found this interesting and it could be a general problem: How can we automatically retrieve all the fields and related documents from the Rust struct definition?

Clap should have this internally. I didn't find any public interface. But there is another lib [facet](https://github.com/facet-rs/facet). I suppose we can use this one to extract all the options from [`Config`](https://github.com/lycheeverse/lychee/blob/e743ea3f5f02cd989452229f51b2b0c139037ff9/lychee-bin/src/options.rs#L354) and dump to a local file.

mre:
@thomas-zahner made a lot of improvements adressing the issues mentioned here in #1906. Can we close this or is there more work left to be done?

thomas-zahner:
[...]

PR #1843, opened 2025-09-07
fix: add missing fields to `fold_in` macro

katrinafyi:
the [`fold_in!` macro](https://github.com/lycheeverse/lychee/blob/b40fc2c32a62919f185dd85ec3c52813b8b2d2ae/lychee-bin/src/options.rs#L199) relies on the listed fields in order to know which fields to merge from the toml file. if fields were missing from this list, as they were previously, then the toml file values for those fields would be silently ignored.

this pr adds those missing fields, except for header and github_token which are handled separately. this also sorts the field list alphabetically.

it would be nice if there was a "kitchen sink" toml file which had every option set. maybe this could be auto-generated.

command to list all config fields:
```bash
grep --after-context=10000 'pub(crate) struct Config' \
  lychee-bin/src/options.rs | grep '    pub(crate) ' \
  | cut -d' ' -f6 | grep : | sort
```
getting all fields within `fold_in!` was done by vim `ya{`


mre:
Nice. Yeah, that sort of problem keeps resurfacing. Would be nice to get a compile-time error if we forgot to fold a field.

Or maybe we can look into alternatives like https://github.com/gibfahn/clap_config or similar, but that would be a bigger change. It might not be so easy because we also want to skip a few fields as you mentioned.

> it would be nice if there was a "kitchen sink" toml file which had every option set. maybe this could be auto-generated (related to https://github.com/lycheeverse/lychee/issues/1768).

The closest thing to that is https://github.com/lycheeverse/lychee/blob/master/fixtures/configs/smoketest.toml but I'm sure there are missing options by now. Might be worth investigating.

Either way, thanks for fixing.

Issue #1846, opened 2025-09-09
See if `truncate_source` can be removed


mre:
In #1832, we moved to `Cow` for `InputSource::String`.

We should consider removing this method or alternatively updating the TODO:
https://github.com/lycheeverse/lychee/blob/d6b16bc457c4125eb3d102b6b319fb8f11802122/lychee-lib/src/utils/request.rs#L115-L120

_Originally posted by @katrinafyi in https://github.com/lycheeverse/lychee/issues/1832#issuecomment-3259380754_

PR #1847, opened 2025-09-10
refactor: Remove truncate_source and inline string truncation logic

XmchxUp:
Closes: #1846

### DIFF
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -2,6 +2,7 @@ use log::warn;
 use percent_encoding::percent_decode_str;
 use reqwest::Url;
 use std::{
+    borrow::Cow,
     collections::HashSet,
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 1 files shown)


mre:
Thanks for the PR. It's been a while since I found the time to review it. Looks good overall.

lychee-lib/src/utils/request.rs
```diff
@@ -29,8 +30,23 @@ fn create_request(
...
+                    Cow::Owned(s_owned) => ResolvedInputSource::String(Cow::Owned(s_owned.clone())),
+                }
+            } else {
+                ResolvedInputSource::String(Cow::Owned(
+                    s.chars().take(MAX_TRUNCATED_STR_LEN).collect(),
+                ))
+            }
+        }
+        other => other.clone(),
+    };
```
mre: Do we even need to truncate the input at all? If so, should we add a comment why this is still needed? I did not run any benchmarks, but my thinking was that we wanted to avoid expensive clones in the past, but with `Cow` it shouldn't be a problem anymore.

XmchxUp: You’re right — with `Cow` we no longer need to worry about expensive clones. I can’t think of any other reason for keeping the `truncate_source` logic, so I’d suggest removing it altogether.

mre:
Thanks a lot @XmchxUp! ✌️

PR #1849, opened 2025-09-11
Add WinGet installation instructions

bdukes:
Related to https://github.com/lycheeverse/lycheeverse.github.io/pull/100

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,12 @@ pkg install lychee
 apk add lychee
 ```
 
+### WinGet (Windows)
+
+```sh
+winget install --id lycheeverse.lychee
+```
+
 ### Chocolatey (Windows)
 
 ```sh


README.md
```diff
@@ -119,6 +119,12 @@ pkg install lychee
 apk add lychee
 ```
 
+### WinGet (Windows)
+
+```sh
+winget install --id lycheeverse.lychee
```
thomas-zahner, line 125: I've never used `winget`. Is there a reason for using `--id` or omitting it? In the docs it wasn't specified.

bdukes: If you don't use `--id`, what you pass is a `--query` parameter, which can become ambiguous if there are other packages with similar names. In practice, it's not necessary.

bdukes: See https://learn.microsoft.com/en-us/windows/package-manager/winget/install#multiple-selections

thomas-zahner: Ah I see. Let's leave the `--id` argument then.

thomas-zahner:
@bdukes Thanks for improving the documentation and README :+1:

PR #1850, opened 2025-09-11
fix: do not reuse excluded status from cache

### LINKED ISSUE #1770: --cache-exclude-status still uses existing cache entries that are excluded
I think this might not be what people expect. Let's say you are using a cache, for example in CI, and you are adding the `--cache-exclude-status` flag at some point, for example ignoring status 418. If there are already cache entries for that, they will still be used and pages that previously returned 418 aren't checked again. I think this is counter intuitive and annoying in some cases. You'll either have to clean these cache entries manually or throw away the whole cache.

Instead, lychee could ignore existing cache entries with status that is excluded by `--cache-exclude-status`.

charludo:
Ensures that Uris in the cache whose status has in the meantime been excluded will not be reused.

If any more skips end up in `Cache::load`, it would probably make sense to pass the entire config to it, rather than bits and pieces.

Fixes: #1770

### DIFF
--- a/lychee-bin/src/cache.rs
+++ b/lychee-bin/src/cache.rs
@@ -1,7 +1,7 @@
 use crate::time::{self, Timestamp, timestamp};
 use anyhow::Result;
 use dashmap::DashMap;
-use lychee_lib::{CacheStatus, Status, Uri};
+use lychee_lib::{CacheStatus, Status, StatusCodeExcluder, Uri};
 use serde::{Deserialize, Serialize};
 use std::path::Path;
 
@@ -33,7 +33,11 @@ pub(crate) trait StoreExt {
     fn store<T: AsRef<Path>>(&self, path: T) -> Result<()>;
 
     /// Load cache from path. Discard entries older than `max_age_secs`
-    fn load<T: AsRef<Path>>(path: T, max_age_secs: u64) -> Result<Cache>;
+    fn load<T: AsRef<Path>>(
+        path: T,
+        max_age_secs: u64,
+        excluder: &StatusCodeExcluder,
+    ) -> Result<Cache>;
 }
 
 impl StoreExt for Cache {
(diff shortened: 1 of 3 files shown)


lychee-bin/src/cache.rs
```diff
@@ -58,10 +66,55 @@ impl StoreExt for Cache {
...
-            if current_ts - value.timestamp < max_age_secs {
-                map.insert(uri, value);
+            if current_ts - value.timestamp >= max_age_secs {
+                continue;
             }
+
+            // Discard entries for status codes which have been excluded.
+            // Without this check, an entry might be cached, then its status code is configured as
+            // excluded, and in subsequent runs the cached value is still reused.
+            if Option::<u16>::from(value.status)
```
thomas-zahner: I find this a bit hard to read. I would propose to move the logic into CacheStatus for better readability.

```rust
impl CacheStatus {
    /// Returns `true` if the cache status is excluded by the given [`StatusCodeExcluder`].
    pub fn is_excluded(&self, excluder: &StatusCodeExcluder) -> bool {
        if let Some(status) = Option::<u16>::from(*self) {
            excluder.contains(status)
        } else {
            false
        }
    }
}
```

And then here:

```rust
if value.status.is_excluded(excluder) {
    continue;
}
```

thomas-zahner (changes requested):
Thanks for addressing this bug :rocket:

lychee-bin/src/cache.rs
charludo: Yeah good call. I took the liberty to simplify the `is_excluded` suggestion a bit more.

thomas-zahner: Sure, makes sense :+1:

charludo:
Sorry to ask here, but: did I receive an actual email regarding this PR a couple of hours ago, or was that an incredibly elaborate phishing attempt? 😅

mre:
Nope, it's real. I sent it. :)

thomas-zahner:
@charludo Can confirm that the NLnet funding is indeed real :)
Also see https://github.com/lycheeverse/lychee/commit/24b39b8e2fb67be4ab45501cf0e0c26f745930ff.

PR #1852, opened 2025-09-12
add cargo-msrv to the devshell

charludo:
Just noticed that this was missing and `make verify` is therefore not available ootb.


thomas-zahner:
@charludo Ah right, I forgot about this. Thanks for the PR :+1:

PR #1854, opened 2025-09-15
docs: hard wrap help text for options

katrinafyi:
most of the options already have hard-wrapped `long_help`. however, for a few options which defined their help text in `///` comments, these would get joined into one long line. this was the `[inputs]`, `--scheme`, and `--cookie-jar`.

i've sorted the lines in `--help` by length before and after this change:
[old.txt](https://github.com/user-attachments/files/22330603/old.txt)
[new.txt](https://github.com/user-attachments/files/22330606/new.txt)

previously, the longest line was 353 characters and after this change it is 126.

this change will make the [cli options webpage](https://lychee.cli.rs/usage/cli/) a bit easier to read, as it makes you scroll to the right if the line is long. though, maybe other solutions can be done instead (maybe generate markdown from the options, or maybe enable wrapping on the webpage).

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,10 +328,10 @@ Usage: lychee [OPTIONS] [inputs]...
 
 Arguments:
   [inputs]...
-          Inputs for link checking (where to get links to check from). These can be: files (e.g. `README.md`),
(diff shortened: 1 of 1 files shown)


mre:
Definitely an improvement. Thanks!

Issue #1606, opened 2025-01-06
Support relative paths for `--root-dir` 


mre:
In #1576, we added support for the `--root-dir` parameter.
It is useful for checking absolute links, especially when the rendered documents are located in a subdirectory.

## Common usage scenarios for `--root-dir`

- Static site builds in a public or dist directory
- Documentation sites with absolute links
- Any project where links start with `/`

Here's a usage example

```sh
# Assuming the output html files get rendered into the `public` directory
lychee --root-dir "$(pwd)/public" "**/*.html"
```

Full documentation [here](https://lychee.cli.rs/recipes/root-dir/).

## Proposal

Add support for relative paths, e.g. `.` or `./public`.

That would be very convenient and relatively easy to implement.
We could apply [`std::fs::canonicalize`](https://doc.rust-lang.org/beta/std/fs/fn.canonicalize.html) when we handle the `root-dir`:

```rust
use std::fs::canonicalize;
use std::path::PathBuf;

fn main() {
        let path = PathBuf::from(".");
        println!("{:?}", canonicalize(path));
}
```

Our argumentation for not doing that in the first place was that we wanted to be conservative and avoid running into edge-cases. However, after playing with the feature a bit, I'm beginning to think there is no downside to supporting that.

The above example would be simplified to

```sh
lychee --root-dir public "**/*.html"
```

FYI @trask, since you worked on the initial implementation.

*Edit: Removed my other proposal of making `--root-dir .` the default. I'm no longer confident that this is a good idea.*

YDX-2147483647:
https://lychee.cli.rs/recipes/root-dir/#the-difference-between-root-dir-and-base-url
> If you have a use case for allowing relative root directories, please [open an issue](https://github.com/lycheeverse/lychee/issues).

I have one: specifying `args` for lychee-action, when `pwd` is not available.
`${{ github.workspace }}` works, but that might be too compilcated for beginners.

mre:
Hum, from what I can see, `${{ github.workspace }}` works for your use-case? Shall we focus on better documentation at this point? E.g. we could add a note to the page you mentioned. 

If you like, and you have some time, would you mind opening a pull request to update the page?
Source code is here: https://github.com/lycheeverse/lycheeverse.github.io/edit/master/src/content/docs/recipes/root-dir.mdx

YDX-2147483647:
> would you mind opening a pull request to update the page?

Sure! #94

dkarlovi:
[...]

PR #1857, opened 2025-09-23
feat(base-url)!: disallow relative local base to avoid confusion

katrinafyi:
the previous behaviour accepted relative directory paths as bases but this led to later InvalidBaseJoin errors, because relative bases cannot be used to join relative URLs. this means that relative local bases were *only* useful for resolving root-relative links, and were confusingly problematic with ordinary relative links.

see https://github.com/lycheeverse/lychee/issues/1574 which talks about errors when passing `--base ../network-documentation/` or, in a later comment, `--base build`. it is my opinion that it is better to fail earlier in these cases, so the user is not hit with mysterious InvalidBaseJoin.

also, the previous behaviour would parse something like `--base-url google.com` to a local base pointing to `./google.com`. this would also lead to downstream errors or incorrect URLs and it's better to guard against this.

after this pr, trying to use a non-absolute local base will be a CLI parsing error:
```
error: invalid value 'build' for '--base-url <BASE_URL>': Error with
base dir `build` : Base must either be a URL (with scheme) or an
absolute path. Alternatively, if you want to resolve root-relative links
in local files, see `--root-dir`.
```

in a slightly opinionated touch, i've mentioned --root-dir in the error message, since i think --root-dir is more suitable for most use cases where people try to use relative local bases. this agrees with later comments in https://github.com/lycheeverse/lychee/issues/1574.

however, this does make the error message quite long. so i'm happy to take on feedback and changes about this. also, i'm not quite sure which [conventional commit](https://release-plz.dev/docs/changelog/format#how-should-i-write-my-commits) verb to use in this PR title. i've put "feat" for now.

also, an alernative approach to think about is to ban "local" bases altogether and require use of `file:///` for local paths. this would unify the argument under the URL format and maybe help prod people towards --root-dir for local paths.

also, another alternative instead of doing `path.is_absolute()` is to make a Local base then try to `.join()` with it and ensure the join succeeds. this will more precisely guard against join failures (and hence, InvalidBaseJoin errors)

this pr implements (1) in my outline to reduce InvalidBaseJoin confusion, as described in https://github.com/lycheeverse/lychee/pull/1624#issuecomment-3274485963

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,7 +581,7 @@ Options:
   -b, --base-url <BASE_URL>
           Base URL to use when resolving relative URLs in local files. If specified,
           relative links in local files are interpreted as being relative to the given
-          base URL. A base URL must either be a URL (with scheme) or an absolute path.
+          base URL.
 
           For example, given a base URL of `https://example.com/dir/page`, the link `a`
           would resolve to `https://example.com/dir/a` and the link `/b` would resolve
@@ -595,6 +595,9 @@ Options:
 
           Basically, the base URL option resolves links as if the local files were hosted
           at the given base URL address.
+
+          The provided base URL value must either be a URL (with scheme) or an absolute path.
+          Note that certain URL schemes cannot be used as a base, e.g., `data` and `mailto`.
 
       --root-dir <ROOT_DIR>
           Root directory to use when checking absolute links in local files. This option is
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -737,7 +737,7 @@ separated list of accepted status codes. This example will accept 200, 201,
         value_parser = parse_base,
         long_help = "Base URL to use when resolving relative URLs in local files. If specified,
 relative links in local files are interpreted as being relative to the given
-base URL. A base URL must either be a URL (with scheme) or an absolute path.
+base URL.
 
 For example, given a base URL of `https://example.com/dir/page`, the link `a`
 would resolve to `https://example.com/dir/a` and the link `/b` would resolve
@@ -750,7 +750,10 @@ of the base URL. For example, a base URL of `https://example.com/dir/page/` and
 a link of `a` would resolve to `https://example.com/dir/page/a`.
 
 Basically, the base URL option resolves links as if the local files were hosted
-at the given base URL address."
+at the given base URL address.
+
+The provided base URL value must either be a URL (with scheme) or an absolute path.
+Note that certain URL schemes cannot be used as a base, e.g., `data` and `mailto`."
     )]
     #[serde(default)]
     pub(crate) base_url: Option<Base>,


lychee-lib/src/types/base.rs
```diff
@@ -61,7 +61,24 @@ impl TryFrom<&str> for Base {
...
+        } else {
+            Err(ErrorKind::InvalidBase(
+                value.to_string(),
+                [
+                    "Base must either be a URL (with scheme) or an absolute path.",
+                    "Alternatively, if you want to resolve root-relative links in",
+                    "local files, see `--root-dir`.",
+                ]
+                .join(" "),
+            ))
```
mre: Do you think we should add a few examples or reference to the docs here?

katrinafyi: For adding examples, I think that this message location isn't quite suitable, the current message is already relatively large. Maybe a link to the docs would be okay, but idk which docs. The [base url page](https://lychee.cli.rs/recipes/base-url/) is not super informative.

Maybe it could mention that more info is in --help? But that's also kind of obvious.

mre: You're right.

The base url page is definitely not great and needs a makeover at some point. Besides, hardcoding website URLs is probably not a good idea. Would be quite ironic if that link breaks in the future. 😆 

Setting up the base url and root dir are probably the trickiest things to get right from a user's perspective, so in this case mentioning `--help` might be a good idea.

katrinafyi:
I've made some changes in response to the review comments, but is the CI workflow working? It's failed in a setup step twice in a row.

Also, I'm wondering if this PR's approach is a good idea. It's adding a message to `TryFrom<&str> for Base` that mentions command-line flags, but this is part of lychee-lib and could be used entirely without the CLI.

mre:
> is the CI workflow working? It's failed in a setup step twice in a row.

Yeah, I noticed that in a few other projects as well.
https://github.com/lycheeverse/lychee/pull/1862

mre:
> Also, I'm wondering if this PR's approach is a good idea. 
The approach is fine, I think. Or at least it's a step into the right direction to avoid confusion around these options.

> It's adding a message to TryFrom<&str> for Base that mentions command-line flags, but this is part of lychee-lib and could be used entirely without the CLI.

Yeah, that's indeed not great. Sorry for not noticing this sooner and leading you astray. Maybe we can instead add more `.context` whereever this `TryFrom` gets called in the CLI for a cleaner separation of concerns.

mre:
https://github.com/lycheeverse/lychee/pull/1862 is merged so you can rebase.

katrinafyi:
Thanks for the fix and the hint about context. I've made the changes and it makes a bit more sense now. See the commit comment, though.

lychee-bin/src/parse.rs
```diff
@@ -13,8 +13,20 @@ pub(crate) fn parse_remaps(remaps: &[String]) -> Result<Remaps> {
...
+        Ok(x) => Ok(x),
+        Err(e) => {
+            // if context is defined, clap displays only the context string in
+            // argument parse errors. to keep the message from within InvalidBase,
+            // we need to retain it manually.
+            let message = format!(
+                "{e}. See `--help` for more information. If you want to resolve \
+                root-relative links in local files, also see `--root-dir`."
+            );
+            Err(e).context(message)
```
mre, line 27: So you mean this doesn't work?

```rust
pub(crate) fn parse_base(src: &str) -> Result<Base> {
    Base::try_from(src).context(
        "See `--help` for more information. If you want to resolve \
        root-relative links in local files, also see `--root-dir`."
    )
}
```

katrinafyi: Yeah, that would only print the context message without anything from the original error. it looks like this:
```
error: invalid value 'a:datafdajsio' for '--base-url <BASE_URL>': See `--help` for more information. If you want to resolve root-relative links in local files, also see `--root-dir`.
```

katrinafyi:
I've tweaked the --base-url help text to read a bit better from top to bottom. I'm happy for it to be merged anytime now :)

mre (approved):

mre:
Good work, thanks

thomas-zahner:
[Documentation](https://lychee.cli.rs/github_action_recipes/check-repository/) is outdated:

```
      - name: Link Checker
        id: lychee
        uses: lycheeverse/lychee-action@v2
        with:
          args: --base-url dist --exclude-all-private dist
          fail: false
```

Will now lead to an error as expected. @katrinafyi how would you fix this?

I tried

```
          args: --dump --base-url "$(pwd)/dist/" --exclude-all-private dist
```

but this doesn't work. I guess this is a GitHub workflow related issue. How would you be able to access the absolute path?
I've also tried `${GITHUB_WORKSPACE}` and `${{ github.workspace }}` without success.

See https://github.com/lycheeverse/lycheeverse.github.io/pull/114 and https://github.com/lycheeverse/lycheeverse.github.io/actions/runs/19075734005/job/54490637825?pr=114

katrinafyi:
@thomas-zahner I started a fix here but there's other (unrelated) issues to do with how existing links are excluded: https://github.com/lycheeverse/lycheeverse.github.io/pull/111

In hindsight, maybe we should've been more careful with this change 😅 even if it was slightly buggy, i think base-url was very widely used...

thomas-zahner:
Ah I've missed that one. Thank you. Yes maybe we were a bit quick to release that but at the same time we're still not on v1 so I think it's okay if the docs are up to date.

mre:
> i think base-url was very widely used...

Oh yes, it was (and still is). 😬 

The new syntax is slightly more cumbersome as instead of the old `--base .` we now have to write `--root-dir $(pwd)`. Will be nicer once https://github.com/lycheeverse/lychee/issues/1606 is done.

PR #1858, opened 2025-09-24
Man pages

thomas-zahner:
WIP

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,7 +328,7 @@ See below for a full list.
 ```text
 A fast, async link checker
 
-Finds broken URLs and mail addresses inside Markdown, HTML, `reStructuredText`, websites and more!
+Finds broken URLs and mail addresses inside Markdown, HTML, reStructuredText, websites and more!
 
 Usage: lychee [OPTIONS] [inputs]...
 
--- a/README.md
+++ b/README.md
@@ -406,8 +406,6 @@ Options:
           comma-separated list of excluded status codes. This example will not cache results
           with a status code of 429, 500 and 501.
 
-          [default: ]
-
       --dump
           Don't perform any link checking. Instead, dump all the links extracted from inputs that would be checked
 
--- a/lychee-bin/Cargo.toml
+++ b/lychee-bin/Cargo.toml
@@ -20,6 +20,7 @@ lychee-lib = { path = "../lychee-lib", version = "0.20.1", default-features = fa
 anyhow = "1.0.99"
 assert-json-diff = "2.0.2"
 clap = { version = "4.5.47", features = ["env", "derive"] }
+clap_mangen = "0.2.29"
 console = "0.16.1"
 const_format = "0.2.34"
 csv = "1.3.1"
@@ -58,6 +59,7 @@ tokio = { version = "1.47.1", features = ["full"] }
 tokio-stream = "0.1.17"
 toml = "0.9.5"
 url = "2.5.7"
+chrono = { version = "0.4.42", features = ["alloc", "now", "std", "clock"] }
 
 
 [dev-dependencies]
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -311,10 +311,11 @@ impl HeaderMapExt for HeaderMap {
     }
 }
 
-/// A fast, async link checker
+/// lychee is a tool to detect broken URLs and mail addresses in
+/// local files and websites. It supports Markdown and HTML explicitly
+/// and works well with many plain text file formats.
 ///
-/// Finds broken URLs and mail addresses inside Markdown, HTML,
-/// `reStructuredText`, websites and more!
+/// lychee is powered by lychee-lib, the Rust library to for link checking.
 #[derive(Parser, Debug)]
 #[command(version, about)]
 pub(crate) struct LycheeOptions {
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -938,6 +938,8 @@ impl Config {
 mod tests {
     use std::collections::HashMap;
 
+    use clap::CommandFactory;
+
     use super::*;
 
     #[test]
@@ -1067,5 +1069,26 @@ mod tests {
                 ("X-Test".to_string(), "check=this".to_string()),
             ]
         );
+    }
+
+    #[test]
+    fn man_pages() -> std::io::Result<()> {
+        let out_dir = Path::new(env!("CARGO_MANIFEST_DIR"))
+            .parent()
+            .unwrap()
+            .to_path_buf();
+        // let out_dir = std::path::PathBuf::from(
+        //     std::env::var_os("OUT_DIR").ok_or(std::io::ErrorKind::NotFound)?,
+        // );
+
+        let date = chrono::offset::Local::now().format("%Y-%m-%d");
+        let man = clap_mangen::Man::new(LycheeOptions::command()).date(format!("{}", date));
+
+        let mut buffer: Vec<u8> = Default::default();
+        man.render(&mut buffer)?;
+
+        std::fs::write(out_dir.join("lychee.1"), buffer)?;
+
+        Ok(())
     }
 }
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -193,7 +193,6 @@ default_function! {
     retry_wait_time: usize = DEFAULT_RETRY_WAIT_TIME_SECS;
     method: String = DEFAULT_METHOD.to_string();
     verbosity: Verbosity = Verbosity::default();
-    cache_exclude_selector: StatusCodeExcluder = StatusCodeExcluder::new();
     accept_selector: StatusCodeSelector = StatusCodeSelector::default();
 }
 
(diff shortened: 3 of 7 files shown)


lychee-bin/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "lychee"
-authors = ["Matthias Endler <matthias@endler.dev>"]
+authors = ["Matthias Endler <matthias@endler.dev>", "Thomas Zahner <thomas.zahner@protonmail.ch>"]
```
thomas-zahner, line 3: @mre I've included myself as an author which feels a bit awkward :sweat_smile: But we agreed on this in the last meeting, right?

mre: Of course

lychee-bin/src/options.rs
```diff
@@ -312,12 +312,13 @@ impl HeaderMapExt for HeaderMap {
...
 }
 
-/// A fast, async link checker
+/// lychee is a tool to detect broken URLs and mail addresses in
+/// local files and websites. It supports Markdown and HTML explicitly
+/// and works well with many plain text file formats.
 ///
-/// Finds broken URLs and mail addresses inside Markdown, HTML,
-/// `reStructuredText`, websites and more!
+/// lychee is powered by lychee-lib, the Rust library to for link checking.
```
mre:
```suggestion
/// lychee is powered by lychee-lib, the Rust library for link checking.
```

lychee-bin/src/options.rs
```diff
@@ -312,12 +312,13 @@ impl HeaderMapExt for HeaderMap {
     }
 }
 
-/// A fast, async link checker
+/// lychee is a tool to detect broken URLs and mail addresses in
```
mre:
```suggestion
/// lychee is a fast, async link checker which detects broken URLs and mail addresses in
```

maybe?

lychee-bin/src/commands/generate.rs
```diff
@@ -0,0 +1,175 @@
...
+
+use crate::LycheeOptions;
+
+const CONTRIBUTOR_THANK_NOTE: &str = "\n\nA huge thank you to all the wonderful contributors who helped make this project a success.";
+
+const BUG_SECTION: &str =
+    "Report any bugs or questions to <https://github.com/lycheeverse/lychee/issues/>
+
+Questions can also be asked on <https://github.com/lycheeverse/lychee/discussions>";
+const EXAMPLES_SECTION: &str = "Check all links in supported files by specifying a directory
```
mre: While reading through the examples, I thought that it would be nice to run them such that they don't become outdated.
I was thinking about a "machine-readable" format like 

```rust
const EXAMPLES_SECTION: &[(&str, &[&str])] = &[
    (
        "Check all links in supported files by specifying a directory",
        &["lychee ."],
    ),
    (
        "Check links from stdin",
        &[
            "cat test.md | lychee -",
            "echo 'https://example.com' | lychee -",
        ],
    ),
];

```

So basically a human-readable description and a list of examples per section.

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=5aa2929962765c774504afb068f04353

Not sure if it's worth it, though. Just an idea.

lychee-bin/src/commands/generate.rs
```diff
@@ -0,0 +1,175 @@
...
+0   Success. The operation was completed successfully as instructed.
+
+1   Missing inputs or any unexpected runtime failures or configuration errors
+
+2   Link check failures. At least one non-excluded link failed the check.
+
+3   Encountered errors in the config file.
+";
+
+/// What to generate when provided the --generate flag
```
mre:
```suggestion
/// What to generate when providing the --generate flag
```

mre: Or maybe

```suggestion
/// What to generate, provided the --generate flag
```

lychee-bin/src/commands/generate.rs
```diff
@@ -0,0 +1,175 @@
...
+#[derive(Debug, Deserialize, Clone, Display, EnumIter, EnumString, VariantNames, PartialEq)]
+#[non_exhaustive]
```
mre, line 75: Thanks for thinking of `non_exhaustive` here. 👍

lychee-bin/src/commands/generate.rs
```diff
@@ -0,0 +1,175 @@
...
+#[strum(serialize_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
+pub(crate) enum GenerateMode {
+    /// Generate roff used for the man page
+    Man,
+}
+
+pub(crate) fn generate(mode: &GenerateMode) -> Result<String> {
```
mre, line 84: It's trivial, but can we add a doc-comment?

lychee-bin/src/options.rs
```diff
@@ -829,6 +828,10 @@ followed by the absolute link's own path."
     #[serde(default)]
     pub(crate) format: StatsFormat,
 
+    /// Generate special output (e.g. man page) instead of performing link checking
```
mre: Not sure, but this reads a little nicer?

```suggestion
    /// Generate special output (e.g. the man page) instead of performing link checking
```

lychee-bin/tests/usage.rs
```diff
@@ -67,4 +59,44 @@ mod readme {
...
+        let mut sorted = arguments.clone();
+        sorted.sort_by_key(|arg| arg.to_lowercase());
+
+        assert_eq!(
+            arguments, sorted,
+            "Arguments are not sorted alphabetically: {arguments:?}",
+        );
+
+        Ok(())
+    }
```
mre, line 101: Here's a slightly modified version with a few changes:

- The regex now also supports arguments, which don't have a short flag
- The error output is a bit clearer (I hope) in a sense that it shows the mismatched lines
- I use `or_else` and `map` instead of `unwrap_or_else`.


```rust
	use std::collections::HashSet;

    #[test]
    #[cfg(unix)]
    fn test_arguments_ordered_alphabetically() -> Result<(), Box<dyn std::error::Error>> {
        use regex::Regex;

        let mut cmd = main_command();
        let help_cmd = cmd.env_clear().arg("--help").assert().success();
        let help_text = std::str::from_utf8(&help_cmd.get_output().stdout)?;

        let regex = Regex::new(r"^\s{2,6}(?:-(?<short>[a-zA-Z]),)?\s*--(?<long>[a-zA-Z-]+)")?;

        let arguments: Vec<&str> = help_text
            .lines()
            .filter_map(|line| {
                let captures = regex.captures(line)?;
                captures
                    .name("short")
                    .or_else(|| captures.name("long"))
                    .map(|m| m.as_str())
            })
            .collect();

        // Check for duplicates first
        let mut seen = HashSet::new();
        let duplicates: Vec<_> = arguments.iter().filter(|arg| !seen.insert(*arg)).collect();

        if !duplicates.is_empty() {
            eprintln!("Warning: Duplicate arguments found: {duplicates:?}");
        }

        let mut sorted = arguments.clone();
        sorted.sort_by_key(|arg| arg.to_lowercase());

        if arguments != sorted {
            // Find all positions where order differs
            let mismatches: Vec<_> = arguments
                .iter()
                .zip(&sorted)
                .enumerate()
                .filter(|(_, (a, b))| a != b)
                .map(|(i, (actual, expected))| format!("  [{i}] '{actual}' should be '{expected}'"))
                .collect();

            panic!(
                "\nArguments are not sorted alphabetically!\n\nMismatches:\n{}\n\nFull actual order:\n{:?}\n\nFull expected order:\n{:?}",
                mismatches.join("\n"),
                arguments,
                sorted
            );
        }

        Ok(())
    }
```

README.md
```diff
@@ -326,9 +326,9 @@ There is an extensive list of command line parameters to customize the behavior.
 See below for a full list.
 
 ```text
-A fast, async link checker
+lychee is a tool to detect broken URLs and mail addresses in local files and websites. It supports Markdown and HTML explicitly and works well with many plain text file formats.
```
mre: How about this slight rewording?

```text
lychee is a fast, asynchronous link checker, which detects broken URLs and mail addresses in local files and websites. It supports Markdown and HTML, and works well with many plain text file formats.
```

thomas-zahner: Ah, you already suggested this change in the doc comment and now this part in the README was no longer in sync. I've updated `test_readme_usage_up_to_date` to cover this from now on.

lychee-bin/tests/usage.rs
thomas-zahner: Thanks for the improvements. But there's two things here that I didn't include. Firstly, the regex change from ` ` to `\s*` broke the tests, because it made it include lines like `--index-files logic).` which is part of the description of another flag. Secondly, I didn't include the duplicate check. This is not necessary as it is clap's responsibility to check that. Just tested what happens if I try to have a duplicate and run lychee:

```
thread 'main' panicked at /home/thomas/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.47/src/builder/debug_asserts.rs:97:17:
Command lychee: Long option names must be unique for each argument, but '--header' is in use by both 'foo' and 'header'
```

lychee-bin/src/commands/generate.rs
thomas-zahner: Done :+1:

PR #1859, opened 2025-09-29
feat: improve InvalidIndexFile error to mention index file names

katrinafyi:
this makes it clearer which files are causing the link check to fail. it also lets us customise the message in the special case of `--index-files ''`, which could otherwise be confusing with a generic message.

the messages are:
```
[ERROR] file:///h | Cannot find index file within directory: An index file (index.html) is required
[ERROR] file:///h | Cannot find index file within directory: An index file (index.html, or index.htm) is required
[ERROR] file:///h | Cannot find index file within directory: An index file (index.html, index.htm, or index.md) is required
```

and for the `--index-files ''` case:
```
[ERROR] file:///h | Cannot find index file within directory: No directory links are allowed because index_files is defined and empty
```

i do wonder if this last error message is still confusing because of the "Cannot find index file within directory" text. maybe it should be changed to a more general "Cannot resolve link to local directory" or "Invalid local directory link"?

for comparison, the old message was this:
```
[ERROR] file:///h | Cannot find index file within directory: Index file not found in directory. Check if index.html or other index files exist
```
and it was constant and never changed.

### DIFF
--- a/lychee-lib/src/types/error.rs
+++ b/lychee-lib/src/types/error.rs
@@ -70,9 +70,9 @@ pub enum ErrorKind {
     #[error("Cannot find fragment")]
     InvalidFragment(Uri),
 
-    /// The given directory is missing a required index file
+    /// Cannot resolve local directory link using the configured index files
     #[error("Cannot find index file within directory")]
-    InvalidIndexFile(PathBuf),
+    InvalidIndexFile(Vec<String>),
 
     /// The given path cannot be converted to a URI
     #[error("Invalid path to URL conversion: {0}")]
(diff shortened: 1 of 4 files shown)


lychee-lib/src/types/error.rs
```diff
@@ -330,9 +330,11 @@ impl ErrorKind {
             ErrorKind::StatusCodeSelectorError(status_code_selector_error) => Some(format!(
                 "Status code selector error: {status_code_selector_error}. Check accept configuration",
             )),
-            ErrorKind::InvalidIndexFile(_path) => Some(
-                "Index file not found in directory. Check if index.html or other index files exist".to_string()
-            ),
+            ErrorKind::InvalidIndexFile(index_files) => match &index_files[..] {
+                [] => "No directory links are allowed because index_files is defined and empty".to_string(),
+                [name] => format!("An index file ({name}) is required"),
+                [init @ .., tail] => format!("An index file ({}, or {}) is required", init.join(", "), tail),
```
mre, line 336: I had no clue you could do `init @ ..`

mre:
Definitely much clearer. I don't really have a strong opinion on your `--index-files ''` case, but it does read fine to me.

PR #1861, opened 2025-09-30
Fix CI lint by adding rustfmt in the the toolchain components

JayJayArr:
CI is currently failing with this error:
`'cargo-fmt' is not installed for the toolchain 'stable-x86_64-unknown-linux-gnu'.`

adding rustfmt as a toolchain component fixes is

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,10 +41,7 @@ jobs:
       - name: Run cargo fmt (check if all code is rustfmt-ed)
         run: cargo fmt --all --check
       - name: Run cargo clippy (deny warnings)
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --all-targets --all-features -- -D warnings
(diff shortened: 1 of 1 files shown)


mre:
Sorry, I didn't see your PR before I created and merged https://github.com/lycheeverse/lychee/pull/1862.

However, I think your change is better, because you avoid one GitHub action altogether, which might make the builds a little faster. If only I knew why the diff still shows `rustfmt` as a change, which should now be in `master`. No clue.

Maybe you can rebase and we'll try again?

JayJayArr:
Looks like `actions-rs/cargo@v1` is deprecated by quite some time now (although it works fine I guess?).
Do you think it would be worth replacing in all workflows?

mre:
yes

mre:
I will go ahead and merge this PR already as the work is solid and much appreciated. If you find the time, feel free to create another PR for the rest. Thanks a lot!

PR #1864, opened 2025-10-04
feat: propagate URL parsing/resolving errors to top level 

katrinafyi:
previously, errors which happened while collecting links (e.g., due to invalid base join) would print a terse message to the console and then they would be forgotten. this looked like:
```
[WARN] Error creating request: InvalidPathToUri("/relative-link/")
🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors
```
this is unhelpful because it only shows the internal enum name and it's not reported in the final error count, so it's very easy to miss.

now, these "early" errors get propagated into the final error list and displayed alongside the HTTP failures.
```
[a.md]:
   [ERROR] error:// | Error building URL for "/relative-link/" (Attribute: Some("href")): Cannot convert path '/relative-link/' to a URI: To resolve relative links in local files, provide a root dir

```
this makes them more obvious and lets us make use of the existing functions for displaying error details and suggested fixes.

this implements step (2) of the plan here https://github.com/lycheeverse/lychee/pull/1624#issuecomment-3274485963

**however,** i don't like the way this is implemented. it has to "smuggle" the early errors through a new error case,
```
ErrorKind::CreateRequestItem(RawUri, ResolvedInputSource, Box<ErrorKind>)
```
and all of the commands have to deliberately handle this case by bypassing the check logic and directly constructing a failed Response. it would be very easy for a user to forget to do this. originally, i wanted to make this new error separate from the usual ErrorKind, but the `lychee_lib::Result<T>` type with ErrorKind is so pervasive and it would've needed extensive changes.

maybe, instead, i could embed this into the Request type by making its Uri field into a `Result<Uri, CreatRequestError>` but that seems not great too.

another downside of the current approach is it uses the fake `error://`  URL to display these messages. this is because the ResponseBody needs a "valid" Uri. 

TODO: 
- [ ] tests. fix old and add new.

also, along the way, this changes the `ErrorKind::details` cases for invalid path and invalid base join. these details previously just repeated the error message and would look like "Cannot convert path '/../../' to a URI: Cannot convert path to URI: '/../../'. Check path format".

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -12,7 +12,7 @@ use tokio_stream::wrappers::ReceiverStream;
 
 use lychee_lib::archive::Archive;
 use lychee_lib::{Client, ErrorKind, Request, Response, Uri};
-use lychee_lib::{InputSource};
+use lychee_lib::{InputSource, Result};
 use lychee_lib::{ResponseBody, Status};
 
 use crate::formatters::get_response_formatter;
@@ -23,16 +23,13 @@ use crate::parse::parse_duration_secs;
 use crate::verbosity::Verbosity;
 use crate::{ExitCode, cache::Cache, stats::ResponseStats};
 
-use std::result::Result;
-
-use lychee_lib::collector::CollectError;
 use super::CommandParams;
 
 pub(crate) async fn check<S>(
     params: CommandParams<S>,
 ) -> Result<(ResponseStats, Arc<Cache>, ExitCode)>
 where
-    S: futures::Stream<Item = Result<Result<Request, CollectError>>>
+    S: futures::Stream<Item = Result<Request>>,
 {
     // Setup
     let (send_req, recv_req) = mpsc::channel(params.cfg.max_concurrency);
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -12,7 +12,7 @@ use tokio_stream::wrappers::ReceiverStream;
 
 use lychee_lib::archive::Archive;
 use lychee_lib::{Client, ErrorKind, Request, Response, Uri};
-use lychee_lib::{InputSource, Result};
+use lychee_lib::{InputSource};
 use lychee_lib::{ResponseBody, Status};
 
 use crate::formatters::get_response_formatter;
@@ -23,13 +23,16 @@ use crate::parse::parse_duration_secs;
 use crate::verbosity::Verbosity;
 use crate::{ExitCode, cache::Cache, stats::ResponseStats};
 
+use std::result::Result;
+
+use lychee_lib::collector::CollectError;
 use super::CommandParams;
 
 pub(crate) async fn check<S>(
     params: CommandParams<S>,
 ) -> Result<(ResponseStats, Arc<Cache>, ExitCode)>
 where
-    S: futures::Stream<Item = Result<Request>>,
+    S: futures::Stream<Item = Result<Result<Request, CollectError>>>
 {
     // Setup
     let (send_req, recv_req) = mpsc::channel(params.cfg.max_concurrency);
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -254,9 +254,11 @@ async fn request_channel_task(
                     )
                     .await
                 }
-                Err(ErrorKind::CreateRequestItem(uri, src, e)) => {
-                    Response::new(Uri::try_from("https://google.com").unwrap(), Status::Error(*e), src)
-                }
+                Err(ErrorKind::CreateRequestItem(uri, src, e)) => Response::new(
+                    Uri::try_from("error://").unwrap(),
+                    Status::Error(ErrorKind::CreateRequestItem(uri, src.clone(), e)),
+                    src,
+                ),
                 Err(e) => Err(e).expect("cannot read request"),
             };
 
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -180,15 +180,18 @@ where
 {
     tokio::pin!(requests);
     while let Some(request) = requests.next().await {
-        let request = request?;
+        let request = match request {
+            x @ Ok(_) => x,
+            x @ Err(ErrorKind::CreateRequestItem(_, _, _)) => x,
+            Err(e) => Err(e)?,
+        };
         if let Some(pb) = &bar {
             pb.inc_length(1);
-            pb.set_message(request.to_string());
+            if let Ok(request) = &request {
+                pb.set_message(request.to_string());
+            }
         }
-        send_req
-            .send(Ok(request))
-            .await
-            .expect("Cannot send request");
+        send_req.send(request).await.expect("Cannot send request");
     }
     Ok(())
 }
@@ -240,15 +243,22 @@ async fn request_channel_task(
         ReceiverStream::new(recv_req),
         max_concurrency,
         |request: Result<Request>| async {
-            let request = request.expect("cannot read request");
-            let response = handle(
-                &client,
-                cache.clone(),
-                cache_exclude_status.clone(),
-                request,
-                accept.clone(),
-            )
-            .await;
+            let response = match request {
+                Ok(request) => {
+                    handle(
+                        &client,
+                        cache.clone(),
+                        cache_exclude_status.clone(),
+                        request,
+                        accept.clone(),
+                    )
+                    .await
+                }
+                Err(ErrorKind::CreateRequestItem(uri, src, e)) => {
+                    Response::new(Uri::try_from("https://google.com").unwrap(), Status::Error(*e), src)
+                }
+                Err(e) => Err(e).expect("cannot read request"),
+            };
 
             send_resp
                 .send(response)
--- a/lychee-lib/src/types/input/input.rs
+++ b/lychee-lib/src/types/input/input.rs
@@ -7,7 +7,6 @@ use super::InputResolver;
 use super::content::InputContent;
 use super::source::InputSource;
 use super::source::ResolvedInputSource;
-use crate::collector::CollectError;
 use crate::filter::PathExcludes;
 use crate::types::FileType;
 use crate::types::file::FileExtensions;
--- a/lychee-lib/src/types/input/input.rs
+++ b/lychee-lib/src/types/input/input.rs
@@ -7,6 +7,7 @@ use super::InputResolver;
 use super::content::InputContent;
 use super::source::InputSource;
 use super::source::ResolvedInputSource;
+use crate::collector::CollectError;
 use crate::filter::PathExcludes;
 use crate::types::FileType;
 use crate::types::file::FileExtensions;
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -1,4 +1,3 @@
-use crate::collector::CollectError;
 use log::warn;
 use percent_encoding::percent_decode_str;
 use reqwest::Url;
@@ -124,18 +123,18 @@ pub(crate) fn create(
     root_dir: Option<&PathBuf>,
     base: Option<&Base>,
     extractor: Option<&BasicAuthExtractor>,
-) -> HashSet<std::result::Result<Request, CollectError>> {
+) -> HashSet<Request> {
     let base = base.cloned().or_else(|| Base::from_source(source));
 
     uris.into_iter()
-        .map(|raw_uri| {
-            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor).map_err(|e| {
-                CollectError {
-                    source: source.clone(),
-                    raw_uri: raw_uri,
-                    error: e,
+        .filter_map(|raw_uri| {
+            match create_request(&raw_uri, source, root_dir, base.as_ref(), extractor) {
+                Ok(request) => Some(request),
+                Err(e) => {
+                    warn!("Error creating request: {e:?}");
+                    None
                 }
-            })
+            }
         })
         .collect()
 }
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -128,9 +128,9 @@ pub(crate) fn create(
 
     uris.into_iter()
         .map(|raw_uri| {
-            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor).map_err(|e|
+            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor).map_err(|e| {
                 ErrorKind::CreateRequestItem(raw_uri.clone(), source.clone(), Box::new(e))
-        )
+            })
         })
         .collect()
 }
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -129,13 +129,8 @@ pub(crate) fn create(
 
     uris.into_iter()
         .map(|raw_uri| {
-            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor).map_err(|e| {
-                CollectError {
-                    source: source.clone(),
-                    raw_uri: raw_uri,
-                    error: e,
-                }
-            })
+            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor)
+                .map_err(|e| CollectError::CannotParseUrl(raw_uri, source.clone(), e))
         })
         .collect()
 }
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -129,8 +129,13 @@ pub(crate) fn create(
 
     uris.into_iter()
         .map(|raw_uri| {
-            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor)
-                .map_err(|e| CollectError::CannotParseUrl(raw_uri, source.clone(), e))
+            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor).map_err(|e| {
+                CollectError {
+                    source: source.clone(),
+                    raw_uri: raw_uri,
+                    error: e,
+                }
+            })
         })
         .collect()
 }
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -123,18 +123,14 @@ pub(crate) fn create(
     root_dir: Option<&PathBuf>,
     base: Option<&Base>,
     extractor: Option<&BasicAuthExtractor>,
-) -> HashSet<Request> {
+) -> HashSet<Result<Request>> {
     let base = base.cloned().or_else(|| Base::from_source(source));
 
     uris.into_iter()
-        .filter_map(|raw_uri| {
-            match create_request(&raw_uri, source, root_dir, base.as_ref(), extractor) {
-                Ok(request) => Some(request),
-                Err(e) => {
-                    warn!("Error creating request: {e:?}");
-                    None
-                }
-            }
+        .map(|raw_uri| {
+            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor).map_err(|e|
+                ErrorKind::CreateRequestItem(raw_uri.clone(), source.clone(), Box::new(e))
+        )
         })
         .collect()
 }
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -1,3 +1,4 @@
+use crate::collector::CollectError;
 use log::warn;
 use percent_encoding::percent_decode_str;
 use reqwest::Url;
@@ -123,18 +124,18 @@ pub(crate) fn create(
     root_dir: Option<&PathBuf>,
     base: Option<&Base>,
     extractor: Option<&BasicAuthExtractor>,
-) -> HashSet<Request> {
+) -> HashSet<std::result::Result<Request, CollectError>> {
     let base = base.cloned().or_else(|| Base::from_source(source));
 
     uris.into_iter()
-        .filter_map(|raw_uri| {
-            match create_request(&raw_uri, source, root_dir, base.as_ref(), extractor) {
-                Ok(request) => Some(request),
-                Err(e) => {
-                    warn!("Error creating request: {e:?}");
-                    None
+        .map(|raw_uri| {
+            create_request(&raw_uri, source, root_dir, base.as_ref(), extractor).map_err(|e| {
+                CollectError {
+                    source: source.clone(),
+                    raw_uri: raw_uri,
+                    error: e,
                 }
-            }
+            })
         })
         .collect()
 }
--- a/lychee-bin/src/commands/dump.rs
+++ b/lychee-bin/src/commands/dump.rs
@@ -13,7 +13,7 @@ use super::CommandParams;
 /// Dump all detected links to stdout without checking them
 pub(crate) async fn dump<S>(params: CommandParams<S>) -> Result<ExitCode>
 where
-    S: futures::Stream<Item = Result<Request>>,
+    S: futures::Stream<Item = Result<Result<Request, CollectError>>>
 {
     let requests = params.requests;
     tokio::pin!(requests);
--- a/lychee-bin/src/commands/dump.rs
+++ b/lychee-bin/src/commands/dump.rs
@@ -13,7 +13,7 @@ use super::CommandParams;
 /// Dump all detected links to stdout without checking them
 pub(crate) async fn dump<S>(params: CommandParams<S>) -> Result<ExitCode>
 where
-    S: futures::Stream<Item = Result<Result<Request, CollectError>>>
+    S: futures::Stream<Item = Result<Request>>,
 {
     let requests = params.requests;
     tokio::pin!(requests);
--- a/lychee-bin/src/commands/dump.rs
+++ b/lychee-bin/src/commands/dump.rs
@@ -1,4 +1,5 @@
 use log::error;
+use lychee_lib::ErrorKind;
 use lychee_lib::Request;
 use lychee_lib::Result;
 use std::fs;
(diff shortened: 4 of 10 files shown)


katrinafyi:
I see that this PR contradicts an existing test case which says the current behaviour is intentional. 
https://github.com/lycheeverse/lychee/blob/54e425c6261fc336a77f1c975643bd5c440877bd/lychee-bin/tests/cli.rs#L1458-L1462

Maybe there should be an option about whether to treat absolute links as an error when root-dir is not specified? At the moment, it would not be possible to get the old "ignoring" behaviour after this PR. Or, we just change the test

thomas-zahner:
Thanks for the PR and thanks for the extensive description and thoughts. I like it and this should improve UX quite a bit.

Since we now pass errors through the `Response` struct we might consider making `Uri` an `Option<Uri>` so we could get rid of this sentinel value `error://`. But I'm not sure if it's really worth it.

Then it would also be great to add a CLI test which covers the example you provided in this PR `cargo run -- https://google.com 'noasdjfi/[' non-existing-fjdaifdsjai.com no-perms https://example.com`. To assert that lychee reports the errors as expected etc.

katrinafyi:
I can try to change Response and report back on how it goes. I should add a CLI test as well.

Did you have thoughts about whether absolute links should be flagged as errors in the absence of root-dir/base-url? I noticed in [the feature table](https://github.com/lycheeverse/lychee#features), there is a tick for the "Skip relative URLs" feature in Lychee. Making root-relative URLs into failures by default would also conflict with this, in addition to the unit test I mentioned earlier.

thomas-zahner:
Regarding the handling of absolute URLs I think it probably make sense to flag them as errors, as you proposed. We are now at quite a different stage compared to when `Skip relative URLs` was added to the feature table. This `Skip relative URLs` "feature" in the table predates the root-dir feature by many years so it makes sense to reconsider the behaviour. I think an error is much clearer than a warning that users easily glance over and I personally prefer that behaviour.

> Maybe there should be an option about whether to treat absolute links as an error when root-dir is not specified? At the moment, it would not be possible to get the old "ignoring" behaviour after this PR. Or, we just change the test

So for me it makes sense to just change that behaviour, update/remove the test and update/remove the "feature" in the README.

thomas-zahner:
@mre and I had a quick chat about this. We realised that this PR improves the current situation of panicking and crashing (among improving other things like skipping over relative URLs) but also that this new behaviour is not ideal. In our opinion erroneous user inputs should not be handled the same way as link check errors. For example take a look at the exit codes and how we differentiate between user errors and link check errors. We think user provided erroneous arguments should fail fast and make lychee exit before link checking.

Ideally we handle errors properly (which happens in this PR but not on master) while exiting early upon encountering user errors. The output could look something like this:

```
$ lychee https://google.com 'noasdjfi/[' non-existing-fjdaifdsjai.com no-perms https://example.com 

Problems reading these inputs:
- 'http://non-existing-fjdaifdsjai.com/': Network error: Connection failed. Check network connectivity and firewall settings
- 'no-perms': Cannot traverse input directory: no-perms: IO error for operation on no-perms: Permission denied (os error 13)
- 'noasdjfi/[': UNIX glob pattern is invalid: Pattern syntax error near position 9: invalid range pattern. Check pattern syntax

Please check for typos and try again.
```

So, could you try to keep user errors from making the program fail early instead of moving and converting user errors into link check errors?

katrinafyi:
It's an interesting perspective, and thanks for discussing it and giving your thoughts. I agree it could do with more thinking.

With this PR, however, I think the errors that arise within an input file when resolving a URL are reasonably flagged as link checking errors. I'm interested in getting this part of the feature merged sooner, so would it be okay to revert the input reading errors to the old behaviour and merge the "Error building URL" errors on their own? That said, the changes to the Result type are such that the input reading errors would still have to be passed through the new RequestError::GetInputContent. So, I would make it behave the same on the outside while going through different logic on the inside. 

About the erroneous inputs case, I think there is a scale. While some inputs are obviously wrong, like an invalid glob syntax, I think that the HTTP and filesystem errors are more nuanced. Unlike an invalid glob, these errors are affected by external state and this has some ramifications.

- For one, they take time and resources to check - this may noticeably delay starting up Lychee.
- For two, this means that, even if the inputs are checked preemptively, you must always account for the possibility that the error state happens _after_ the check (for example, a laptop might lose wifi) (this is a TOCTOU-style problem). Lychee must still handle errors sensibly in these situations.
- As another nuance, directories can obviously contain other directories. Since using a directory input walks recursively, it's possibly for a deep subdirectory to have no permissions. Would an early check recurse through all subdirectories? 
- Also, there is the possibility that a Lychee library user does not go through the usual CLI parsing and validating logic.

All this is to say, regardless of whether the inputs are validated beforehand, the link checking phase will need _some_ way to sensibly handle these transient errors. I think this PR achieves that, and that it could complement - rather than replace - any extra early input checking which you want to add. (The glob syntax errors, however, should be handled through stricter checking in the InputSource constructor).

To summarise, I want URL building errors the most. For the input loading errors, I believe the current implementation doesn't conflict with your plans, but I'm also happy to scale it back. About implementing the early checks, I think that is best done in another PR (and I'm not the best person to do it, because I don't think special early checking is super beneficial (aside from the glob syntax)).

Sorry for the long text, let me know what you want :))

thomas-zahner:
Thanks for the thoughtful comment.

> With this PR, however, I think the errors that arise within an input file when resolving a URL are reasonably flagged as link checking errors. I'm interested in getting this part of the feature merged sooner, so would it be okay to revert the input reading errors to the old behaviour and merge the "Error building URL" errors on their own?

Yes, splitting this PR up into two separate PRs would be great, if that's possible. Errors that arise within an input file should definitely be flagged as link check errors, as indicated by my previous comment. So it's just the direct user input which we are discussing about and as this might be getting somewhat complicated this would deserve a separate issue and PR. This would allow us to quickly merge the part you are most interested in and then we could continue on discussions with user input handling where you didn't need to participate when you mention you aren't as interested in that area.

> About the erroneous inputs case, I think there is a scale. While some inputs are obviously wrong, like an invalid glob syntax, I think that the HTTP and filesystem errors are more nuanced. Unlike an invalid glob, these errors are affected by external state and this has some ramifications.

Thanks, that makes sense. Maybe I was focusing too much on "early exit" in my previous comment. It's true that some errors are impossible to catch "early". However, it is still possible to differentiate between user errors and link-check errors, which makes sense from a UX perspective. For example running `lychee https://invalid.example.com` should result in a clear user error with exit code 1 because the user supplied an invalid argument to lychee. (note that this is not the same as `echo https://invalid.example.com | lychee -`) But this PR moves/converts the error into a link check error with exit code 2 which means that the currently documented exit codes are no longer accurate. We would prefer to keep the separation between the two different types of errors with their respective exit codes.

I realise that the output example of my previous comment might not be worth the effort. It would be much simpler to exit on the first user error, as is the current behaviour, rather than listing all user errors.

In any case it would make sense to split up this PR if possible and potentially create a new issue.

katrinafyi:
> For example running lychee https://invalid.example.com should result in a clear user error with exit code 1 because the user supplied an invalid argument to lychee. 

Right, yeah, I see. The old code would've exited with 1 because rust uses 1 for an Err returned from the main function. I think the error handling just looked so noisy that I hadn't considered that some aspects of it could be desired.

I'll see what I can do. It'll need some way to distinguish input input errors from discovered input errors - maybe an extra case in RequestError.

katrinafyi:
Bloop! I think the review comments should be addressed now. I've brought back the early exit with code 1 for failing to load CLI inputs, and I've updated the tests and readme to match the new error behaviour for InvalidBaseJoin. 

In the implementation, the early user errors should behave the same as before, but the path they go through is changed. Now, they are packaged into a RequestError::UserInputContent and they get sent to the request channel task before being sent back to the main thread as an error. 

Some other misc notes:
- the Response inputsource is changed from ResolvedInputSource to InputSource, slightly reverting https://github.com/lycheeverse/lychee/pull/1840. this is because Response has to represent both real HTTP responses and fake "responses" which arise from failed input resolution. this is kind of undesirable
- similar to above, i find it's too hard to change ResponseBody's uri field to an Option<Uri>, just because of how often it's used. so, we have to use the fake error:// URL to represent URL resolving failures

examples/collect_links/collect_links.rs
```diff
@@ -1,10 +1,10 @@
-use lychee_lib::{Collector, Input, InputSource, Result};
+use lychee_lib::{Collector, Input, InputSource, RequestError};
 use reqwest::Url;
-use std::{collections::HashSet, path::PathBuf};
+use std::{collections::HashSet, path::PathBuf, result::Result};
```
mre: I think `result::Result` is in the prelude, so no need to import it explicitly?

lychee-bin/src/commands/check.rs
```diff
@@ -29,7 +30,7 @@ pub(crate) async fn check<S>(
     params: CommandParams<S>,
 ) -> Result<(ResponseStats, Arc<Cache>, ExitCode)>
 where
-    S: futures::Stream<Item = Result<Request>>,
+    S: futures::Stream<Item = std::result::Result<Request, RequestError>>,
```
mre: The `std::result::Result` looks a bit noisy throughout that file. Maybe we should call the other one `RequestResult` or so? Or maybe we use `StdResult` as an alias. I don't know, might be fine.

lychee-bin/src/commands/check.rs
```diff
@@ -176,36 +177,35 @@ async fn suggest_archived_links(
...
 {
     tokio::pin!(requests);
     while let Some(request) = requests.next().await {
-        let request = request?;
         if let Some(pb) = &bar {
             pb.inc_length(1);
-            pb.set_message(request.to_string());
+            if let Ok(request) = &request {
+                pb.set_message(request.to_string());
+            }
```
mre: So we don't print errors in the progress bar anymore? Not sure I understand. 🤔

lychee-bin/tests/cli.rs
```diff
@@ -2965,4 +2965,28 @@ mod cli {
...
+        // maybe test with a directory with no write permissions? but there
+        // doesn't seem to be an equivalent to chmod on the windows API:
+        // https://doc.rust-lang.org/std/fs/struct.Permissions.html
+
+        main_command!()
+            .arg("invalid-glob[")
+            .assert()
+            .failure()
+            .code(2);
+        // TODO: change above exit code to 1 after https://github.com/lycheeverse/lychee/pull/1869
```
mre: https://github.com/lycheeverse/lychee/pull/1869 is done. Can be updated.

lychee-lib/src/types/input/input.rs
```diff
@@ -170,20 +170,49 @@ impl Input {
...
+            let user_input_error =
+                move |e: ErrorKind| RequestError::UserInputContent(source.clone(), Box::new(e));
+            let discovered_input_error =
+                |e: ErrorKind| RequestError::GetInputContent(self.source.clone(), Box::new(e));
+
+            // Handle simple cases that don't need resolution. Also, perform
+            // simple *stateful* checks for more complex input sources.
+            //
+            // However, stateless well-formedness checks (e.g., checking glob
+            // syntax) should be done in Input::new.
```
mre: Should we mention here that we check the glob already?

lychee-lib/src/types/input/input.rs
```diff
@@ -170,20 +170,49 @@ impl Input {
...
                 }
+                InputSource::FsPath(ref path) => {
+                    let is_readable = if path.is_dir() {
+                        path.read_dir()
+                            .map(|_| ())
+                            .map_err(|e| ErrorKind::DirTraversal(ignore::Error::Io(e)))
+                    } else {
+                        // this checks existence without requiring an open. opening here,
+                        // then re-opening later, might cause problems with pipes. this
+                        // does not validate permissions.
```
mre:
```suggestion
                        // This checks existence without requiring an open. Opening here,
                        // then re-opening later, might cause problems with pipes. This
                        // does not validate permissions.
```

lychee-lib/src/types/request.rs
```diff
@@ -1,8 +1,57 @@
...
 
-use super::ResolvedInputSource;
+/// An error which occurs while trying to construct a [`Request`] object.
+/// That is, an error which happens while trying to load links from an input
+/// source.
+#[derive(Error, Debug, PartialEq, Eq, Hash)]
+pub enum RequestError {
+    /// Unable to construct a URL for a link appearing within the given source.
+    #[error("Error building URL for {0}: {2}")]
+    CreateRequestItem(RawUri, ResolvedInputSource, #[source] Box<ErrorKind>),
```
mre: Did you consider to use struct variants here?
I find that it makes the error strings a bit more readable and less error-prone because it's harder to mix up the 0,1,2... indices.

E.g. 

```rust
#[error("Error building URL for {uri}: {error}")]
struct CreateRequestItem {
  uri: RawUri,
  source: ResolvedInputSource,
  #[source]
  error: Box<ErrorKind>
}
```

lychee-lib/src/utils/request.rs
```diff
@@ -197,6 +192,19 @@ mod tests {
 
     use super::*;
 
+    fn create_ok_only(
```
mre, line 212: I understand that this is an internal function for testing only, but can we still add a comment about what it does and why?

clippy.toml
mre, line 1: Just checking: this file is still needed even after your recent refactoring, right?

mre:
> Maybe there should be an option about whether to treat absolute links as an error when root-dir is not specified? At the moment, it would not be possible to get the old "ignoring" behaviour after this PR. Or, we just change the test

Yeah, we can simply change the test. As Thomas mentioned, it's not really a "feature" to skip relative URLs, and the code predates many of our recent changes.

mre:
I'm done with my review. Looks good. Added a few minor comments, after which we can merge this.

lychee-bin/src/commands/check.rs
katrinafyi: The old code would not have printed errors either, because there was a ? before this line. So, I excluded errors just to match the old code.

Now, I suppose we have the option to print these errors if we want. Do we want?

mre: Yes, we want. 😆

katrinafyi:
We have a new problem! Preprocessor errors are being turned into request errors and display like this:
```

    [/home/runner/work/lychee/lychee/fixtures/TEST.md]:
     [ERROR] error:// | Error reading input \'/home/runner/work/lychee/lychee/fixtures/TEST.md\': Preprocessor command \'program does not exist\' failed: could not start: No such file or directory (os error 2): Command \'program does not exist\' failed could not start: No such file or directory (os error 2). Check value of the pre option

```
In particular, this includes when the preprocessor command does not exist. We'd probably want this to be a CLI error rather than a link check error. However, since a command might be found through `$PATH`, it's hard to validate that the command exists aside from actually trying to start the command - which could have side-effects. There are libraries like `which`, but using a crate also seems a bit heavy for this small check.

Thoughts? @thomas-zahner ?

clippy.toml
katrinafyi: Yeah, it is. It's needed because request::create now returns `HashSet<std::result::Result<Request, RequestError>>` and there's some mutability deep inside RequestError.

Honestly, I'm not too happy with ignoring this because I think it _is_ weird to put errors inside a hashset. Maybe we could return a vec rather than a hashset? This hashset is only for a single input file, so I think its deduplication is not super impactful anyway.

lychee-lib/src/types/request.rs
katrinafyi: I hadn't thought about that! I'm _lightly_ against struct variants because it's a bit verbose and I think the types are distinct enough. The index problem is only significant for the #error strings because those are untyped.

Let me know if you'd greatly prefer struct variants :)

lychee-bin/src/commands/check.rs
katrinafyi: I changed Result to the std result instead of using the lychee alias. Let me know if that works

thomas-zahner:
> We have a new problem! Preprocessor errors are being turned into request errors and display like this:

@katrinafyi Honestly, I'm okay with this. Conceptually, it's not wrong to treat preprocessing as part of link extraction and therefore the link check process. This PR then also has the benefit that if preprocessing only fails on one/some file lychee doesn't abort, which is an improvement.

**Before**

```
cargo run Cargo.* --preprocess aa
Error: Preprocessor command 'aa' failed: could not start: No such file or directory (os error 2)
```

**PR**

```
cargo run Cargo.* --preprocess aa
  2/2 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links
Issues found in 2 inputs. Find details below.

[Cargo.lock]:
   [ERROR] error:// | Error reading input 'Cargo.lock': Preprocessor command 'aa' failed: could not start: No such file or directory (os error 2): Command 'aa' failed could not start: No such file or directory (os error 2). Check value of the pre option

[Cargo.toml]:
   [ERROR] error:// | Error reading input 'Cargo.toml': Preprocessor command 'aa' failed: could not start: No such file or directory (os error 2): Command 'aa' failed could not start: No such file or directory (os error 2). Check value of the pre option

🔍 2 Total (in 0s) ✅ 0 OK 🚫 2 Errors
```

We could think about a simple prevalidation step, as you mentioned which checks if the value of `--preprocess` itself is valid (program exits in path) but this is not very important. For reference ripgrep doesn't do this either and this PR makes lychee behave more similar to ripgrep, which I like. As you mention it is impossible to validate if the preprocess commands exit successfully, without running them first. This would be another reason why it makes sense to conceptually move preprocessing into the link check process.

```
rg 'some text' Cargo.* --pre aa
rg: Cargo.toml: preprocessor command could not start: '"aa" "Cargo.toml"': No such file or directory (os error 2)
rg: Cargo.lock: preprocessor command could not start: '"aa" "Cargo.lock"': No such file or directory (os error 2)
```

Just two things I noticed:

* In what cases do we use the `error://` dummy value? Doesn't it make sense to make it optional and not display any dummy value?
* The error message with preprocess is repetitive and then mentions the `pre` option which was the name before I renamed it to preprocess. Definitely my mistake :sweat_smile:

clippy.toml
thomas-zahner: Yeah, using `Vec` might even make more sense, right? If the same error happened twice it would also make sense to report it twice. If it resolve the clippy issue that's even better.

katrinafyi: It's a bit funny because you kind of want a set of requests and a vec of errors. How do we feel about returning them separately like `(HashSet<Request>, Vec<Error>)`? I also wouldn't mind a vec of Results.

katrinafyi:
@thomas-zahner Thanks for your reply. Yeah, I can see the logic in that approach. I'm happy to leave preprocess errors as link check errors.

Re the error://, it appears for every error arising before a HTTP request is made. I also don't like it very much, but it's passed through the stats and response formatter logic, so avoiding it would need a lot of changes there. But maybe this is something I should dig deeper into. The error:// url could also be confusing.

clippy.toml
mre: Yeah, that might be the way to go. Not a huge fan of tuples for things that we pass around between different parts of the codebase, as it's hard to track down what `.0` and `.1` refer to, so maybe we can introduce a new type for that? `RequestBatch` or `CheckResults` or `CheckedRequests` perhaps?

katrinafyi: I just made it a Vec with manual deduplication [`50e91d1` (#1864)](https://github.com/lycheeverse/lychee/pull/1864/commits/50e91d1936082e4ec2c7a55f3b005feafa24d427). A custom class like RequestBatch is probably the _right_ solution, but it's not super useful because it immediately gets deconstructed and put into a `Stream<Result<Request, RequestError>>`.

katrinafyi:
I don't want to pursue making the error URL optional in this PR. I think it will be a simple but extensive change. You just have to allow `format_response` to take an `Option<Uri>` instead of a ResponseBody (that must contain Uri). If you're already looking at changing these functions in https://github.com/lycheeverse/lychee/pull/1914, maybe that is a better place to investigate it.

In the meantime, it could be changed to `error:` instead of `error://`. That might be less weird for a user to see.

lychee-lib/src/types/request_error.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// input, the underlying cause of the `RequestError` will be returned
+    /// as an Err. This allows the error to be propagated back to the user.
+    #[allow(clippy::missing_panics_doc)]
+    pub fn into_response(self) -> Result<Response, ErrorKind> {
+        match self {
+            RequestError::UserInputContent(_, e) => Err(*e),
+            e => {
+                let src = e.input_source();
+                Ok(Response::new(
+                    Uri::try_from(ERROR_URL).unwrap(),
```
thomas-zahner: This seems like a good use-case for a static variable:

```
static ERROR_URI: LazyLock<Uri> = LazyLock::new(|| Uri::try_from("error:").unwrap());
```

Then `#[allow(clippy::missing_panics_doc)]` should no longer be required. E.g. see `DEFAULT_ACCEPTED_STATUS_CODES` for inspiration and how to update the test.

lychee-lib/src/utils/request.rs
```diff
@@ -113,30 +110,41 @@ fn create_uri_from_file_path(
...
+                source.clone(),
+                Box::new(e),
+            )),
+        }
+    }
+
+    let errs_iter = errors.into_iter().map(Result::Err);
+    let reqs_iter = requests.into_iter().map(Result::Ok);
+
+    reqs_iter.chain(errs_iter).collect()
```
thomas-zahner: Personally prefer

```suggestion
    requests
        .into_iter()
        .map(Result::Ok)
        .chain(errors.into_iter().map(Result::Err))
        .collect()
```

lychee-lib/src/types/request_error.rs
```diff
@@ -0,0 +1,89 @@
...
+const ERROR_URL: &str = "error:";
+
+/// An error which occurs while trying to construct a [`Request`] object.
+/// That is, an error which happens while trying to load links from an input
+/// source.
+#[derive(Error, Debug, PartialEq, Eq, Hash)]
+pub enum RequestError {
+    /// Unable to construct a URL for a link appearing within the given source.
+    #[error("Error building URL for {0}: {2}")]
+    CreateRequestItem(RawUri, ResolvedInputSource, #[source] Box<ErrorKind>),
```
thomas-zahner: Just double-checked - these `Box`es for each variant's ErrorKind are unnecessary

```diff --git a/lychee-lib/src/types/input/input.rs b/lychee-lib/src/types/input/input.rs
index 9fd80992..eadd404b 100644
--- a/lychee-lib/src/types/input/input.rs
+++ b/lychee-lib/src/types/input/input.rs
@@ -100,9 +100,9 @@ impl Input {
             let source = self.source.clone();
 
             let user_input_error =
-                move |e: ErrorKind| RequestError::UserInputContent(source.clone(), Box::new(e));
+                move |e: ErrorKind| RequestError::UserInputContent(source.clone(), e);
             let discovered_input_error =
-                |e: ErrorKind| RequestError::GetInputContent(self.source.clone(), Box::new(e));
+                |e: ErrorKind| RequestError::GetInputContent(self.source.clone(), e);
 
             // Handle simple cases that don't need resolution. Also, perform
             // simple *stateful* checks for more complex input sources.
diff --git a/lychee-lib/src/types/request_error.rs b/lychee-lib/src/types/request_error.rs
index 5fd55092..008f51d4 100644
--- a/lychee-lib/src/types/request_error.rs
+++ b/lychee-lib/src/types/request_error.rs
@@ -13,15 +13,15 @@ const ERROR_URL: &str = "error:";
 pub enum RequestError {
     /// Unable to construct a URL for a link appearing within the given source.
     #[error("Error building URL for {0}: {2}")]
-    CreateRequestItem(RawUri, ResolvedInputSource, #[source] Box<ErrorKind>),
+    CreateRequestItem(RawUri, ResolvedInputSource, #[source] ErrorKind),
 
     /// Unable to load the content of an input source.
     #[error("Error reading input '{0}': {1}")]
-    GetInputContent(InputSource, #[source] Box<ErrorKind>),
+    GetInputContent(InputSource, #[source] ErrorKind),
 
     /// Unable to load an input source directly specified by the user.
     #[error("Error reading user input '{0}': {1}")]
-    UserInputContent(InputSource, #[source] Box<ErrorKind>),
+    UserInputContent(InputSource, #[source] ErrorKind),
 }
 
 impl RequestError {
@@ -41,7 +41,7 @@ impl RequestError {
         match self {
             Self::CreateRequestItem(_, _, e)
             | Self::GetInputContent(_, e)
-            | Self::UserInputContent(_, e) => *e,
+            | Self::UserInputContent(_, e) => e,
         }
     }
 
@@ -65,7 +65,7 @@ impl RequestError {
     #[allow(clippy::missing_panics_doc)]
     pub fn into_response(self) -> Result<Response, ErrorKind> {
         match self {
-            RequestError::UserInputContent(_, e) => Err(*e),
+            RequestError::UserInputContent(_, e) => Err(e),
             e => {
                 let src = e.input_source();
                 Ok(Response::new(
diff --git a/lychee-lib/src/utils/request.rs b/lychee-lib/src/utils/request.rs
index d00fe988..2f8bc0af 100644
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -136,7 +136,7 @@ pub(crate) fn create(
             Err(e) => errors.push(RequestError::CreateRequestItem(
                 raw_uri.clone(),
                 source.clone(),
-                Box::new(e),
+                e,
             )),
         }
     }

```

lychee-lib/src/utils/request.rs
katrinafyi: I think  that format makes the symmetry between errors and requests harder to see, because one side is vertical in the dot chain and the other is horizontal inside the .chain() call.

What do you think about this?
```rust
    (requests.into_iter().map(Result::Ok))
        .chain(errors.into_iter().map(Result::Err))
        .collect()
```

thomas-zahner: Ah yes really like that :+1:

lychee-lib/src/types/request_error.rs
```diff
@@ -80,10 +80,11 @@ impl RequestError {
...
 #[cfg(test)]
 mod tests {
-    use super::{ERROR_URL, Uri};
+    use super::ERROR_URI;
+    use std::sync::LazyLock;
 
     #[test]
     fn test_error_url_parses() {
-        assert!(Uri::try_from(ERROR_URL).is_ok());
+        let _ = LazyLock::force(&ERROR_URI);
```
thomas-zahner, line 88: ah cool, that's seems like the proper way of doing it

thomas-zahner:
Many thanks for this PR!
It's quite an improvement I think. In the long run hiding the `error:` dummy value would be nice but that's not urgent of course.

Issue #1809, opened 2025-08-20
[Feature] Make warning when no access to folder


AndreiCherniaev:
I install on Ubuntu 25.04 from snap `sudo snap install --edge lychee`
Test with 0 Errors just because no access to outside of $HOME folder [problem](https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1643706)
```
cd /tmp
git clone https://github.com/mxmlnkn/rapidgzip
lychee "rapidgzip/*"
```
Out
```
  0/0 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                                                                                                    🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors
```

But when I work in $HOME (sub)folder, lychee works like a charm... I suggest to make warning when no access. For now for user it looks like no problem, but it is not right...

mre:
Don't think it's possible to detect the problem from lychee's side because snap just hides those files from lychee. To the lychee process, it looks like these files simply don't exist.

I don't know much about snap, but there's probably a way to ask for permission to access the certain folders. Asking for permission to access all files would be the fix. However, that is quite invasive and these folder permissions are there for a reason.

I wondered how ripgrep did it and sure enough, they have the same problem. From their `README.md`:

> (N.B. Various snaps for ripgrep on Ubuntu are also available, but none of them seem to work right and generate a number of very strange bug reports that I don't know how to fix and don't have the time to fix. Therefore, it is no longer a recommended installation option.)

and from their issue template:

> If you installed ripgrep with snap and are getting strange file permission or file not found errors, then please do not file a bug. Instead, use one of the GitHub binary releases.

So I guess there is not much we can do here, sorry. You can use one of the pre-built binaries [here](https://github.com/lycheeverse/lychee/releases) instead.

PR #1865, opened 2025-10-05
feat: print warning if input source matches no files

katrinafyi:
it looks like this:
```
$ lychee 'non-existing/*' '*.fdsamifdsa' 'no-matches?????' empty-dir
   [WARN ] *.fdsamifdsa: No files found for this input source
   [WARN ] no-matches?????: No files found for this input source
   [WARN ] non-existing/*: No files found for this input source
   [WARN ] empty-dir: No files found for this input source
  0/0 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                                        🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors
```

this is implemented as a sneaky `log::warn!` which is probably undesirable. however, the code isn't set up yet to properly pass errors from the input resolving stage up to the top level.

even if it was, the "error on empty input" error should probably be put behind a flag. it would also be tricky because the reporting expects a ResolvedInputSource rather than InputSource, and these warnings arise in the process of resolving. (maybe we could slightly wind back #1840 to make this work)

anyway, much to think about.

related to https://github.com/lycheeverse/lychee/issues/1809, but we don't make any effort to try and detect paths which are hidden by snap. instead, we just emit the warning for any inputs which appear to match no files. i _think_ this will include the snap case, but i haven't tested it.

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -2977,7 +2977,7 @@ mod cli {
             .arg("non-existing-file-name???")
             .assert()
             .success()
-            .stdout(contains("No files found").count(5));
+            .stderr(contains("No files found").count(5));
 
         Ok(())
     }


thomas-zahner:
Thanks for the PR I think it's appropriate to simply print a warning here. Exiting with an error when the user provides a valid path/glob resolving to no files would be unexpected IMO. So I think this is an improvement :+1:

thomas-zahner:
The only thing that bothers me a bit is the formatting of the warning prefix, which obviously is unrelated to this PR.

```
[WARN ]
    ^^^
```

katrinafyi:
Indeed, I think the formatting is just from the logging macro. Interestingly, it looks like it did change sometime between [this issue](https://www.github.com/lycheeverse/lychee/issues/1574) being posted and its second comment. It's obviously to pad out the brackets in case of `[ERROR]` but it does look a bit weird, and Lychee rarely prints error logs anyway so maybe it's worth changing.

Anyway, thanks for the review!

thomas-zahner:
No problem :+1: 
I've create a new issue for that weird padding now.

PR #1869, opened 2025-10-09
refactor: check glob validity when parsing input source arguments

katrinafyi:
now, when providing an input with glob characters but an invalid glob pattern, this will cause a "cannot parse inputs" error rather than erroring out later during link checking. this makes use of the existing ErrorKind::InvalidGlobPattern error case.

for example, a glob with unbalanced brackets is invalid and now looks like this:
```console
$ cargo run -- 'noasdjfi/['
Error: Cannot parse inputs from arguments

Caused by:
    0: UNIX glob pattern is invalid
    1: Pattern syntax error near position 9: invalid range pattern
```

i've also added some TODO notes which would make the logic neater. i didn't do those refactors in this PR because i am mindful of another currently-open (and much bigger) PR which touches these lines (https://www.github.com/lycheeverse/lychee/pull/1837)

i wonder if the reason FsGlob stores strings rather than a Pattern is for flexibility to switch to different glob packages.

this PR was prompted by findings in https://github.com/lycheeverse/lychee/pull/1864#issuecomment-3385187203

### DIFF
--- a/lychee-lib/src/types/input/source.rs
+++ b/lychee-lib/src/types/input/source.rs
@@ -214,24 +214,18 @@ impl Display for InputSource {
 mod tests {
     use super::*;
 
-    fn to_json<T>(value: &T) -> Result<String, String>
-    where
-        T: serde::Serialize + ?Sized,
-    {
-        serde_json::to_string(value).map_err(|e| e.to_string())
-    }
-
     /// Serialization of `FsGlob` relies on [`glob::Pattern::to_string`].
     /// Here, we check that the `to_string` works as we require.
     #[test]
     fn test_pattern_serialization_is_original_pattern() {
         let pat = "asd[f]*";
         assert_eq!(
-            to_json(&InputSource::FsGlob {
+            serde_json::to_string(&InputSource::FsGlob {
                 pattern: Pattern::new(pat).unwrap(),
                 ignore_case: false,
-            }),
-            to_json(pat),
+            })
+            .unwrap(),
+            serde_json::to_string(pat).unwrap(),
         );
     }
 }


thomas-zahner:
> i wonder if the reason FsGlob stores strings rather than a Pattern is for flexibility to switch to different glob packages.

Good question, I wouldn't think so but then I'm not the author. @mre might know the answer to that.
Have you tried to change the type to `Pattern`? I would much prefer that if it is possible without too much effort.

Otherwise I have no objections to this PR.

katrinafyi:
> Have you tried to change the type to Pattern? 

If you're asking me, I haven't tried. I think the change would be pretty easy. But it's the kind of thing which is so obvious that if it hasn't been done already, then there's probably a logical reason why it hasn't. That's my reasoning at least. 

If you're asking @ mre, then we have to wait for him :)

mre:
Sorry for the late reply. 

> If you're asking me, I haven't tried. I think the change would be pretty easy. But it's the kind of thing which is so obvious that if it hasn't been done already, then there's probably a logical reason why it hasn't. That's my reasoning at least.

That's a sensible approach in general. In this specific case however, I have to disappoint you. It might just have been an oversight on my side. 😆 Well, perhaps one issue might be serializing/deserializing those values when writing/reading the config. The `glob` crate [does not support this](https://github.com/rust-lang/glob/issues/136). We'd have to find a workaround.

mre:
> i've also added some TODO notes which would make the logic neater. i didn't do those refactors in this PR because i am mindful of another currently-open (and much bigger) PR which touches these lines ([github.com/lycheeverse/lychee/pull/1837](https://www.github.com/lycheeverse/lychee/pull/1837))

Don't worry about this. My PR is stale anyway, so I'd have to rebase it. So feel free to make any changes you like.

katrinafyi:
Thanks for your comments! I've done the two extra TODOs within this PR. The InputSource parsing is now in InputSource::new and the FsGlob field is a Pattern. As @mre foreshadowed, we have to do some custom serialisation/deserialisation, but it's not too much trouble.

lychee-lib/src/types/input/source.rs
```diff
@@ -105,10 +202,36 @@ impl Display for InputSource {
...
+        serde_json::to_string(value).map_err(|e| e.to_string())
+    }
+
+    /// Serialization of `FsGlob` relies on [`glob::Pattern::to_string`].
+    /// Here, we check that the `to_string` works as we require.
+    #[test]
+    fn test_pattern_serialization_is_original_pattern() {
+        let pat = "asd[f]*";
+        assert_eq!(
+            to_json(&InputSource::FsGlob {
```
thomas-zahner: and then `unwrap` both results here when we expect two `Ok`s

lychee-lib/src/types/input/source.rs
```diff
@@ -105,10 +202,36 @@ impl Display for InputSource {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn to_json<T>(value: &T) -> Result<String, String>
+    where
+        T: serde::Serialize + ?Sized,
+    {
+        serde_json::to_string(value).map_err(|e| e.to_string())
```
thomas-zahner: Here it would actually be more explicit IMO to unwrap and return `Result<String, ()>`

```suggestion
        Ok(serde_json::to_string(value).unwrap())
```

lychee-lib/src/types/input/source.rs
```diff
@@ -41,6 +46,98 @@ pub enum InputSource {
...
+        }
+
+        // This seems to be the only way to determine if this is a glob pattern
+        let is_glob = glob::Pattern::escape(input) != input;
+
+        if is_glob {
+            return Ok(Pattern::new(input).map(|pattern| InputSource::FsGlob {
+                pattern,
+                ignore_case: glob_ignore_case,
+            })?);
```
thomas-zahner: I prefer

```suggestion
            return Ok(InputSource::FsGlob {
                pattern: Pattern::new(input)?,
                ignore_case: glob_ignore_case,
            });
```

lychee-lib/src/types/input/source.rs
katrinafyi: Unwrapping in the helper function is a bit less convenient because if it panics, it'll only show the helper function's line unless you RUST_BACKTRACE=1. If I unwrap in the assert_eq!, I think we can get rid of the helper function

lychee-lib/src/types/input/source.rs
katrinafyi: it's done ~

lychee-lib/src/types/input/source.rs
thomas-zahner: Ah yes, inlining the function in that case makes sense. Thanks :rocket:

PR #1806, opened 2025-08-19
Extract spans for URIs.

Akida31:
Got nerd-sniped in #1304 and implemented the extraction of line and column numbers of links.
Note that it doesn't display them in the cli output for now, and I currently don't have plans to implement that.

Feel free to push to this PR or subsume it.

Issue #1304, opened 2023-11-21
Feature request: Line numbers and columns in output


mihaigalos:
Hi,

Awesome stuff!

Can you please support line numbers in the output? - ideally they can come immediately after a colon after the filename (i.e.: `[./README.md:123:54]:` - this enables easy access to the file on line `123`, column `54` via other editors.

mre:
Unfortunately, that is not possible due to limitations of the underlying libraries, namely https://github.com/servo/html5ever/issues/48 and https://github.com/untitaker/html5gum/issues/10. I think you need to try your luck there first. 😕 
[pulldown-cmark supports spans already](https://docs.rs/pulldown-cmark/latest/pulldown_cmark/struct.BrokenLink.html#structfield.span) and the plaintext library, linkify, [supports it, too](https://github.com/robinst/linkify/pull/3).  So the issue is with the HTML libraries at the moment. You could add a comment in the linked issues.

mihaigalos:
If I understand this correctly, lychee creates a list of URLs to test, makes an async call to servo which then test if they're up or not.
You're implying that servo should return both the line number and the status of the link.

I think they should be separate, and lychee should know the line number for a URL which is up or down.

This separates concerns cleanly: servo, a web broser engine should not care about the line number in a file that produced the URL it needs to check. That is part of the presentation layer in the actuall app - lychee.

mre:
> If I understand this correctly, lychee creates a list of URLs to test, makes an async call to servo which then test if they're up or not.

Not quite. lychee uses servo's html parser, html5ever, to extract URLs and makes an async call with reqwest, a http client, to test if they are up or not.

html5ever does not return the line number. That information is lost and there is no way to get the line numbers except for parsing the html in some other way. As far as I'm aware, there currently is no html parser in Rust that gives me line numbers (or spans to be precise).

mre:
Closing this as there is currently no way to support this feature due to upstream crate limitations. We can revsisit at a later point in case the tools support it.

untitaker:
html5gum 0.8.0 supports this feature thanks to @akida31, so it could be integrated in lychee now

thomas-zahner:
With merging #1870 we are now a big step closer to complete this. @Akida31 implemented span extraction in lychee-lib. So now we only need to make use of the newly introduced `RawUriSpan` in lychee-bin and show them to users. This should be fairly easy compared to the extraction itself. So any contributions are appreciated.

PR #1870, opened 2025-10-09
Extract spans

thomas-zahner:
Replaces  #1806. Rebased on master and tidied up a few small things. This moves us one big step closer to completing #1304.

Thank you very much @Akida31

### DIFF
--- a/lychee-lib/src/extract/html/html5ever.rs
+++ b/lychee-lib/src/extract/html/html5ever.rs
@@ -2,7 +2,7 @@ use std::cell::RefCell;
 
 use html5ever::{
     buffer_queue::BufferQueue,
-    tendril::StrTendril,
+    tendril::{StrTendril, Tendril, fmt::UTF8},
     tokenizer::{Tag, TagKind, Token, TokenSink, TokenSinkResult, Tokenizer, TokenizerOpts},
 };
 
@@ -48,27 +48,21 @@ impl TokenSink for LinkExtractor {
     #[allow(clippy::match_same_arms)]
     fn process_token(&self, token: Token, line_number: u64) -> TokenSinkResult<()> {
         debug_assert_ne!(line_number, 0);
+        let line_number =
+            usize::try_from(line_number).expect("Unable to convert u64 line_number to usize");
+
         match token {
             Token::CharacterTokens(raw) => {
                 if self.current_verbatim_element_name.borrow().is_some() {
                     return TokenSinkResult::Continue;
                 }
                 if self.include_verbatim {
-                    // offset line number by line breaks included in the raw text,
-                    // as we are provided multiline `Tendril`s
-                    let line_number = line_number.saturating_sub(
-                        raw.chars()
-                            .filter(|c| *c == '\n')
-                            .count()
-                            .try_into()
-                            .unwrap(),
-                    );
                     self.links
                         .borrow_mut()
                         .extend(extract_raw_uri_from_plaintext(
                             &raw,
                             &LineOffsetSpanProvider {
-                                lines_before: line_number.try_into().unwrap(),
+                                lines_before: respect_multiline_tendril(line_number, &raw),
                                 inner: &SourceSpanProvider::from_input(&raw),
                             },
                         ));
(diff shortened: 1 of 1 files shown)


thomas-zahner:
@Akida31 as you can see in CI `test_include_verbatim` fails with:

```
thread 'extract::html::html5ever::tests::test_include_verbatim' panicked at lychee-lib/src/extract/html/html5ever.rs:349:9:
    assertion `left == right` failed
      left: [RawUri { text: "https://example.com/", element: None, attribute: None, span: RawUriSpan { line: 4, column: None } }, RawUri { text: "https://example.org/", element: Some("a"), attribute: Some("href"), span: RawUriSpan { line: 4, column: None } }, RawUri { text: "https://foo.com/", element: None, attribute: None, span: RawUriSpan { line: 10, column: Some(9) } }, RawUri { text: "http://bar.com/some/path", element: None, attribute: None, span: RawUriSpan { line: 10, column: Some(29) } }, RawUri { text: "https://baz.org/", element: Some("a"), attribute: Some("href"), span: RawUriSpan { line: 9, column: None } }]
     right: [RawUri { text: "https://example.com/", element: None, attribute: None, span: RawUriSpan { line: 4, column: None } }, RawUri { text: "https://example.org/", element: Some("a"), attribute: Some("href"), span: RawUriSpan { line: 4, column: None } }, RawUri { text: "https://foo.com/", element: None, attribute: None, span: RawUriSpan { line: 7, column: None } }, RawUri { text: "http://bar.com/some/path", element: None, attribute: None, span: RawUriSpan { line: 7, column: None } }, RawUri { text: "https://baz.org/", element: Some("a"), attribute: Some("href"), span: RawUriSpan { line: 9, column: None } }]
```

So the line numbers and columns for `https://foo.com/` and `http://bar.com/some/path` are no longer correct.
This is caused by rebasing to master where html5ever was updated from version 0.31.0 to 0.35.0. I've debugged the tokens in `process_token` and noticed the following difference between the old and new version:

```
<     Tendril<UTF8>(shared: "        Some random text"),
< )
< [lychee-lib/src/extract/html/html5ever.rs:51:15] token = CharacterTokens(
<     Tendril<UTF8>(inline: "\n"),
< )
< [lychee-lib/src/extract/html/html5ever.rs:51:15] token = CharacterTokens(
<     Tendril<UTF8>(shared: "        https://foo.com and http://bar.com/some/path"),
< )
< [lychee-lib/src/extract/html/html5ever.rs:51:15] token = CharacterTokens(
<     Tendril<UTF8>(inline: "\n"),
< )
< [lychee-lib/src/extract/html/html5ever.rs:51:15] token = CharacterTokens(
<     Tendril<UTF8>(shared: "        Something else"),
< )
< [lychee-lib/src/extract/html/html5ever.rs:51:15] token = CharacterTokens(
<     Tendril<UTF8>(inline: "\n"),
< )
< [lychee-lib/src/extract/html/html5ever.rs:51:15] token = CharacterTokens(
<     Tendril<UTF8>(inline: "        "),
---
>     Tendril<UTF8>(shared: "        Some random text\n        https://foo.com and http://bar.com/some/path\n        Something else\n        "),
```

So in summary the newer html5ever version produces a single `Tendril` containing multiple lines, whereas the older version produced a single `Tendril` for each line. I wasn't yet able to figure out adapt to this new behaviour. The problem might have to do with `SourceSpanProvider::from_input`. Do you know how to fix this?

Akida31:
I think the problem is that the line number given by html5ever is the number of the last line of the multi-line tendril. So I think the following patch should fix that by giving the span provider the line number of the first line of the tendril. I didn't test this patch, so I don't know if this fixes the issue.

```patch
diff --git a/lychee-lib/src/extract/html/html5ever.rs b/lychee-lib/src/extract/html/html5ever.rs
index d7fa67bbf1..2c70d00865 100644
--- a/lychee-lib/src/extract/html/html5ever.rs
+++ b/lychee-lib/src/extract/html/html5ever.rs
@@ -58,6 +58,14 @@
                     return TokenSinkResult::Continue;
                 }
                 if self.include_verbatim {
+                    // offset line number by line breaks included in the raw text
+                    let line_number = line_number.saturating_sub(
+                        raw.chars()
+                            .filter(|c| *c == '\n')
+                            .count()
+                            .try_into()
+                            .unwrap(),
+                    );
                     self.links
                         .borrow_mut()
                         .extend(extract_raw_uri_from_plaintext(
```

If you have questions how my larger changes work, I'm happy to answer them. Sadly, documenting code isn't a strength of mine currently.

thomas-zahner:
This indeed fixed the problem :+1:
I'm now finally merging this. Note that we don't make use of the spans yet in lychee-bin as requested in #1304.
But I think the hardest part in getting there is completed with this PR so thanks again @Akida31!

Issue #1875, opened 2025-10-20
Streamline release for Windows & macOS


thomas-zahner:
Currently, `linux`, `windows` and `macos` are separate steps in release-binary.yml. It should not be necessary to distinguish at job level between these three targets. Incorporate `windows` and `macos` into the `linux` job and rename this job `release` or similar.

Take a look at https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/release.yml for inspiration.

PR #1876, opened 2025-10-20
Update release workflow

thomas-zahner:
With 9f23103 I unfortunately broke the release-binary.yml workflow. Because we now want to generate the man pages in CI we need to run lychee itself, which in turn requires qemu. This PR is heavily inspired by [ripgrep's release workflow](https://github.com/BurntSushi/ripgrep/blob/master/.github/workflows/release.yml).

The qemu invocations with docker are actually really fast, which surprised me. They take about one second to run.
You see that I also update the target matrix, which for now I've cloned from ripgrep. Would you like to keep the original matrix? Having this one seems to be fine to me, since ripgrep has probably a much broader user base it should cover enough users.

I've notice that Windows and macOS are completely separated in this workflow. IMO it makes sense to streamline the release for all three targets, as is done in by ripgrep. I've opened #1875 for that.

See https://github.com/thomas-zahner/lychee/actions/runs/18653717357 for a workflow run.

### DIFF
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -43,47 +43,36 @@ jobs:
     strategy:
       matrix:
         include:
-          - build: linux
-            os: ubuntu-latest
-            rust: nightly
+          - build: x86_64-gnu
+            target: x86_64-unknown-linux-gnu
+            strip: x86_64-linux-gnu-strip
+          - build: x86_64-musl
             target: x86_64-unknown-linux-musl
             strip: x86_64-linux-musl-strip
-          - build: stable-x86
-            os: ubuntu-latest
-            rust: stable
-            target: i686-unknown-linux-gnu
-            strip: x86_64-linux-gnu-strip
-            qemu: i386
-          - build: stable-aarch64
-            os: ubuntu-latest
-            rust: stable
-            target: aarch64-unknown-linux-gnu
-            strip: aarch64-linux-gnu-strip
-            qemu: qemu-aarch64
-          - build: stable-arm-gnueabihf
-            os: ubuntu-latest
-            rust: stable
+          - build: arm-gnueabihf
             target: armv7-unknown-linux-gnueabihf
             strip: arm-linux-gnueabihf-strip
             qemu: qemu-arm
-          - build: stable-arm-musleabihf
-            os: ubuntu-latest
-            rust: stable
-            target: armv7-unknown-linux-musleabihf
-            strip: arm-linux-musleabihf-strip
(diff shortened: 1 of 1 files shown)


.github/workflows/release-binary.yml
```diff
@@ -30,6 +30,7 @@ jobs:
       - name: Get release
         id: get_release
         uses: bruceadams/get-release@v1.3.2
+        if: ${{ github.repository == 'lycheeverse/lychee' && github.ref == 'refs/heads/master' }}
```
thomas-zahner: Without this line, the workflow always fails here, if not on master on the official lychee repository. ([see here](https://github.com/thomas-zahner/lychee/actions/runs/18646155804/job/53153785334)) This was the reason, why I only properly tested the workflow in master in the official repository :sweat_smile: 

With this change I can verify that all the targets still build and it will only fail in the upload step in the end.

mre: Should we add that as a comment to the workflow?

.github/workflows/release-binary.yml
```diff
@@ -40,17 +41,48 @@ jobs:
     needs: prepare
     strategy:
       matrix:
-        target:
-          - x86_64-unknown-linux-gnu
-          - x86_64-unknown-linux-musl
-          - arm-unknown-linux-gnueabihf
-          - arm-unknown-linux-musleabi
-          - arm-unknown-linux-musleabihf
```
mre, line 48: These arm builds don't exist in the new build matrix anymore. That's probably correct, right? I just don't want us to skip any previously supported build targets. 😅 

Same for `aarch64-unknown-linux-musl` below.

.github/workflows/release-binary.yml
mre: We might want to switch to `main` at some point. Do you think it makes sense to add `main` to these if-conditions already?

thomas-zahner: Good points :+1:

.github/workflows/release-binary.yml
thomas-zahner: True, that makes sense. I've restructured the matrix so it should be identical to the current master version. Only differences are:

- added i686-unknown-linux-gnu (32bit x86)
- replaced `arm-unknown-linux-gnueabihf` with `armv7-unknown-linux-gnueabihf`. Without `v7` the `Strip binary (cross)` step fails. This means armv7 instead of armv6 which is a difference. However, armv6 users are still able to use the results of `arm-musleabi` and `arm-musleabihf` as to my understanding.

PR #1881, opened 2025-10-24
Use 'match' for version pattern matching and not 'pattern'

eread:
This issue: https://github.com/lycheeverse/lychee/issues/1833 wasn't solved by: https://github.com/lycheeverse/lychee/pull/1834.

I think it's because the action was trying to pattern match the version with the wrong setting.

After reviewing: https://github.com/docker/metadata-action/issues/56, I think the action needs to use `match`, rather than `pattern`.


thomas-zahner:
Thanks, let's see if this works :)

PR #1882, opened 2025-10-25
 fix: inverted gitignore behaviour for --dump-inputs 

katrinafyi:
the use of `no_ignore` was missing a `!` before being passed as skip_ignored. i figure there was some confusion about the meaning of "no ignore", it can be either:

- skip ignored files, or
- skip the ignore process itself.

the actual meaning is the second one. the no_ignore value was used inconsistently between the main logic and the dump-inputs logic.

related to https://github.com/lycheeverse/lychee/issues/1767

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -332,7 +332,8 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
             &opts.config.exclude_path,
             &opts.config.extensions,
             !opts.config.hidden,
-            opts.config.no_ignore,
+            // be aware that "no ignore" means do *not* ignore files
+            !opts.config.no_ignore,
         )
         .await?;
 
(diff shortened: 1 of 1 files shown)


thomas-zahner:
Thanks for catching that. Doesn't this fully resolve https://github.com/lycheeverse/lychee/issues/1767? Did you still observe any notable differences? In the issue I only mentioned the ignore inconsistency. :sweat_smile: I should probably double check to see if I spot any other problems.

katrinafyi:
Maybe! It seemed like you and mre had already covered most of the issues, so I didn't look very hard for other differences. But nothing else stood out to me.

thomas-zahner:
Coincidentally, while resolving clippy suggestions (too many function parameters) in #1891 I extracted the three boolean flags into a `Skip` struct. While, doing so I also solved this bug. I now cherry-picked your test commit of this PR.

Could you take a look at https://github.com/lycheeverse/lychee/pull/1891/commits/96eae641eb688591286c292e282a99e55e3bdb88? If you agree with the changes I would like to close this PR in favour of #1891. Reviews of my PR are welcome.

PR #1887, opened 2025-10-27
test: require exact version number in README's --help output

katrinafyi:
previously, the README test would ignore any line with `lychee/` so that the README's lychee/x.y.z could still match the lychee --help output of lychee/0.21.0 (or similar).

however, this scrubbing is too coarse, because it also passes the test when the README contains a line which is lychee/0.99.0 (or any other version number). this has a *very small* possibility of confusing users and is a minor inaccuracy.

this PR is one fix to this problem, by requiring that the README contain the exact version number.

an alternative fix would be to keep `remove_lychee_version_line` but change it to require `x.y.z` on the README side. this would assert that the README text uses the `lychee/x.y.z` form.

### DIFF
--- a/lychee-bin/tests/usage.rs
+++ b/lychee-bin/tests/usage.rs
@@ -10,15 +10,6 @@ mod readme {
         Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("Couldn't get cargo package name")
     }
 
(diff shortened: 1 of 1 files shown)


katrinafyi:
This would complicate the release proceess because you'd need to manually update the README within the release branch in order to get a green CI.

I guess it has to be the alternative fix. But maybe another time

PR #1889, opened 2025-10-27
Match more than one digit for version numbers

eread:
As suggested here: https://github.com/lycheeverse/lychee/issues/1833#issuecomment-3449970411 by @thomas-zahner, let's test the result of https://github.com/lycheeverse/lychee/pull/1881.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,8 +38,8 @@ jobs:
             type=schedule
             type=ref,event=branch
             type=ref,event=pr
-            type=semver,pattern={{version}},value=${{ github.ref_name }},match=lychee-v(\d.\d.\d)$
-            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},match=lychee-v(\d.\d.\d)$
+            type=semver,pattern={{version}},value=${{ github.ref_name }},match=lychee-v(\d+.\d+.\d+)$
+            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},match=lychee-v(\d+.\d+.\d+)$
             type=sha
             # set latest tag for default branch
             type=raw,value=latest,enable={{is_default_branch}}
@@ -59,8 +59,8 @@ jobs:
             type=schedule
             type=ref,event=branch
             type=ref,event=pr
-            type=semver,pattern={{version}},value=${{ github.ref_name }},match=lychee-v(\d.\d.\d)$
-            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},match=lychee-v(\d.\d.\d)$
+            type=semver,pattern={{version}},value=${{ github.ref_name }},match=lychee-v(\d+.\d+.\d+)$
+            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},match=lychee-v(\d+.\d+.\d+)$
             type=sha
             # set latest tag for default branch
             type=raw,value=latest,enable={{is_default_branch}}


eread:
@thomas-zahner Ok, I think we need one more minor change based on this commit: https://github.com/lycheeverse/lychee/pull/1889/commits/68cbed079e97665b2a218377440421f0fec1cbe0, and these runs:

- https://github.com/lycheeverse/lychee/actions/runs/18862068098/job/53822174152#step:3:56
- https://github.com/lycheeverse/lychee/actions/runs/18862068098/job/53822174152#step:4:85

Basically, `/d` will match only a single digit, whereas the minor version number for Lychee right now is more than one digit, so the regular expression as it currently is doesn't quite catch it.

I'll refactor this PR to just make that change, and I think the Docker images will be properly tagged for the next release!

eread:
@thomas-zahner Could you review? I believe this change is required (see: https://github.com/lycheeverse/lychee/pull/1889#issuecomment-3454266641).

.github/workflows/docker.yml
```diff
@@ -38,8 +38,8 @@ jobs:
             type=schedule
             type=ref,event=branch
             type=ref,event=pr
-            type=semver,pattern={{version}},value=${{ github.ref_name }},match=lychee-v(\d.\d.\d)$
-            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},match=lychee-v(\d.\d.\d)$
+            type=semver,pattern={{version}},value=${{ github.ref_name }},match=lychee-v(\d.\d+.\d)$
+            type=semver,pattern={{major}}.{{minor}},value=${{ github.ref_name }},match=lychee-v(\d.\d+.\d)$
```
thomas-zahner: It's quite likely that we reach multiple digits in the patch or major version at some point in the future. So we should add `+` to major, minor and patch

.github/workflows/docker.yml
```diff
@@ -59,8 +59,8 @@ jobs:
...
```
thomas-zahner: Same thing applies here, of course

eread:
> Glad to hear that you could test the changes in your local repository and thank you for testing and submitting this PR.

Thanks for the review @thomas-zahner 🙇🏻 

I thought you might ask for all three parts of the version number to be similarly handled.

Here's a commit for testing: <https://github.com/lycheeverse/lychee/pull/1889/commits/6c48cfc95f8d9c1c819d7d4b8fcd32f031512690>

Here's the test run with results:

- <https://github.com/lycheeverse/lychee/actions/runs/18891603745/job/53920314460#step:3:56>
- <https://github.com/lycheeverse/lychee/actions/runs/18891603745/job/53920314460#step:4:85>

I'll now refactor the PR to just have the required changes.

.github/workflows/docker.yml
eread: @thomas-zahner Done!

.github/workflows/docker.yml
eread: @thomas-zahner Done!

nobkd:
Normally when I see `.` in a regex, I would expect it to match anything. Is this applicable here, and has to be escaped to be a real dot (like `\.`)?
(just, if this is applicable here, the regex dot would probably have coincidentally matched the real dot beforehand)

(edit: It doesn't really matter that much, as the match is not used very often, but the group would match something like `123-1/1`, not only `123.1.1`, and this might not be the desired match)

eread:
> Normally when I see `.` in a regex, I would expect it to match anything. Is this applicable here, and has to be escaped to be a real dot (like `\.`)? (just, if this is applicable here, the regex dot would probably have coincidentally matched the real dot beforehand)
> 
> (edit: It doesn't really matter that much, as the match is not used very often, but the group would match something like `123-1/1`, not only `123.1.1`, and this might not be the desired match)

Thanks for chiming in @nobkd. I agree that `.` normally acts as an operator of some kind. I didn't dig in to what sort of regex parser that the action uses, just looked at the documentation: https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver.

nobkd:
They use the unescaped version of the dots in many patterns, so this probably works as intended :shrug:

thomas-zahner:
True, for correctness we could replace `.` with `\.`. However, it shouldn't pose a problem as we only create tags with version numbers in the format of `lychee-v1.2.3`.

Issue #1190, opened 2023-07-31
new srcset parsing breaks on sentry-docs


untitaker:
#1160 broke running lychee on sentry-docs. Found by bisecting `434baa8..HEAD`.

In order to repro, [check out sentry-docs](https://github.com/getsentry/sentry-docs), run `yarn && yarn build` and run `lychee --offline public/`, or use the attached archive.

I tried to make a minimal example, but I failed to do so because I ran into a different, much older bug.

Apparently lychee ignores absolute paths under some circumstances:

```
$ cat index.html
<img src="/trace-view.png" srcSet="/trace-view.png 300w" />

$ lychee index.html  # or lychee .
  0/0 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                      🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors
```

this is mysterious to me because sentry-docs uses plenty of absolute URLs.

Is there some nuance to how lychee treats absolute paths?

----

## About the archive.

I had to figure out how to attach 500MB of html to a github issue when github would only allow 25 MB uploads.

The solution I came up with was to make all non-html files empty (which does not affect link-checking), and compressing the file with `zstd -9` instead of gzip to get from 50MB to 7MB.

since github does not allow any upload other than `.tgz`, the file extension is incorrect too. you can't extract this with a regular archival tool, you need to use:

```
zstdcat public.tgz | tar xz
```

this will create a `public/` folder in the cwd.

Download archive here: [public.tgz](https://github.com/lycheeverse/lychee/files/12216164/public.tgz)

mre:
Ouch, sorry to hear that.

> Is there some nuance to how lychee treats absolute paths?

You need to set the base directory with `--base`:

```sh
> echo '<img src="/trace-view.png" srcSet="/trace-view.png 300w" />' > index.html
> lychee --dump index.html
> lychee --dump --base . index.html
file:///private/tmp/trace-view.png
file:///private/tmp/trace-view.png
```

I managed to reproduce the bug locally.

Before (latest released version):

```
> lychee --offline --base . --dump platforms/apple/dsym/index.html | grep static | sort                                                                                                                                                                        
file:///private/tmp/sentry/static/49827f66243ac136dced551c15e979aa/0a47e/integration-tokens.png
file:///private/tmp/sentry/static/49827f66243ac136dced551c15e979aa/5a46d/integration-tokens.png
[...]

PR #1890, opened 2025-10-28
fix: parsing of srcset URLs after the first URL

katrinafyi:
previously, the first character of srcset URLs past the first URL had their first character dropped. this led to absolute URLs becoming relative and other problems. this happened when the srcset did not have spaces after its commas: `/300.png 300w,/600.png 600w`

really, this could be fixed by simply deleting the `index += 1;` line. this increment causes a mismatch between the index and the remaining string which causes the off-by-one error.

this pr makes some extra changes to avoid this duplicated logic by removing the index variable entirely, since the remaining string is all you need. this avoids needing to think about indices and should avoid similar bugs in future.

this pr also changes some pattern match of space to accept other ASCII whitespace, following the spec.

for those curious, the new test case previously failed with this error:
```
      left: ["/300.png", "600.png", "900.png", "ttps://x.invalid/a.png", "elative.png"]
     right: ["/300.png", "/600.png", "/900.png", "https://x.invalid/a.png", "relative.png"]
```

related to https://github.com/lycheeverse/lychee/issues/1190

### DIFF
--- a/lychee-lib/src/extract/html/srcset.rs
+++ b/lychee-lib/src/extract/html/srcset.rs
@@ -50,40 +50,6 @@ where
 // This state-machine is a bit convoluted, but we keep everything in one place
 // for simplicity so we have to please clippy.
 pub(crate) fn parse(input: &str) -> Vec<&str> {
-    /// Implements one iteration of the "splitting loop" from the reference algorithm.
-    /// This is intended to be repeatedly called until the remaining string is empty.
-    ///
-    /// Returns a tuple of remaining string and an optional parsed URL, if successful.
-    /// Otherwise, in case of srcset syntax errors, returns Err.
-    ///
-    /// <https://html.spec.whatwg.org/multipage/images.html#parsing-a-srcset-attribute>
-    fn parse_one_url(remaining: &str) -> Result<(&str, Option<&str>), String> {
-        let (start, remaining) = split_at(remaining, |c| *c == ',' || c.is_ascii_whitespace());
-
-        if start.find(',').is_some() {
-            return Err("srcset parse error (too many commas)".to_string());
-        }
-
-        if remaining.is_empty() {
-            return Ok(("", None));
-        }
-
-        let (url, remaining) = split_at(remaining, |c| !c.is_ascii_whitespace());
-
-        let comma_count = url.chars().rev().take_while(|c| *c == ',').count();
-        if comma_count > 1 {
-            return Err("srcset parse error (trailing commas)".to_string());
-        }
-
-        let url = url.get(..url.len() - comma_count);
-
-        let (_spaces, remaining) = split_at(remaining, char::is_ascii_whitespace);
-
-        let remaining = skip_descriptor(remaining);
-
-        Ok((remaining, url))
-    }
-
     let mut candidates: Vec<&str> = Vec::new();
     let mut remaining = input;
     while !remaining.is_empty() {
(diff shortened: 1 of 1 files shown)


lychee-lib/src/extract/html/srcset.rs
```diff
@@ -46,78 +50,90 @@ where
...
+/// Helper function to skip over a descriptor. Returns the string remaining
+/// after the descriptor (i.e. a string beginning after the next comma or an
+/// empty string).
+#[allow(clippy::single_match)]
+fn skip_descriptor(remaining: &str) -> &str {
     let mut state = State::InsideDescriptor;
 
-    for c in remaining.chars() {
-        index += 1;
+    // ASCII whitespace is U+0009 TAB, U+000A LF, U+000C FF, U+000D CR, or U+0020 SPACE.
```
thomas-zahner: IMO you can remove this comment and replace the checks below with [is_ascii_whitespace](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_whitespace). It perfectly matches your comment:

```
Rust uses the WhatWG Infra Standard’s [definition of ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).
```

lychee-lib/src/extract/html/srcset.rs
```diff
@@ -46,78 +50,90 @@ where
...
+    // https://infra.spec.whatwg.org/#ascii-whitespace
 
+    let mut it = remaining.chars();
+    while let Some(c) = it.next() {
         match state {
             State::InsideDescriptor => match c {
-                ' ' => state = State::AfterDescriptor,
+                ' ' | '\t' | '\n' | '\u{000C}' | '\r' => state = State::AfterDescriptor,
```
thomas-zahner: so it becomes something like

```rust
c if c.is_ascii_whitespace() =>
```

lychee-lib/src/extract/html/srcset.rs
```diff
@@ -46,78 +50,90 @@ where
...
-                _ => {}
+                ',' => return it.as_str(), // returns string after this comma
+                _ => (),
+            },
+            State::InsideParens => match c {
+                ')' => state = State::InsideDescriptor,
+                _ => (),
+            },
+            State::AfterDescriptor => match c {
+                ' ' | '\t' | '\n' | '\u{000C}' | '\r' => (),
```
thomas-zahner: same here

thomas-zahner (changes requested):
Ah thank you very much for this bugfix :rocket:

lychee-lib/src/extract/html/srcset.rs
katrinafyi: Yeah so true, I've changed it to use is_ascii_whitespace

lychee-lib/src/extract/html/srcset.rs
```diff
@@ -46,78 +50,86 @@ where
...
-        let position = &input[index..];
-        let (start, remaining) = split_at(position, |c| *c == ',' || c.is_whitespace());
+    /// Implements one iteration of the "splitting loop" from the reference algorithm.
+    /// This is intended to be repeatedly called until the remaining string is empty.
+    ///
+    /// Returns a tuple of remaining string and an optional parsed URL, if successful.
+    /// Otherwise, in case of srcset syntax errors, returns Err.
+    ///
+    /// <https://html.spec.whatwg.org/multipage/images.html#parsing-a-srcset-attribute>
+    fn parse_one_url(remaining: &str) -> Result<(&str, Option<&str>), String> {
```
thomas-zahner: Could you extract this function, just above `skip_descriptor`. Less nesting makes it more readable IMO.

katrinafyi: Sure! I like less indentation too

thomas-zahner: Thanks :+1:

Issue #1672, opened 2025-03-31
File format preprocessing


thomas-zahner:
lychee can currently extract and check links from plain text file formats. In addition to "simple" plaintext link extraction lychee internally tries to detect HTML and Markdown formats to parse them with more specific parser.

This way link extraction on websites (HTML) and local plaintext file formats works quite well most of the time. But some plain text files will not work well, such as CSV files without whitespace and quotes. (https://github.com/lycheeverse/lychee/issues/1299) This is because there is no way of knowing when a URL ends as subsequent CSV columns are separated with `,` which could be part of the path in the URL. Additionally, it is impossible to use lychee on non-plaintext files. It was previously suggested that lychee should add support for the PDF (https://github.com/lycheeverse/lychee/issues/1583), EPUB (https://github.com/lycheeverse/lychee/issues/202) and AsciiDoc (https://github.com/lycheeverse/lychee/issues/291)  file format.

I think that officially supporting the above examples with lychee for every user would be the wrong approach. It should not be lychee's responsibility to understand every single file format and every edge-case of file format types. (as with the CSV example) Attempting to do so would bloat lychee and make it hard to maintain.

Instead it should be the user's responsibility to preprocess file formats in such a way that lychee can understand them. To help users do this we should provide documentation on how this is done and we should probably introduce a new command line flag called `--preprocess`:

```
lychee --preprocess "csv: cat {} | csvq 'select *'" --preprocess "pdf: pdftotext {} -" my-project-to-check/
```

This flag is comparable to the `exec` flag for find and fd.

Additionally, there were discussions about depending on pandoc to support more file formats. (https://github.com/lycheeverse/lychee/issues/202#issuecomment-1080771637) However, I think this should not be necessary when we introduce this new flag and provide enough user documentation. I think depending on pandoc would be a very opinionated choice. And for certain file formats or specific use-cases pandoc might not be suitable. In the EPUB example, a user suggested another program that was probably better suited to their needs. (https://github.com/lycheeverse/lychee/issues/202#issuecomment-879634519) If the demand for a "all file type knowing" link checker really would be that big we could still create a new repository called "lychee-all" akin to [ripgrep-all](https://github.com/phiresky/ripgrep-all).

thomas-zahner:
The rough work steps might include:

- [x] CLI flag for file preprocessing
- [x] lychee-lib option for file preprocessing (for library users and a potential "lychee-all")
- [x] User documentation on how to handle common file formats

[...]

PR #1891, opened 2025-10-29
feat: file preprocessing

thomas-zahner:
Addresses #1672

I've created a new repository for testing and documenting compatibility with other file formats: https://github.com/thomas-zahner/lychee-all
We might want to merge this information into the official docs later.

As mentioned [in the issue](https://github.com/lycheeverse/lychee/issues/1672#issuecomment-3432088247) this PR is heavily inspired by [ripgrep's preprocessor](https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#preprocessor).

### DIFF
--- a/README.md
+++ b/README.md
@@ -621,17 +621,15 @@ Options:
 
           #!/usr/bin/env bash
           case "$1" in
-          *.epub|*.odt|*.docx|*.ipynb)
-              exec pandoc "$1" --to=html --wrap=none --markdown-headings=atx
-              ;;
-          *.adoc|*.asciidoc)
-              asciidoctor -a stylesheet! "$1" -o -
-              ;;
           *.pdf)
-              exec pdftotext "$1" -
+              exec pdftohtml -i -s -stdout "$1"
+              ;;
+          *.odt|*.docx|*.epub|*.ipynb)
+              exec pandoc "$1" --to=html --wrap=none
               ;;
           *)
-              exec cat # identity function, output input without changes
+              # identity function, output input without changes
+              exec cat
               ;;
           esac
 
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -871,17 +871,15 @@ wrapper program such as a shell script. An example script looks like this:
 
 #!/usr/bin/env bash
 case "$1" in
-*.epub|*.odt|*.docx|*.ipynb)
-    exec pandoc "$1" --to=html --wrap=none --markdown-headings=atx
-    ;;
-*.adoc|*.asciidoc)
-    asciidoctor -a stylesheet! "$1" -o -
-    ;;
 *.pdf)
-    exec pdftotext "$1" -
+    exec pdftohtml -i -s -stdout "$1"
+    ;;
+*.odt|*.docx|*.epub|*.ipynb)
+    exec pandoc "$1" --to=html --wrap=none
     ;;
 *)
-    exec cat # identity function, output input without changes
+    # identity function, output input without changes
+    exec cat
     ;;
 esac"#
     )]


Makefile
```diff
@@ -56,7 +56,7 @@ test: ## Run tests
 
 .PHONY: bench
 bench: ## Run benchmarks
-	cargo bench
+	cargo bench --profile=dev
```
thomas-zahner: This is necessary for it to compile. Otherwise we get multiple:

```
ErrorKind::TestError
           ^^^^^^^^^ variant or associated item not found in `ErrorKind`
```

The same happens if you run `cargo test --release`.
Which is why I adjusted https://github.com/NixOS/nixpkgs/blob/2d5a7306d86d4ebfba8b65515260523353745c45/pkgs/by-name/ly/lychee/package.nix#L53 as well. Maybe it would be better to address the problem differently, maybe to remove the `TestError` completely?

fixtures/pre/error_message.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+>&2 echo "Some error message"
```
katrinafyi, line 3: i didn't know you could do that

lychee-bin/src/options.rs
```diff
@@ -853,6 +854,38 @@ and existing cookies will be updated."
...
     #[serde(default)]
     pub(crate) include_wikilinks: bool,
+
+    /// Preprocess input files.
+    #[arg(
+        short,
+        long,
+        value_name = "COMMAND",
+        long_help = r#"Preprocess input files.
+For each input file, this flag causes lychee to process the standard output of COMMAND PATH instead of the contents of PATH.
```
katrinafyi:
```suggestion
For each input file, this flag causes lychee to process the standard output of COMMAND instead of the contents of PATH.
```
typo?

lychee-bin/src/options.rs
```diff
@@ -853,6 +854,38 @@ and existing cookies will be updated."
...
+*.pdf)
+    exec pdftotext "$1" -
+    ;;
+*)
+    exec cat # identity function, output input without changes
+    ;;
+esac"#
+    )]
+    #[serde(default)]
+    pub(crate) pre: Option<Preprocessor>,
```
katrinafyi: I would like a more explicit name like --preprocess. "pre" is a bit non-specific. For instance, in the html world, pre is a HTML tag. But I don't feel strongly about this

lychee-lib/src/types/file.rs
```diff
@@ -121,6 +121,28 @@ pub enum FileType {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Skip {
+    /// Skip input files which do not exist. If `false` missing inputs will lead to an error.
+    pub missing_inputs: bool,
+    /// Skip files which are ignored by git and `.ignore` files.
+    /// See [`ignore`] for more details.
+    pub ignored: bool,
+    /// Skip files that are hidden.
+    pub hidden: bool,
+}
```
katrinafyi: "Skip" is a nebulous concept. I'd like the field names to be skip_x just to be more explicit. Also, the obvious odd one out here is missing_inputs which is about handling errors rather than enumerating files. Although the names are the same, the values are not really related.

If the purpose of this refactor is to collect related arguments into a struct, I think it would make more sense to omit missing_inputs and include file_extensions. This would no longer be "Skip" but it would make sense as a collection of arguments to WalkBuilder. Idk what the name would be, though.

README.md
```diff
@@ -609,6 +609,31 @@ Options:
...
           Only check local files and block network requests
 
+  -p, --pre <COMMAND>
+          Preprocess input files.
+          For each input file, this flag causes lychee to process the standard output of COMMAND PATH instead of the contents of PATH.
+          This allows you to convert files that would otherwise not be understood by lychee.
+          The preprocessor COMMAND is only run on input files, not on standard input or URLs.
+
+          To invoke programs with custom arguments or to use multiple preprocessors use a wrapper program such as a shell script.
+          An example script looks like this:
```
katrinafyi: hard wrapping here is a bit inconsistent.

katrinafyi (commented):
Looks good! The most significant comment is the one about Skip's concept. I think a different collection of values would make more sense.

The rest of this comment is just my own commentary. Feel free to ignore it, and I certainly don't expect anything in this PR to change because of it.

When reading this PR, I had the thought that the new preprocess value has to be handled a lot and passed repeatedly to get to where it needs to be used. I think this is due to an architecture where it's all very hierarchical. It looks something like this (conceptually):
```ocaml
inputs
|> collector(basic_auth, skip, include_verbatim, client, preprocess, ...)
```
Here,  `collector` calls other helper functions and has to amalgamate all their arguments. It is responsible for a lot of functionality, from resolving inputs all the way to link extraction and request building. 

If the architecture was more like a flat pipeline, it would reduce the need for this argument injection. Instead, of one big "collector",  it might look like this:
```ocaml
inputs
|> resolve_inputs(skip, glob_ignore_case)
|> preprocess_inputs(pre_cmd)
|> get_input_contents(basic_auth, retries, max_redirect)
|> extract_links(root_dir, base_url)
```
Hopefully, you can see how this reduces the parameters needed - each step only needs the parameters for its own functionality. A clear pipeline makes it much easier to implement features like --dump or --dump-inputs, which are just stopping at certain points in the pipeline (I started thinking about this because of the dumping issues). It also makes testing easier.

Anyway, this is all theoretical at the moment. I don't know if this is possible or how hard it would be. There is Chain in the codebase, but it's limited to homogenous pipeline functions. Anyway, as I said, nothing that needs to affects this PR right now.

lychee-bin/src/options.rs
katrinafyi: or not a typo. maybe this meand invoking command with path as argument. but it's not clear from the text formatting

fixtures/pre/error_message.sh
thomas-zahner: Me neither until recently :)

lychee-bin/src/options.rs
thomas-zahner: Yes, I want to say that instead of e.g. the contents of README.md lychee will process stdout of `cat README.md` if the command is `cat`. I previously forgot to mention in this PR's description that this is heavily inspired by ripgrep. While writing the help message I was also inspired by ripgrep, so it's quite similar :smile:

Note that `COMMAND` refers to the value of the flag which might be more obvious in the help message output:

```
  -p, --pre <COMMAND>
          Preprocess input files.
          For each input file, this flag causes lychee to process the standard output of COMMAND PATH instead of the contents of PATH.
          This allows you to convert files that would otherwise not be understood by lychee.
          The preprocessor COMMAND is only run on input files, not on standard input or URLs.
```

How would you adjust the message? Would you just like to remove the `COMMAND PATH` part?

lychee-bin/src/options.rs
thomas-zahner: True, we can do that. @mre [once proposed](https://github.com/lycheeverse/lychee/issues/1672#issuecomment-2867169100) `--process-ext` but back then we were still thinking about including extensions in the argument's value. Again here I just copied the name that ripgrep uses.

lychee-bin/src/options.rs
katrinafyi: maybe this? The main changes are backticks and the verb execute in front to signal that what follows is a command line.

> For each file input, this flag causes lychee to execute `COMMAND PATH` and process its standard output instead of the original contents of PATH.

lychee-lib/src/types/file.rs
thomas-zahner: Yes, my intent was exactly to collect related arguments into a struct, mainly because of a clippy warning of passing too many arguments. I'll give this some more thought.

thomas-zahner:
@katrinafyi Thanks for your thoughts.

> If the architecture was more like a flat pipeline, it would reduce the need for this argument injection. Instead, of one big "collector", it might look like this

I really do like this idea and I totally agree. It would probably simplify quite a lot. IMO we could open up an issue to tackle that separately.

Edit: opened up https://github.com/lycheeverse/lychee/issues/1898

lychee-lib/src/types/file.rs
thomas-zahner: Reverted this (by rebasing). See https://github.com/lycheeverse/lychee/pull/1891/files#diff-2937c96b9d2c5c3719710f52cbbd825caf3bd4887a716667dbfa11f4e75a2c63R162

lychee-bin/src/options.rs
thomas-zahner: @mre Do you have any preferences?

lychee-lib/src/types/file.rs
katrinafyi: I see, yeah it makes sense to revert for now

lychee-bin/src/options.rs
mre: I like 'em both, `pre` but also `preprocess`. The argument about pre being an HTML tag makes sense and slightly tips me towards `preprocess` now. While I like short argument names, the command will probably be read more often than it gets written, and in that scenario, `preprocess` is easier to grasp.

Makefile
mre: Ouch. If I understand the cargo bench docs correctly, this would disable optimizations, making the benchmark sort of inaccurate?

We can remove `TestError`. I introduced it in https://github.com/lycheeverse/lychee/pull/1731 to make testing the formatters easier. The idea was that we can create an arbitrary "test error" and check that it gets properly rendered.

So either we change that in this PR, or we add a note to look into it later.

lychee-bin/src/options.rs
```diff
@@ -853,6 +854,39 @@ and existing cookies will be updated."
...
+        short,
+        long,
+        value_name = "COMMAND",
+        long_help = r#"Preprocess input files.
+For each file input, this flag causes lychee to execute `COMMAND PATH` and process
+its standard output instead of the original contents of PATH. This allows you to
+convert files that would otherwise not be understood by lychee. The preprocessor
+COMMAND is only run on input files, not on standard input or URLs.
+
+To invoke programs with custom arguments or to use multiple preprocessors use a
```
mre:
```suggestion
To invoke programs with custom arguments or to use multiple preprocessors, use a
```

lychee-lib/src/types/input/resolver.rs
```diff
@@ -52,6 +54,29 @@ impl InputResolver {
...
+        skip_hidden: bool,
+        skip_gitignored: bool,
+    ) -> Result<Walk> {
+        Ok(WalkBuilder::new(path)
+            .git_ignore(skip_gitignored)
+            .git_global(skip_gitignored)
+            .git_exclude(skip_gitignored)
+            .ignore(skip_gitignored)
+            .parents(skip_gitignored)
+            .hidden(skip_hidden)
```
mre: Can we retain some of the old comments and/or write new ones to explain what each option means? I understand that it's just a command-click (or `gd`) away, but I find it easier if we add some inline docs for when I look at code outside my IDE.

lychee-lib/src/types/preprocessor/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn from(s: String) -> Self {
+        Self(s)
+    }
+}
+
+impl Preprocessor {
+    /// Try to invoke the preprocessor command with `path` as single argument
+    /// and return the resulting stdout.
+    pub(crate) fn process(&self, path: &PathBuf) -> Result<String> {
+        let pre = &self.0;
```
mre: That's kind of the reason why I prefer structs for newtypes recently. E.g. with

```rust
pub struct Preprocessor {
  command: String,
}
```

this would read nicely: `let pre = &self.command`.

I personally would also do the borrow where the command gets used, so I'd write

```rust
let command = self.command;
let output = Command::new(&command)...
```

or maybe even

```rust
let output = Command::new(&self.command)...
```


But that's definitely just a stylistic choice.

lychee-lib/src/types/error.rs
```diff
@@ -170,6 +170,10 @@ pub enum ErrorKind {
     #[error("Status code range error")]
     StatusCodeSelectorError(#[from] StatusCodeSelectorError),
 
+    /// Preprocessor command error
+    #[error("Preprocessor command '{0}' failed: {1}")]
+    PreprocessorError(String, String),
```
mre: Since both arguments are strings, they can easily get mixed up during refactoring. I find this easier to read:

```rust
PreprocessorError {
  command: String,
  reason: String,
}
```

lychee-lib/src/types/preprocessor/mod.rs
thomas-zahner: I totally agree :+1:
Only thing that doesn't work is `let command = self.command;` because it can't be borrowed here.

Makefile
thomas-zahner: Yeah, my thoughts as well. Removed `TestError` in dc800ea and replaced references with the `EmptyUrl` variant which does not take any parameters

PR #1893, opened 2025-10-30
refactor: remove unimportant public library function

thomas-zahner:
I noticed the public library function `collect_sources_with_file_types` which is not used by lychee-bin. There could be library users depending on this function but IMO this is unlikely. The reason I would like to remove this function is because it's not used by ourselves and I think it is not very useful, it's not a feature that library users of a link checker would expect.

Note that all the function does, is collect the provided inputs, convert them into the path string representation. If library users really need to do this they are still able to so on their own when we remove the function.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2562,7 +2562,6 @@ dependencies = [
  "cached",
  "check-if-email-exists",
  "cookie_store 0.22.0",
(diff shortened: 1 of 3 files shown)


mre:
Looks good. Can you fix the merge conflicts? After that, feel free to merge. Thanks! 👍

PR #1895, opened 2025-10-30
Fix typos, and configure the typos tool

linkmauve:
[typos](https://github.com/crate-ci/typos) is a nice tool which finds and fixes typos in a codebase.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,12 @@ env:
   RUSTFLAGS: -D warnings
 
 jobs:
+  typos:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: crate-ci/typos@v1.38.1
+
   test:
     runs-on: ubuntu-latest
     steps:


thomas-zahner:
Thank you for the PR. I agree that this is a nice addition.
Can you also add a check to CI, potentially with the dedicated [GitHub action](https://github.com/crate-ci/typos/blob/master/docs/github-action.md)? This way we won't miss new typos.

Also, I've just addressed the new linting issues on master. So by rebasing on master CI should pass again.

linkmauve:
Yay, the added CI job seems to work just fine! https://github.com/lycheeverse/lychee/actions/runs/18968249929/job/54169626137?pr=1895

I’ll remove the testing commit then.

Edit: wow, that job finishes in just four seconds, that’s fast!

Edit 2: #1897 fixes one small issue your clippy fix uncovered.

thomas-zahner:
Really cool, thanks for the PR. And yes four seconds is surprisingly quick. :)
Also, did a quick test if all works like I would expect and it does. See https://github.com/thomas-zahner/lychee/actions/runs/18972333688/job/54182928624 (ah, just noticed you already did the same, thanks :+1: )

PR #1897, opened 2025-10-31
Remove an unneeded clone

linkmauve:
In 8cc0cf0c7e78a8c07f8b62995ad3873011e79f6d, clippy suggested to replace a `String::to_string()` with a `Clone::clone()`, without noticing that the value was passed to `Hash::hash()` anyway which borrows the value.


thomas-zahner:
Ah thanks for glancing over my commit. :smile:
Noticed one very similar case which I did address but missed this one.

PR #1906, opened 2025-11-06
Config file up to date

thomas-zahner:
This PR addresses the following (also see commit history)

- Reject unknown TOML keys, which is quite a nice addition, as I've struggled in the past with this. Also, I found quite a few mistakes in the fixtures because of this.
- Ensure full documentation coverage in lychee.example.toml
- `files_from` is now usable within the config file
- Refactor to simplify CLI tests

Addresses https://github.com/lycheeverse/lychee/issues/1768

### DIFF
--- a/fixtures/configs/files_from/example.md
+++ b/fixtures/configs/files_from/example.md
@@ -1,0 +1,1 @@
+https://wikipedia.org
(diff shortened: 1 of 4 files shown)


mre (approved):
Great! I don't see any issues. 👍

Issue #1903, opened 2025-11-04
Better documentation and error messages for verbose option in `lychee.toml`


user27182:
I can use the `--verbose` flag with CLI as  `lychee . --verbose`.

But if I want to move this to `lychee.toml`, it's not currently obvious how to do this. I tried `verbose = true`, but got this error:

```
% lychee .          
   [ERROR] Error while loading config: Cannot load default configuration file `lychee.toml`: Failed to parse configuration file

Caused by:
    TOML parse error at line 17, column 11
       |
    17 | verbose = true
       |           ^^^^
    invalid type: boolean `true`, expected a string
```

I had to look at the source code to figure out that `--verbose` is equivalent to `verbose = 'info'`. Maybe `true` should be supported as an alternative to `'info'`? Or maybe the error message should tell me what the actual string options are, rather than a generic string type arg.

thomas-zahner:
Thanks for reporting.

@katrinafyi already opened up https://github.com/lycheeverse/lychee/pull/1907 to address the problem.
With this PR passing an invalid value will lead to:

```
unknown variant `test`, expected one of `ERROR`, `WARN`, `INFO`, `DEBUG`, `TRACE`
```

Note that there is already an issue about creating/generating a schema: https://github.com/lycheeverse/lychee/issues/1382
This would enable IDE support and suggestions.

thomas-zahner:
@user27182 The issue was resolved by @katrinafyi in https://github.com/lycheeverse/lychee/pull/1907.
It is now in master & nightly and should be shipped with the next release.

PR #1907, opened 2025-11-07
 refactor: use log::Level's deserialiser, and link example TOML

katrinafyi:
use the serde functionality of the log crate. this will make the error
message print the accepted strings if an incorrect string is provided.

    unknown variant `asd`, expected one of `ERROR`, `WARN`, `INFO`, `DEBUG`, `TRACE`

however, if you use a wrong type (like bool), it will just say
"wanted string or table".

to try and direct to docs, we link to the example toml in github.
alternatively, this could link to https://lychee.cli.rs/guides/config/
instead?

the table format from log::Level is a bit mysterious and could be
a source for confusion, but idk how to prevent mentioning - can you somehow
assert that what is being deserialised is a string?

related to https://github.com/lycheeverse/lychee/issues/1903

### DIFF
--- a/lychee-lib/Cargo.toml
+++ b/lychee-lib/Cargo.toml
@@ -30,7 +30,7 @@ hyper = "1.6.0"
 ignore = "0.4.24"
 ip_network = "0.4.1"
 linkify = "0.10.0"
-log = { version = "0.4.28", features = ["serde"] }
+log = "0.4.28"
 octocrab = "0.47.0"
 openssl-sys = { version = "0.9.110", optional = true }
 path-clean = "1.0.1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2485,7 +2485,6 @@ version = "0.4.28"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34080505efa8e45a4b816c349525ebe327ceaa8559756f0356cba97ef3bf7432"
 dependencies = [
- "serde",
  "value-bag",
 ]
 
--- a/lychee-bin/Cargo.toml
+++ b/lychee-bin/Cargo.toml
@@ -34,7 +34,7 @@ humantime = "2.3.0"
 humantime-serde = "1.1.1"
 numeric-sort = "0.1.5"
 indicatif = "0.18.1"
-log = { version = "0.4.28", features = ["serde"] }
+log = "0.4.28"
 openssl-sys = { version = "0.9.110", optional = true }
 pad = "0.1.6"
 regex = "1.12.2"


thomas-zahner:
Thank you very much for this PR. Making use of the `log::Level::deserialize` is much better in many respects.

> however, if you use a wrong type (like bool), it will just say
"wanted string or table".

I've played around with deserialisation and figured that your PR is ideal. I didn't find an easy way to add more context to the error message. (anyhow cannot be used because of the return types, the only possibility would probably be to implement `Visitor` ourself) The error message hints that we no longer enforce deserialisation from a string. It's now also possible to use tables in various formats. Example:

```toml
[verbose.DEBUG]
```

So I think the error message is perfectly accurate. There is just the limitation that the valid variants are only listed as soon as the toml file is valid in terms of types. (verbose is string or table) Trying to list the variants before this step wouldn't make too much sense IMO.

Linking in the help message to GitHub is great and I also prefer this over lychee.cli.rs, as our docs are not versioned yet.

katrinafyi:
Yeah it is a bit hard to make the error messages show what you want. 

The other idea I was thinking about was that you deserialise a string, then use FromStr of log::Level. You'd have to map the errors and manually list the acceptable values - I'd wanted to avoid writing the list manually, but maybe it's not a big deal.

I'm also not sure about whether we want to broaden the accepted input types, it was really a side-effect here. I don't know if the table format is something that we'd want to accept and continue accepting in the future.

thomas-zahner:
Hmm, I must admit that I preferred the version at 0514883. It's less code, we don't have to manually spell out the variants and construct the error ourself. I think the error message was already great with that version. Also I don't see any problems with allowing TOML table format.

Or do you think the error message wasn't good enough in 0514883?

katrinafyi:
Yeah, I think that the message when you put a wrong type is not clear enough if it only says "wanted string or table". From the message alone, it's not obvious that if you put a wrong string, it will tell you which strings are valid. The "or table" raises even more questions - what is the table's format and why can verbosity be a table?

About the table format, does it have any benefits over a string? Does it let a user have more detailed customisation, e.g., enable debug but disable info? I just don't know, and unless we want to document and maintain this format, I don't think it should be added.

I guess I'm not strictly against having table format, but I definitely think the initial error message should mention all the valid strings. Once this is in place, it could (additionally) mention "or table" with a docs reference to the table format.

lychee-lib/Cargo.toml
```diff
@@ -30,7 +30,7 @@ hyper = "1.6.0"
 ignore = "0.4.24"
 ip_network = "0.4.1"
 linkify = "0.10.0"
-log = "0.4.28"
+log = { version = "0.4.28", features = ["serde"] }
```
thomas-zahner: the feature is no longer required I think?

PR #1908, opened 2025-11-10
Update Ubuntu package list before installing tools

eread:
Several jobs for nightly builds are failing because additional tooling can't be installed:

- <https://github.com/lycheeverse/lychee/actions/runs/19096465366/job/54557655997#step:3:22>
- <https://github.com/lycheeverse/lychee/actions/runs/19096465366/job/54557655999#step:3:22>
- <https://github.com/lycheeverse/lychee/actions/runs/19096465366/job/54557656055#step:3:22>
- <https://github.com/lycheeverse/lychee/actions/runs/19096465366/job/54557656047#step:4:22>
- <https://github.com/lycheeverse/lychee/actions/runs/19096465366/job/54557656019#step:4:22>

It looks like the package list that comes in the latest Ubuntu container image is out-of-date or non-existent.

At any rate, updating the package list before trying to install anything should fix the problem for these jobs.

### DIFF
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -78,19 +78,19 @@ jobs:
     steps:
       - name: Install musl tools
         if: ${{ contains(matrix.target, 'musl') }}
-        run: sudo apt-get install -y musl-tools
+        run: apt-get update && apt-get install -y musl-tools
 
       - name: Install arm tools
         if: ${{ contains(matrix.target, 'arm') }}
         run: |
           echo "GNU_PREFIX=arm-linux-gnueabihf-" >> $GITHUB_ENV
-          sudo apt-get install -y binutils-arm-linux-gnueabihf
+          apt-get update && apt-get install -y binutils-arm-linux-gnueabihf
 
       - name: Install aarch64 tools
         if: ${{ contains(matrix.target, 'aarch64') }}
         run: |
           echo "GNU_PREFIX=aarch64-linux-gnu-" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


eread:
@thomas-zahner Could you review? I noticed that the list of binaries available for a nightly build is smaller than the list of binaries available for a release build, and I believe I've contributed to fixing the problem with this PR.

.github/workflows/release-binary.yml
```diff
@@ -78,19 +78,19 @@ jobs:
     steps:
       - name: Install musl tools
         if: ${{ contains(matrix.target, 'musl') }}
-        run: sudo apt-get install -y musl-tools
+        run: sudo apt-get update && apt-get install -y musl-tools
```
thomas-zahner: The `sudo` looks a bit suspicious to me. We most probably should have it for both `apt-get` commands or none. The same holds for the other changes below.

So if we have enough privileges already

```suggestion
        run: apt-get update && apt-get install -y musl-tools
```

otherwise

```suggestion
        run: sudo apt-get update && sudo apt-get install -y musl-tools
```

thomas-zahner (changes requested):
Thank you very much for investigating the problem, appreciate it :+1:

.github/workflows/release-binary.yml
eread: @thomas-zahner That makes sense. It seemed like, based on some reading, that my original suggestion would work, but we shouldn't need `sudo` within running containers. I've updated!

thomas-zahner: Thanks :rocket:

Issue #1905, opened 2025-11-05
Lychee does not find links after <pre><code></code></pre>


konstantin-tkachuk:
I'm using lychee 0.21.0 on the following example html file:
```
<body>
    <div>
        See <a href="https://www.google.com">My working link</a>
    </div>
    <pre>
        <code></code>
    </pre>
    <div>
        See <a href="https://foobar.baz.com">My broken link</a>
    </div>
</body>
```
When you run `lychee` on this file, it will see that the first link is valid, but will not see the second link at all. If you switch the links around, it will discover the broken link, but not the valid one.

Note that this is valid html, without any errors.

You _can_ use      
>  --include-verbatim
>           Find links in verbatim sections like `pre`- and `code` blocks

to force lychee to find all links. However, to my understanding it is only supposed to be necessary for text **within** a pre block. 
It also actually starts checking links within the pre blocks, as it is supposed to, which in my case it very much undesired and forces me to add a lot of explicit excludes for "fake" links.

The expected behaviour should be that lychee only ignores the text **within** the ` <pre><code></code></pre>`, not everything **after** the first such block.

thomas-zahner:
Ouch, that's quite annoying. So thanks for reporting!
Using the following script to bisect I found that 1f97165d93749909d4fda84f3bc7808dd2777908 is responsible for introducing this bug. I will take a closer look when I find the time.

```bash
#!/usr/bin/env bash

file="/tmp/example.html"

echo "
  <body>
    <div>
        See <a href="https://www.google.com">My working link</a>
    </div>
    <pre>
        <code></code>
    </pre>
    <div>
        See <a href="https://foobar.baz.com">My broken link</a>
    </div>
</body>" > $file

count=$(cargo run $file --dump | wc -l)
rm $file


if [ "$count" == "2" ]; then
  exit 0
else
  echo "unexpected count"
  exit 1
fi
```

mre:
This should be fixed in `master` now @konstantin-tkachuk. As @thomas-zahner already mentioned, thanks for reporting. Always super helpful when people mention parsing issues. 👍

PR #1911, opened 2025-11-11
Fix extracting links after `<pre><code></code></pre>`

mre:
Fixes https://github.com/lycheeverse/lychee/issues/1905

The problem was that the EndTag event handler was checking if `last_verbatim == &self.current_element`, but `self.current_element` was never updated during EndTag events - it only got set during OpenStartTag events. This meant when `</code>` was processed, it couldn't match against the verbatim stack to pop it, leaving the verbatim flag incorrectly active for subsequent content.


thomas-zahner:
Cool, I don't see any problems :+1:

mre:
Nice. Thanks for the quick review.

Issue #1736, opened 2025-06-18
Don't show progress bar when waiting for user input


thomas-zahner:
Running `lychee -` (or `cargo run -`) shows the progress bar without ever finishing. This should not happen.

This doesn't seem to be new. It also happens on v0.10.0 from 3 years ago.

mre:
That's expected behavior. Other UNIX tools like `cat` and `grep` also block on `stdin` if no file paths are given.
`curl` does the same.

```
curl -d@- # waits forever
```

(The `-d@-` option tells curl to read the data from stdin. [Source](https://stackoverflow.com/a/59690957
))

The "-" is a convention as well.

> [Ken Thompson](https://en.wikipedia.org/wiki/Ken_Thompson) (designer and implementer of the original Unix operating system) modified [sort](https://en.wikipedia.org/wiki/Sort_(Unix)) in [Version 5 Unix](https://en.wikipedia.org/wiki/Version_5_Unix) to accept "-" as representing standard input, which spread to other utilities and became a part of the operating system as a [special file](https://en.wikipedia.org/wiki/Special_file) in [Version 8](https://en.wikipedia.org/wiki/Version_8_Unix).
> -- (Source: [Wikipedia](https://en.wikipedia.org/wiki/Standard_streams))


So unless I'm missing something, I'd keep it that way. 😅

thomas-zahner:
Oh, okay I see 😄
I totally agree with you. I realise it was the progress bar that caused my confusion.

Using `lychee -v -` interestingly enough behaves like I would expect.

Typing `lychee -` means you can now enter URLs to check and "run" it with Ctrl+D. This is congruent with `sort`, `cat` and `curl @-`. (`curl -d@-` seems to not behave this way)

So I would suggest to just remove the endlessly spinning progress bar with `lychee -` because this seems like a bug to me. lychee is not checking links while waiting for user input so no progress bar should be visible. Do you agree?

mre:
Ah, yes, let's do that.

rodic:
why these two produce different outputs?
```sh
➜  lychee git:(master) ✗ cargo run https://github.com/lycheeverse/lychee/issues/1736
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.52s
     Running `target/debug/lychee 'https://github.com/lycheeverse/lychee/issues/1736'`
[...]

PR #1914, opened 2025-11-11
refactor: progress bar

thomas-zahner:
Clone of https://github.com/lycheeverse/lychee/pull/1835 with one additional commit with the review comments resolved. Then the main work - more refactoring. The PR retains the current behaviour as much as possible. The only notable behavioural difference is that the bar no longer uses `enable_steady_tick` and the spinner animation has been removed as described in the summary comment.

Closes https://github.com/lycheeverse/lychee/issues/1736

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -169,7 +169,7 @@ async fn collect_responses(
     while let Some(response) = recv_resp.recv().await {
         let response = response?;
         let out = formatter.format_response(response.body());
-        progress.show(out)?;
+        progress.update(Some(out));
         stats.add(response);
     }
     Ok(stats)
--- a/lychee-bin/src/progress.rs
+++ b/lychee-bin/src/progress.rs
@@ -1,6 +1,5 @@
 use indicatif::{ProgressBar as Bar, ProgressStyle};
-use lychee_lib::Result;
-use std::{io::Write, sync::LazyLock};
+use std::sync::LazyLock;
 
 #[derive(Clone)]
 struct ProgressConfig {
@@ -20,9 +19,11 @@ static STYLE: LazyLock<ProgressStyle> = LazyLock::new(|| {
 });
 
 #[derive(Clone)]
-/// Report progress to the CLI.
+/// Report link check progress on stderr.
 pub(crate) struct Progress {
+    /// Optional progress bar to visualize progress
     bar: Option<Bar>,
+    /// Show detailed progress information when `true`
     detailed: bool,
 }
 
@@ -41,25 +42,20 @@ impl Progress {
         Progress { bar, detailed }
     }
 
-    pub(crate) fn show(&self, out: String) -> Result<()> {
-        // progress is reported on stderr and NOT on stdout
-        self.show_to_buffer(&mut std::io::stderr(), out)
-    }
-
-    fn show_to_buffer(&self, buffer: &mut dyn Write, out: String) -> Result<()> {
-        if self.detailed {
-            writeln!(buffer, "{}", &out)?;
+    /// If a bar is configured it is advanced by one and optionally updated with `message`.
+    /// If reporting is `detailed` `message` is printed.
+    pub(crate) fn update(&self, message: Option<String>) {
+        if self.detailed
+            && let Some(message) = message.as_ref()
+        {
+            // progress is reported on stderr and NOT on stdout
+            eprintln!("{message}");
         }
 
-        self.update(Some(out));
-        Ok(())
-    }
-
-    pub(crate) fn update(&self, message: Option<String>) {
         self.with_bar(|bar| {
             bar.inc(1);
-            if let Some(msg) = message {
-                bar.set_message(msg);
+            if let Some(message) = message {
+                bar.set_message(message);
             }
         });
     }
@@ -83,35 +79,5 @@ impl Progress {
         if let Some(bar) = &self.bar {
             action(bar);
         }
-    }
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    use log::info;
-
-    #[test]
-    fn test_skip_cached_responses_in_progress_output() {
-        let mut buf = Vec::new();
-        Progress::new("", false, false)
-            .show_to_buffer(&mut buf, "I checked a link!".into())
-            .unwrap();
-
-        info!("{:?}", String::from_utf8_lossy(&buf));
-        assert!(buf.is_empty());
-    }
-
-    #[test]
-    fn test_show_cached_responses_in_progress_debug_output() {
-        let mut buf = Vec::new();
-        Progress::new("", false, true)
-            .show_to_buffer(&mut buf, "I checked a link!".into())
-            .unwrap();
-
-        assert!(!buf.is_empty());
-        let buf = String::from_utf8_lossy(&buf);
-        assert_eq!(buf, "I checked a link!\n");
     }
 }


lychee-bin/src/commands/check.rs
```diff
@@ -16,6 +14,7 @@ use lychee_lib::{InputSource, Result};
 use lychee_lib::{ResponseBody, Status};
 
 use crate::formatters::get_response_formatter;
+use crate::formatters::progress::ProgressBar;
```
thomas-zahner: Do you think this is the right location?

mre: Good point, I would move it to `crate::progress::ProgressBar`.

lychee-bin/src/commands/check.rs
```diff
@@ -84,35 +76,35 @@ where
         &formatter_default
     });
 
-    let show_results_task = tokio::spawn(progress_bar_task(
+    let hide_progress = params.cfg.no_progress;
+    let detailed = params.cfg.verbose.log_level() >= log::Level::Info;
+
+    let progress = Progress::new("Extracting links", hide_progress, detailed);
```
mre: It's kinda strange that we pass `hide_progress` to the constructor. Maybe `hide_progress_bar`?

lychee-bin/src/commands/check.rs
```diff
@@ -198,37 +177,17 @@ where
...
+) -> Result<ResponseStats> {
     while let Some(response) = recv_resp.recv().await {
-        show_progress(
-            &mut io::stderr(),
-            pb.as_ref(),
-            &response,
-            formatter.as_ref(),
-            &verbose,
-        )?;
+        progress.show(&mut io::stderr(), &response, formatter.as_ref())?;
```
mre: Was wondering if we should provide a `show_error` which would be an alias to `rogress.show(&mut io::stderr(), ...)?;`.

lychee-bin/src/progress.rs
```diff
@@ -0,0 +1,142 @@
...
+        &self,
+        output: &mut dyn Write,
+        response: &Response,
+        formatter: &dyn ResponseFormatter,
+    ) -> Result<()> {
+        let out = if self.detailed {
+            formatter.format_detailed_response(response.body())
+        } else {
+            formatter.format_response(response.body())
+        };
```
mre: From a type-perspective, I find it a bit weird that we "know" so much about responses and formatters here.

If that was a library, we'd expect a formatted string (or some richer type) as an input.

So the signature would be more like

```rust
 pub(crate) fn show(
        &self,
        output: &mut dyn Write,
        msg: &str,
    ) -> Result<()> {
```

Or so.

Maybe we can refactor that a bit and construct the message on the outside? But perhaps you already considered that before, or maybe you want to land the original version earlier, and we can always refactor later, so we can also keep it as is. Just some food for thought. 😉

lychee-bin/src/commands/check.rs
thomas-zahner: As it's only used in testing I can make `show` write to stderr and create a private function for using a custom buffer :+1:

lychee-bin/src/progress.rs
thomas-zahner: Thanks for the idea.
I think there are two goals here. One, which I focused most on, is to move progress logic out of check.rs. Then as you say it would also be great to move any logging/formatting logic out of progress. I've now added 76e5dedea678114834b5f26df84ec60099dcf25b which achieve both.

For this specific snippet I could do as you say, but as a first step that would simply move the logic back to check.rs
I think some additional refactoring would be required to solve this nicely.

One approach that pops to mind that might solve the issue is updating `ResponseFormatter` from

```
pub(crate) trait ResponseFormatter: Send + Sync {
    fn format_response(&self, body: &ResponseBody) -> String;

    fn format_detailed_response(&self, body: &ResponseBody) -> String {
        self.format_response(body)
    }
}
```

to

```
pub(crate) trait ResponseFormatter: Send + Sync {
    fn format_response(&self, body: &ResponseBody) -> String;
}
```
and then refactor the ResponseFormatter which have detailed implementations `ColorFormatter` -> `ColorFormatter` & `ColorFormatterDetailed`. This way we don't have to decide which format method to use at the bottom level (in progress/check but we can decide at the beginning/top level.

Would this make sense?

thomas-zahner: I took a closer look and realised that a distinction of `format_response` and `format_detailed_response` is unnecessary. See 292bbc63 where I make `format_detailed_response` the default. The only difference in practice is that the progress bar message also shows detailed messages which means that it includes the part e.g. ` | Unknown status (999 <unknown status code>)` which IMO is fine. Also most of the time you don't even see the progress bar message because it changes so fast. Then I extracted it as you requested.

thomas-zahner:
# Summary

@mre I realised that not showing the progress bar as requested in https://github.com/lycheeverse/lychee/issues/1736 is not feasible. What bothered me the most, is that the progress bar was clogging up stdin when waiting for user input with `lychee -`. This way the UX is pretty bad when manually typing something or pasting content, as the progress bar overwrites any input.

So I mostly explored 2 ideas when working on this PR.

1. Change the timing when the progress bar was displayed as it seemed wrong to me initially.
2. Fully hide the progress bar if waiting for stdin.

As the link collection and checking is so highly async it means that we had to introduce some special handling for stdin. We would have to make sure that the stdin element would always be the first input that is checked. This would require a huge effort for such a minimal issue as everything is so async and it would increase complexity in the codebase. For example running `cargo run - README` you can see how it is nondeterministic whether README is checked before blocking for stdin or the other way around.

With this in mind I thought about the second approach. But this doesn't work either because of the concurrency. We would basically have to hide the progress as with `--no-progress` as soon as stdin is present as input. But this wouldn't make sense because it would fully hide the progress for all inputs, e.g. also for README with `cargo run - README`

This is why I came up with the simple solution of removing `enable_steady_tick` for the progress bar. This means the progress bar is shown the same way but there is no longer a thread in the background which clutters the terminal if we block on stdin. Simple solution for a simple problem. :sweat_smile: I've also disabled the spinner animation as it might look like lychee got stuck if the animation doesn't steadily tick.

I've spent more time on this than I should admit but I've also refactored and simplified the code around `progress` quite a bit.

mre:
> This is why I came up with the simple solution of removing `enable_steady_tick` for the progress bar. This means the progress bar is shown the same way but there is no longer a thread in the background which clutters the terminal if we block on stdin. Simple solution for a simple problem.

Sorry, I don't quite understand what you mean by "thread in the background." 😅  What is the architectural advantage of that approach? One thread less I guess? But how does it avoid cluttering the terminal?

> I've spent more time on this than I should admit but I've also refactored and simplified the code around `progress` quite a bit.

Thanks for that! It looks much cleaner.

One more rebase and we can merge this I guess. Good work!

thomas-zahner:
> Sorry, I don't quite understand what you mean by "thread in the background." 😅 What is the architectural advantage of that approach? One thread less I guess? But how does it avoid cluttering the terminal?

It's not really about an architectural advantage (though one thread less is nice as well) but about UX. Try running `cargo run -`. Conceptually, it should behave the same as running `cat` or `sort`. lychee waits for user input to be link checked until pressing CTRL+D. Compare it with the current release and this PR. Without this PR the progress bar continuously overwrites any user input because of the background thread. See the difference in this recording.

[asciicast](https://asciinema.org/a/yK2x4CxfdcKEchvCqkO6GhG3G)

PR #1919, opened 2025-11-17
Add sudo to apt-get commands

eread:
Following on from https://github.com/lycheeverse/lychee/pull/1908, it appears we need `sudo` after all.

Most of the jobs in this run are failing because of insufficient permissions: https://github.com/lycheeverse/lychee/actions/runs/19403112211.


eread:
@thomas-zahner Could you review?

thomas-zahner:
Ah yes, thanks for double-checking :+1:

PR #1924, opened 2025-11-18
Fix parsing larger HTML blocks in MDX files

### LINKED ISSUE #1923: Certain links in mdx files are skipped
If I put this in my `.mdx` file:
```mdx
<Card
  title="Invalid Link"
  href="https://an-invalid-url.com"
>
  Some text
</Card>
```
then lychee correctly reports the error.

However, if it appears in a nested structure:
```mdx
<CardGroup cols={1}>
  <Card
    title="Invalid Link"
    href="https://an-invalid-url.com"
  >
    Some text
  </Card>
</CardGroup>
```
then lychee fails to check the URL and incorrectly reports success.

I'm using lychee 0.21.0 on Mac OS 15.

mre:
Previously, each HTML line inside a Markdown file would be parsed independently. This worked reasonably well, but in the case where individual lines are not valid HTML, we would skip link checking.

The fix is to accumulate an HTML block until it gets closed and only _then_ parsing it.

Fixes #1923

### DIFF
--- a/fixtures/mdx/test.mdx
+++ b/fixtures/mdx/test.mdx
@@ -1,0 +1,8 @@
+<CardGroup cols={1}>
+  <Card
+    title="Example"
+    href="https://example.com"
+  >
+    Some text
+  </Card>
+</CardGroup>
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -3103,4 +3103,15 @@ The config file should contain every possible key for documentation purposes."
                 script.as_os_str().to_str().unwrap()
             )));
     }
+
+    #[test]
+    fn test_mdx_file() {
+        let file = fixtures_path!().join("mdx").join("test.mdx");
+        cargo_bin_cmd!()
+            .arg("--dump")
+            .arg(&file)
+            .assert()
+            .success()
+            .stdout(contains("https://example.com"));
+    }
 }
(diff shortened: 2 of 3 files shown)


mre:
@katrinafyi, @thomas-zahner FYI

katrinafyi:
Very quick! I'm wondering whether this affects parsing markdown inside HTML, which should work [according to the spec](https://spec.commonmark.org/0.18/#example-108)? I think this should already be handled by pulldown-cmark, but maybe add a test case with a markdown link inside a html block?

mre:
Your concern was valid to check, but our implementation handles it correctly. 👍 
I've added the test.

Some notes for future reference:
- pulldown_cmark treats block-level HTML tags (`<div>`, `</div>`) as separate HTML blocks
- Markdown content between HTML blocks is parsed as normal Markdown paragraphs
- We only accumulate HTML chunks within a single HTML block (between `Start(HtmlBlock)` and `End(HtmlBlock)`)
- Markdown events like `Start(Link)` are processed normally and separately

mre:
Huh, looks like the tests failed, but for an unrelated issue. We should move all tests to "example URLs" at some point and not depend on external servers. 😖

mre:
Okay, the test fixed itself. My favorite kind of fix. Merging.

thomas-zahner:
Awesome I see no problems. That was really quick :rocket:

Issue #1920, opened 2025-11-17
Clarify `Network error` reporting


ArthurFlag:
Hi and thanks for this tool. 

I'm trying to run lychee on https://docs.talon.one with:

```
lychee --verbose https://docs.talon.one
```

And I'm getting this output/error:

```
Error: Network error

Caused by:
    0: error sending request for url (https://docs.talon.one/)
    1: client error (Connect)
    2: bad protocol version
```

I'm not sure how to diagnose this error, is there a way to get more information about what went wrong?

mre:
I investigated several related issues:
- [reqwest TLS discussion](https://github.com/seanmonstar/reqwest/discussions/2428#discussioncomment-10710130)
- [Rust users forum thread](https://users.rust-lang.org/t/reqwest-tls-error-with-use-native-tls/49627)

First, I tried setting a different user agent:

```sh
lychee --user-agent "curl/8.4.0" https://docs.talon.one
```

This didn't resolve the issue.

Next, I attempted to use different TLS features available in [lychee](https://github.com/lycheeverse/lychee?tab=readme-ov-file#feature-flags):
- `native-tls` enables the platform-native TLS crate
- `vendored-openssl` compiles and statically links a copy of OpenSSL
- `rustls-tls` enables the alternative rustls TLS crate

All configurations produced the same error.

I tested the SSL certificate using [SSL Labs](https://www.ssllabs.com/ssltest/analyze.html?d=docs.talon.one&hideResults=on). The certificate appears valid overall, though there are minor chain issues: "Incorrect order, Extra certs." I'm unsure if these issues are related to the lychee failures.

The site runs TLS 1.3, which lychee supports, so there shouldn't be any compatibility problems.

I'm unable to determine why lychee is failing with this particular URL. For additional debugging information, I typically run:

```
RUST_LOG=trace lychee -vvv https://docs.talon.one
```

However, this doesn't provide any additional insights in this case.

To help troubleshoot further, could you share your operating system and lychee version?

ArthurFlag:
Thanks a lot for the quick answer and investigation 🙌 ⭐ ! 

Here's the info you requested:
- OS: MacOs 15.7
- lychee: 0.21.0

Let me know if you need anything!

mre:
Thanks.

What's interesting is that this works as expected:

```
lychee https://talon.one
```

So it has to have something to do with https://docs.talon.one, which I'm assuming runs on totally different infrastructure.

I can see some differences in the SSL setup:
[...]

PR #1927, opened 2025-11-19
Provide a more helpful error message in case of TLS protocol issues

mre:
The new error message looks like this:

```sh
> lychee https://docs.talon.one   

Error: Network error: TLS protocol version mismatch. The client and server cannot agree on a TLS version. This is often due to outdated system TLS libraries or the server TLS settings. (error sending request for url (https://docs.talon.one/))

Caused by:
    0: error sending request for url (https://docs.talon.one/)
    1: client error (Connect)
    2: bad protocol version
```

Related to https://github.com/lycheeverse/lychee/issues/1920

FYI @ArthurFlag

### DIFF
--- a/lychee-lib/src/utils/reqwest.rs
+++ b/lychee-lib/src/utils/reqwest.rs
@@ -346,7 +346,7 @@ fn analyze_generic_error_string(error_msg: &str) -> Option<String> {
     // Try to match using our rule-based system
     let result = ErrorRules::new()
         .rule(
-            &["bad protocol version", "protocol version"],
+            &["protocol version"],
             "TLS protocol version mismatch. The client and server cannot agree on a TLS version. This is often due to outdated system TLS libraries or the server TLS settings.",
         )
         .rule(


mre:
It's a longish error message, and I regret that. However, in this case it might actually be justified because the aim is to provide as much user support as we can.

lychee-lib/src/utils/reqwest.rs
```diff
@@ -345,6 +345,10 @@ fn analyze_generic_error_string(error_msg: &str) -> Option<String> {
 
     // Try to match using our rule-based system
     let result = ErrorRules::new()
+        .rule(
+            &["bad protocol version", "protocol version"],
```
thomas-zahner: This is our rule matching logic.

```rust
/// Check if any of the patterns match the given text
fn matches(&self, text: &str) -> bool {
    self.patterns.iter().any(|pattern| text.contains(pattern))
}
```

So I think the overlapping rules are redundant and this has exactly the same meaning:

```suggestion
            &["protocol version"],
```

mre: Nice catch!

Issue #1673, opened 2025-04-01
`--insecure` flag does not work with self-signed cert hosted in a Docker container


tomcodes:
Hi!

Thanks for the great project, Lychee is a delight to use.

I wanted to test some local app that runs in Docker, pretty standard PHP / nginx stack with self signed cert.

When I launch Lychee with the insecure flag I got this error:

```
lychee --insecure https://super.app.test
Error: Network error

Caused by:
    0: error sending request for url (https://super.app.test/)
    1: client error (Connect)
    2: error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:2104: (self-signed certificate)
    3: error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:2104:
```

This local app is reachable from my browser (after I allow the self-signed cert in Firefox).

Is there a way to bypass this error?

mre:
Thanks! According to the [docs](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.danger_accept_invalid_certs) we already allow accepting invalid certificates:

https://github.com/lycheeverse/lychee/blob/e185e90acbdd3b083f28de493e343de31b450832/lychee-lib/src/client.rs#L343-L349

Maybe there's another setting in `reqwest` that we miss.

MB175:
I can reproduce this behaviour, as soon as a self signed certificate is in the chain the insecure flag does not work

thomas-zahner:
Thank you for reporting and confirming. We intend to switch over from OpenSSL to Rustls in https://github.com/lycheeverse/lychee/pull/1928.

@MB175 @tomcodes Would you be able to test if you get the same behaviour when using the branch we intend to merge soon. You can find instructions [over here](https://github.com/lycheeverse/lychee/issues/1920#issuecomment-3554334576) were the switch also solves another problem, also having to do with certificates.

thomas-zahner:
I ran across this problem coincidentally with failing tests. `example.org` seems to use an invalid certificate at this moment (could maybe be fixed very soon) which led to some failing tests. Because of that I took a closer look.

It basically seems like we don't respect `--insecure` for direct URL inputs. This is very different from URLs contained in the content. But I do agree that this is a bug and `--insecure` should also allow insecure URL arguments.

# What works

```
$ echo 'https://example.org' | cargo run --all-features -
[...]

PR #1928, opened 2025-11-19
Replace OpenSSL with rustls 

### LINKED ISSUE #1721: Uses openssl by default
cargo install fails to find openssl and it falls in a heap on my arm mac. rustls is a good default for things to compile on lots of machines. More rust = less problems.

### LINKED ISSUE #1920: Clarify `Network error` reporting

### LINKED ISSUE #1970: Make TLS errors more descriptive
I got a `TLS protocol version mismatch` error for a link but according to `sslscan`, the site uses `TLSv1.3` and there are not much details in the error message about what the exact issue is.

Command to repro the error:
```
 ❯ lychee <(echo 'https://duc.zevv.nl')
1/1 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links

[/dev/fd/20]:
   [ERROR] https://duc.zevv.nl/ | Network error: TLS protocol version mismatch. The client and server cannot agree on a TLS version. This is often due to outdated system TLS libraries or the server TLS settings. (error sending request for url (https://duc.zevv.nl/)): TLS protocol version mismatch. The client and server cannot agree on a TLS version. This is often due to outdated system TLS libraries or the server TLS settings.

```
`sslscan` says the supported protocols are:
```
SSLv2     disabled
SSLv3     disabled
TLSv1.0   disabled
TLSv1.1   disabled
TLSv1.2   disabled
TLSv1.3   enabled
```
Supported ciphers:
```
Preferred TLSv1.3  128 bits  TLS_AES_128_GCM_SHA256        Curve 25519 DHE 253
Accepted  TLSv1.3  256 bits  TLS_AES_256_GCM_SHA384        Curve 25519 DHE 253
Accepted  TLSv1.3  256 bits  TLS_CHACHA20_POLY1305_SHA256  Curve 25519 DHE 253
```
Setting
```
verbose = "trace"
format = "detailed"
```
doesn't give much extra info either, except the detailed format also says `Unsupported` in the summary.

### LINKED ISSUE #2024: `GITHUB_TOKEN` environment variable causes panics with "Could not automatically determine the process-level CryptoProvider from Rustls crate features"
When the `GITHUB_TOKEN` environment variable is set, Lychee panics.

```console
$ docker run --env GITHUB_TOKEN=abc --pull always lycheeverse/lychee:latest-alpine https://example.com
latest-alpine: Pulling from lycheeverse/lychee
Digest: sha256:04252bc5d545b726d6417e46942c817045ff14cef7a5e6d9b06d5750beb2d620
Status: Image is up to date for lycheeverse/lychee:latest-alpine

thread 'main' (1) panicked at /cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rustls-0.23.36/src/crypto/mod.rs:249:14:

Could not automatically determine the process-level CryptoProvider from Rustls crate features.
Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.
See the documentation of the CryptoProvider type for more information.
```

thomas-zahner:
Closes #1721

Closes #1970

Closes #1920

Hopefully closes #2024

## Background

Using Rustls instead of openSSL should simplify many things. Depending on platform specific external openSSL libraries seems to be more brittle than using Rustls for all targets. [Very recently](https://github.com/lycheeverse/lychee/commit/892b1ddb821a7c82bda002a41e2f4682d1de2452) I've updated lychee to use reqwest 0.13.1 which coincidentally also uses Rustls by default now. This shows how Rustls should now be fully production ready and well tested. So now seems like a good moment to make the full switch.

As it is a breaking change anyhow I'd like to seize the opportunity to fully remove openSSL. If it works out well it should make maintenance easier. If it turns out that some people really need openSSL (which I think is unlikely) or run into issues we could still add a new feature flag before the next release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,10 @@ Lychee supports the following feature flags:
 - `check_example_domains` allows checking example domains such as `example.com`. This feature is useful for testing.
 
 By default, `email-check` is enabled.
+Note that in the past lychee could be configured to use either OpenSSL or Rustls.
+[It was decided](https://github.com/lycheeverse/lychee/pull/1928)
+to fully switch to Rustls and drop OpenSSL support.
+Please tell us if this this negatively affects you in any way.
 
 ## Features
 


thomas-zahner:
@mre If we see that this PR works in CI and on different machines for users, do you agree that it makes sense to fully get rid of the OpenSSL approach? Or do you think we should keep it and only change the default?

mre:
TBH, for now I would keep it and change the default. At least for one version. Then we can tell people to switch back to OpenSSL if there are any problems. We could mention that in the release-notes. On the other side, I'm flexible here. That's just what I would do, but we can also go all-in on rustls. 😆 Worst case, we release a patch version with the OpenSSL option available again. So whatever you believe is the best tradeoff between simplicity and user experience.

thomas-zahner:
This unfortunately is blocked by https://github.com/reacherhq/check-if-email-exists/issues/1625. The problem is that the latest version of check-if-email-exists on crates.io uses openSSL without an option to use ruslts.

kemingy:
I just found that the `check-if-email-exists` is dual licensed under [AGPL-3 or Reacher Commercial license](https://github.com/reacherhq/check-if-email-exists/blob/master/LICENSE.md).

But it's enabled by default for lychee CLI: https://github.com/lycheeverse/lychee/blob/caf63cc9e583844a93d636a4f9d7f63d93f65151/lychee-bin/Cargo.toml#L96

As far as I know, it's not compatible with the Apache2 + MIT license.

Maybe we need to open another issue to discuss this. What do you think? @mre

thomas-zahner:
@kemingy Thanks for pointing it out. This is known since 2022 see https://github.com/lycheeverse/lychee/issues/594. Unfortunately, we never really prioritised the issue. But you are right that this is quite problematic and now as it even blocks the transition to rustls I will try to resolve it as soon as possible.

lychee-lib/Cargo.toml
```diff
@@ -35,21 +35,12 @@ linkify = "0.10.0"
 log = "0.4.28"
 mailify-lib = { version = "0.2.0", optional = true }
 octocrab = "0.49.5"
-openssl-sys = { version = "0.9.111", optional = true }
 path-clean = "1.0.1"
 percent-encoding = "2.3.1"
 pulldown-cmark = "0.13.0"
 regex = "1.12.2"
 # Disable default features to avoid rustls being enabled by default
```
katrinafyi:
```suggestion
```

thomas-zahner:
> would it be worthwhile to mention somewhere, maybe near the readme feature flags, that we use rustls and to report any issues or shortcomings?

Sure, that makes sense.

> also, idk much about TLS, but does this make it harder to use system-installed root certificates?

I'm also no expert on that. But it [seems like self-signed certificates are currently broken](https://github.com/lycheeverse/lychee/issues/1673), so this PR shouldn't worsen the situation. I've asked on the issue if this branch resolves the problems, which is not entirely unlikely.

examples/builder/builder.rs
```diff
@@ -36,7 +36,7 @@ async fn main() -> Result<()> {
         .build()
         .client()?;
 
-    let response = client.check("https://example.com").await?;
+    let response = client.check("https://rust-lang.org").await?;
```
mre, line 39: Was this change intentional? It seems to be unrelated to rustls and as far as I remember, the goal was to show how `includes` overrides `excludes`. By changing the URL this is no longer clear.

thomas-zahner: Oh thanks for pointing out. Yes it was intentional but I forgot to update the includes and excludes. Fixed now.
I've replaced `example.com` with `rust-lang.org` because it might be confusing to people. (at least it confused me) Running `cargo run --example builder` makes the example panic. The URL is still excluded because we explicitly exclude `example.com` from checking.

We need to include the feature flag to make the example not panic: `cargo run --example builder -F check_example_domains`. With f4d1a389 the example will now pass independently of the provided feature flags.

thomas-zahner:
> Feels good. 😃

Thanks, I agree :)
Less dependencies, shorter workflow files, and most probably less runtime issues and more coherent behaviour across different machines.

> would it be worthwhile to mention somewhere, maybe near the readme feature flags, that we use rustls and to report any issues or shortcomings?

Done in e9d1e97d5426f8f46cc6031738883d4d7cd6c328

PR #1844, opened 2025-09-07
feat: implement per-host rate limiting and statistics

mre:
This PR implements a comprehensive per-host rate limiting system with adaptive backoff, statistics tracking, and configurable concurrency controls for lychee.

**Fixes #1605** - Add Per-Host Rate Limiting and Caching  
**Fixes #989** - Add custom delay between requests (prevent ban)  
**Addresses #367** - 429 Too Many Requests  
**Addresses #1593** - The cache is ineffective with the default concurrency
**Addresses #1815** - All duplicates should get removed, and no link ever gets checked more than once 

### Key Features

- **Per-host rate limiting**: Uses token bucket algorithm with governor crate for precise rate control
- **Adaptive backoff**: Automatically adjusts request timing based on server responses (429, 5xx errors)
- **Host-specific controls**: Configurable concurrency limits and request intervals per host
- **Comprehensive statistics**: Tracks requests, success rates, response times, and cache performance
- **Multiple output formats**: Compact, detailed, markdown, and JSON formats for host statistics
- **Cache integration**: Per-host caching with TTL and hit/miss tracking
- **Configuration flexibility**: Both CLI options and config file support for per-host settings

### CLI Changes

- New `--host-stats` flag to display per-host statistics at the end of runs
- New `--default-host-concurrency` option to set per-host concurrency limits
- New `--default-request-interval` option to set minimum intervals between requests
- Host-specific configuration support via config file

#### CLI Usage
```bash
# Basic usage with host statistics
lychee --host-stats README.md

# Configure global defaults
lychee --default-host-concurrency 5 --default-request-interval 200ms --host-stats docs/

# Different output formats
lychee --host-stats --format json docs/
lychee --host-stats --format markdown docs/
```

#### Config File Support (`lychee.toml`)
```toml
# Global defaults (can be overridden by CLI)
default_host_concurrency = 10
default_request_interval = "100ms"
host_stats = true

# Per-host overrides

[hosts."api.example.com"]  
concurrency = 1                # One request at a time
request_interval = "1s"     # Very conservative for sensitive APIs

[hosts."github.com"]
concurrency = 20                     # Higher concurrency for GitHub API
request_interval = "100ms"     # Optimistic interval
```

### Breaking Changes

- Replaced global cache with per-host caching architecture

### Other Changes

- Updated Client API to include host statistics methods  
- Added new dependencies: `governor`, `humantime-serde`

### Performance Impact

- **Positive**: More efficient request distribution and reduced server load
[...]

Issue #1815, opened 2025-08-23
Duplicate links still checked


tooomm:
Running the most recent version (v0.20.0), I still see a lot of duplicates:
https://github.com/Cockatrice/cockatrice.github.io/actions/runs/17156704956/job/48675191268#step:3:72

mre:
If a link occurs multiple times within one file, this shouldn't happen. Duplicates can still occur if a link appears on two different pages and the link is not in the global cache yet. So there's a race to cache the link between two concurrent tasks. Per-host request handling should fix this. See #1605. Can you try https://github.com/lycheeverse/lychee/pull/1844 and see if the issue still occurs when using that version?

mre:
This should be fixed with the next release, thanks to @thomas-zahner's work! It's already fixed in `master` in case you could give it a spin, @tooomm. This would help catch any open issues.

Issue #1298, opened 2023-11-14
Security: allow restrict custom HTTP request header to specific URL patterns


sanmai-NL:
Currently, the `--header` CLI and configuration parameter does not distinguish between target URLs. If the user where to supply a crucial secret in there, in order to make some URLs accessible during CI, then this secret would leak to all other hosts for which URLs are found.

## Solution design

Rather than implementing URL/header mapping logic in Lychee, I propose to separate this concern into an, if you so choose, external tool like a proxy. If Lychee were to have proxying support, complex logic, mappings, analysis, flows, etc. can be configured through a proxy. If common use cases are documented in a how-to style within the Lychee docs, the value for the user would not be much less than with a native implementation by Lychee (one could argue, much more indeed).

sanmai-NL:
An example of a tool that could support this functionality: https://docs.mitmproxy.org/stable/overview-features/#modify-headers

mre:
I like the `modify_headers` syntax you linked to. We could add something like this.
Just to clarify, this doesn't require a proxy, but rather a way to pass these headers to `reqwest`, our HTTP request client, right?

thomas-zahner:
I definitely agree that there should be the possibility to scope headers to specific hosts. The mentioned syntax looks really interesting. But I didn't find a proper specification for that syntax. It might also be a bit overkill to start with that syntax. For the beginning it would suffice if we can somehow associate host names with header values. We could still use that syntax but only implement the first example.

> Just to clarify, this doesn't require a proxy, but rather a way to pass these headers to reqwest, our HTTP request client, right?

@mre I also understand it that way.

thomas-zahner:
Also for the person implementing this: try to make use of [the internal request chain](https://docs.rs/lychee-lib/latest/lychee_lib/chain/type.RequestChain.html) if possible.

thomas-zahner:
[...]

PR #1929, opened 2025-11-21
feat: implement per-host rate limiting and statistics

### LINKED ISSUE #989: Add custom delay inbetween requests (prevent ban)
```
lychee v0.10.3
```

I'm seeing many `[429] ... | Failed: Network error: Too Many Requests` errors.

The issue is that the links we want to check are actually queries to an API which then return URLs again.
We are doing well over 1000 requests.

From their docs:
>**Rate Limits and Good Citizenship**
>
>We kindly ask that you insert 50 – 100 milliseconds of delay between the requests you send to the server
>
>Submitting excessive requests to the server may result in a HTTP 429 Too Many Requests status code. Continuing to overload the API after this point may result in a temporary or permanent ban of your IP address.

Is there any default delay? Could not find information on that.

The lychee docs list various mitigations to circumvent rate limits: https://lychee.cli.rs/#/troubleshooting/rate-limits?id=rate-limits
But none really help here in order to prevent getting banned.

### LINKED ISSUE #1605: Add Per-Host Rate Limiting and Caching
Currently, lychee faces challenges with rate limiting and cache effectiveness when checking links, particularly when dealing with multiple requests to the same hosts. This leads to several issues that need to be addressed:

## Current Problems
- Multiple concurrent requests to the same host trigger rate limits (429 errors) (See #989)
- Cache is ineffective with high concurrency due to race conditions (See https://github.com/lycheeverse/lychee/issues/1593#issuecomment-2552233507)
- Global concurrency settings are too coarse-grained
- Different hosts have different rate limit requirements

## Proposed Solution
We should implement a smart per-host rate limiting and caching system that would:

1. Track rate limits per host using a concurrent HashMap:
```rust
use std::collections::HashMap;
use time::OffsetDateTime;

struct HostConfig {
    rate_limit_reset: Option<OffsetDateTime>,
    request_delay: Option<Duration>,
    max_concurrent_requests: Option<u32>,
}
```

2. Implement smart caching:
- Maintain separate cache states per host

3. Stretch goal: Add configuration options per host:

```bash
lychee --max-concurrency-per-host github.com=10 --delay-per-host github.com=100ms
```

## Implementation Notes
- Use the existing [rate-limits](https://github.com/mre/rate-limits) crate, which is mostly useful for APIs
- Handle 429 responses with proper backoff using response headers when available

## Benefits
- Prevents IP bans from aggressive checking
- More efficient resource usage
- Better compliance with API rate limits
- Improved cache effectiveness. Since the cache is per host, there would be no synchronization issues
- Faster overall execution by avoiding unnecessary retries

## Examples
Configuration file example:
```toml
[hosts."github.com"]
max_concurrent_requests = 10
request_delay = "100ms"

[hosts."api.example.com"]
max_concurrent_requests = 1
request_delay = "1s"
```

CLI usage example:
```bash
lychee --max-concurrency-per-host github.com=10 --delay-per-host github.com=100ms
```

This is just a proposal. I'm not 100% certain about the naming yet.

Related issues: #989, #1593

thomas-zahner:
Supersedes #1844. Rebased master to resolve conflicts. Since there were many conflicts and I spent some time resolving them I didn't force push to the existing branch in case I messed something up. (it seems like I didn't)

Things to be done:

- [x] Don't return `Client` from `check` & update `Arc`s (https://github.com/lycheeverse/lychee/pull/1844/files#r2360356264)
- [x] CLI test for host-specific headers
- [x] Fix the behavior when setting the request interval to 0
- [x] Maybe CLI test to verify expected timing?
- [x] Do through manual testing and potentially reconsider new default settings

**Fixes #1605** - Add Per-Host Rate Limiting and Caching  
**Fixes #989** - Add custom delay between requests (prevent ban)  
**Addresses #1298** - restrict custom HTTP request headers to specific URL patterns  
**Addresses #367** - 429 Too Many Requests  
**Addresses #1593** - The cache is ineffective with the default concurrency
**Addresses #1815** - All duplicates should get removed, and no link ever gets checked more than once


thomas-zahner:
@mre I cleaned up the PR quite a bit. I think the PR is almost ready to be merged. I'm just considering some additional testing and changing some defaults, and maybe address the errors when `request_interval` is set to `0`. If you have time, could you review my additions?

# Removed things

While cleaning up I encountered a few things which I didn't understand why you added them. I've removed them to test and see if it made a difference. I think we're better of without them, unless you correct me.

I've removed `cache_max_age`, see 47d8b65d.
As explained in the commit message I don't see why it is needed.
I've understood the cache to be only host-specific and in memory.
It has nothing to do with the lychee cache exposed to users. (.lycheecache)
Why would we need cache expiration and why would we set up a expiration time of one hour,
when a link check run normally takes a few seconds/minutes?

In c9578b9a I've removed `max_concurrency` and `global_semaphore`.
As explained in the message I didn't see any use of them either.
Did I miss something?

See 727f2092 where I replace `Window` with `Vec`. Was there a good reason for using a `Window`? IMO it only distorts the statistics. Memory shouldn't be an issue, unless we had millions of requests, `Duration` is 16 bytes.

lychee-bin/tests/cli.rs
```diff
@@ -2461,7 +2459,8 @@ The config file should contain every possible key for documentation purposes."
             .arg("--verbose")
             .arg("--config")
             .arg(config)
-            .arg(server.uri())
+            .arg("-")
+            .write_stdin(server.uri())
```
thomas-zahner, line 2505: @mre Host-specific headers are currently not sent to URLs for CLI args, as the host pool is not used in that case. That's the reason for this arg change. Not sure if it would be feasible to change that behaviour. We could create an issue for this if you think this is a problem, I would prefer to not make this PR even bigger.

mre: That's fine. Can you add a TODO comment to the code? Should be enough for now.

mre:
> I've removed `cache_max_age`, see [47d8b65](https://github.com/lycheeverse/lychee/commit/47d8b65ddbae4c72cec6998e8ae5cc09285eb705). As explained in the commit message I don't see why it is needed. 

You're right. Makes no sense. Similar to limiting the window-size, I was a bit worried about memory usage for long-running processes. My thought was that we could limit the in-memory cache in case link checking takes a long time. This might become relevant once we add recursion support, but even then I'm not 100% sure. So it's unlikely that this causes a problem, and I'm thankful that it got removed.

> In [c9578b9](https://github.com/lycheeverse/lychee/commit/c9578b9a800bf10d1d6ac9ae5e50b1e0d20275d5) I've removed `max_concurrency` and `global_semaphore`. As explained in the message I didn't see any use of them either. Did I miss something?

The global semaphore was to enforce an overall concurrency limit, but I somehow missed that our mpsc channel already limits that. So that's unnecessary. Thanks for the cleanup.

> See [727f209](https://github.com/lycheeverse/lychee/commit/727f20926e741cd8cf70d0df34510c4895a76b0d) where I replace `Window` with `Vec`. Was there a good reason for using a `Window`? IMO it only distorts the statistics. Memory shouldn't be an issue, unless we had millions of requests, `Duration` is 16 bytes.

Yeah, I was a bit worried about the memory usage, actually, in the case where we handle millions of requests (which is unlikely but not impossible). But it's probably a case of premature optimization and before that part becomes a problem, other issues might arise before (such as input handling). So it's fine to use a `Vec` for consistency.

mre:
Great work!

Can we set a different default for GitHub? As per their docs, they allow [up to 100 concurrent requests](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits) and many of our users use lychee to check GitHub links specifically, which would reduce friction as it closely matches the prior behavior.

If we decide to change GitHub's default, we should mention that in the `README` somewhere and maybe in the extended help text.

lychee-lib/src/client.rs
```diff
@@ -427,6 +402,72 @@ impl ClientBuilder {
...
+        }
+
+        if let Some(timeout) = self.timeout {
+            builder = builder.timeout(timeout);
+        }
+
+        Ok(builder)
+    }
+
+    fn headers(&self) -> Result<HeaderMap> {
```
mre: Guess we should rename that to `default_headers` to avoid confusion. Alternatively, we could add a comment about what it returns.


```suggestion
    fn default_headers(&self) -> Result<HeaderMap> {
```

lychee-lib/src/types/error.rs
```diff
@@ -138,7 +138,7 @@ pub enum ErrorKind {
     #[error("Cannot send/receive message from channel")]
     Channel(#[from] tokio::sync::mpsc::error::SendError<InputContent>),
 
-    /// An URL with an invalid host was found
+    /// An URL with no host was found
```
mre:
```suggestion
    /// A URL without a host was found
```

lychee-lib/src/ratelimit/pool.rs
```diff
@@ -0,0 +1,367 @@
...
+        match self.hosts.entry(host_key) {
+            dashmap::mapref::entry::Entry::Occupied(entry) => {
+                // Another thread created it, use theirs
+                Ok(entry.get().clone())
+            }
+            dashmap::mapref::entry::Entry::Vacant(entry) => {
+                // We're first, insert ours
+                Ok(entry.insert(host).clone())
+            }
+        }
```
mre: Thinking about this a little more, the part where we create a new host and then try to insert it into the dashmap seems silly. 😆  If we see `Vacant` wouldn't that be where we could create the client?

Something like that should work I guess:

```rust
fn get_or_create_host(&self, host_key: HostKey) -> Result<Arc<Host>> {
    self.hosts
        .entry(host_key.clone())
        .or_try_insert_with(|| {
            let host_config = self
                .host_configs
                .get(&host_key)
                .cloned()
                .unwrap_or_default();

            let client = self
                .client_map
                .get(&host_key)
                .unwrap_or(&self.default_client)
                .clone();

            Host::new(
                host_key,
                &host_config,
                &self.global_config,
                client,
            )
            .map(Arc::new)
        })
        .map(|entry| entry.value().clone())
}
```

Then we don't needlessly create ton of clients. Especially in the beginning, I would assume that occurs a lot.

lychee-lib/src/ratelimit/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+//! server-provided rate limit headers.
+//!
+//! # Architecture
+//!
+//! - [`HostKey`]: Represents a hostname/domain for rate limiting
+//! - [`Host`]: Manages rate limiting, concurrency, caching, and cookies for a specific host
+//! - [`HostPool`]: Coordinates multiple hosts and routes requests appropriately
+//! - [`HostConfig`]: Configuration for per-host behavior
+//! - [`HostStats`]: Statistics tracking for each host
+//! - [`Window`]: Rolling window data structure for request timing
```
mre: `Window` doesn't exist anymore, right? Weird that it didn't catch that.

lychee-lib/src/ratelimit/headers.rs
```diff
@@ -0,0 +1,98 @@
...
+                    // if date is in the past, we can use ZERO
+                    .unwrap_or(Duration::ZERO)
+            })
+            .map_err(|_| RetryAfterParseError::ValueError(value.into()))
+    })
+}
+
+/// Parse the common "X-RateLimit" header fields.
+/// Unfortunately, this is not standardised yet, but there is an
+/// [IETF draft](https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/).
```
mre, line 39: Can we add a TODO here to replace this with https://github.com/mre/rate-limits at some point?

lychee-lib/src/ratelimit/config.rs
```diff
@@ -0,0 +1,200 @@
...
+    /// Get the effective request interval, falling back to the global default
+    #[must_use]
+    pub fn effective_request_interval(&self, global_config: &RateLimitConfig) -> RequestInterval {
+        self.request_interval
+            .unwrap_or(global_config.request_interval)
+    }
+}
+
+/// Default number of concurrent requests per host
+const DEFAULT_HOST_CONCURRENCY: usize = 10;
```
mre: Can we move that to the top of the file? It's quite an important constant.

lychee-lib/src/ratelimit/host/host.rs
```diff
@@ -0,0 +1,375 @@
...
+use super::stats::HostStats;
+use crate::types::Result;
+use crate::{CacheStatus, Status, Uri};
+use crate::{
+    ErrorKind,
+    ratelimit::{HostConfig, RateLimitConfig},
+};
+
+/// Cap retry-after to reasonable limits
+const MAXIMUM_BACKOFF: Duration = Duration::from_secs(10 * 60);
```
mre: 10 minutes is a long time. Should we cap it to 60 seconds instead? 🤔 Not sure.

lychee-bin/src/options.rs
```diff
@@ -528,6 +537,33 @@ with a status code of 429, 500 and 501."
...
+    ///   --host-concurrency 2   # Conservative for slow APIs
+    ///   --host-concurrency 20  # Aggressive for fast APIs
+    #[arg(long = "host-concurrency", verbatim_doc_comment)]
+    #[serde(default)]
+    pub(crate) host_concurrency: Option<usize>,
+
+    /// Minimum interval between requests to the same host (default: 100ms)
+    ///
+    /// Sets a baseline delay between consecutive requests to prevent
+    /// hammering servers. The adaptive algorithm may increase this based
```
mre: My choice of wording here wasn't perfect. Maybe we should replace the term "hammering" with "overloading"?


```suggestion
    /// overloading servers. The adaptive algorithm may increase this based
```

lychee-bin/src/formatters/stats/mod.rs
```diff
@@ -13,17 +13,38 @@ pub(crate) use raw::Raw;
...
 
 pub(crate) trait StatsFormatter {
     /// Format the stats of all responses and write them to stdout
     fn format(&self, stats: ResponseStats) -> Result<Option<String>>;
 }
 
+/// If configured to do so, output response statistics to stdout or the specified output file.
+pub(crate) fn output_response_statistics(stats: ResponseStats, config: &Config) -> Result<()> {
+    let is_empty = stats.is_empty();
+    let formatted_stats = get_stats_formatter(&config.format, &config.mode).format(stats)?;
```
mre: It's fine, but perhaps we should break that up into two lines:

```suggestion
    let formatter = get_stats_formatter(&config.format, &config.mode);
    let stats = formatter.format(stats)?;
```

I find this slightly easier to read.

lychee-bin/src/commands/check.rs
```diff
@@ -260,18 +267,27 @@ async fn handle(
...
+
     let response = check_url(client, request).await;
 
-    // - Never cache filesystem access as it is fast already so caching has no
-    //   benefit.
-    // - Skip caching unsupported URLs as they might be supported in a
-    //   future run.
-    // - Skip caching excluded links; they might not be excluded in the next run.
-    // - Skip caching links for which the status code has been explicitly excluded from the cache.
+    // Apply the same caching rules as before
```
mre: Sorry about removing these comments. Let's bring them back.

```suggestion
    // - Never cache filesystem access as it is fast already so caching has no    //   benefit.
    // - Skip caching unsupported URLs as they might be supported in a
    //   future run.
    // - Skip caching excluded links; they might not be excluded in the next run.
    // - Skip caching links for which the status code has been explicitly excluded from the cache.
```

mre:
One more note: have we checked that caching still works? E.g. do the links end up in the cache file at the end?

lychee-lib/src/ratelimit/headers.rs
thomas-zahner: Ah that's funny :joy: I seem to have copied some parts of your crate without knowing..

Is there a reason for not using your crate?

lychee-lib/src/ratelimit/host/host.rs
thomas-zahner: Agree. Also thought about that but you had 1 hour previously so I thought maybe there's a good reason for not going to low. But I think it's a better default. We could still make it configurable in the future.

lychee-lib/src/ratelimit/mod.rs
thomas-zahner: Yeah that's weird..

lychee-lib/src/ratelimit/pool.rs
thomas-zahner: Ah yes, that's much better :+1:

lychee-lib/src/ratelimit/headers.rs
mre: It's mostly to avoid the extra dependency right now. We could move the crate into our workspace at some point to maintain it inline. People could still use it independently of lychee. But that said, I'm not too happy with the design anymore, so a rewrite would be better anyway.

thomas-zahner: Okay, I've added a note to the module in 151b6d4347ea59b2b04cfb41d7edcd477e293fd3.

lychee-bin/tests/cli.rs
thomas-zahner: Sure: bac8533d

thomas-zahner:
> One more note: have we checked that caching still works? E.g. do the links end up in the cache file at the end?

Did a quick manual test and seems to work as expected. Also this should be covered by integration test in cli.rs.

thomas-zahner:
I just did some small adjustments which I think are improvements. Do you agree with the changes? If so, I would now like to merge this PR. As part of fc3023cd I've also adjusted the default request interval value because I think it was a bit too pessimistic, especially considering that we previously had none and we now also respect rate limit headers.

> Can we set a different default for GitHub? As per their docs, they allow [up to 100 concurrent requests](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits) and many of our users use lychee to check GitHub links specifically, which would reduce friction as it closely matches the prior behavior.

I would like to do this separately, as it might still be some effort. For this I've created #1960.

lychee-lib/src/ratelimit/host/interval.rs
```diff
@@ -1,68 +0,0 @@
-use governor::Quota;
```
thomas-zahner: Huh, forgot to delete this now unused file in my original commit, only just noticed it now before merging. @mre Surprises me that clippy isn't bothered by this.

thomas-zahner:
> Thanks for the high-quality additions and for following through. This is a massive step forward and I'm happy to finally see it land. A lot of our current issues will go away once merged. 😊 ⭐

Thank you for the core work! Totally agree that is quite a milestone for lychee. I did some tests where I compared lychee v0.21.0 with this branch and the link process was more than twice as fast. (so intentional slower requests leads to an overall speedup) Will do some follow up testing and this might be worth a report and blog post.

PR #1936, opened 2025-11-25
fix: make file matcher respect the hidden option

### LINKED ISSUE #1935: Hidden files aren't automatically ignored
The Lychee [documentation](https://lychee.cli.rs/recipes/excluding-paths/) states:

```
Additionally, lychee ignores hidden files by default. To traverse hidden directories and check hidden files use --hidden.
```

However, this is not the behaviour I notice when using the GitHub Action; our hidden files (files that begin with a `.`) are still checked for dead links. I had to manually exclude these paths with `--exclude-path "(^|\/)\."`.

thomas-zahner:
Closes #1935

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -685,23 +685,19 @@ mod cli {
             .arg("--hidden")
             .assert()
             .success()
-            .stdout(contains("1 Total"));
+            .stdout(contains("2 Total"));
 
-        cargo_bin_cmd!()
+        let result = cargo_bin_cmd!()
             .arg("--dump")
             .arg("--hidden")
             .arg(fixtures_path!().join("hidden/"))
             .assert()
-            .stdout(contains("wikipedia.org"))
             .success();
 
-        cargo_bin_cmd!()
-            .arg("--dump-inputs")
-            .arg("--hidden")
-            .arg(fixtures_path!().join("hidden/"))
-            .assert()
-            .stdout(contains(".hidden"))
-            .success();
+        assert_lines_eq(
+            result,
+            vec!["https://rust-lang.org/", "https://rust-lang.org/"],
+        );
     }
 
     #[test]


lychee-lib/src/types/file.rs
```diff
@@ -46,21 +46,19 @@ impl FileExtensions {
...
     /// # Errors
     ///
     /// Fails if an extension is `all` or otherwise contains any character that
     /// is not a Unicode letter or number.
-    fn try_from(value: FileExtensions) -> super::Result<Self> {
+    pub fn build(self, skip_hidden: bool) -> super::Result<Types> {
         let mut types_builder = TypesBuilder::new();
-        for ext in value.0.clone() {
-            types_builder.add(&ext, &format!("*.{ext}"))?;
+        let prefix = if skip_hidden { "[!.]" } else { "" };
```
mre, line 58: Question: Wouldn't `[!.]*.{ext}` only match the filename component, not the full path? I.e. do we handle this case?

`hidden/.hidden/file.md` (file in hidden directory)

When `skip_hidden` is `false`, the Types filter with `[!.]` would still exclude hidden files, right? Doesn't that contradict with the intent? 🤔 I clearly don't have a good overview yet, sorry.

mre: Never mind, I think I get it now. Your fix looks good.

thomas-zahner: Well I do think it's a great question. It's because file type matchers are ran against the file names and not the full path. Which makes sense intuitively, but the docs don't really mention that.

The relevant internals:

```rust
        // File types don't apply to directories, and we can't do anything
        // if our glob set is empty.
        if is_dir || self.set.is_empty() {
            return Match::None;
        }
        // We only want to match against the file name, so extract it.
        // If one doesn't exist, then we can't match it.
        let name = match file_name(path.as_ref()) {
            Some(name) => name,
            None if self.has_selected => {
                return Match::Ignore(Glob::unmatched());
            }
            None => {
                return Match::None;
            }
        };
```

PR #1938, opened 2025-11-27
Hide progress bar if input is `stdin` only

mre:
In https://github.com/lycheeverse/lychee/pull/1914, work was done to hide the progress bar on `stdin`. The simple solution of removing enable_steady_tick for the progress bar. This worked well for avoiding cluttering the output while the user typed their input.

However, the initial progress bar still gets shown, which is not ideal. This commit slightly imporves the behavior by waiting for all inputs before showing the progress bar. We do this by iterating over the inputs and checking if any of them is `stdin`.

This way, the progress bar does no longer get shown. To start the processing, the user has to press Ctrl^D, which is customs for other Unix tools as well. Other behavior remains unchanged.

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -59,7 +59,7 @@
 #![deny(missing_docs)]
 
 use std::fs::{self, File};
-use std::io::{self, BufRead, BufReader, ErrorKind, IsTerminal};
+use std::io::{self, BufRead, BufReader, ErrorKind, IsTerminal, stdin};
 use std::path::PathBuf;
 
 use anyhow::{Context, Error, Result};
@@ -311,11 +311,11 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
     //
     // When stdin is piped (`cat links.txt | lychee -`), `is_terminal()` returns
     // false, so the progress bar is shown normally.
-    let stdin_input = inputs.len() == 1
+    let is_stdin_input = inputs.len() == 1
         && inputs
             .iter()
             .any(|input| matches!(input.source, lychee_lib::InputSource::Stdin))
-        && std::io::stdin().is_terminal();
+        && stdin().is_terminal();
 
     // TODO: Remove this section after `--base` got removed with 1.0
     let base = match (opts.config.base.clone(), opts.config.base_url.clone()) {
@@ -384,7 +384,7 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
         cache,
         requests,
         cfg: opts.config.clone(),
-        stdin_input,
+        is_stdin_input,
     };
 
     let exit_code = if opts.config.dump {
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -59,7 +59,7 @@ where
         accept,
     ));
 
-    let hide_bar = params.cfg.no_progress || params.stdin_input;
+    let hide_bar = params.cfg.no_progress || params.is_stdin_input;
     let level = params.cfg.verbose().log_level();
 
     let progress = Progress::new("Extracting links", hide_bar, level, &params.cfg.mode());
--- a/lychee-bin/src/commands/mod.rs
+++ b/lychee-bin/src/commands/mod.rs
@@ -22,7 +22,7 @@ pub(crate) struct CommandParams<S: futures::Stream<Item = Result<Request, Reques
     pub(crate) cache: Cache,
     pub(crate) requests: S,
     pub(crate) cfg: Config,
-    pub(crate) stdin_input: bool,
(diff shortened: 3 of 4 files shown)


thomas-zahner:
Personally, I'm okay with the current behaviour. As mentioned in https://github.com/lycheeverse/lychee/pull/1914#issuecomment-3532526933 and https://github.com/lycheeverse/lychee/issues/1736#issuecomment-3206966681 I find it unexpected to fully hide the bar if stdin is one of the inputs.

The main reason is that this might confuse users a lot. I assume the more common use case is to pipe something to lychee, rather than plain `lychee -`. With this PR if we now do `cat README.md | lychee -` the progress bar disappears without any apparent reason and there is no way for users to enable it again.

mre:
> `cat README.md | cargo run -- -`

Oh, I hadn't thought of that. Nice catch!

Yeah, it's a bit of a tradeoff. In [your recording](https://github.com/lycheeverse/lychee/pull/1914#issuecomment-3562135681) (now archived), I got a bit irritated by the progress bar. It looked like a rendering bug, to be honest. But as you pointed out, there's no easy solution here. 😕 

Or so I thought!

Maybe we should only hide the progress bar when stdin is not a TTY (i.e., it's being piped or redirected). If someone runs `lychee -` interactively and types into the terminal, stdin is still a TTY.

I tweaked the behavior a bit. The implementation now detects when stdin is piped or redirected and hides the progress bar only in those cases. What do you think? 🤩

thomas-zahner:
Ooh I didn't know this was possible :+1: 

It still feels a bit like a hack but I'm okay with it, if we are even more strict with the predicate and set `stdin_input` to `true` if stdin is the only input. (`cargo run -`) Because running `cargo run - README.md` also checks README.md. The order is non-deterministic so sometimes the file is checked while lychee also waits for stdin and sometimes it is only checked after the user finished writing the contents to stdin. If we do this then I'm okay with the PR.

mre:
Quick heads-up: When I tested this, I think I found a few issues. That means, we should test that thoroughly before merging. Don't have time to re-run the tests right now, but if someone wants to check out this branch and try it on a few inputs, that would be helpful.

mre:
I took another look and decided to avoid hiding the progress bar if there are more inputs than just `-`. So in the case of 

```sh
lychee - README.md
``` 

It would show the progress bar.

@thomas-zahner, wdyt?

lychee-bin/src/main.rs
```diff
@@ -301,15 +301,32 @@ fn underlying_io_error_kind(error: &Error) -> Option<io::ErrorKind> {
...
+    // interactive (TTY).
+    //
+    // We restrict this to the sole-input case because with mixed inputs like
+    // `lychee - README.md` the file is processed concurrently with stdin, so
+    // the order of completion is non-deterministic. Hiding the bar there would
+    // be confusing rather than helpful.
+    //
+    // When stdin is piped (`cat links.txt | lychee -`), `is_terminal()` returns
+    // false, so the progress bar is shown normally.
+    let stdin_input = inputs.len() == 1
```
thomas-zahner: I'd personally find `is_stdin_input` a bit clearer.

lychee-bin/src/main.rs
```diff
@@ -301,15 +301,32 @@ fn underlying_io_error_kind(error: &Error) -> Option<io::ErrorKind> {
...
+        && inputs
+            .iter()
+            .any(|input| matches!(input.source, lychee_lib::InputSource::Stdin))
+        && std::io::stdin().is_terminal();
```
thomas-zahner: Oh nice! Nitpick; maybe we can import `std::io::stdin` instead of fully qualifying it?

It bothered me in the past that ` [DEBUG] Reading content from stdin` was shown even if the content was piped. I think checking for `stdin().is_terminal()` and only printing the message then, would be much nicer! Could you maybe also quickly do this?

lychee-bin/src/main.rs
mre: true; changed

lychee-bin/src/main.rs
mre: >  I think checking for stdin().is_terminal() and only printing the message then, would be much nicer! Could you maybe also quickly do this?

yes, done

>  maybe we can import std::io::stdin instead of fully qualifying it?

mh, not that easy because we also import `tokio::io::stdin` so I didn't want to use an alias for that to disambiguate the two. Tried:

```rust
use std::io::{IsTerminal, stdin as std_stdin};
use tokio::io::{AsyncReadExt, stdin as tokio_stdin};
```

but it became messy.

thomas-zahner: Thank you! Sure that was a nitpick anyways. Ah, I see we use two different stdin's :sweat_smile:

PR #1956, opened 2025-12-15
Extract URLs from CSS

### LINKED ISSUE #1485: Unable to check url inside <style> tag and css file
There are two broken links on this page https://testing826.wpenginepowered.com/, if you inspect it, both image's url miss the "g" at the end of url
```
lychee -vv https://testing826.wpenginepowered.com/
✔ [200] https://testing826.wpenginepowered.com/assets/css/styles.css

🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors
```
it did not catch those two broken urls.

mre:
In #1485 it was requested to extract URLs from CSS files and inline styles.

While lychee proper is generally becoming more conservative about the file types we support, the places where links can occur in CSS are pleasantly limited to `url()` directives (with the exception of inline URLs in `font-face` and `@import` rules, which we conveniently ignore for now).

As such, integrating CSS support into lychee can be done by means of a simple extractor utilizing a bunch of regular expressions. It's not quite as exhaustive as a proper CSS parser (such as Servo's rust-cssparser crate), but on the plus side, we don't need to add additional dependencies.

Fixes #1485

### DIFF
--- a/lychee-lib/src/extract/css.rs
+++ b/lychee-lib/src/extract/css.rs
@@ -45,7 +45,7 @@ use crate::types::uri::raw::{RawUri, SourceSpanProvider, SpanProvider};
 /// - `src: url(../fonts/font.woff2);`
 static CSS_URL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(
-        r#"(?x)
+        r#"(?x)                     # Enable extended mode for whitespace and comments
         url\s*\(                    # Match 'url(' with optional whitespace
         \s*                         # Optional whitespace
         (?:                         # Non-capturing group for the URL


mre:
I briefly considered adding SCSS/SASS file support as well, but decided against it. Mainly for the following reasons:
  - SCSS/SASS are source files, not final CSS, so the compiled output might differ
  - Variables make URLs dynamic: url($cdn-path + '/image.png'), which we can't validate

Besides, users can opt-in with `--include "*.scss"` if needed.

lychee-lib/src/extract/css.rs
```diff
@@ -0,0 +1,439 @@
...
+/// - Escaped quotes within the URL
+///
+/// Examples:
+/// - `background-image: url("./image.png");`
+/// - `background: url('/path/to/image.jpg');`
+/// - `@import url(https://example.com/style.css);`
+/// - `src: url(../fonts/font.woff2);`
+static CSS_URL_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(
+        r#"(?x)
```
thomas-zahner: What's `(?x)`? Probably also worth a comment.

lychee-lib/src/extract/css.rs
```diff
@@ -0,0 +1,439 @@
...
+/// * `input` - The CSS content to extract URLs from
+/// * `span_provider` - Provides source location information for extracted URLs
+///
+/// # Returns
+///
+/// A vector of `RawUri` objects representing the extracted URLs
+///
+/// # Examples
+///
+/// CSS input:
```
thomas-zahner, line 80: Would be great to make this a doctest where we invoke the function and have an assert_eq.

thomas-zahner: Never mind, tried to do it and realised that it doesn't work because it isn't public.

lychee-lib/src/extract/css.rs
```diff
@@ -0,0 +1,439 @@
...
+/// ```css
+/// .example {
+///     background-image: url("./image.png");
+///     background: url('/absolute/path.jpg');
+/// }
+/// @import url(https://example.com/style.css);
+/// ```
+///
+/// Extracts 3 URLs: `./image.png`, `/absolute/path.jpg`, and `https://example.com/style.css`
+pub(crate) fn extract_css<S: SpanProvider>(input: &str, span_provider: &S) -> Vec<RawUri> {
```
thomas-zahner, line 90: I find the name a bit weird. I would prefer `extract` with a parameter `css: &str`. Otherwise `extract_from_css`? But that should be clear from the context already. (`css::extract`)

lychee-lib/src/extract/css.rs
mre: Yeah, it's "verbose mode." See [here](https://docs.rs/regex/latest/regex/#example-verbose-mode). It allows me to use those fancy `#` comments. Definitely worth a comment.

mre: done

lychee-lib/src/extract/css.rs
mre: exactly

lychee-lib/src/extract/css.rs
mre: Fair. My intention was to stick to the existing naming scheme. In the other extractors we have functions like `extract_html` and `extract_markdown`. The reason is that it makes it easier to reason about the function call in `match` expressions.

mre:
> I've added https://github.com/lycheeverse/lychee/commit/5409ff9dd972764c45654561d07a617385043705. Do you agree with that? It's less code and yields better coverage

Absolutely! I forgot we had that. Much better, thanks.

mre:
AFAICT, the only open question is what we do with https://github.com/lycheeverse/lychee/pull/1956#discussion_r2623627064.

lychee-lib/src/extract/css.rs
thomas-zahner: Ah I see, didn't check that. Let's leave it then :+1:

thomas-zahner:
I see no problems, so we're good to merge

mre:
Thanks for the swift review.

PR #1957, opened 2025-12-15
Unify Releases Across Operating Systems

### LINKED ISSUE #1875: Streamline release for Windows & macOS

mre:
This merges the Windows/macOS/Linux build into one job in the release workflow. It follows the same principle as the release build for ripgrep.

Fixes #1875

### DIFF
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -6,6 +6,8 @@
 #   - https://github.com/lycheeverse/lychee-action/blob/master/action.yml
 
 name: Release Binary
+
+# Only do the release on x.y.z tags or manual workflow dispatch
 on:
   release:
     types:
@@ -14,11 +16,17 @@ on:
       - prereleased
   workflow_dispatch:
 
+# We need this to be able to upload releases
+permissions:
+  contents: write
+
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
-  prepare:
+  # The create-release job runs purely to initialize the GitHub release itself
+  create-release:
+    name: create-release
     runs-on: ubuntu-latest
     outputs:
       tag_name: ${{ steps.get_release.outputs.tag_name }}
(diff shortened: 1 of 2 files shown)


mre:
On the plus-side, it's shorter. On the "minus-side" I couldn't test it. 😅

thomas-zahner:
Thank you for addressing the issue. I would really like to test it before blindly merging it. :sweat_smile: And it also should be possible to do so by manually triggering the workflow on a non-tagged version. Maybe we would have to temporarily update the `on:` value to do so. But we should be able to run it. It's just `Upload release archive` which should fail then but everything before that should, i.e. building, should work anywhere. Maybe some small other adjustments would be necessary.

If you don't find the time to do so, you can let me know and I'll try do it. It's just that I don't want to debug the workflow during the next release but before merging this PR.

mre:
After rebasing onto master, I noticed that the goal of this PR has already been achieved: the current master already has a unified `build-release` matrix job covering Linux, macOS, and Windows in a single job. I believe we should be closing this as superseded. Thoughts?

mre:
Will close for now. I don't think there's anything that needs to be done here. As mentioned, we already have a matrix job.

mre:
Erm, sorry, diffed against the wrong branch. This wasn't fixed in `master` yet. Reopening.

.github/workflows/release-binary.yml
```diff
@@ -32,105 +40,188 @@ jobs:
         uses: bruceadams/get-release@v1.3.2
         # Skip step if not in the official repository as it would fail.
         # Instead fail at the end of this workflow.
-        if: ${{ github.repository == 'lycheeverse/lychee' }}
+        if: ${{ github.repository == 'lycheeverse/lychee' && github.event_name == 'release' }}
```
mre: I can remove this after the build is verified.

.github/workflows/release-binary.yml
```diff
@@ -14,15 +16,21 @@ on:
...
-  prepare:
+  # The create-release job runs purely to initialize the GitHub release itself
+  create-release:
+    name: create-release
     runs-on: ubuntu-latest
     outputs:
-      tag_name: ${{ steps.get_release.outputs.tag_name }}
-      upload_url: ${{ steps.get_release.outputs.upload_url }}
+      tag_name: ${{ steps.get_release.outputs.tag_name || 'test-tag' }}
+      upload_url: ${{ steps.get_release.outputs.upload_url || '' }}
```
mre: Can remove this. Was needed for the test runs.

mre:
OK, this should be good to go. Thanks, Thomas for the suggestion to run a "test release."

This way, I uncovered an issue with our new file-limits support (https://github.com/lycheeverse/lychee/pull/2106). 

On Windows, I got

```rust
error[E0432]: unresolved imports `rlimit::Resource`, `rlimit::getrlimit`
   --> lychee-bin\src\main.rs:143:18
    |
143 |     use rlimit::{Resource, getrlimit, increase_nofile_limit};
    |                  ^^^^^^^^  ^^^^^^^^^ no `getrlimit` in the root
    |                  |
    |                  no `Resource` in the root
For more information about this error, try `rustc --explain E0432`.
error: could not compile `lychee` (bin "lychee") due to 1 previous error
Caused by:
  process didn't exit successfully:
```

Inside the `rlimit` crate, `increase_nofile_limit(u64::MAX)` evaluates internally via `#[cfg(unix)]` to attempt bumping to the max hard limit. If it fails, it still gracefully returns `Ok(current_soft_limit)`. And on Windows, via `#[cfg(windows)]`, it simply returns `Ok(lim)`. Source code [here](https://github.com/Nugine/rlimit/blob/94c9e171e9d11690756cb15dc7dac07ad9067478/src/tools.rs#L44).

This means `rlimit` is already entirely cross-platform and already provides the *exact fallback semantics* we want internally! The entire `limit_result` match statement with `cfg(unix)` and `cfg(not(unix))` was just redundant boilerplate. :)

This run shows that all binaries build successfully: https://github.com/lycheeverse/lychee/actions/runs/23962659010/job/69895885187

actions/upload-release-asset@v1 is deprecated and relies on older, outdated versions of Node.js. The `gh` CLI tool is natively pre-installed on all GitHub-hosted runners (Ubuntu, macOS, and Windows). This makes it universally available without needing to download external dependencies or rely on third-party actions.

lychee-bin/src/main.rs
```diff
@@ -140,15 +140,14 @@ fn read_lines(file: &File) -> Result<Vec<String>> {
...
 fn handle_fd_limits(opts: &mut LycheeOptions) {
-    use rlimit::{Resource, getrlimit, increase_nofile_limit};
+    use rlimit::increase_nofile_limit;
+
     /// Baseline overhead estimate to account for file descriptors
     /// which are opened before lychee actually does any link checking.
     /// Estimate is based on observing `lsof -p $(pgrep lychee)`.
     const BASELINE_OVERHEAD: u64 = 20;
 
-    if let Ok(soft_limit) = increase_nofile_limit(u64::MAX)
```
thomas-zahner, line 149: Oh right, thanks for simplifying this. My wrong assumption was that `getrlimit` is available on Windows. If this would have been the case it would still have lowered lychee's concurrency on Windows if the limit was low. But realising that this is Unix specific it makes sense to simplify it, and do "nothing" on Windows.

.github/workflows/release-binary.yml
```diff
@@ -32,105 +40,188 @@ jobs:
...
+          ${{ env.CARGO }} build --verbose --release ${{ env.TARGET_FLAGS }}
+          if [[ "${{ matrix.os }}" == windows-* ]]; then
+            bin="target/${{ matrix.target }}/release/lychee.exe"
+          else
+            bin="target/${{ matrix.target }}/release/lychee"
+          fi
           echo "BIN=$bin" >> $GITHUB_ENV
 
-      - name: Strip binary (cross)
+      - name: Strip release binary (macOS)
```
thomas-zahner, line 186: Why is macOS separate from the rest? Shouldn't we either use the "cross" approach below for all or alternatively the native approach for all as with macOS? (in which case we would remove the docker invocation below)

.github/workflows/release-binary.yml
```diff
@@ -32,105 +40,188 @@ jobs:
...
+      - name: Creating directory for archive
+        shell: bash
         run: |
-          mkdir out out/docs out/complete
-          cp $BIN out
-          cp README.md docs/*.md out/docs
+          mkdir -p "$ARCHIVE/docs" "$ARCHIVE/complete"
+          cp "$BIN" "$ARCHIVE"/
+          cp README.md "$ARCHIVE"/
+          cp docs/*.md "$ARCHIVE/docs/" || true
```
thomas-zahner: Why would this fail?

```suggestion
          cp docs/*.md "$ARCHIVE/docs/"
```

.github/workflows/release-binary.yml
```diff
@@ -140,146 +231,99 @@ jobs:
...
-          cd out
-          tar -czf lychee.tar.gz lychee docs/ complete/
+          7z a "$ARCHIVE.zip" "$ARCHIVE"
+          certutil -hashfile "$ARCHIVE.zip" SHA256 > "$ARCHIVE.zip.sha256"
+          echo "ASSET=$ARCHIVE.zip" >> $GITHUB_ENV
+          echo "ASSET_SUM=$ARCHIVE.zip.sha256" >> $GITHUB_ENV
+
+      - name: Build archive (Unix)
+        shell: bash
+        if: ${{ !startsWith(matrix.os, 'windows') }}
```
thomas-zahner, line 287: The `matrix.os` part confuses me a bit. What does it mean? Maybe worth a comment?

.github/workflows/release-binary.yml
```diff
@@ -140,146 +231,99 @@ jobs:
...
+        run: |
+          tar czf "$ARCHIVE.tar.gz" "$ARCHIVE"
+          shasum -a 256 "$ARCHIVE.tar.gz" > "$ARCHIVE.tar.gz.sha256"
+          echo "ASSET=$ARCHIVE.tar.gz" >> $GITHUB_ENV
+          echo "ASSET_SUM=$ARCHIVE.tar.gz.sha256" >> $GITHUB_ENV
+
+      - name: Build macOS DMG (aarch64 only)
```
thomas-zahner, line 294: What's with non-aarch64 macOS? Does it fall in the Unix category? Then it would be related to my other comment.

.github/workflows/release-binary.yml
```diff
@@ -140,146 +231,99 @@ jobs:
...
+          version="${{ needs.create-release.outputs.tag_name }}"
+          gh release upload "$version" ${{ env.ASSET }} ${{ env.ASSET_SUM }}
 
-          mkdir dmg
-          mv lychee dmg/
-          hdiutil create -fs HFS+ -srcfolder dmg -volname lychee lychee.dmg
-
-      - name: Upload .dmg
-        uses: actions/upload-release-asset@v1
+      - name: Upload macOS DMG
```
thomas-zahner, line 321: I don't think this section is necessary. We could just use the same variable names in both cases then we don't need to differentiate in this last step here.

thomas-zahner (approved):
Thanks for the PR, looks good! I've left a few comments & questions. Just don't forget to remove the temporary changes you intend to revert before merging.

lychee-bin/src/main.rs
mre: Haha, glad we'll soon have the Windows job to automate those checks. 👍

.github/workflows/release-binary.yml
mre: I can't remember. Think I ran into some problems on macOS, but I'll try again. Might just be that I forgot to merge that part. Thanks for pointing that out.

.github/workflows/release-binary.yml
mre: Yeah, I added this because `docs` might not contain md files and would return an error in that case. Not sure if we want to make that a hard build error instead, but come to think about it, I think we should.

.github/workflows/release-binary.yml
mre: Sure. Basically I want to say "Only run this step on non-Windows builds. I use `startsWith` because GitHub supports different Windows operating systems. Any suggestions for the comment?

.github/workflows/release-binary.yml
mre: Good point. Will check.

.github/workflows/release-binary.yml
mre: You're right.

.github/workflows/release-binary.yml
thomas-zahner: Oh wait, never mind. I was only confused by the order of the arguments. I would have expected something like `matrix.os.startsWith('windows')` which probably makes no sense. Maybe I've been doing too much Rust recently :sweat_smile: 

It should actually be quite obvious, so no need to comment.

.github/workflows/release-binary.yml
mre: Removed in 229deb1d (the build is verified).

.github/workflows/release-binary.yml
mre: Removed in 229deb1d.

.github/workflows/release-binary.yml
mre: The separation is intentional: macOS uses the native `strip` on the runner, while Linux cross-compiled targets (arm, aarch64, musl, i686) have to go through the cross Docker image because the host `strip` cannot handle foreign architectures/libc combos reliably.

I also noticed while working on this that the native `x86_64-unknown-linux-gnu` build was not being stripped at all (it skips cross, so the "cross" strip step did not match). Added a dedicated native Linux strip step in 229deb1d.

.github/workflows/release-binary.yml
mre: Intentional. x86_64 macOS is on its way out and a tar.gz is enough for that target; only Apple Silicon gets a DMG. Added a comment in 229deb1d documenting this and inviting folks to open an issue if they really need an x86_64 DMG.

.github/workflows/release-binary.yml
mre: Consolidated in 229deb1d: the DMG step now appends to `ASSET`/`ASSET_SUM` and the dedicated `Upload macOS DMG` step is gone. Single upload step handles both.

mre:
Failing test is the flaky WayBackMachine test again...

thomas-zahner:
> Failing test is the flaky WayBackMachine test again...

@katrinafyi Addressed this in 434ae96560fd331363a016e9e55ec4df17778822 :)

Thanks again for the PR, I'll test this with the new release

thomas-zahner:
Seems to have [worked](https://github.com/lycheeverse/lychee/releases/)! :tada:
Now it's just the [docker step failing](https://github.com/lycheeverse/lychee/actions/runs/24880946622/job/72858998976). It's no news that this is unstable but now it fails even after multiple retries.
I'll investigate this as part of https://github.com/lycheeverse/lychee/issues/2037 and hope to come up with a stable long-term solution. I feel like docker releases were unreliable since a very long time.

PR #1972, opened 2025-12-28
Adds shell completions for lychee

mre:
This follows ripgrep's lead and uses the same naming conventions for the CLI args. `clap_complete` does all the heavy-lifting, so there's no need to write our own generator. Completions get added to the binary releases.

@thomas-zahner, I tried to integrate the feature similar to your man page support (thanks for that!). 

~⚠️ TODO: Remove test-completions script before merging.~

Also, we should probably update the homebrew recipe after the next release. Example from ripgrep: https://github.com/BurntSushi/ripgrep/blob/0a88cccd5188074de96f54a4b6b44a63971ac157/pkg/brew/ripgrep-bin.rb#L20

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,5 +100,3 @@ jobs:
         run: cargo check --workspace --all-targets
       - name: Run cargo check with all features
         run: cargo check --workspace --all-targets --all-features
-      - name: Run cargo check with rustls-tls feature
-        run: cargo check --workspace --all-targets --no-default-features --features rustls-tls


.github/workflows/test-completions.yml
mre, line 1: Just a test file to see if the workflow passes and generates those completion files.

lychee-bin/src/commands/generate.rs
```diff
@@ -72,20 +73,43 @@ const EXIT_CODE_SECTION: &str = "
 /// What to generate when providing the --generate flag
 #[derive(Debug, Deserialize, Clone, Display, EnumIter, EnumString, VariantNames, PartialEq)]
 #[non_exhaustive]
-#[strum(serialize_all = "snake_case")]
-#[serde(rename_all = "snake_case")]
+#[strum(serialize_all = "kebab-case")]
+#[serde(rename_all = "kebab-case")]
```
mre, line 77: That's because I want to type `complete-bash` instead of `complete_bash`. Doesn't change anything for `man`.

lychee-bin/complete/README.md
```diff
@@ -0,0 +1,150 @@
...
+lychee --generate complete-zsh         # Zsh
+```
+
+## Installation
+
+### Bash
+
+**User installation:**
+```bash
+dir="${XDG_CONFIG_HOME:-$HOME/.config}/bash_completion"
```
thomas-zahner, line 16: This looks a bit weird (the `:-` part) but I guess it works as intended?

lychee-bin/complete/_lychee
```diff
@@ -0,0 +1,117 @@
+#compdef lychee
```
thomas-zahner: I think we shouldn't track any of these files (lychee-bin/complete/* expect readme), as they are generated either manually as described in the docs or via release workflow. Do you agree? We could then also consider moving lychee-bin/complete/README.md to docs/AUTOCOMPLETE.md

lychee-bin/src/commands/generate.rs
thomas-zahner: Fine with that :+1:

.github/workflows/test-completions.yml
thomas-zahner: So we remove this workflow before merging, right?

lychee-bin/complete/README.md
mre: yeah, that's just the syntax for the fallback

lychee-bin/complete/_lychee
mre: [Ripgrep tracks them.](https://github.com/BurntSushi/ripgrep/tree/0a88cccd5188074de96f54a4b6b44a63971ac157/crates/core/flags/complete). In my opinion, it's great to see the diff for future troubleshooting between versions of `clap_complete`.

.github/workflows/test-completions.yml
mre: yes yes

mre:
@thomas-zahner fyi, I tested it on bash, zsh and fish on macOS.

Todo before merging:

- [x] Test locally (@thomas-zahner)
- [ ] update nix package (@thomas-zahner)
- [ ] remove test-completions workflow (@mre)

lychee-bin/complete/_lychee
thomas-zahner: Huh that surprises me..
Wouldn't we then also need a workflow, a test or some sort of automation to keep these generated files up to date? Otherwise, we constantly forget about keeping these files up-to-date.

mre: Technically yes. That would be a nice addition, but it would also mean some additional work. These things tend to get tricky, especially with committing those files, which might trigger other workflows etc. But it's not strictly necessary in the beginning, in my opinion.

mre: To expand on that a bit, I believe most people wouldn't care and for the few people who do, they get it out of the box as part of their release builds. We always bundle the latest versions. And the files in the repo are for troubleshooting, for the most part. I could add a `make` target to update them manually for now.

mre: Added the `make completions` target.

thomas-zahner: Yes that's why personally I would not track those generated files at all. They should always be fully reproducible and as you say they are generated and included with every release. It would be less code, less files and we wouldn't have this problem of outdated files at all.

But if you really prefer to include them for debugging/troubleshooting we can do so.

Also if most people don't care, why should we care to add these files to the repository in the first place? :sweat_smile:

thomas-zahner: I think it would be interesting to know why ripgrep tracks them and if they have a mechanism to keep the files up-to-date.

mre: Did some research. 

1. They don't update the completions through CI, but they added a test to check that the completions are up-to-date, which is quite clever, because it would fail if contributors forget to update the file when new options get added: https://github.com/BurntSushi/ripgrep/blob/master/ci/test-complete

2. One reason (perhaps the only one?) why they track the completion files is that the ZSH completions are handwritten. https://github.com/BurntSushi/ripgrep/blob/0a88cccd5188074de96f54a4b6b44a63971ac157/crates/core/flags/complete/zsh.rs#L4-L5

This way, they get a lot of confidence in their completions with little maintenance overhead. We could do the same.

mre: Added the check

mre:
The completion check makes the build slower (by ~30 seconds) because we have to build the binary first to run `lychee --help`.  Maybe that's acceptable, but I'm open for suggestions. I would assume that it gets faster with subsequent runs because of the build cache.

thomas-zahner:
> The completion check makes the build slower

Thanks for adding the test.

You do build specifically for the test script.

```
      - name: Build binary
        run: cargo build
```

I don't think this should be necessary.
See Test [the "zsh shell completions (Unix, sans cross)" step](https://github.com/BurntSushi/ripgrep/blob/0a88cccd5188074de96f54a4b6b44a63971ac157/.github/workflows/ci.yml#L182) which is part of the normal ci workflow, where the binary was built in a previous step already. This would probably save those 30 additional seconds?

mre:
Good idea. Let's try!

lychee-bin/complete/README.md
thomas-zahner: ah okay

lychee-bin/complete/_lychee
thomas-zahner: > One reason (perhaps the only one?) why they track the completion files is that the ZSH completions are handwritten.

Ah that would make sense. Then one could argue that we don't need to track them, as we have no customisation or handwritten scripts. Also we could do the new CI shell completion tests by using the output of `--generate` and not track the generated files. I personally would prefer that. I think tracking the generated files only has any benefit, if we make manual adjustments to those files or if we intend to do so in the future.

thomas-zahner:
Cool. I've tested it with ZSH and it works perfectly fine. So it's fine to merge from my side. Only my comment about not tracking the generated files which I personally prefer. But I'm also fine with tracking them if you prefer it, or if you think we will soon do some manual adjustments to the scripts.

mre:
Yup! That did the trick. Checking completion takes all of 1 second now. Thanks for the tip!

lychee-bin/complete/README.md
```diff
@@ -0,0 +1,138 @@
+# Shell Completions for lychee
+
+lychee comes with built-in support for generating shell completions.
+This unlocks tab-completion for commands, options, and arguments in your shell!
+
+## Installation
+
+### Bash
+
+**User installation:**
```
katrinafyi, line 14: Maybe just a sentence here that if a user has installed lychee through a package manager, then *hopefully* completions should work out of the box. Otherwise, they can try the manual steps and/or ping the package maintained.

.github/workflows/release-binary.yml
```diff
@@ -185,10 +223,15 @@ jobs:
...
           chmod +x lychee
-          mkdir docs
+          mkdir docs complete
           cp ../../README.md ../../docs/TROUBLESHOOTING.md docs
           ./lychee --generate man > docs/lychee.1
-          tar -czf lychee.tar.gz lychee docs/
+          ./lychee --generate complete-bash > complete/lychee.bash
+          ./lychee --generate complete-elvish > complete/lychee.elv
+          ./lychee --generate complete-fish > complete/lychee.fish
+          ./lychee --generate complete-powershell > complete/_lychee.ps1
```
katrinafyi, line 232: This block is repeated 3 times amongst the workflows. Can we use the makefile task instead? I'm just worried about subtle typos in the workflows that run less often, like release.

scripts/check-completions.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
```
katrinafyi: The script is lovely and colourful, but is it needed if completions aren't checked in anymore? If `make completions` succeeds then it should all be up to date.

.github/workflows/ci.yml
```diff
@@ -35,6 +35,9 @@ jobs:
         run: cargo check --locked
       - name: Test
         run: make test
+      # Shell completions are generated from CLI help and should be platform-independent
+      - name: Check completions are up to date
+        run: ./scripts/check-completions.sh
```
katrinafyi: Remove this to fix CI?

lychee-bin/complete/README.md
katrinafyi: Also, maybe add a gitignore in this folder for the generated files.

mre:
This is ready for another review. I think I addressed all comments now.

katrinafyi (approved):

PR #1974, opened 2025-12-29
feat: per-host caching

thomas-zahner:
This actually implements per-host caching.
Previously there were some structures and methods which made it look like it was implemented. But upon closer inspection I realised this was not the case.

In the process I've also tried to clarify the difference between persistent inter-process caching (lychee cache file) and this process internal per-host caching in some places. Note that the requests which are cached in this "new" way do not appear as "cached" in the progress output. (from lychee cache `[200] http://localhost:9000/ | OK (cached)` vs per-host cached `[200] http://localhost:9000/`) I've considered doing that as well, but that would probably require quite an additional effort without really much UX improvement. The per-host statistics cache rate is accurate and does count the per-host cache hits as such.

Closes https://github.com/lycheeverse/lychee/issues/1815 and https://github.com/lycheeverse/lychee/issues/1593

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1484,12 +1484,12 @@ The config file should contain every possible key for documentation purposes."
 
     #[tokio::test]
     async fn test_process_internal_host_caching() -> Result<()> {
-        // Note that this process internal per-host caching
+        // Note that this process-internal per-host caching
         // has no direct relation to the lychee cache file
         // where state can be persisted between multiple invocations.
         let server = wiremock::MockServer::start().await;
 
-        // Return one rate limited response to make sure that
+        // Return one rate-limited response to make sure that
         // such a response isn't cached.
         wiremock::Mock::given(wiremock::matchers::method("GET"))
             .respond_with(ResponseTemplate::new(429))
@@ -1505,7 +1505,7 @@ The config file should contain every possible key for documentation purposes."
 
         let temp_dir = tempfile::tempdir()?;
         for i in 0..9 {
-            let test_md1 = temp_dir.path().join(format!("test{}.md", i));
+            let test_md1 = temp_dir.path().join(format!("test{i}.md"));
             fs::write(&test_md1, server.uri())?;
         }
 
--- a/lychee-lib/src/ratelimit/mod.rs
+++ b/lychee-lib/src/ratelimit/mod.rs
@@ -38,13 +38,13 @@ pub(crate) struct CacheableResponse {
 }
 
 impl CacheableResponse {
-    async fn try_from(value: Response) -> Result<Self> {
-        let status = value.status();
-        let headers = value.headers().clone();
-        let url = value.url().clone();
-        let text = value.text().await.map_err(ErrorKind::ReadResponseBody)?;
+    async fn try_from(response: Response) -> Result<Self> {
+        let status = response.status();
+        let headers = response.headers().clone();
+        let url = response.url().clone();
+        let text = response.text().await.map_err(ErrorKind::ReadResponseBody)?;
 
-        Ok(CacheableResponse {
+        Ok(Self {
             status,
             text,
             headers,


lychee-bin/src/formatters/host_stats/compact.rs
```diff
@@ -70,12 +70,8 @@ impl Compact {
 }
 
 impl HostStatsFormatter for Compact {
-    fn format(&self, host_stats: HashMap<String, HostStats>) -> Result<Option<String>> {
-        if host_stats.is_empty() {
-            return Ok(None);
-        }
-
+    fn format(&self, host_stats: HashMap<String, HostStats>) -> Result<String> {
         let compact = CompactHostStats { host_stats };
```
mre, line 74: Thanks for cleaning that up.

lychee-bin/tests/cli.rs
```diff
@@ -1482,6 +1482,51 @@ The config file should contain every possible key for documentation purposes."
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_process_internal_host_caching() -> Result<()> {
+        // Note that this process internal per-host caching
```
mre:
```suggestion
        // Note that this process-internal per-host caching
```

lychee-bin/tests/cli.rs
```diff
@@ -1482,6 +1482,51 @@ The config file should contain every possible key for documentation purposes."
...
+        // has no direct relation to the lychee cache file
+        // where state can be persisted between multiple invocations.
+        let server = wiremock::MockServer::start().await;
+
+        // Return one rate limited response to make sure that
```
mre:
```suggestion
        // Return one rate-limited response to make sure that
```

lychee-bin/tests/cli.rs
```diff
@@ -1482,6 +1482,51 @@ The config file should contain every possible key for documentation purposes."
...
+
+        wiremock::Mock::given(wiremock::matchers::method("GET"))
+            .respond_with(ResponseTemplate::new(200))
+            .expect(1)
+            .mount(&server)
+            .await;
+
+        let temp_dir = tempfile::tempdir()?;
+        for i in 0..9 {
+            let test_md1 = temp_dir.path().join(format!("test{}.md", i));
```
mre:
```suggestion
            let test_md1 = temp_dir.path().join(format!("test{i}.md"));
```

lychee-lib/src/ratelimit/host/host.rs
```diff
@@ -108,80 +99,64 @@ impl Host {
...
 
     /// Cache a request result
-    pub fn cache_result(&self, uri: &Uri, status: &Status) {
-        let cache_value = HostCacheValue::from(status);
-        self.cache.insert(uri.clone(), cache_value);
+    fn cache_result(&self, uri: &Uri, response: CacheableResponse) {
+        // Do not cache responses that are potentially retried
+        if !response.status.should_retry() {
+            self.cache.insert(uri.clone(), response);
+        }
```
mre, line 119: Should we add a `debug!` or `trace!` entry if it wasn't cached?

lychee-lib/src/ratelimit/host/host.rs
```diff
@@ -365,7 +334,7 @@ mod tests {
         let host = Host::new(key.clone(), &host_config, &global_config, Client::default());
 
         assert_eq!(host.key, key);
-        assert_eq!(host.available_permits(), 10); // Default concurrency
```
mre, line 368: I kinda liked that method. The semaphore is an implementation detail. But I don't feel too strongly about it, either. 😆

lychee-lib/src/ratelimit/mod.rs
```diff
@@ -19,4 +19,36 @@ mod pool;
...
+        let text = value.text().await.map_err(ErrorKind::ReadResponseBody)?;
+
+        Ok(CacheableResponse {
+            status,
+            text,
+            headers,
+            url,
+        })
+    }
+}
```
mre, line 54: It's probably a bit nicer to use `response: Response` here? And if you use `Self` at the end, it doesn't need to be updated if we change the struct name. And I moved the text parsing to the top, as this avoids the unnecessary clones on headers and url if it fails.

```suggestion
impl CacheableResponse {
    async fn try_from(response: Response) -> Result<Self> {
        let text = response.text().await.map_err(ErrorKind::ReadResponseBody)?;
        let status = response.status();
        let headers = response.headers().clone();
        let url = response.url().clone();

        Ok(Self {
            status,
            text,
            headers,
            url,
        })
    }
}
```

mre (approved):
Sorry about the mess. That means that we kept checking repeated URLs to the same hosts. On the plus side, with this being merged, we should see another performance bump for websites with many identical links. Only added a few very minor syntactic suggestions. Semantically, it's all good, so feel free to ignore. Great work!

lychee-lib/src/ratelimit/host/host.rs
thomas-zahner: Hmm, I think the less common case is actual cache hits. So if we log anything I would probably rather log cache hits than cache misses.

Going to merge now, we could still add that if you want.

lychee-lib/src/ratelimit/host/host.rs
thomas-zahner: True. I've removed it because it was a `pub(crate)` method but it wasn't used anywhere except the test below. IMO it's fine to have that implementation detail in a unit test

PR #1975, opened 2025-12-30
Combine host stats with response stats

thomas-zahner:
Previously the two statistics were treated separately. This especially lead to an invalid JSON output, where we showed two separate top-level objects which weren't separated with a comma. Clean up and remove unnecessary abstractions in the process.

# Previous

```json
{
  "total": 12,
  "successful": 8,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 0,
  "excludes": 4,
  "errors": 0,
  "cached": 0,
  "success_map": {},
  "error_map": {},
  "suggestion_map": {},
  "redirect_map": {},
  "excluded_map": {},
  "duration_secs": 0,
  "detailed_stats": false
}
{
  "host_statistics": {
    "endler.dev": {
      "cache_hit_rate": 0.0,
      "cache_hits": 0,
      "cache_misses": 1,
      "client_errors": 0,
      "median_request_time_ms": 72,
      "rate_limited": 0,
      "server_errors": 0,
      "status_codes": {
        "200": 1
      },
      "success_rate": 1.0,
      "successful_requests": 1,
      "total_requests": 1
    }
  }
}
```

# New

```json
{
  "total": 12,
  "successful": 8,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 0,
  "excludes": 4,
  "errors": 0,
  "cached": 0,
  "success_map": {},
  "error_map": {},
  "suggestion_map": {},
  "redirect_map": {},
  "excluded_map": {},
  "duration_secs": 0,
  "detailed_stats": false,
  "host_stats": {
    "endler.dev": {
      "total_requests": 1,
      "successful_requests": 1,
      "success_rate": 1.0,
      "rate_limited": 0,
      "client_errors": 0,
      "server_errors": 0,
      "median_request_time_ms": 87,
      "cache_hits": 0,
      "cache_misses": 1,
      "cache_hit_rate": 0.0,
      "status_codes": {
        "200": 1
      }
    }
  }
}
```

### DIFF
--- a/lychee-bin/src/formatters/host_stats/compact.rs
+++ b/lychee-bin/src/formatters/host_stats/compact.rs
@@ -4,14 +4,14 @@ use crate::formatters::color::{DIM, NORMAL, color};
 use lychee_lib::ratelimit::HostStatsMap;
 
 pub(crate) struct CompactHostStats {
-    pub(crate) host_stats: HostStatsMap,
+    pub(crate) host_stats: Option<HostStatsMap>,
 }
 
 impl Display for CompactHostStats {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.host_stats.is_empty() {
+        let Some(host_stats) = &self.host_stats else {
             return Ok(());
-        }
+        };
 
         writeln!(f)?;
         writeln!(f, "📊 Per-host Statistics")?;
@@ -20,7 +20,7 @@ impl Display for CompactHostStats {
         color!(f, DIM, "{}", separator)?;
         writeln!(f)?;
 
-        let sorted_hosts = self.host_stats.sorted();
+        let sorted_hosts = host_stats.sorted();
 
         // Calculate optimal hostname width based on longest hostname
         let max_hostname_len = sorted_hosts
--- a/lychee-bin/src/formatters/host_stats/detailed.rs
+++ b/lychee-bin/src/formatters/host_stats/detailed.rs
@@ -3,19 +3,19 @@ use std::fmt::{self, Display};
 use lychee_lib::ratelimit::HostStatsMap;
 
 pub(crate) struct DetailedHostStats {
-    pub(crate) host_stats: HostStatsMap,
+    pub(crate) host_stats: Option<HostStatsMap>,
 }
 
 impl Display for DetailedHostStats {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if self.host_stats.is_empty() {
+        let Some(host_stats) = &self.host_stats else {
             return Ok(());
-        }
+        };
 
         writeln!(f, "\n📊 Per-host Statistics")?;
         writeln!(f, "---------------------")?;
 
-        for (hostname, stats) in self.host_stats.sorted() {
+        for (hostname, stats) in host_stats.sorted() {
             writeln!(f, "\nHost: {hostname}")?;
             writeln!(f, "  Total requests: {}", stats.total_requests)?;
             writeln!(
--- a/lychee-bin/src/formatters/host_stats/markdown.rs
+++ b/lychee-bin/src/formatters/host_stats/markdown.rs
@@ -7,18 +7,18 @@ use tabled::{
 };
 
 pub(crate) struct MarkdownHostStats {
-    pub(crate) host_stats: HostStatsMap,
+    pub(crate) host_stats: Option<HostStatsMap>,
 }
 
 impl Display for MarkdownHostStats {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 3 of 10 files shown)


lychee-bin/src/formatters/mod.rs
```diff
@@ -42,19 +42,6 @@ pub(crate) fn get_progress_formatter(mode: &OutputMode) -> Box<dyn ResponseForma
     get_response_formatter(&mode)
 }
 
-/// Create a host stats formatter based on the given format and mode options
-pub(crate) fn get_host_stats_formatter(
-    format: &StatsFormat,
-    _mode: &OutputMode,
-) -> Box<dyn HostStatsFormatter> {
-    match format {
-        StatsFormat::Compact | StatsFormat::Raw => Box::new(host_stats::Compact::new()), // Use compact for raw
```
thomas-zahner, line 51: Previously we said "Raw should behave the same as Compact", I've changed it now. See other comment.

lychee-bin/src/formatters/stats/raw.rs
```diff
@@ -12,7 +12,7 @@ impl Raw {
 
 impl StatsFormatter for Raw {
     /// Don't print stats in raw mode
-    fn format(&self, _stats: ResponseStats) -> Result<Option<String>> {
-        Ok(None)
+    fn format(&self, _: OutputStats) -> Result<String> {
+        Ok(String::new())
```
thomas-zahner, line 16: Raw does not currently output anything at all on latest release. With the new big per-host feature, on master host statistics are shown as with Compact, but all other stats are hidden, I feel like this doesn't make sense? Is the purpose of the raw mode really to hide all output?

mre: Yes, the purpose of the raw mode is that the output could be piped into another program for processing. As such, it should not apply any formatting. My current understanding is that we should only print the outcome of the link check itself in raw mode, because the stats would conflate the output and make parsing harder. (I don't suggest that users should parse the raw output, but rather that it could be done as a last resort in case there is no better way to do the data handling.)

lychee-bin/src/formatters/mod.rs
mre: Thanks, yes, that's a footgun.

mre (approved):
Big improvement.

lychee-bin/src/formatters/stats/raw.rs
thomas-zahner: > My current understanding is that we should only print the outcome of the link check itself in raw mode

So you mean it should only print `🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors`? That would be compact mode. (the default mode) Raw currently does not print anything at all (both with latest release and this PR), except the progress which has nothing to do with the output format.

I currently do not see any use in this raw mode. If the purpose is to print nothing at all, I think we could say that's the users responsibility. (` > /dev/null`) Also the name doesn't make sense. Raw doesn't sound like "print nothing at all".

thomas-zahner:
@mre Thanks for the review. During refactoring I forgot to use the `host_stats` option, now fixed in 5fb8fe325da004d7595107338ad2cb48a68e9553. Now I'm only confused about the raw mode, as mentioned in the other comment.

lychee-bin/src/formatters/stats/raw.rs
mre: > Raw currently does not print anything at al

Oh! My bad. That's certainly unexpected. In this case, your change makes sense. I don't know if it broke or if it has always been the case. Too lazy to check, but my hope is that it did something useful at some point. ^^ So the original idea of was "print all the check results without any special formatting; just 'plaintext'."

mre:
Yeah, my bad. It looks like raw mode is broken. See my other comment. The gist of it is, we can and should fix it and I'm okay with your change. 😊

lychee-bin/src/formatters/stats/raw.rs
thomas-zahner: Ok, I've just double-checked. It was introduced in 8c0a32d81d9e3890b7df37c85bc907b5bc9ccb80. Even back then final stats were hidden with raw mode. (`format_stats` returned `None`) The only difference was in the `ResponseFormatter`  (`write_response`) which is used for printing progress (not the final result).

This is different by now. We now have `mode` to specify how progress formatting is done. (e.g. emoji)
Because of this I would like to fully remove raw mode. It serves no purpose except hiding the final output. But this equivalent with e.g. piping to `/dev/null`.

thomas-zahner: I've created https://github.com/lycheeverse/lychee/issues/1976

This PR retains the current behaviour of hiding the results. We can then discuss what to do about the raw format on the new issue.

Issue #1980, opened 2026-01-06
`data:` URLs rejected by latest


haines:
With the `lycheeverse/lychee:latest-alpine` image I get

```
[ERROR] data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAASCAYAAABb0P4QAAAABHNCSVQICAgIfAhkiAAAAAlwSFlzAAAK8AAACvABQqw0mAAAABx0RVh0U29mdHdhcmUAQWRvYmUgRmlyZXdvcmtzIENTM5jWRgMAAAAVdEVYdENyZWF0aW9uIFRpbWUAMy8xNC8wOeNZPpQAAAE2SURBVDiNrZTBccIwEEXfelIAHUA6CZ24BGaWO+FuzZAK4k6gg5QAdGAq+Bxs2Yqx7BzyL7Llp/VfzZeQhCTc/ezuGzKKnKSzpCxXJM8fwNXda3df5RZETlIt6YUzSQDs93sl8w3wBZxCCE10GM1OcWbWjB2mWgEH4Mfdyxm3PSepBHibgQE2wLe7r4HjEidpnXMYdQPKEMJcsZ4zs2POYQOcaPfwMVOo58zsAdMt18BuoVDPxUJRacELbXv3hUIX2vYmOUvi8C8ydz/ThjXrqKqqLbDIAdsCKBd+Wo7GWa7o9qzOQHVVVXeAbs+yHHCH4aTsaCOQqunmUy1yBUAXkdMIfMlgF5EXLo2OpV/c/Up7jG4hhHcYLgWzAZXUc2b2ixsfvc/RmNNfOXD3Q/oeL9axJE1yT9IOoUu6MGUkAAAAAElFTkSuQmCC | URL is missing a host: URL missing hostname. Check URL format
```

With `lycheeverse/lychee:0.22.0-alpine` I don't get an error.

katrinafyi:
bisected to f03b1c9acbb0d5cee52899858d2296c390623cc1 as first bad commit.

i also note that there _is_ a test dealing with data URIs, but it only tests that the URIs are detected in --dump mode. they don't check whether they are accepted by the link checker.

after changing `TryFrom<&Url> for HostKey` to support data URIs, a different error is returned when it gets to
https://github.com/lycheeverse/lychee/blob/03bbbd7c539770eb707366e3bc7bf276fb632781/lychee-lib/src/ratelimit/pool.rs#L85-L90
[...]

PR #1982, opened 2026-01-06
fix: restore request::Error builder errors as Unsupported

katrinafyi:
idk if this was an intentional change in the rate limiting change. the  change happened because `enum Status` has special logic for detecting  error kinds, but this only happens when `From<reqwest::Error>` is  invoked, and not `From<ErrorKind>`. the logic says that some reqwest  errors get treated as different status depending on their type. in  particular, the data URI case got an Unsupported status.
  
  the rate limiting change did an extra .into() (via `?`) which converted a  reqwest error into an ErrorKind, and this happened _before_ the into  Status conversion.

this PR fixes the bug by changing `From<ErrorKind> for Status` to dispatch to `From<reqwest::Error> for Status` in the BuildRequestClient case. this allows ErrorKind::BuildRequestClient to also be treated as Unsupported
  
  along the way, i also turn `TryFrom<&Url> for HostKey` into `From<&Url> for HostKey`,  by setting a fallback hostname when the URL has none of its own. this  should be conceptually fine, but maybe it will be surprising if it turns  up in user-facing output. this could be worked around another way if desired.
  
  Fixes https://github.com/lycheeverse/lychee/issues/1980

### DIFF
--- a/lychee-bin/tests/data_uris.rs
+++ b/lychee-bin/tests/data_uris.rs
@@ -65,5 +65,4 @@ mod cli {
             .success()
             .stdout(contains("3 Unsupported"));
     }
-
 }


katrinafyi:
it seems like setting a fallback hostname is not the way to go? at least, the tests don't like it. feedback would be appreciated.

that said, i do feel like it shouldn't be an error if there's no hostname... not sure

thomas-zahner:
Thank you for the investigation and PR! It wasn't really intentional I think, but to fix some failing tests IIRC. I guess we didn't consider and test URLs which do not contain any hostnames. I will give this some thought.

lychee-bin/tests/data_uris.rs
```diff
@@ -53,4 +53,17 @@ mod cli {
...
+    #[test]
+    fn test_check_data_uris() {
+        let input = fixtures_path().join("TEST_DATA_URIS.html");
+
+        cargo_bin_cmd!()
+            .arg(input)
+            .arg("--exclude-loopback")
+            .assert()
+            .success()
+            .stdout("3 Unsupported");
```
thomas-zahner:
```suggestion
            .stdout(contains("3 Unsupported"));
```

katrinafyi:
closing in favour of #1990 :)

Issue #1896, opened 2025-10-31
Using a local --base-url breaks fragments in relative links in local files


katrinafyi:
```bash
echo '[a](b.html#a)
[a](/b.html#a)
[a](https://a.com/#asd)' > a.md
cargo run --  a.md --base-url $(pwd) --dump
cargo run --  a.md  --dump     
```
The first command shows this bug. The hash (which should be a fragment) is treated as part of the file name. This only happens on relative links.
```
file:///home/x/progs/lychee/b.html%23a
https://a.com/#asd
file:///b.html%23a
```
For comparison, the second command without --base-url leaves the hashes as hashes so they will get picked up as fragments.
```
    [WARN] Error creating request: InvalidPathToUri("/b.html#a")
https://a.com/#asd
file:///home/x/progs/lychee/b.html#a
```

I've used --dump for demonstration, but the bug also happens in normal link checking.
```
$ cargo  run --  a.md --base-url $(pwd)          
[a.md]:
   [ERROR] file:///b.html%23a | Cannot find file: File not found. Check if file exists and path is correct
   [ERROR] file:///home/x/progs/lychee/b.html%23a | Cannot find file: File not found. Check if file exists and path is correct
   [ERROR] https://a.com/#asd | Network error: Connection failed. Check network connectivity and firewall settings
```

The bug doesn't happen when we use a file:// base because that gets treated as a remote base for the purposes of parsing.
```
$ cargo run --  a.md --base-url file://$(pwd) --dump 
file:///b.html#a
https://a.com/#asd
file:///home/x/progs/b.html#a
```
Very fun

wade-cheng:
I'm a bit confused at what this issue means, but small note: I'm also getting links in local files being broken, ones that seem more simple than yours. This is for the contents of a static site (that is, HTML, not md). I'm running lychee on the file tree obtainable here https://github.com/wade-cheng/wade-cheng.github.io/tree/gh-pages.

That is, `lychee .` gives a bajillion warn messages like:
```
    [WARN] Error creating request: InvalidPathToUri("/")
    [WARN] Error creating request: InvalidPathToUri("/blog/")
    [WARN] Error creating request: InvalidPathToUri("/garden/")
    [WARN] Error creating request: InvalidPathToUri("/blog/")
    [WARN] Error creating request: InvalidPathToUri("/assets/eb-garamond/eb-garamond.css")
    [WARN] Error creating request: InvalidPathToUri("/style.css")
[...]

PR #1983, opened 2026-01-07
refactor!: remove Base::Local and use file:// base URLs instead

katrinafyi:
This is mostly a behind-the-scenes refactor, but it has a minor behaviour change when a local base URL is used with root dir. You can see the changes in this test case: https://github.com/lycheeverse/lychee/commit/9ecf45536d3068f72fda26ccdcffa2899d341b4d.

I cannot tell if this would be a positive or negative change, or whether it would be okay to make. It does bring the local base URL behaviour more in line with remote base URL, with root-dir being injected directly after the base URL.

  Of course, it will also be a breaking change for users of lychee-lib.
  
  Removing Base::Local means that some special code paths handling local bases can be removed. This fixes some odd bugs that happened with local bases and not with remote bases.
  
  The functionality of local bases is still available through a file:// "remote" base. The Base::try_from function smooths this over by detecting strings which look like absolute paths and mapping these to file:// URLs. There is a new utils::url::parse_url_or_path helper function to help with this. 
  
  Fixes https://github.com/lycheeverse/lychee/issues/1896

This doesn't need to be merged if you don't want to

### DIFF
--- a/fixtures/resolve_paths/index2.html
+++ b/fixtures/resolve_paths/index2.html
@@ -1,0 +1,24 @@
+<html>
+    <head>
+        <title>Index</title>
+    </head>
+    <body>
+        <h1>Index Title</h1>
+        <p>
+            <ul>
+                <li>
+                    <a href="/">home</a>
+                </li>
+                <li>
+                    <a href="/about">About</a>
+                </li>
+                <li>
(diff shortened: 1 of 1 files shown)


katrinafyi:
see:

- https://github.com/lycheeverse/lychee/pull/2005

PR #1989, opened 2026-01-14
Accept ranges

### LINKED ISSUE #1769: `--accept` doesn't accept the same values as `--cache-exclude-status`
>       --cache-exclude-status <CACHE_EXCLUDE_STATUS>
>           A list of status codes that will be ignored from the cache
> 
>           The following exclude range syntax is supported: [start]..[[=]end]|code. Some valid
>           examples are:
> 
>           - 429 (excludes the 429 status code only)
>           - 500.. (excludes any status code >= 500)

looks like this should work

```shell-session
$ lychee -- --config tools/lychee/config.toml --cache --cache-exclude-status '500..' .
warning: Git tree '/home/katexochen/edgelesssys/contrast' is dirty
error: invalid value '500..' for '--cache-exclude-status <CACHE_EXCLUDE_STATUS>': failed to parse accept range: no range pattern found

For more information, try '--help'.
error: Recipe `check-links` failed on line 385 with exit code 2

$ lychee -- --version
lychee 0.18.1
```

thomas-zahner:
Fixes #1769. Previously `lychee example.com --accept 123..` lead to `Error: invalid status code`.

In the process I've eliminated some primitive obsession with `u16` where we actually meant `StatusCode`. This also improved error handling with invalid values in the lychee cache file. See commit messages for more information.

Improves the error messages overall. For example prviously:

`error: invalid value '123..1234' for '--cache-exclude-status <CACHE_EXCLUDE_STATUS>': failed to parse accept range: no range pattern found`

Now:

`error: invalid value '123..1234' for '--cache-exclude-status <CACHE_EXCLUDE_STATUS>': failed to parse range: values must represent valid status codes between 100 and 999 (inclusive)`

### DIFF
--- a/lychee-lib/src/types/accept/range.rs
+++ b/lychee-lib/src/types/accept/range.rs
@@ -13,8 +13,13 @@ const MIN: u16 = 100;
 /// Biggest accepted value
 const MAX: u16 = 999;
 
-static RANGE_PATTERN: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^([0-9]+)?\.\.((=?)([0-9]+))?$|^([0-9]+)$").unwrap());
+/// Regex to compute range values
+static RANGE_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(
+        r"^(?<start>[0-9]+)?\.\.(?<until>(?<inclusive>=)?(?<end>[0-9]+))?$|^(?<single>[0-9]+)$",
+    )
+    .unwrap()
+});
 
 /// Indicates that the parsing process of an [`StatusRange`]  from a string
 /// failed due to various underlying reasons.
@@ -33,7 +38,7 @@ pub enum StatusRangeError {
     InvalidRangeIndices,
 
     /// The u16 values must be representable as status code
-    #[error("values must represent valid status codes between 100 and 999 (inclusive)")]
+    #[error("values must represent valid status codes between {MIN} and {MAX} (inclusive)")]
     InvalidStatusCodeValue,
 }
 
@@ -52,20 +57,20 @@ impl FromStr for StatusRange {
             .captures(s)
             .ok_or(StatusRangeError::NoRangePattern)?;
 
-        if let Some(value) = captures.get(5) {
+        if let Some(value) = captures.name("single") {
             let value: u16 = value.as_str().parse()?;
             Self::new(value, value)
         } else {
-            let start: u16 = match captures.get(1) {
+            let start: u16 = match captures.name("start") {
                 Some(start) => start.as_str().parse().unwrap_or_default(),
                 None => MIN,
             };
-            if captures.get(2).is_none() {
+            if captures.name("until").is_none() {
                 return Self::new(start, MAX);
             }
 
-            let inclusive = !captures[3].is_empty();
-            let end: u16 = captures[4].parse()?;
+            let inclusive = captures.name("inclusive").is_some();
+            let end: u16 = captures["end"].parse()?;
 
             if inclusive {
                 Self::new(start, end)


lychee-lib/src/types/accept/range.rs
```diff
@@ -1,10 +1,20 @@
...
 
+/// Smallest accepted value
+const MIN: u16 = 100;
+
+/// Biggest accepted value
+const MAX: u16 = 999;
+
 static RANGE_PATTERN: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^([0-9]{3})?\.\.((=?)([0-9]{3}))?$|^([0-9]{3})$").unwrap());
+    LazyLock::new(|| Regex::new(r"^([0-9]+)?\.\.((=?)([0-9]+))?$|^([0-9]+)$").unwrap());
```
thomas-zahner: This is for better error messages, so that we can yield `InvalidStatusCodeValue`

lychee-lib/src/types/accept/range.rs
```diff
@@ -21,117 +31,114 @@ pub enum AcceptRangeError {
...
+            Self::new(value, value)
         } else {
             let start: u16 = match captures.get(1) {
                 Some(start) => start.as_str().parse().unwrap_or_default(),
-                None => 0,
+                None => MIN,
             };
             if captures.get(2).is_none() {
-                return Self::new_from(start, u16::MAX);
+                return Self::new(start, MAX);
```
thomas-zahner, line 69: This is the bugfix with the `123..` example, it now ends at 999

thomas-zahner:
@HadrienG2 & @katrinafyi feel free to take a look and review

lychee-lib/src/types/accept/range.rs
katrinafyi: I'd comment the regex, outlining the cases it's meant to cover and labelling each group number. 

_optionally_, I personally try very hard to avoid regex - it could be a string split and manual processing. but I think the complexity is just enough for a regex too

lychee-lib/src/types/accept/range.rs
```diff
@@ -21,117 +31,114 @@ pub enum AcceptRangeError {
     /// The start index is larger than the end index.
     #[error("invalid range indices, only start < end supported")]
     InvalidRangeIndices,
+
+    /// The u16 values must be representable as status code
+    #[error("values must represent valid status codes between 100 and 999 (inclusive)")]
```
katrinafyi: any chance of using the constants to make this string, rather than hard coding?

lychee-lib/src/types/accept/range.rs
thomas-zahner: > I personally try very hard to avoid regex

I like this mindset :) But I think I won't change it in this PR. It's well tested so that we know it works as expected but I agree that regex is often not the best way. I've documented it by using named capture groups in 45e54fea

lychee-lib/src/types/accept/range.rs
thomas-zahner: Great idea, was easier than expected. 45e54fea

lychee-lib/src/types/cache.rs
```diff
@@ -37,15 +64,24 @@ impl<'de> Deserialize<'de> for CacheStatus {
...
+                        // classify successful status codes as cache status success
+                        // Does not account for status code overrides passed through
+                        // the 'accept' flag. Instead, this is handled at a higher level
+                        // when the cache status is converted to a status.
+                        Ok(CacheStatus::Ok(code))
+                    } else {
+                        // classify redirects, client errors, & server errors as cache status error
+                        Ok(CacheStatus::Error(Some(code)))
+                    }
+                }
```
mre, line 84: Nicely done

mre:
image

Superb job and much cleaner. Neat!

PR #1990, opened 2026-01-15
Rejected urls

thomas-zahner:
Fixes https://github.com/lycheeverse/lychee/issues/1980

Supersedes https://github.com/lycheeverse/lychee/pull/1982

@katrinafyi This change should suffice to achieve the same thing as in https://github.com/lycheeverse/lychee/pull/1982 but without changing error handling much or introducing the magic/sentinel value.

### DIFF
--- a/lychee-lib/src/checker/website.rs
+++ b/lychee-lib/src/checker/website.rs
@@ -158,10 +158,7 @@ impl WebsiteChecker {
                     status
                 }
             }
-            Err(e) => match e {
-                ErrorKind::NetworkRequest(error) => Status::from(error),
-                _ => e.into(),
-            },
+            Err(e) => e.into(),
         }
     }
 
--- a/lychee-lib/src/types/status.rs
+++ b/lychee-lib/src/types/status.rs
@@ -298,21 +298,20 @@ impl From<ErrorKind> for Status {
     fn from(e: ErrorKind) -> Self {
         match e {
             ErrorKind::InvalidUrlHost => Status::Unsupported(ErrorKind::InvalidUrlHost),
+            ErrorKind::NetworkRequest(e)
+            | ErrorKind::ReadResponseBody(e)
+            | ErrorKind::BuildRequestClient(e) => {
+                if e.is_timeout() {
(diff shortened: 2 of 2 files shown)


katrinafyi:
yeah, seems simpler! 

edit: what do you think about adding a case so ErrorKind::BuildRequestClient can get mapped to timeout and unsupported statuses? it would match the old behaviour in cases aside from data URIs

basically, just
```diff
+     ErrorKind::BuildRequestClient(e) => Status::from(e),
```

thomas-zahner:
@katrinafyi Hmm that might make sense.

> it would match the old behaviour

Do you have an example of that, or could you point me to a specific line? I don't remember having changed that :sweat_smile:
We would ideally add a test which covers that case too, as documentation.

katrinafyi:
@thomas-zahner I tried a few things and don't have a concrete example, sadly. But it's the same premise as I wrote in the description of https://github.com/lycheeverse/lychee/pull/1982, and the line change would look like the change to lychee-lib/src/types/status.rs here: https://github.com/rina-forks/lychee/commit/5bcd7006195ef55fe6dab729eb68567161c7f698

The idea is that if you have a reqwest::Error, there are different ways to end up at a lychee Status. At the moment, some reqwest::Error will lead to a different Status depending on what path you take. I think this can be subtle and surprising, especially given implicit conversions within `?` operator.

I think that either the two paths should be made to be logically the same (by adding this extra case), or the reqwest::Error &rarr; Status edge should be deleted and it all goes through ErrorKind first.

```mermaid
flowchart LR

reqwest::Error -->|Into| ErrorKind
reqwest::Error -->|Into| Status
ErrorKind -->|Into| Status
```

thomas-zahner:
@katrinafyi Thank you very much for the explanation. The diagram really helps!

What do you think if we just remove (or rather inline) the conversion from `reqwest:Error` to `Status`, this feels pretty clean. Also I realised that this conversion wasn't even used except for one place.

See 136913b490f467e21a76bf22698b8e415b546c1d

katrinafyi:
@thomas-zahner looks good!

Issue #1992, opened 2026-01-16
Why is ErrorKind::details not formatted?


katrinafyi:
The indentation is way too much in some places and it doesn't get picked up by `cargo fmt` or detected by the formatting check. Removing the clippy annotation doesn't seem to change the behaviour.

https://github.com/lycheeverse/lychee/blob/db63437eb41d1c6ae1dadce7823976d1a7c18484/lychee-lib/src/types/error.rs#L199-L205

 Other functions in the same file get formatted as you'd expect.

I know that macros can interfere with rustfmt, but it doesn't look like this code is within any macros.

katrinafyi:
Due to strings being too long and running into https://github.com/rust-lang/rustfmt/issues/3863

PR #1993, opened 2026-01-16
Fix source code formatting in ErrorKind::details()

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/1992. Doesn't change anything aside from formatting.

It looks like rustfmt gives up on a function if it has strings which exceed the max_width, because rustfmt doesn't make any changes to string literals.

After some long strings are manually broken up, rustfmt can format the rest of the function.

Example of manual change, old:
```
            ErrorKind::MissingGitHubToken => Some(
                "GitHub token required. Use --github-token flag or GITHUB_TOKEN environment variable".to_string()
            ),
```
new:
```
            ErrorKind::MissingGitHubToken => Some(format!(
                "GitHub token required. {}",
                "Use --github-token flag or GITHUB_TOKEN environment variable",
            )),
```

Unfortunately, this PR doesn't do anything to prevent this happening again. Some options I looked into:
- The rustfmt options error_on_line_overflow and error_on_unformatted seemed promising. However, I tried both and they do  not report the long strings as problems, but they _do_ report different unrelated errors across the codebase.
- Enable rustfmt option format_strings. Does fix the problem, but sometimes introduces `\` inside string literals. Also, the option is unstable and has some known bugs. Could be chosen as a trade-off.

### DIFF
--- a/lychee-lib/src/types/error.rs
+++ b/lychee-lib/src/types/error.rs
@@ -8,7 +8,7 @@ use tokio::task::JoinError;
 
 use super::InputContent;
 use crate::types::StatusCodeSelectorError;
-use crate::{basic_auth::BasicAuthExtractorError, utils, Uri};
+use crate::{Uri, basic_auth::BasicAuthExtractorError, utils};
 
 /// Kinds of status errors
 /// Note: The error messages can change over time, so don't match on the output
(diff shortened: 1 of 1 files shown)


thomas-zahner:
Thank you! I didn't know about this limitation until now, but I also remember being confused why rustfmt didn't work int that section and thought it was an issue on my side. (blaming my IDE probably)

`format_strings` sounds promising, but I guess we need to wait until it's stabilised. I really wouldn't use it while unstable.

mre:
Nice catch. It looks much better now.

Issue #1934, opened 2025-11-24
Cannot suppress redirect output


bmcfads:
It seems that some things have changed around redirects recently, and now my output (using the GitHub Action) is littered with redirect information. I cannot figure out how to suppress the redirects; I only want to see the error results.

For example, the output shows something like the following for every redirect:

```
* [200] <https://some-link-that-was-checked.com> | Redirect: Followed 1 redirect resolving to the final status of: OK. 
Redirects: https://some-link-that-was-checked.com --> https://the-link-it-was-redirected-to.com
```

How do I suppress these? I tried using the `--quiet` argument x2 with no luck.

bmcfads:
It's the `## Redirects per input` section that I am trying to suppress. 

The `# Summary` and `## Errors per input` sections appear above the redirects, so I have have to scroll up through hundreds of redirect outputs to find the actual errors.

mre:
Right now we only print redirects in verbose mode (`-v` and above). @thomas-zahner, maybe we should introduce a dedicated flag for that?

bmcfads:
@mre - I might be doing something wrong then. 

I tried adding `--quiet` to the list of args for the GitHub Action and it didn't suppress the redirect output. I then tried adding the quiet flag a second time in the list of args, and that didn't work either. 

Do you have an example of how to suppress the redirects per input info that comes after the summary, when using the GitHub Action?

mre:
The default lychee-action args contain `--verbose`, so maybe that's the problem:

https://github.com/lycheeverse/lychee-action/blob/79de8811d696d79a713d425d265c9870b929fbf2/action.yml#L6

You'd have to overwrite the args.

```yaml
- name: Link Checker
  uses: lycheeverse/lychee-action@v2
  with:
    # Check all markdown, html and reStructuredText files in repo (same as default, but no `--verbose`)
    args: --no-progress './**/*.md' './**/*.html' './**/*.rst'
```

If that doesn't work, can you show me your lychee workflow? Just paste the relevant parts here.

bmcfads:
Yes, the args are overwritten. These are the args I had been using:

```
- name: Check links
      uses: lycheeverse/lychee-action@v2
      with:
        args: >-
          --root-dir "${{ github.workspace }}"
          --no-progress
          --fallback-extensions mdx
          --include-fragments
          --include-mail
          --exclude-all-private
          --exclude-path "(^|\/)\."
          --accept 100..=103,200..=299,403,429
[...]

Issue #1312, opened 2023-12-03
Only show 200 OK links in -vv output


nathany:
Hi Matthias. I'm running lychee on a book with 574 total links reported, which is making it a little difficult to know which links I need to correct. The book is written with AsciiDoctor, so I first generated an HTML file. I don't mind that *.adoc isn't supported (#291) directly by lychee.

## Redirects

Running lychee nightly (or rather, compile from source), the main issue I'm seeing is that there are a number of **redirects**, but it's difficult to tell which redirect belongs to which link.

```
~/src/github.com/lycheeverse/lychee/target/debug/lychee output/get-programming-with-go-manuscript.html -vv
```

![redirect-nightly](https://github.com/lycheeverse/lychee/assets/4566/6fc89646-71a4-43d8-9eb4-0e00b4bd1a5c)

The redirect to `https://www.red-gate.com`... is from the `https://www.simple-talk.com` link at the bottom of the screenshot shown, but they aren't tied together. I assume this is related to concurrency and the redirect being a debug log (#1117) rather than designed into the user experience.

Ideally I'd like a report of which links required a redirect along with the final URL they redirected to. That way I could easily search for the original links and update them. (the intermediate redirects being a debug log is fine, as that doesn't concern me unless troubleshooting)

## Excluded

A related concern is the "6 Excluded" when looking at a list this large. If I scroll through the verbose version I can track them down, but there are hundreds of `✔ [200] ` lines that I want to ignore, even with a single `-v`. Even piping to `grep` doesn't help. 

Most of the excluded in my example are email addresses, which is fine (2 repeated a few times). But one excluded link is:

```
? [EXCLUDED] https://gobyexample.com/ | Excluded
```

From the HTML `<a href="https://gobyexample.com" class="bare">gobyexample.com</a>`. I'm not entirely sure why. So that's the sort of thing I'd like listed out.

## Suggestion

I think that the good links should only be shown with `-vv` and the corrections should be clearly shown with a single `-v` or even without.

Also may want to double check if there's an example.com default exclusion that's a little too aggressive.

mre:
All good suggestions. Changing the behavior of -v and -vv is a breaking change, but it should be fine as we haven't released 1.0 yet. 
The redirects steps should be more easily traceable. Let's see if we can add the redirect source and destination to the output. 
For excluded links, I think you could try different output formats (see `--format`) to choose a preferred style. We could add more styles.

[...]

PR #1994, opened 2026-01-19
Improve progress reporting

thomas-zahner:
- Fix untracked bug with -qq in f8756796 (bf8caed5 in master)
- Implement [suggestion](https://github.com/lycheeverse/lychee/issues/1934#issuecomment-3675453600) from @katrinafyi in 1852db35 (36ffa445 in master)
- Fix https://github.com/lycheeverse/lychee/issues/1934 in 33171c1c (91cda356 in master)
- Fix https://github.com/lycheeverse/lychee/issues/1312 in a157e51fffbbebe2641cfc8461a4817afdced3e7 (839e5a2f in master)

The PR is preferably merged without squashing. I tried to keep the history clean.


lychee-bin/src/main.rs
```diff
@@ -393,6 +393,7 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
     } else {
         let (response_stats, cache, exit_code, host_pool) = commands::check(params).await?;
         github_warning(&response_stats, &opts.config);
+        redirect_warning(&response_stats, &opts.config);
```
mre, line 396: We could consider introducing a `Warnings` struct, where we can push warnings collected throughout the run. Just an idea. Definitely overkill for now.

lychee-bin/src/progress.rs
```diff
@@ -39,27 +51,43 @@ impl Progress {
...
+            let show_success = self.log_level > Level::Info;
+            let is_success = response.status.is_success();
+
+            // hide successful responses unless log level is high enough
+            if !is_success || show_success {
+                // progress is reported on stderr and NOT on stdout
+                eprintln!("{message}");
+            }
+        }
+    }
```
mre, line 89: I found this a bit harder to read than necessary. Might just be me, but the double nesting confused me for a moment. You could use let-else and different boolean terms instead. What do you think?

```rust
fn print_progress(&self, response_message: Option<&(&ResponseBody, String)>) {
    if self.log_level < Level::Info {
        return;
    }

    let Some((response, message)) = response_message else {
        return;
    };

    let should_show_success = self.log_level > Level::Info;
    let is_error = !response.status.is_success();

    if is_error || should_show_success {
        eprintln!("{message}");
    }
}
```

lychee-bin/src/verbosity.rs
```diff
@@ -57,9 +57,10 @@ impl Verbosity {
         level_enum(self.verbosity()).to_level_filter()
     }
 
-    #[allow(clippy::cast_possible_wrap)]
-    const fn verbosity(&self) -> i8 {
-        level_value(log::Level::Warn) - (self.quiet as i8) + (self.verbose as i8)
+    const fn verbosity(&self) -> u8 {
+        level_value(log::Level::Warn)
+            .saturating_sub(self.quiet)
+            .saturating_add(self.verbose)
```
mre, line 63: Nice!

lychee-lib/src/types/redirect_history.rs
```diff
@@ -7,42 +10,58 @@ use std::{
...
+        }
     }
 
-    /// Represents zero redirects
+    /// Count how many times a redirect was followed.
     #[must_use]
-    pub const fn none() -> Self {
-        Redirects(vec![])
+    pub const fn count(&self) -> usize {
+        self.redirects.len()
```
mre, line 60: Ah, so much cleaner.

lychee-bin/src/main.rs
thomas-zahner: Yeah, might be worth a consideration in the long run. However, collecting the warnings during the run should not be necessary as of now. The response/stats contain all data needed to produce the current warnings. So currently I see no use of keeping track of additional data during the run when we can just map the final response/stats into the warnings.

lychee-bin/src/progress.rs
thomas-zahner: Good suggestion. I also like the name `should_show_success`. However, `is_error` is misleading and I want to keep `is_success`. Because if it's not a successful status it can still be a redirect, timeout, unknown, excluded, ...

PR #1997, opened 2026-01-20
Update comparison table

thomas-zahner:
Update comparison. We support the two cases since f03b1c9a


mre:
Thanks. I checked the entries and they all seem to be correct. Let's merge it. 😊

PR #2002, opened 2026-01-25
Ensure lychee is installed and in PATH before running

mre:
The auto-install hook installs `lychee` via `cargo binstall` / `cargo install`, but the binary is not discoverable in pre-commit's isolated environment because Cargo's bin directory is not added to `PATH`.

This PR fixes the issue by explicitly exporting the Cargo bin path before execution. AFAIK it has no impact on global user environment as the hook gets executed in its own subshell.

@Borda FYI

### DIFF
--- a/scripts/lychee_pre_commit.sh
+++ b/scripts/lychee_pre_commit.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 export CARGO_HOME="${CARGO_HOME:-$HOME/.cargo}"


Borda:
I can try it later but very appropriate your help!

scripts/lychee_pre_commit.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
```
thomas-zahner: Not really the topic of this PR, but increases portability (wouldn't run on my machine without this change)

```suggestion
#!/usr/bin/env bash
```

thomas-zahner:
@mre Looks good. On my machine the change does not seem to make any difference, master and this PR behave the same. That might be a good sign.

scripts/lychee_pre_commit.sh
mre: Indeed. Thanks a lot for the suggestion. Those small improvements add up to much better software in the long run, so thanks.

mre:
> @mre Looks good. On my machine the change does not seem to make any difference, master and this PR behave the same. That might be a good sign.

Yeah I think that's probably because you have lychee in your `$PATH`? I think my mistake was not understanding that those pre-commit-hooks spawn their own env with a clean path and so it won't work for people that don't install it on their machine. My mental model about commit hooks is still very shaky, though.

altendky:
This fix would be very useful for my project. I'm running pre-commit on Windows in CI and hitting this issue:

```
lychee..................................................................Failed
- hook id: lychee
- exit code: 1

Executable `/bin/bash` not found
```

pre-commit parses shebangs via the [`identify` library](https://github.com/pre-commit/identify/blob/main/identify/identify.py#L78-L83), which has special handling to strip `#!/usr/bin/env` and search PATH for the command. This makes `#!/usr/bin/env bash` work on Windows (finding `bash` via Git Bash), while `#!/bin/bash` fails.

CI failure for reference: https://github.com/altendky/onshape-mcp/actions/runs/21538951953/job/62070149665?pr=30

Thanks so much for the work here. I look forward to the fixes getting merged. If relevant, I could make a separate PR for just the shebang change.

cheers,
-kyle

PR #2004, opened 2026-01-25
Fix broken path in the extract example

seri:
Awesome repo that really helped me learn idiomatic Rust. I look forward to contributing more.

This PR fixes this command:

```
cargo run --example extract
```

It was broken on master.


mre:
Oh, very nice. Thanks for sending in the PR.

Issue #2000, opened 2026-01-25
Base URL breaks all local links in local files, except for root-relative


katrinafyi:
Bizarrely, using a base-url causes lychee to fail to resolve local links _except_ for root-relative links.

We will make a test file which has 3 local links, all of which should be valid.
```bash
touch a.html b.html
echo "[a](a.html)
[b](/b.html)
[b](file://$(pwd)/test.md)
[b](https://github.com)
" > test.md
cargo  run -- --verbose --base-url $(pwd) test.md
```

The output is doubly surprising because the file:// URLs which are errors actually exist on disk, and the one which does not exist on disk is counted as valid.
```
   [ERROR] file:///home/rina/progs/lychee/a.html | Cannot find file: File not found. Check if file exists and path is correct
     [200] file:///b.html
     [200] https://github.com/
   [ERROR] file:///home/rina/progs/lychee/test.md | Cannot find file: File not found. Check if file exists and path is correct
```

I think that there's some double-handling of base-url [in file checker](https://github.com/lycheeverse/lychee/blob/34199a407de0c22fba7f8f6198d8dfa29a59da33/lychee-lib/src/checker/file.rs#L117), which is re-inserting base URL to the beginning of all file:// URLs. This makes the `file:///b.html` valid and it makes the other ones into broken links like
```
file:///home/rina/progs/lychee/home/rina/progs/lychee/test.md
```

Issue #1964, opened 2025-12-20
root-dir incorrectly applies to links in remote HTML files


katrinafyi:
i have made a test .md file  at: https://gist.githubusercontent.com/katrinafyi/daefc003e04b7c2f73cb54615510dce0/raw/fe1b03f7a328e169b7d6445ecbac34053b526e8f/lychee-url-mapping-test-file.md

this has a number of links, one of which is `[root](/root)`. if we point lychee at this URL with the root-dir option, it looks does this:
```console
$ lychee --dump --root-dir /tmp https://gist.githubusercontent.com/katrinafyi/daefc003e04b7c2f73cb54615510dce0/raw/fe1b03f7a328e169b7d6445ecbac34053b526e8f/lychee-url-mapping-test-file.md 
[...]
https://gist.githubusercontent.com/tmp/root
```

this is merging the root-dir into the implicit base url from the file's URL location. this is the same as it does with `--root-dir` + `--base-url` and it is always wrong.

two possible fixes to make this make sense:
- only apply root-dir to local files
- `/root` link in remote files should become `/tmp/root`

i _strongly suggest_ the first way. in fact, i wrote the [`--root-dir` help](https://lychee.cli.rs/guides/cli/#--root-dir) assuming it already worked like that:

> `lychee --root-dir <ROOT_DIR>`: Root directory to use when checking absolute links **in local files.**

katrinafyi:
if you do `--root-dir /`, it gets even "crazier":

```
https://root/
```

mre:
> only apply root-dir to local files

I agree that this is the way to go. @thomas-zahner?

thomas-zahner:
Yes, to me it sounds very sensible to apply root-dir only to local files.

> if you do --root-dir /, it gets even "crazier":

Ha, that's weird 😅
And thank you for really thinking this through. This really makes lychee better. It's a topic I've personally never dealt much with in the codebase and the two options evolved quite a bit historically without much planning and without the use-cases in mind.

PR #1965, opened 2025-12-21
tests: add tests for the current state of "Checking a Local Folder with URL Remapping"

katrinafyi:
this adds tests for the use case of "Checking a Local Folder with URL Remapping", as motivated in https://github.com/lycheeverse/lycheeverse.github.io/pull/140. it makes no behaviour changes.

to review this PR, i suggest first understanding the use-case, then you can read the input files and expected URLs to check that they make sense. there are two new .md files inside the fixtures folder, one for whole domain mapping and one for subpath mapping. the markdown text explains the reasoning for each URL.

the remaps are deeply unpleasant and you shouldn't try too hard to understand them. they're also different from https://github.com/lycheeverse/lycheeverse.github.io/pull/140 because i found new edge cases and new bugs. i'll update the docs pr after...

issues arising:
- https://github.com/lycheeverse/lychee/issues/1953
- https://github.com/lycheeverse/lychee/issues/1964

Issue #1953, opened 2025-12-14
Relative parent links can traverse outside of root-dir


katrinafyi:
If I have a link inside `a.md` like
```md
[a](/../../../b.html)
```
and run lychee with root-dir, then lychee will search for b.html outside of root-dir
```console
$ pwd
/home/x/progs/lychee
$ cargo  run -- a.md --root-dir . --dump
file:///home/b.html
```

This is probably not a big deal, but it is _a_ difference between `--root-dir` and being hosted at the root of a website domain. 

For instance, a link `/../../../b.html` from a page at `https://example.com/a.html` would resolve to `https://example.com/b.html`. In local files, this would correspond to `/home/x/progs/lychee/b.html` rather than `/home/b.html`.

mre:
You're right. This should probably throw an error instead.

katrinafyi:
I'd want it to not be an error, but be "anchored" at the root dir. so `<root-dir>/.. == <root-dir>` basically.

this matches what happens if you have a `..` link in the root page of an online website.

mre:
Even better.

katrinafyi:
Having thought about it some more, I now think that for most users of --root-dir, preventing traversal outside of the root-dir could be surprising behaviour. They might not be thinking of a root-dir as a website root and they just think of it as a path on disk - in which case, you should be able to use .. to go to a parent.

So, I think the current behaviour is the more "natural" behaviour which most users would assume. Users wanting different behaviour will know to go looking for some option, and we could accommodate them with an option if needed.

This is probably a niche issue anyway :)

mre:
So shall we close this?

PR #2005, opened 2026-01-26
refactor!: add BaseInfo for resolving links, delete Base, cleanup utils

katrinafyi:
At the moment, lychee has _a lot_ of separately-implemented code paths which try to handle permutations of base-url and root-dir for both local and remote links. The separate code paths make it hard to reason about and a lot of this code is also manually handling strings. This leads to unclear semantics and, eventually, bugs.

In this PR, I try to unify a lot of this logic and standardise on using functions from [url::Url](https://docs.rs/reqwest/0.12.23/reqwest/struct.Url.html) for the bulk of the parsing and resolving. Local base URLs and root dirs, both being filesystem paths, are treated as `file://` URLs and this works nicely with the Url functions. The only thing we have to keep in mind is the special behaviour of URLs without trailing slash---always using Url::from_directory_path to convert paths to URLs will ensure they have trailing slashes.

This change (almost) entirely removes string operations on URLs. It avoids the need to manually percent encode/decode or to manually process anchors and query parameters. Unifying on URL rather than URL+PathBuf (like old Base) also avoids and fixes bugs where a path was incorrectly combined with a URL.

The biggest change in this PR is the new BaseInfo replaces the old Base type. It has three cases (more detailed comments are in the code):
```rust
pub enum BaseInfo {
    /// No base information is available.
    None,
    /// A base which cannot resolve root-relative links.
    NoRoot(Url),
    /// A full base made up of `origin` and `path`. This can resolve
    /// all kinds of relative links.
    Full(Url, String),
}
```
Rather than being separated like Base::Local and Base::Remote, the different BaseInfo cases represent capabilities of the base URL. We also add a None case to represent no base information.

This means that `BaseInfo` can be used in place of `Option<Base>`, which simplifies handling of the no-base case by moving it inside the BaseInfo functions. Much of the code changes are just making this replacement.

The second biggest change in this PR is the new `join_rooted` function in utils::url. This does a pretty straightforward adjustment of URLs so that root-relative file links go to the given root directory. (Previously: ~~This does some shuffling of URLs so we can use the URL library functions to implement root-dir in the way we want---namely, that a link of `/b` with a root-dir of `file:///a/` should go to `file:///a/b` and not `file:///b`.~~ )

This PR deletes `base.rs` and it deletes most functions in lychee-lib/src/utils/request.rs. All of those functionalities are now provided by BaseInfo and join_rooted.

This PR should have very few user-visible behaviour changes, *unless* the user was using local base URL or using base-url + root-dir.

Bugs which I think might be fixed by this PR:
- https://github.com/lycheeverse/lychee/issues/1896
- https://github.com/lycheeverse/lychee/issues/1964
- https://github.com/lycheeverse/lychee/issues/2000
- https://github.com/lycheeverse/lychee/issues/1953 (edit: i've undone the fix to this "bug" because i think it doesn't need fixing)

This PR supersedes https://github.com/lycheeverse/lychee/pull/1983. This PR would be complemented by https://github.com/lycheeverse/lychee/pull/1965 as we would gain more test cases which cover the mentioned bugs. Working on this PR motivated https://github.com/lycheeverse/lychee/pull/2001.

You can also read the rustdocs for this PR's code at https://lychee-docs-katrinafyi.pages.dev/lychee_lib/enum.BaseInfo. You can also see the unsquashed commit history over at https://github.com/lycheeverse/lychee/compare/master...rina-forks:lychee:base-info

Fixes https://github.com/lycheeverse/lychee/issues/1896
Fixes https://github.com/lycheeverse/lychee/issues/1964
Fixes https://github.com/lycheeverse/lychee/issues/2000

### DIFF
--- a/lychee-lib/src/types/base_info.rs
+++ b/lychee-lib/src/types/base_info.rs
@@ -42,21 +42,18 @@ pub enum BaseInfo {
     /// but you cannot jump to the "root".
     NoRoot(Url),
 
-    /// A full base which can resolve all kinds of relative links. `file:` URLs
-    /// with a known root directory fall into this case, as do all non-`file:`
-    /// URLs which *can be a base*.
+    /// A full base made up of `origin` and `path`. This can resolve
+    /// all kinds of relative links.
     ///
-    /// The `origin` and `path` fields are most significant for `file:` URLs.
-    /// For these, `origin` represents the root used for root-relative links,
-    /// and `path` is the subpath to the current input source within `origin`.
-    /// Keeping `path` separate allows locally-relative links to be resolved
-    /// relative to `origin.join(path)`.
+    /// All non-`file:` URLs which *can be a base* fall into this case. For these,
+    /// `origin` and `path` are obtained by dividing the source URL into its
+    /// origin and path. When joined, `${origin}/${path}` should be equivalent
+    /// to the source's original URL.
     ///
-    /// For non-`file:` URLs, the distinction between `origin` and `path` is
-    /// less important, as the "root" is always taken to be the natural root
-    /// of the origin website domain. For these URLs, all kinds of relative
-    /// links are resolved relative to `origin.join(path)` using the normal
-    /// rules.
+    /// This also represents `file:` URLs with a known root. The `origin` field
+    /// records the `file:` URL which will be used to resolve root-relative links.
+    /// The `path` field is the subpath to a particular input source within the
+    /// root. This is retained to resolve locally-relative links.
     ///
     /// In all cases, the fields should satisfy `origin.join(path) == input_source_url`
     /// where `input_source_url` is the URL of the originating input source.


lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// to resolve root-relative links (i.e., it's the root dir). The `path`
+    /// field is the subpath to a particular file within the root dir. This
+    /// is retained to resolve locally-relative links.
+    Full(Url, String),
+}
+
+impl BaseInfo {
+    /// Constructs [`BaseInfo::None`].
+    #[must_use]
+    pub const fn no_info() -> Self {
```
thomas-zahner: Maybe? Info is already part of `BaseInfo`

```suggestion
    pub const fn none() -> Self {
```

lychee-lib/src/checker/wikilink/resolver.rs
```diff
@@ -12,21 +12,22 @@ pub(crate) struct WikilinkResolver {
...
-                ))?,
-            },
+            base => base,
+        };
+        let base = match base.to_file_path() {
+            Some(p) => p,
+            None => Err(ErrorKind::WikilinkInvalidBase(
+                "Base cannot be remote".to_string(),
+            ))?,
         };
```
thomas-zahner: Could be simplified

```suggestion
        let base = base.to_file_path().ok_or(ErrorKind::WikilinkInvalidBase(
            "Base cannot be remote".to_string(),
        ))?;
```

lychee-lib/src/checker/wikilink/resolver.rs
```diff
@@ -12,21 +12,22 @@ pub(crate) struct WikilinkResolver {
...
 
         Ok(Self {
```
thomas-zahner, line 30: The base clone below is now unnecessary. Maybe that was the case already on master?

So instead of

`checker: WikilinkIndex::new(base.clone()),`

`checker: WikilinkIndex::new(base),`

lychee-lib/src/types/input/source.rs
```diff
@@ -158,6 +159,31 @@ pub enum ResolvedInputSource {
     String(Cow<'static, str>),
 }
 
+impl ResolvedInputSource {
+    /// Converts an [`ResolvedInputSource::RemoteUrl`] or
+    /// [`ResolvedInputSource::FsPath`] to a [`Url`] pointing to the source.
+    ///
+    /// The outer result indicates whether the operation succeeded.
```
thomas-zahner: I think we can remove this single line. There is only one result, no inner and outer, and if `Result` was used to indicate something else it wouldn't be very idiomatic code.

lychee-lib/src/utils/url.rs
```diff
@@ -1,77 +1,168 @@
...
+
 use linkify::LinkFinder;
+use url::ParseError;
 
-static LINK_FINDER: LazyLock<LinkFinder> = LazyLock::new(LinkFinder::new);
+/// Returns whether the text represents a relative link that is
+/// relative to the domain root. Textually, it looks like `/this`.
+pub(crate) fn is_root_relative(text: &str) -> bool {
+    let text = text.trim_ascii_start();
+    text.starts_with('/') && !text.starts_with("//")
```
thomas-zahner: What is `text.starts_with("//")` covering? It has to do with `file://`, right? Might be worth an explanation. You could also extract it into a variable with a descriptive name, like `is_absolute_path` if that's what it is about.

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,440 @@
...
+
+        Ok(url)
+    }
+
+    /// Parses the given URL text into a fully-qualified URL, including
+    /// resolving relative links if supported by the current [`BaseInfo`]
+    /// and applying the given root-dir if necessary.
+    ///
+    /// The root-dir is applied if the current `BaseInfo` is [`BaseInfo::None`]
+    /// or has a `file:` URL and and if the given text is a root-relative link.
```
thomas-zahner:
```suggestion
    /// or has a `file:` URL and if the given text is a root-relative link.
```

lychee-lib/src/utils/url.rs
```diff
@@ -1,77 +1,168 @@
...
-        );
-
-        assert_eq!(
-            remove_get_params_and_separate_fragment("https://example.com/index.html#anchor"),
-            ("https://example.com/index.html", Some("anchor"))
-        );
-        assert_eq!(
-            remove_get_params_and_separate_fragment(
-                "https://example.com/index.html?foo=bar#anchor"
+    fn test_join_rooted() {
```
thomas-zahner: Would be a good candidate for rstest, so you don't need to write a loop and we get good error messages. E.g. see `test_empty_wikilinks_are_detected`

Same with the below test.

lychee-lib/src/utils/url.rs
```diff
@@ -1,77 +1,168 @@
...
+            ("https://a.com/x?q#anchor", vec![""], "https://a.com/x?q"),
+            ("https://a.com/x#anchor", vec!["?x"], "https://a.com/x?x"),
+            // scheme relative link - can traverse outside of root
+            ("file:///root/", vec!["///new-root"], "file:///new-root"),
+            ("file:///root/", vec!["//a.com/boop"], "file://a.com/boop"),
+            ("https://root/", vec!["//a.com/boop"], "https://a.com/boop"),
+        ];
+
+        for (base, subpaths, expected) in test_urls_and_expected {
+            println!("base={base}, subpaths={subpaths:?}, expected={expected}");
```
thomas-zahner: Then the println could also be removed

lychee-lib/src/utils/url.rs
```diff
@@ -1,77 +1,168 @@
...
+/// parsed as URLs---are not parsed as URLs.
+///
+/// # Errors
+///
+/// Returns an [`Err`] if the given text is not a valid URL, or if the given text
+/// *could* be interpreted as a filesystem path. The string is returned within
+/// the error to allow for easier subsequent processing.
+pub(crate) fn parse_url_or_path(input: &str) -> Result<Url, &str> {
+    match Url::parse(input) {
+        Ok(url) if cfg!(windows) && url.scheme().len() == 1 => Err(input),
```
thomas-zahner: Hmm, not sure how bullet proof this is.. But I guess it's fine and makes sense. Not really criticising it, just thinking out loud. I don't have a better idea.

lychee-lib/src/utils/url.rs
```diff
@@ -1,77 +1,168 @@
...
+    /// to the `file:` base URL.
+    ///
+    /// Other relative links and links with non-`file:` bases are joined
+    /// normally, matching the behaviour of [`Url::join`].
+    fn join_rooted(&self, subpaths: &[&str]) -> Result<Url, ParseError>;
+}
+
+impl ReqwestUrlExt for Url {
+    fn join_rooted(&self, subpaths: &[&str]) -> Result<Url, ParseError> {
+        let mut url = Cow::Borrowed(self);
```
thomas-zahner: I admit that I'm really inexperienced with using `Cow`s. Is there a point in using it here? I guess the only difference occurs when subpaths is of length zero? Or is it that we safe one initial clone? (`let mut url = self.clone();`)

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,440 @@
...
+    /// A base which cannot resolve root-relative links. This is for
+    /// `file:` URLs where the root directory is not known. As such, you can
+    /// traverse relative to the current URL (by traversing the filesystem),
+    /// but you cannot jump to the "root".
+    NoRoot(Url),
+
+    /// A full base made up of `origin` and `path`. This can resolve
+    /// all kinds of relative links.
+    ///
+    /// All fully-qualified non-`file:` URLs fall into this case. For these,
```
thomas-zahner: `fully-qualified non-file: URLs` to me sounds like this would also include `mailto:`, `tel:` etc. but that might not make sense, right? Or are they not "fully qualified"?

What would happen to these schemas? Are they currently never used in the construction of `BaseInfo` as they are filtered out at an earlier stage?

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,440 @@
...
+impl BaseInfo {
+    /// Constructs [`BaseInfo::None`].
+    #[must_use]
+    pub const fn no_info() -> Self {
+        Self::None
+    }
+
+    /// Constructs [`BaseInfo::Full`] with the given fields.
+    #[must_use]
+    pub const fn full_info(origin: Url, path: String) -> Self {
```
thomas-zahner: And when applying the above suggestion
```suggestion
    pub const fn full(origin: Url, path: String) -> Self {
```

lychee-lib/src/utils/request.rs
```diff
@@ -118,17 +65,28 @@ fn create_uri_from_file_path(
...
+        Ok(None) => BaseInfo::no_info(),
+        Ok(Some(url)) => BaseInfo::from_source_url(&url),
+        Err(e) => {
+            // TODO: GetInputContent is not quite the right error.
+            return vec![Err(RequestError::GetInputContent(source.clone().into(), e))];
+        }
+    };
+
+    // TODO: avoid use_fs_root_as_origin once base-url semantics are clarified
+    let fallback_base = fallback_base.use_fs_root_as_origin();
```
thomas-zahner, line 106: Hmm, is this an ongoing discussion on an issue? Or is this the default behaviour of master?

lychee-bin/tests/cli.rs
```diff
@@ -448,12 +448,33 @@ mod cli {
...
             .assert()
             .success()
-            .stdout(contains("3 Total"))
-            .stdout(contains("3 OK"));
+            .stdout(contains("5 Total"))
+            .stdout(contains("5 OK"));
+    }
+
+    #[test]
+    fn test_resolve_paths_from_root_dir_and_local_base_url() {
```
thomas-zahner, line 470: Great test which acts as useful documentation :+1:

lychee-bin/tests/cli.rs
```diff
@@ -3333,4 +3354,103 @@ The config file should contain every possible key for documentation purposes."
...
+        let dir = tempdir()?;
+
+        // without root-dir, fails to resolve in all cases
+        cargo_bin_cmd!()
+            .arg("-")
+            .arg("--default-extension=md")
+            .write_stdin("[a](a)")
+            .assert()
+            .failure()
+            .stdout(contains("relative URL without a base"));
```
thomas-zahner: I think it could actually be helpful and make the test more comprehensible when asserting a bigger part of the error message. If we update the error/detail texts it would also force us to think more about the end user experience. And I really think the error message is great.

```suggestion
            .stdout(contains("Cannot parse `a` into a URL: relative URL without a base: This relative link was found inside an input source that has no base location"));
```

lychee-bin/tests/cli.rs
```diff
@@ -3333,4 +3354,103 @@ The config file should contain every possible key for documentation purposes."
...
+
+        cargo_bin_cmd!()
+            .arg("-")
+            .arg("--default-extension=md")
+            .write_stdin("[a](/a)")
+            .assert()
+            .failure()
+            .stdout(contains("Cannot resolve root-relative link"));
```
```suggestion
            .stdout(contains("Cannot resolve root-relative link '/a': To resolve root-relative links in local files, provide a root dir"));
```

lychee-bin/tests/cli.rs
```diff
@@ -3333,4 +3354,103 @@ The config file should contain every possible key for documentation purposes."
...
+        // there is no current base URL
+        cargo_bin_cmd!()
+            .arg("-")
+            .arg("--root-dir")
+            .arg(dir.path())
+            .arg("--default-extension=md")
+            .write_stdin("[a](a)")
+            .assert()
+            .failure()
+            .stdout(contains("relative URL without a base"));
```
```suggestion
            .stdout(contains("Cannot parse `a` into a URL: relative URL without a base: This relative link was found inside an input source that has no base location"));
```

lychee-bin/tests/cli.rs
```diff
@@ -3333,4 +3354,103 @@ The config file should contain every possible key for documentation purposes."
...
+
+    #[test]
+    fn test_base_url_applies_to_local() {
+        cargo_bin_cmd!()
+            .arg("--base-url=https://lychee.cli.rs/")
+            .arg(fixtures_path!().join("resolve_paths/index.html"))
+            .assert()
+            .failure()
+            .stdout(contains("https://lychee.cli.rs/another%20page"))
+            .stdout(contains("https://lychee.cli.rs/about"));
```
thomas-zahner: Maybe it's not entirely unlikely that this exact page is created in the future :sweat_smile: but might be quite hypothetical

katrinafyi: No yeah the thought crossed my mind too. I've removed that one :)

lychee-lib/src/checker/wikilink/resolver.rs
katrinafyi: done

lychee-lib/src/utils/request.rs
katrinafyi: I've updated the comment. The code atm is to match master, but I think that the current base-url is a bit weird. I also opened https://github.com/lycheeverse/lychee/issues/2009

lychee-lib/src/utils/url.rs
katrinafyi: It's making sure it's not a [scheme-relative URL](https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Web_mechanics/What_is_a_URL#absolute_urls_vs._relative_urls). I've updated the comments and extracted it into `is_scheme_relative_link`.

lychee-lib/src/utils/url.rs
katrinafyi: Yeah it avoids the initial clone for non-empty subpaths. Empty subpaths still need a clone because we return owned Url. It's a micro-optimisation for sure but it's not too burdensome, I think.

lychee-lib/src/utils/url.rs
katrinafyi: Yeah idk. I've added some tests to show how it works, but I'm also not sure what is best.

The idea is borrowed from one of my comments in https://github.com/lycheeverse/lychee/pull/1837 ([commit](https://github.com/lycheeverse/lychee/pull/1837/changes/c8fb0a1987f88a9ead80095d6fd021edd5569e8b)).

lychee-lib/src/utils/url.rs
katrinafyi: Thanks!! I didn't know about that and it's is really cool! I've made the change but kept the print because the error message only shows the expected value vs the actual value, but not the value of the arguments like subpaths. Is there an option to print all the arguments, or is this okay?

lychee-lib/src/types/base_info.rs
katrinafyi: Yeah this is a really good question. The "fully qualified" thing is a bit unclear. Those schemes are called "cannot be a base" by the URL library and they do not fall into the Full case. I've updated the comment.

As for what happens to them, it depends. They get converted to None by from_source_url, which is used when building a BaseInfo from an input source. However, the --base-url argument is parsed using from_base_url which is more strict. That will return an error if you pass one of those schemes to --base-url.

lychee-lib/src/utils/url.rs
thomas-zahner: Ok makes sense. It's slightly harder to read (for me personally) but it's definitely fine.

lychee-lib/src/utils/url.rs
thomas-zahner: Cool :+1: 

I think the better approach is to move the message into the 3rd `assert_eq` parameter, as in `assert_eq!(a, b, "we are testing addition with {} and {}", a, b);` this wouldn't clutter stdout.

Also, I think it would be slightly better to not name every single test case but use `#[case(...)]`

lychee-lib/src/utils/url.rs
```diff
@@ -1,77 +1,192 @@
...
+    #[case::ok3("http://a.com", Ok("http://a.com/"))]
+    // path-looking things
+    #[case::err1("", Err(""))]
+    #[case::err2(".", Err("."))]
+    #[case::err3("C:", Err("C:"))]
+    #[case::err4("/unix", Err("/unix"))]
+    #[case::err5("C:/a", Err("C:/a"))]
+    #[case::err6(r"C:\a\b", Err(r"C:\a\b"))]
+    #[case::err7("**/*.md", Err("**/*.md"))]
+    #[case::err8("something", Err("something"))]
```
thomas-zahner: And same idea with the other tests using `case`. Better comments (my suggested comment could probably be improved) to remove the verbose test case names.

```suggestion
    #[rstest]
    // OK URLs
    #[case("tel:1", Ok("tel:1"))]
    #[case("file:///a", Ok("file:///a"))]
    #[case("http://a.com", Ok("http://a.com/"))]
    // Invalid URLs
    #[case("", Err(""))]
    #[case(".", Err("."))]
    #[case("C:", Err("C:"))]
    #[case("/unix", Err("/unix"))]
    #[case("C:/a", Err("C:/a"))]
    #[case(r"C:\a\b", Err(r"C:\a\b"))]
    #[case("**/*.md", Err("**/*.md"))]
    #[case("something", Err("something"))]
```

lychee-lib/src/utils/url.rs
thomas-zahner: Cool, great documentation :+1:

lychee-lib/src/utils/url.rs
katrinafyi: Hmm I am sensitive to readability worries. This is what it looks like w/o using Cow but keeping the micro-optimisation: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ee28c71d18b6df558302524ba1ecedc4

Which is interesting because it reveals an alternative version of join_rooted which I was considering, where the root-dir is explicitly given rather than being inferred from the base. Maybe we could just use the `join_one` function instead of `join_rooted`? It would avoid the need for the list of subpaths.

lychee-lib/src/utils/url.rs
katrinafyi: Both changes done. Originally, the named test cases were to make it easier to go from an assertion failure to the test case line, but the assertion message should be clear enough.

lychee-lib/src/utils/url.rs
thomas-zahner: Oh awesome. Yes if `join_one` can fully replace `join_rooted` I would prefer that. Seems more explicit to me, three specific arguments instead of a more generic list/slice.

lychee-lib/src/utils/request.rs
thomas-zahner: Awesome, thank you!

lychee-lib/src/utils/url.rs
katrinafyi: Idk, I went in circles about this.

It's nice in theory, but the additional argument means you have to think about more permutations which shouldn't ever need to happen. Like with `base.join_rooted(path, root)`, what happens when `root` is file but `base` is HTTP, or vice-versa. Does it ever make sense to send root links to somewhere completely unrelated, like `"https://google.com".join_rooted("/", "https://yahoo.com")`? I don't think it does, and I don't like having an API that lets users ask that question.

In BaseInfo, you end up having to do things like
```rust
Self::Full(origin, subpath) => origin
  .join_rooted(subpath, &origin)
  .and_then(|x| x.join_rooted(text, &origin))
```
because the design of BaseInfo is (intentionally) such that the root is encoded "within" the origin. This is exactly to avoid asking weird questions like the last paragraph.

Anyway, I think the best way would be if there was a "join_relative_links" function, then we could use this to join multiple subpaths into one instead of passing as a slice.
```rust
Self::Full(origin, subpath) => origin.join_rooted(join_relative_links(subpath, text))
```
`join_relative_links` just has to join two relative links into one, e.g. `join_relative_links("/a/", "b") == "/a/b"`. The Rust crate doesn't have this, but you could port it from another library's crate, maybe.

katrinafyi: I've had a big think and just deleted the `join_rooted` function. See https://github.com/lycheeverse/lychee/commit/31166e57ee80c7958213996ca079e8c1ed72f068 :))

thomas-zahner: Awesome this seems so much cleaner!

lychee-bin/tests/cli.rs
```diff
@@ -458,12 +458,33 @@ mod cli {
...
+            .arg(dir.join("resolve_paths").join("index2.html"))
+            .env_clear()
+            .assert()
+            .success()
+            .stdout(contains("file:///base/root"))
+            .stdout(contains("file:///base/root/about"))
+            .stdout(contains("file:///base/resolve_paths/index.html"))
+            .stdout(contains("file:///base/root/another%20page#y"))
+            .stdout(contains("file:///base/resolve_paths/same%20folder.html#x"));
     }
```
mre, line 488: If both,  `--base-url` + `--root-dir` is set, who should win for root-relative links?

AFAIK, that's the case where the it silently ignores `root-dir` (because the scheme isn't `file:`), and a future refactor could accidentally change that.

Should we document that by adding another test?

```rust
#[test]
fn test_root_relative_with_remote_base_url_and_root_dir() {
    // When both are set and base-url is remote, root-relative links
    // should resolve against the remote base, not the local root-dir.
    cargo_bin_cmd!()
        .arg("-")
        .arg("--dump")
        .arg("--base-url=https://example.com/docs/")
        .arg("--root-dir=/tmp")
        .arg("--default-extension=md")
        .write_stdin("[a](/page)")
        .assert()
        .success()
        .stdout(contains("https://example.com/page"));
}
```

Also, I wonder if we already have a warning for that? 🤔

lychee-lib/src/types/input/source.rs
```diff
@@ -158,6 +159,29 @@ pub enum ResolvedInputSource {
     String(Cow<'static, str>),
 }
 
+impl ResolvedInputSource {
+    /// Converts a [`ResolvedInputSource::RemoteUrl`] or
+    /// [`ResolvedInputSource::FsPath`] to a [`Url`] pointing to the source.
+    ///
+    /// For other variants (i.e., those without a URL), `Ok(None)` is returned.
```
mre: > `Ok(None)` is returned.

I wonder about the semantics of that. Nothing wrong with using `Option::None`, but it might be hard to understand the meaning outside of that method. What do you think about introducing a custom enum, which encapsulates the output?

```rust
#[derive(Debug)]
enum ResolverOutput {
  Resolved(Url),
  NotResolved,
}
```

It's essentially the same, just closer to what we are computing here, semantically.

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,589 @@
...
+use std::borrow::Cow;
+use std::path::{Path, PathBuf};
+use url::ParseError;
+
+use crate::ErrorKind;
+use crate::Uri;
+use crate::utils;
+use crate::utils::url::is_root_relative_link;
+
+/// Information used for resolving relative URLs within a particular
```
mre, line 15: As always, thanks so much for the amazingly written docs. 👍

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,589 @@
...
+    /// Propagates errors from [`BaseInfo::parse_url_text`].
+    pub fn parse_url_text_with_root_dir(
+        &self,
+        text: &str,
+        root_dir: Option<&Url>,
+    ) -> Result<Url, ErrorKind> {
+        // HACK: if root-dir is specified, apply it by fudging around with
+        // file:// URLs. eventually, someone up the stack should construct
+        // the BaseInfo::Full for root-dir and this function should be deleted.
+
```
mre, line 332: We could log a debug message here.

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,589 @@
...
+            base.parse_url_text_with_root_dir("/a", Some(&root_dir)),
+            Ok(Url::parse("file:///root/a").unwrap())
+        );
+
+        // root-dir can be traversed out of
+        assert_eq!(
+            base.parse_url_text_with_root_dir("/../../", Some(&root_dir)),
+            Ok(Url::parse("file:///").unwrap())
+        );
+    }
```
mre, line 518: I could think of a few more tests. Mostly just as a safeguard for future refactoring. Apologies if some tests are already covered by some logic above.

```rust
#[test]
fn test_none_rejects_relative_but_accepts_absolute() {
    // Ensures BaseInfo::None doesn't silently swallow relative links
    let none = BaseInfo::none();
    // Absolute URLs still work
    assert!(none.parse_url_text("https://a.com").is_ok());
    // Relative links fail
    assert!(none.parse_url_text("relative").is_err());
    assert!(none.parse_url_text("/root-relative").is_err());
}

#[test]
fn test_no_root_rejects_root_relative() {
    // A file:// source without --root-dir can resolve siblings but not root-relative links
    let no_root = BaseInfo::try_from("file:///some/path/").unwrap();
    assert_eq!(
        no_root.parse_url_text("sibling.html").unwrap(),
        Url::parse("file:///some/path/sibling.html").unwrap()
    );
    assert!(no_root.parse_url_text("/root-relative").is_err());
}

#[test]
fn test_or_fallback_prefers_more_capable_variant() {
    // Pins the fallback priority that drives base selection in the collector
    let none = BaseInfo::none();
    let no_root = BaseInfo::NoRoot(Url::parse("file:///a/").unwrap());
    let full = BaseInfo::full(Url::parse("https://a.com/").unwrap(), String::new());

    assert_eq!(none.or_fallback(&full), &full);
    assert_eq!(full.or_fallback(&none), &full);
    assert_eq!(none.or_fallback(&no_root), &no_root);
    assert_eq!(no_root.or_fallback(&full), &full);
    assert_eq!(none.or_fallback(&none), &none);
}

#[test]
fn test_try_from_rejects_invalid_bases() {
    // Prevent data: URLs and relative paths from silently becoming a base in the future
    assert!(BaseInfo::try_from("data:text/plain,hello").is_err());
    assert!(BaseInfo::try_from("relative/path").is_err());
    assert!(BaseInfo::from_path(&PathBuf::from("relative")).is_err());
}
```

lychee-lib/src/utils/request.rs
```diff
@@ -48,65 +47,23 @@ fn create_request(
...
-    };
-    let Ok(constructed_url) =
-        resolve_and_create_url(file_path, &target_path, ignore_absolute_local_links)
-    else {
-        return Err(ErrorKind::InvalidPathToUri(target_path));
-    };
-    Ok(Uri {
-        url: constructed_url,
-    })
+    // BACKWARDS COMPAT: delete trailing slash for file urls
```
mre, line 59: Should we make this a log message?

lychee-lib/src/utils/request.rs
```diff
@@ -118,17 +75,30 @@ fn create_uri_from_file_path(
...
+    fallback_base: &BaseInfo,
     extractor: Option<&BasicAuthExtractor>,
 ) -> Vec<Result<Request, RequestError>> {
-    let base = base.cloned().or_else(|| Base::from_source(source));
+    let source_base = match source.to_url() {
+        Ok(None) => BaseInfo::none(),
+        Ok(Some(url)) => BaseInfo::from_source_url(&url),
+        Err(e) => {
+            // TODO: GetInputContent is not quite the right error.
+            return vec![Err(RequestError::GetInputContent(source.clone().into(), e))];
```
mre: We can introduce a new error variant? E.g. `RequestError::InvalidBaseInfo`.

fixtures/resolve_paths/index2.html
```diff
@@ -0,0 +1,27 @@
+<html>
+    <head>
+        <title>Index</title>
+    </head>
```
katrinafyi: fdsa

lychee-bin/tests/cli.rs
katrinafyi: Added the test, thanks! There's no warning atm I don't think. I had https://github.com/lycheeverse/lychee/pull/2001 but I don't think it's needed anymore.

lychee-lib/src/types/input/source.rs
katrinafyi: I've changed the function to return a BaseInfo, so this case can be a `Ok(BaseInfo::None)` instead which is much more meaningful I think.

lychee-lib/src/types/base_info.rs
katrinafyi: I love docs!!!

lychee-lib/src/types/base_info.rs
katrinafyi: Mmm maybe, but it would be logged on every root-relative link in local files. I think it's a bit much and not very useful to us or the user.

lychee-lib/src/types/base_info.rs
katrinafyi: I've added all of these tests, thanks for the really detailed tests! It's good to test the none cases, they were definitely missing from my own tests.

lychee-lib/src/utils/request.rs
katrinafyi: Added a warning :)
```
            warn!(
                "Removing trailing slash from file URL: {url}. {} {}",
                "This lets the URL match both files and folders.",
                "In future, a file URL ending in / may be required to be a directory."
            );
```

lychee-lib/src/utils/request.rs
katrinafyi: added InputSourceError

lychee-lib/src/types/base_info.rs
mre: I know. 😆 That's great!

lychee-lib/src/types/base_info.rs
mre: Oh yeah, true. Could maybe be a case for #2021 in the future. Not sure. Let's keep it simple for now.

lychee-lib/src/types/base_info.rs
mre: Cool!

lychee-lib/src/utils/request.rs
mre: Yeah, even better. Thanks!

lychee-lib/src/types/input/source.rs
mre: Indeed. I like it. Thanks.

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,633 @@
...
+    /// to the source's original URL.
+    ///
+    /// This also represents `file:` URLs with a known root. The `origin` field
+    /// records the `file:` URL which will be used to resolve root-relative links.
+    /// The `path` field is the subpath to a particular input source within the
+    /// root. This is retained to resolve locally-relative links.
+    ///
+    /// In all cases, the `path` field should be a (possibly-empty) locally- or
+    /// root-relative link and should not be a full URL or a scheme-relative link.
+    Full(Url, String),
```
thomas-zahner: Oh wow, I feel like the docs are much better since I last checked :)

Wouldn't it make sense to give it same names in the code too? I was searching for the above mentioned fields for a few seconds..

```suggestion
    Full{ origin: Url, path: String },
```

lychee-lib/src/types/base_info.rs
```diff
@@ -0,0 +1,633 @@
...
+impl TryFrom<String> for BaseInfo {
+    type Error = ErrorKind;
+    fn try_from(value: String) -> Result<Self, ErrorKind> {
+        BaseInfo::try_from(value.as_ref())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
```
thomas-zahner:
```suggestion
```

lychee-lib/src/types/base_info.rs
katrinafyi: Yeah true, I'll do that. Also, I'm still not 100% satisfied with this doc comment so I might tweak it

katrinafyi: Done~ I ended up changing the comment back. I think this version has a good balance between low and high level details.

thomas-zahner: Sure, I find the docs really good :+1:

thomas-zahner:
@katrinafyi Thanks again for the PR :rocket:

katrinafyi:
Thanks for your review too, as always! It is very satisfying to fix all of those issues.

Now we have to wait and see if we broke anything for users :")

Issue #1959, opened 2025-12-19
Config merging mishandles default values


katrinafyi:
generally, any option specified on the command line should have precedence over options in the toml file (which have precedence over the default values).

however, the fold_in! macro logic for merging two config structs is slightly flawed. this can cause an option specified at the command line to be deprioritised in favour of a toml config value.

the macro assumes that if the CLI has a default value for an argument, then that should be overridden by the .toml's value (if any). but that's not necessarily correct, because the user may have manually specified a CLI argument that happens to be equal to the default. in these cases, the CLI value should still take precedence over the toml file.

this bug will become more important if lychee wants to do more config file things. in particular, if lychee wants to process multiple config files with precedence between them.

mentioned in https://github.com/lycheeverse/lychee/pull/1848

repro: 
```console
$ echo 'mode = "task"' > a.toml
$ echo 'https://mock.httpstatus.io/404' | cargo run -- - -v --mode color -c a.toml
     [404] https://mock.httpstatus.io/404 | Rejected status code (this depends on your "accept" configuration): Not Found

Issues found in 1 input. Find details below.

[stdin]:
- [ ] [404] https://mock.httpstatus.io/404 | Rejected status code (this depends on your "accept" configuration): Not Found
```
note that the `mode = task` from the config is used even though we specified `mode = color` on the command line. this is because color coincides with the default value. if you specify a different `--mode` argument (e.g. emoji), then the CLI value will be correctly used.

thomas-zahner:
Ah that's a funny bug. 😅 Thank you for reporting.
Speaking of merging configs we might also do this for #1298. So the bug might also affect the work on that issue.

PR #2007, opened 2026-01-26
Secrets

### LINKED ISSUE #1952: RFE: support a global config file
Would be nice to see lychee add support for a global (per user) config file. A good location for it would be `~/.config/lychee.toml`.

Based on my short experience with lychee, I find myself hitting a lot of rate limits. I also do one-off runs against various projects, and having to configure lychee for each of them would be a chore. On the other hand remembering the command line options needed is also one, and e.g. for GITHUB_TOKEN, I don't want that in my global env as it might end up being used inadvertently by something.

So a per user config would nicely fit the bill. It could for example be used as the basis, merged with the per project one if found, with the latter having precedence.

A shell alias with the desired options written in it serves as a workaround meanwhile.

thomas-zahner:
Allow specifying an additional config files for managing secrets by manually implementing `Config` merging.

Fixes https://github.com/lycheeverse/lychee/issues/1298
Fixes https://github.com/lycheeverse/lychee/issues/1959
Fixes #1952

# Questions

@mre The only problem I see with this PR right now is that we loose the benefit of clap default which is the automatically appended text to the help message & man pages which I've now appended manually to the comments. Ideally we'd do some macro magic so that we don't have to manually add these defaults in such an error prone way, or alternatively have a test which verifies that `[default: X]` matches the actual values of `Config::default().my_option()`. But both ideas aren't easy to implement. Do you see any alternatives or do you think merging it in the current state is fine?

### DIFF
--- a/lychee-bin/src/options.rs
+++ b/lychee-bin/src/options.rs
@@ -952,6 +952,7 @@ impl Config {
                     // `false` in the merge chain, as there is no way to distinguish
                     // between "default" `false` and user-provided `false`.
                     // We would have to use `Option<bool>` in order to do that.
+                    // See: https://github.com/lycheeverse/lychee/issues/2051
                     $( $bool: self.$bool || other.$bool, )*
                 }
             };
@@ -1071,6 +1072,8 @@ This convention also simplifies our default value testing."
         let contents = read_this_source_file();
 
         let default_value_annotation = Regex::new(r"\s*\[default: (?<value>.*)\]").unwrap();
+        // Matches last line of rustdoc comment, then skips a line (expected to be `#[arg(...)]`),
+        // then matches a *private* field of type Option.
         let default_field =
             Regex::new(r"(?m)^\s+///(?<comment>.*)\n.*\n\s+(?<ident>\w+):\s*Option<.*>,?$")
                 .unwrap();


katrinafyi:
About the clap defaults, I *think* it might be possible to keep using them if we use [`value_source`](https://docs.rs/clap/latest/clap/struct.ArgMatches.html#method.value_source) to detect if it was a default or CLI-provided argument. On the TOML side, you'd have to something like parse into a toml::Table, record the keys, then parse the table into a lychee Config.

But if all this works (and I don't know if it will, it's all just a theory from reading docs), then the `merge` function would take two configs and lists of the user-defined keys for each, then it merge and prioritise using that information.

Idk how much work this would be, but it would be really nice to keep using the clap default system.

thomas-zahner:
@katrinafyi Oh wow, thank you so much for the suggestion. I really missed that, or rather never consulted the docs properly enough :sweat_smile: I will definitely explore this path, thank you for pointing it out.

thomas-zahner:
So I've updated this PR with 4d4fc894. We can now specify `--config` multiple times, which is much cleaner than the confusing and overly-specific `--secrets` option.

# Clap defaults

Clap defaults are removed in the current state of the PR. This is because leaving them leads to incorrect behaviour. We want clap arguments to take precedence over config values, otherwise UX is quite confusing IMO. It would be much easier if this wasn't the case, but I think it really is unexpected to have CLI flags which are overwritten by values in config files.

```mermaid
graph LR;
    Config-A-->Config-B;
    Config-B-->Clap;
```

There can be any amount of configs, this is just an example with two configs. `Config-A` might be the main config, tracked with git. `Config-B` might contain only a few sensitive values (e.g. header values), being specifically excluded from git tracking.

If we decided to leave the clap default values, then if `Config-A` or `Config-B` specifies e.g. `format: Some(Json)` then `Clap` would overwrite it with the default value of `Some(Compact)`. Previously we just ignored any values coming from `Clap` if they were equal to the default value, meaning if a user uses `--format compact` it was ignored. This is #1959.

I've experimented with macros in https://github.com/thomas-zahner/lychee/commit/605fe31f1e1d54f1506533008a9261a7ce8d67eb where I try to auto generate the "getter" methods and assert that the right default value is mentioned in the comment. But unfortunately it's not possible to evaluate const expressions in macros so that it's impossible to assert the default value is contained in the comment when it is a const literal. Also this seems to be quite overkill.

Then I've tried your approach @katrinafyi in https://github.com/thomas-zahner/lychee/commit/dbd36cebcc7362d618514859db259c1a9bd952f8#diff-6168b1b6485695baf19bdea9ca3993b2c1f705821ab33f53775ad279ce2f6badR146. But this does not seem to work well. I see `None`, even though I specified the argument via CLI flag. Also this approach seems to be very error prone and tedious, as the arguments are referred to by string and we get runtime panics if we try to obtain infos for IDs which don't correspond to any argument.

I'm open for ideas. Currently, I feel like the state of this PR is pretty much the most realistic version. Ideally I will write some additional tests, which might be able to assert that manually typed out default values (e.g. `/// [default: compact]`) are actually accurate. Not sure how easy this is.

katrinafyi:
@thomas-zahner with value source, was it particular flags which caused that problem? I was just testing it and it seems to work. I don't know what causes None vs Some(DefaultValue), but Some(CommandLine) does seem to be returned correctly as far as I can tell.

But maybe I'm just rewalking the path you already tried. The API could definitely be safer and less stringy..

thomas-zahner:
@katrinafyi Hmm, it basically worked for not a single argument in my test. See https://github.com/thomas-zahner/lychee/tree/clap-value-source-v2

Could you share your code if you got it working?

katrinafyi:
Coincidentally, I just discovered the same issue :laughing: It's really confusing. `value_source` stops working (and always returns None) after `from_arg_matches_mut` is used. I figure it mutates the arg matches to lose the value source information?? It works if you call `value_source` before parsing the ArgMatches into LycheeOptions.

My code is at https://github.com/lycheeverse/lychee/compare/master..rina-forks:lychee:value-source. It does the merging by lowering to TOML and then it can use the TOML map to represent presence/absence. But it's a real headache with all the string keys and untyped toml AST. 

I think your approach is probably the least bad. Even if `value_source` was working, the string keys are a real bother. I also tried writing macros to make it more type-safe, but you inevitably have to deal with strings at the boundaries. https://github.com/lycheeverse/lychee/commit/ce600361858994f0c6db1a7d7c9c7db279b57493

edit: or maybe the macro is ok. the hard part is just defining `is_defined` predicates for the CLI and TOML. this could be done with helper function that maps field enum to string name. we can check that is safe by taking the image and checking subset of serde names or clap names. then we can keep merging at the type-safe level too........

thomas-zahner:
> My code is at https://github.com/lycheeverse/lychee/compare/master..rina-forks:lychee:value-source. It does the merging by lowering to TOML and then it can use the TOML map to represent presence/absence. But it's a real headache with all the string keys and untyped toml AST.

Oh wow, that looks like you put quite some work into it. But it also seems really complicated :sweat_smile: 

I've just come up with three tests in 3303cc061567f50b89aa80890fc2b08a41178c71. What do you think? It seems much simpler and pragmatic to go with such conventions. And now as we test them it shouldn't be a problem. It also has the benefit that we show even more default values to the user. (e.g. as with archive)

katrinafyi:
> But it also seems really complicated 😅

Yeah... I was following different threads at the same time so could be a little bit simpler, but it's definitely more complicated than I wanted or expected.

It's just... I really don't like all the getters to apply the defaults - it's just so Java 😭 But it's something we'll have to deal with.

Anyway, the tests you wrote look good and make sense. Let me know when it's ready and you want me to look through, I had some small comments queued.

I might or might not rebase my changes after your PR is merged to see what it's like. Or maybe not if it's too much work, we'll see..

thomas-zahner:
> It's just... I really don't like all the getters to apply the defaults - it's just so Java 😭 But it's something we'll have to deal with.

Haha yeah, I see your point. But at the same time I think it's an overall improvement, as we document more default values (accept & archive) than before and any mistakes should be prevented through the new tests. Also it's really low complexity, no macro magic.

Overall the approach is pretty good now in my opinion. If you look at other bigger CLI applications you will see funny things, how they implement CLI parsing fully on their own without any library (ripgrep which switched from clap to their own implementation) or how they have duplicated structs one for CLI args and one for the config file. ([fd](https://github.com/sharkdp/fd/tree/master/src))

The PR is now ready for review.

lychee-bin/src/options.rs
```diff
@@ -567,13 +500,13 @@ with a status code of 429, 500 and 501."
...
     #[arg(short = 'T', long)]
-    #[serde(default)]
     pub(crate) threads: Option<usize>,
 
     /// User agent
-    #[arg(short, long, default_value = DEFAULT_USER_AGENT)]
-    #[serde(default = "user_agent")]
-    pub(crate) user_agent: String,
+    ///
+    /// [default: lychee/0.22.0]
```
thomas-zahner: This should probably be updated to be `lychee/x.y.z` and the test updated as well. Otherwise the auto-generated PR by release-plz will lead to failing tests.

lychee-bin/src/main.rs
```diff
@@ -135,40 +135,23 @@ fn read_lines(file: &File) -> Result<Vec<String>> {
...
-                        default_config.display()
-                    );
-                }
-            }
+            opts.config = opts.config.merge_file(&default_config)?;
+        }
+    } else {
+        for config_file in &opts.config_files {
+            opts.config = opts.config.merge_file(config_file)?;
         }
```
katrinafyi: There's a bug here. If I put `mode = "task"` into a toml, this command still uses task mode instead of emoji:
```bash
cargo  run -- -c b.toml a.md --mode emoji   
```
(A similar issue to https://github.com/lycheeverse/lychee/issues/1959)

The relevant comments are:
```rust
    /// Try to load configuration from a file and merge into `self`.
    /// `config_file` has precedence over `self`.
    pub(crate) fn merge_file(self, config_file: &Path) -> Result<Config> {
        ...
        Ok(config.merge(self))
    }


    /// Merge `self` with another `Config` where the fields of `self` take precedence
    /// over `other`.
    pub(crate) fn merge(self, other: Config) -> Config {
```
I believe these comments are true.  In this case, the highlighted code means that (any) config file takes precedence over the initial opts.config containing the CLI options.

lychee-bin/src/options.rs
```diff
@@ -856,18 +766,19 @@ followed by the absolute link's own path."
 
     /// Output file of status report
     #[arg(short, long, value_parser)]
-    #[serde(default)]
     pub(crate) output: Option<PathBuf>,
```
katrinafyi, line 769: Why are all the `#[serde(default)]` removed? Is this no longer needed or is it implied?

lychee-bin/src/options.rs
```diff
@@ -896,34 +803,30 @@ and existing cookies will be updated."
     pub(crate) include_wikilinks: bool,
 
     /// Preprocess input files.
```
katrinafyi: Line break so short help is a bit more concise.
```suggestion
    /// Preprocess input files with the given command.
    ///
```
Atm, it looks like this in `lychee -h`:
```
  -p, --preprocess <COMMAND>
          Preprocess input files.
          For each file input, this flag causes lychee to execute `COMMAND PATH` and process
          its standard output instead of the original contents of PATH. This allows you to
          convert files that would otherwise not be understood by lychee. The preprocessor
          COMMAND is only run on input files, not on standard input or URLs.
```

lychee-bin/src/options.rs
```diff
@@ -933,126 +836,326 @@ esac"#
...
+                    // Merge chainable fields (e.g. `Vec` and `HashMap`)
+                    $( $chainable: self.$chainable.into_iter().chain(other.$chainable).collect(), )*
+                    // Use self if present, otherwise use other
+                    $( $optional: self.$optional.or(other.$optional), )*
+                    // Use `true` when self or other is `true`.
+                    // Note that this has the drawback, that a value cannot be overwritten with
+                    // `false` in the merge chain, as there is no way to distinguish
+                    // between "default" `false` and user-provided `false`.
+                    // We would have to use `Option<bool>` in order to do that.
+                    $( $bool: self.$bool || other.$bool, )*
```
katrinafyi, line 956: I like this macro, it's nicer than the old ones and chainable is really nice.

Not being able to override bool fields is unfortunate. I can imagine a user has `cache=true` in their toml but might want to use  cache=false on the command line for testing. Is it worth noting this somewhere? (Genuine question, "no" is a valid answer)

lychee-bin/src/options.rs
```diff
@@ -933,126 +836,326 @@ esac"#
...
     use super::*;
 
+    #[test]
+    fn test_no_clap_default_used() {
+        if Regex::new(r"(?ms)#\[arg\([^\]]*default_value")
+            .unwrap()
+            .is_match(&read_file_to_string())
+        {
+            panic!(
+                r"As per convention we avoid clap's default values.
```
katrinafyi: Is the intention of "As per convention" that this convention is broadly accepted amongst clap users? If you mean as a convention within this project, I'd use this instead.
```suggestion
                r"In lychee, we avoid clap's default values.
```

lychee-bin/src/options.rs
```diff
@@ -933,126 +836,326 @@ esac"#
...
+    }
+
+    #[test]
+    fn test_no_clap_long_help_used() {
+        if Regex::new(r"(?ms)#\[arg\([^\]]*long_help")
+            .unwrap()
+            .is_match(&read_file_to_string())
+        {
+            panic!(
+                r"As per convention we avoid clap's long_help.
```
katrinafyi: as above
```suggestion
                r"In lychee, we avoid clap's long_help.
```

lychee-bin/src/options.rs
```diff
@@ -933,126 +836,326 @@ esac"#
...
+    }
+
+    #[test]
+    fn test_default_values() {
+        let contents = read_file_to_string();
+
+        let default_value_annotation = Regex::new(r"\s*\[default: (?<value>.*)\]").unwrap();
+        let default_field =
+            Regex::new(r"(?m)^\s+///(?<comment>.*)\n.*\n\s+(?<ident>\w+):\s*Option<.*>,?$")
+                .unwrap();
```
katrinafyi, line 1079: Hold on, how on earth does this work 😅 I put some prints in and i see these as ident and comment:
```
extensions  [default: md,mkd,mdx,mdown,mdwn,mkdn,mkdown,markdown,html,htm,css,txt]
archive  [default: wayback]
max_redirects  [default: 5]
max_retries  [default: 3]
max_concurrency  [default: 128]
user_agent  [default: lychee/0.22.0]
accept  [default: 100..=103,200..=299]
timeout  [default: 20]
retry_wait_time  [default: 1]
mode  [default: color]
format  [default: compact]
```
This really confuses me. What causes the comment group to match only the last comment line? What guarantees that the default comment is matched into  `(?<comment>.*)` rather than the following `.*`? What causes this regex to _not_ match something like preprocess?
```
    /// ```
    #[arg(short, long, value_name = "COMMAND", verbatim_doc_comment)]
    pub(crate) preprocess: Option<Preprocessor>,
```

lychee-bin/src/options.rs
```diff
@@ -933,126 +836,326 @@ esac"#
...
+
+        let default = parse_default_options();
+
+        let mut remove = |identifier: &str| {
+            let position = default_values
+                .iter()
+                .position(|(ident, _)| *ident == identifier)
+                .unwrap_or_else(|| panic!("Option with name '{identifier}' not found. Is it a non-public field of type `Option<T>`?"));
+            default_values.remove(position)
+        };
```
katrinafyi: collect into HashMap instead? It's doing lookups and removes by key which would be easier with hashmap's entry interface. Or TreeMap if you want a stable order.

lychee-bin/src/options.rs
```diff
@@ -933,126 +836,326 @@ esac"#
...
+                (ident, default_value)
+            }).collect::<Vec<_>>();
+
+        let default = parse_default_options();
+
+        let mut remove = |identifier: &str| {
+            let position = default_values
+                .iter()
+                .position(|(ident, _)| *ident == identifier)
+                .unwrap_or_else(|| panic!("Option with name '{identifier}' not found. Is it a non-public field of type `Option<T>`?"));
```
katrinafyi: Something more detailed and direct, since this should only be seen by developers anyway.
```suggestion
                .unwrap_or_else(|| panic!("Option with name '{identifier}' was expected due to check_default_values!. Make sure it is exists as a private field of type `Option<T>`, or update the call to check_default_values!."));
```

lychee-bin/src/options.rs
```diff
@@ -1134,23 +1237,86 @@ mod tests {
...
+
+    fn read_file_to_string() -> String {
+        fs::read_to_string("./src/options.rs")
+            .expect("Unable to read this source code file to string")
+    }
+
+    fn parse_default_options() -> LycheeOptions {
+        parse_options(vec!["lychee", "-"])
+    }
+
```
katrinafyi: I think this doesn't need to be a helper function and it's clearer if it's inlined.

lychee-bin/src/options.rs
```diff
@@ -1134,23 +1237,86 @@ mod tests {
...
+                    .unwrap()
+                }
+            )])
         );
     }
+
+    fn read_file_to_string() -> String {
+        fs::read_to_string("./src/options.rs")
+            .expect("Unable to read this source code file to string")
+    }
```
katrinafyi: fn name too generic

```suggestion
    fn read_this_source_file() -> String {
        fs::read_to_string("./src/options.rs")
            .expect("Unable to read this source code file to string")
    }
```

lychee-bin/tests/cli.rs
```diff
@@ -1004,6 +1004,51 @@ mod cli {
             .stderr(contains("TOML parse error"));
     }
 
+    #[tokio::test]
+    async fn test_configs_precedence() {
```
katrinafyi, line 1008: Add test for precedence of CLI flags over any toml config.

README.md
```diff
@@ -342,11 +342,14 @@ Options:
...
           202, 203, 204, 429, and 500 as valid status codes.
-          Defaults to '100..=103,200..=299' if the user provides no value.
+
+          [default: 100..=103,200..=299]
 
       --archive <ARCHIVE>
           Specify the use of a specific web archive. Can be used in combination with `--suggest`
 
+          [default: wayback]
+
```
katrinafyi, line 352: "*Can* be used with --suggest" suggests that it has a use _without_ --suggest which I don't think it does. Maybe...
```suggestion
         Web archive to use to provide suggestions for `--suggest`.

          [default: wayback]
```

README.md
```diff
@@ -376,8 +379,11 @@ Options:
       --basic-auth <BASIC_AUTH>
           Basic authentication support. E.g. `http://example.com username:password`
 
-  -c, --config <CONFIG_FILE>
-          Configuration file to use
+  -c, --config <FILE_PATH>
+          Configuration file to use.
+          This option can be specified multiple times.
+          Multiple configs are merged into a single config.
+          Later occurrences take precedence over previous occurrences.
```
katrinafyi, line 386: I like paragraphs rather than one short sentence per line. Optional though.

```suggestion
  -c, --config <FILE_PATH>
          Configuration file to use. Can be specified multiple times.

          If given multiple times, the configs are merged and later
          occurrences take precedence over previous occurrences.
```

katrinafyi (changes requested):
It's good. verbatim_doc_comment is good actually because the old strings had weird indentation. Also, adding more `merge` functions to structs is a good idea. It's also good to delete the more odd macros like `default_function!`.

Unfortunately, there's a bug but, aside from that issue, the comments are small.

lychee-bin/src/options.rs
thomas-zahner: Much better 👍

README.md
thomas-zahner: Thanks, I really like this approach 👍

lychee-bin/src/options.rs
mre: I like this macro as well!

One more observation (and sorry for hijacking this thread; wasn't sure if I should start a new thread) is that we should future-proof this a bit by checking that we handled all config fields. This bit us in the past. 😅 

So we can put something like this at the top of the merge method:

```rust
        // Compile-time exhaustiveness check.
        // If this fails, remember to add your new field to the macro below!
        let _ = {
            let Config {
                hosts: _,
                accept: _,
                include: _,
                header: _,
                cache: _,
            } = &self;
        };
```

Note that this won't prevent us from adding a new field without merging it, but it's a bit of a "smoke test" which can serve as a reminder to update the code when we add a new field. So it's better than nothing. A better check is not possible afaict because it depends on our merging business logic.

lychee-bin/src/main.rs
thomas-zahner: Thank you for taking such a close look! See 5c14769578dea79a49c8e88aa7dc9d457c4d854e

I tried to avoid `rev()` by swapping the order in `merge_file` but forgot about the CLI case. It's now tested as well.

lychee-bin/src/options.rs
thomas-zahner: @mre Thank you! Hmm, not sure if I understand you. The macro is already forcing exhaustiveness. Try removing one field and you see that we get compilation errors.

For example removing the first field yields:

```
error[E0063]: missing field `accept` in initializer of `options::Config`
    --> lychee-bin/src/options.rs:944:17
     |
 944 |                   Config {
     |                   ^^^^^^ missing `accept`
...
 960 | /         merge!(
 961 | |             option {
 962 | |                 archive,
 963 | |                 base,
...    |
1025 | |             },
1026 | |         )
     | |_________- in this macro invocation
     |
     = note: this error originates in the macro `merge` (in Nightly builds, run with -Z macro-backtrace for more info)
```

lychee-bin/src/options.rs
thomas-zahner: That makes much more sense :+1: See 360f54cbca996cce7e870e9f606ef744136df2d7

lychee-bin/src/options.rs
thomas-zahner: > What causes the comment group to match only the last comment line? What guarantees that the default comment is matched into (?<comment>.*) rather than the following .*?

I think that's because regex is greedily matching things.

> What causes this regex to not match something like preprocess?

`\n\s+(?<ident>\w+):\s*Option<.*>,?$` is deliberately missing `pub` or `pub(crate)` between the newline and the identifier. Publicly exposed fields such as `preprocess` are intentionally skipped because they have no default values.

This regex is surely not bullet proof, but the alternative of writing a huge complicated macro would be very overkill in my opinion. If something goes wrong the test will most probably fail, which is fine.

lychee-bin/src/options.rs
thomas-zahner: @katrinafyi Thank you!

> Not being able to override bool fields is unfortunate. I can imagine a user has cache=true in their toml but might want to use cache=false on the command line for testing. Is it worth noting this somewhere? (Genuine question, "no" is a valid answer)

Good question. Actually, your specific use-case is impossible. `bool`s with clap cannot be specified to be false in this case. (it would be possible with some additional arg option) The boolean flags can only be present or not. If they are present, we override with `true`, which works. `--hidden=false` does not work, it yields `unexpected value 'false' for '--hidden' found; no more were expected`.

So it only is a problem when having multiple config files where a later one wants to override a value false. Knowing this I think its fine to keep the `bool`s and not convert them all into `Option<bool>`.

lychee-bin/src/options.rs
thomas-zahner: 360f54cbca996cce7e870e9f606ef744136df2d7

lychee-bin/src/options.rs
thomas-zahner: Yes, `Option`al fields do not have to be specified, as serde seems to be falling back on `None` automatically.

This is different from e.g.

```
    #[arg(long)]
    #[serde(default)]
    pub(crate) include_wikilinks: bool,
```

where loading a config would fail with `missing field include_wikilinks` if we removed `#[serde(default)]`.

lychee-bin/src/main.rs
katrinafyi: Thanks for looking into it and adding the test!

lychee-bin/src/options.rs
katrinafyi: Sounds good to me, thanks for clearing it up.

lychee-bin/src/options.rs
katrinafyi: Yeah fair point, I just assumed and I didn't know that you couldn't do that.

Maybe just file the bool overriding as a future improvement issue, then. It would be nice to have but it's not super critical.

lychee-bin/src/options.rs
katrinafyi: Yeah it all makes sense once explained. Just a comment recording what you've told me.
```suggestion
        // Matches last line of rustdoc comment, then skips a line (expected to be `#[arg(...)]`),
        // then matches a *private* field of type Option.
        let default_field =
            Regex::new(r"(?m)^\s+///(?<comment>.*)\n.*\n\s+(?<ident>\w+):\s*Option<.*>,?$")
                .unwrap();
```

lychee-bin/src/options.rs
mre: > The macro is already forcing exhaustiveness. 

All good then. 👍

thomas-zahner: @katrinafyi https://github.com/lycheeverse/lychee/issues/2051

thomas-zahner:
@katrinafyi Thank you for the thorough review!

katrinafyi:
A really small thing but were these two comments meant to make it in https://github.com/lycheeverse/lychee/pull/2007#discussion_r2807451729 https://github.com/lycheeverse/lychee/pull/2007#discussion_r2807453862? Or did you choose to leave them out?

thomas-zahner:
@katrinafyi Oh, that's what happens when you apply all review suggestions and then address the test failures by copying the diff blindly from the test output :smile: (I assumed there were no suggestions in the README)

See 55008797292983a268ffcc0a4d8b38d14b834d39

katrinafyi:
Aha I see, that's so fair. I should put the suggestions in the rust side next time :))

PR #2008, opened 2026-01-26
Use absolute instead of canonicalize for relative root-dir

katrinafyi:
it's a rare thing but canonicalise replaces all symbolic links with their destinations. this can unexpectedly change the root-dir location, which is a problem given relative link resolution works very much based on location, and users will be expecting that the root-dir they give is the root-dir that is used.

in this change, we have to do a separate check for file existence - canonicalize did that implicitly.

### DIFF
--- a/lychee-lib/src/collector.rs
+++ b/lychee-lib/src/collector.rs
@@ -78,7 +78,7 @@ impl Collector {
             Some(root_dir) => {
                 let root_dir_exists = root_dir.read_dir().map(|_| ());
                 let root_dir = root_dir_exists
-                    .and_then(|_| std::path::absolute(&root_dir))
+                    .and_then(|()| std::path::absolute(&root_dir))
                     .map_err(|e| ErrorKind::InvalidRootDir(root_dir, e))?;
                 Some(root_dir)
             }


lychee-lib/src/collector.rs
```diff
@@ -75,11 +75,13 @@ impl Collector {
...
-                    .canonicalize()
-                    .map_err(|e| ErrorKind::InvalidRootDir(root_dir, e))?,
-            ),
+            Some(root_dir) => {
+                let root_dir_exists = root_dir.read_dir().map(|_| ());
+                let root_dir = root_dir_exists
+                    .and_then(|()| std::path::absolute(&root_dir))
+                    .map_err(|e| ErrorKind::InvalidRootDir(root_dir, e))?;
+                Some(root_dir)
+            }
```
thomas-zahner, line 84: Very nitpicky, but I find the mapping to `()` a bit weird, though I'd also be fine with it.
If you agree we could use the following, otherwise I'm fine with your version.

```suggestion
            Some(root_dir) => Some(
                root_dir
                    .read_dir()
                    .and_then(|_| std::path::absolute(&root_dir))
                    .map_err(|e| ErrorKind::InvalidRootDir(root_dir, e))?,
            ),
```

katrinafyi: Ya it's a bit weird. I like it, though, because it makes it clear that we're ignoring the value of read_dir aside from the error. So the reader of the code doesn't think we're _actually_ reading the directory at this point.

If you chain read_dir directly into and_then, it looks like it's doing something with the result of read_dir.

thomas-zahner: True it might be more explicit and obvious

Issue #1958, opened 2025-12-17
Increase value for --max-redirects


katrinafyi:
[302 Found](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/302) redirects are treated as a final status code rather than a redirect to be followed.

```console
$ echo 'http://dx.doi.org/10.1007/978-3-540-74061-2_11'  | cargo  run - -v      
     [302] http://dx.doi.org/10.1007/978-3-540-74061-2_11 | Rejected status code (this depends on your "accept" configuration): Found
```

If you follow the suggestion and add 302 to the accept set, this just terminates the link check at the 302.

mre:
I wonder if that's an issue with lychee or reqwest.

We have this redirect policy. 
https://github.com/lycheeverse/lychee/blob/8d51c6003ded4ee5151755b5be28704aa30c1404/lychee-lib/src/client.rs#L434

Not sure if there's anything more to that. Didn't dig much deeper.

katrinafyi:
Yeah I couldn't figure out what causes this. It looks like reqwest and tower should correctly handle Found.
- https://docs.rs/tower-http/latest/src/tower_http/follow_redirect/mod.rs.html#273
- https://github.com/seanmonstar/reqwest/blob/ef5b239cbaa5e93607415bcf943c8f571da13e42/src/redirect.rs#L405

thomas-zahner:
This actually is not a bug. It's just the website which redirects 6 times and the default value for `--max-redirects` happens to be 5. If we reach this maximum amount of redirects we just resolve to the status code of the last redirect, which is intended behaviour. We can discuss increasing the value for `--max-redirects`. Should we increase it to 10?

```
echo 'http://dx.doi.org/10.1007/978-3-540-74061-2_11' | lychee - -vv --max-redirects 6
   [DEBUG] Following redirect to https://dx.doi.org/10.1007/978-3-540-74061-2_11
   [DEBUG] Following redirect to http://link.springer.com/10.1007/978-3-540-74061-2_11
   [DEBUG] Following redirect to https://link.springer.com/chapter/10.1007/978-3-540-74061-2_11
   [DEBUG] Following redirect to https://idp.springer.com/authorize?response_type=cookie&client_id=springerlink&redirect_uri=https%3A%2F%2Flink.springer.com%2Fchapter%2F10.1007%2F978-3-540-74061-2_11
[...]

PR #2012, opened 2026-01-28
 feat: add max-redirects hint to error message for 3xx codes

katrinafyi:
This helps a little bit with https://github.com/lycheeverse/lychee/issues/1958.

```
echo 'https://mock.httpstatus.io/301 https://mock.httpstatus.io/404' | cargo run -- - -v --max-redirects 0
```
```
[stdin]:
     [301] https://mock.httpstatus.io/301 | Rejected status code: 301 Moved Permanently (configurable with "accept" option"): Redirects may have been limited by "max-redirects".
     [404] https://mock.httpstatus.io/404 | Rejected status code: 404 Not Found (configurable with "accept" option")
```

### DIFF
--- a/lychee-lib/src/utils/reqwest.rs
+++ b/lychee-lib/src/utils/reqwest.rs
@@ -405,9 +405,13 @@ mod tests {
     /// Test that `ErrorKind::details()` properly uses the new analysis
     #[test]
     fn test_error_kind_details_integration() {
+        // Test rejected status code
+        let status_error = ErrorKind::RejectedStatusCode(http::StatusCode::NOT_FOUND);
+        assert!(status_error.to_string().contains("Not Found"));
+
         // Test redirected status code
-        let status_error = ErrorKind::RejectedStatusCode(http::StatusCode::MOVED_PERMANENTLY);
-        assert!(status_error.details().is_some_and(|x| x.contains(
+        let redir_error = ErrorKind::RejectedStatusCode(http::StatusCode::MOVED_PERMANENTLY);
+        assert!(redir_error.details().is_some_and(|x| x.contains(
             "Redirects may have been limited by \"max-redirects\""
         )));
 


thomas-zahner:
That's a great idea, thanks.

Isn't the reason text always displayed alongside the resulting status code?
If so, maybe we could remove the repetition of the code and only display the canonical reason.

```
[301] https://mock.httpstatus.io/301 | Rejected status code: 301 Moved Permanently
 ^^^                                                         ^^^
```

katrinafyi:
That's true and it is like that in main atm:
```
[200] http://127.0.0.1:34195/ | Rejected status code (this depends on your "accept" configuration): OK
```

But personally, I don't mind repeating the number.

The [200] comes from the response formatter, but it's possible that the ErrorKind would be printed in different places without that. Also, it's just convenient to have the number and text reason next to each other. In the past, I have forgotten to look at the start of the line and I've had to google the status reason to find the number 😅 And as a small nit, it says "Rejected status code:" so it makes sense to have the code there anyway.

I'm happy to go either way with this, let me know :-)

thomas-zahner:
Okay, lets leave it then

lychee-lib/src/utils/reqwest.rs
```diff
@@ -405,9 +405,11 @@ mod tests {
     /// Test that `ErrorKind::details()` properly uses the new analysis
     #[test]
     fn test_error_kind_details_integration() {
-        // Test rejected status code
-        let status_error = ErrorKind::RejectedStatusCode(http::StatusCode::NOT_FOUND);
-        assert_eq!(status_error.details(), Some("Not Found".to_string()));
+        // Test redirected status code
+        let status_error = ErrorKind::RejectedStatusCode(http::StatusCode::MOVED_PERMANENTLY);
+        assert!(status_error.details().is_some_and(|x| x.contains(
+            "Redirects may have been limited by \"max-redirects\""
```
thomas-zahner, line 415: This additional `contains` is great. But I can we keep testing the `NOT_FOUND` path as well?

lychee-lib/src/types/error.rs
```diff
@@ -202,12 +206,9 @@ impl ErrorKind {
                 // Get detailed, actionable error analysis
                 Some(utils::reqwest::analyze_error_chain(e))
             }
-            ErrorKind::RejectedStatusCode(status) => Some(
-                status
-                    .canonical_reason()
-                    .unwrap_or("Unknown status code")
-                    .to_string(),
-            ),
+            ErrorKind::RejectedStatusCode(status) => status
```
thomas-zahner, line 209: Doesn't this hide the details for non-redirected status code? Is there a good reason for doing so? 

Is that why `test_error_kind_details_integration` failed?

katrinafyi: The old details() just had the text reason. I've moved the text reason into the main to_string in #[error] because I think it makes more sense to put it there.

I don't really know what test_error_kind_details_integ is doing because it's inside the reqwest utils file and it's commented to test analyze_error_chain, but it doesn't seem to do that.

lychee-lib/src/utils/reqwest.rs
katrinafyi: Sure

mre:
Sorry if this isn't the right place to discuss this, but I always found the message `(this depends on your "accept" configuration)` on every line a bit noisy. Maybe we should put it at the end, as we do with the GitHub token hint.

```sh
[stdin]:
     [301] https://mock.httpstatus.io/301 | Rejected status code: 301 Moved Permanently: Redirects may have been limited by "max-redirects".
     [404] https://mock.httpstatus.io/404 | Rejected status code: 404 Not Found
     
Hint: You can configure accepted response codes with the "accept" option".
```

We could introduce a `Hint` struct that collects all of those hints (GitHub token, accept codes, redirects detected).

thomas-zahner:
Ah, I totally agree with your idea @mre :+1:

katrinafyi:
Yeah I definitely agree too. A separate PR, maybe

lychee-lib/src/types/error.rs
thomas-zahner: Hmm, I guess that test has undergone quite some changes..
It used to cover some of the reqwest error analysis but now does no longer.

@mre Do you know what happened with `test_error_kind_details_integration`? Is it hard to test? Should we try to restore the test and cover it again?

Relevant commit: https://github.com/thomas-zahner/lychee/commit/41b739715bb6716c844e65e9661f05d18c02923f#diff-d40124ee2f6ccff017999a8fd0c9d935c529627d3bed836ff90eba06c6b0d2f6R405

If not, then it might make sense to move this test into a different place or even remove it, the same path might be covered already.

thomas-zahner:
@katrinafyi I've created https://github.com/lycheeverse/lychee/pull/2020 since I wasn't able to push my small addition to your branch. Thank you for the PR.

katrinafyi:
Yeah sure. I'll try to remember to tick the box next time

PR #2016, opened 2026-01-29
Try to fix rustls setup after recent reqwest change

mre:
IDK if those are the right flags we should use. Attempt to fix https://github.com/lycheeverse/lychee-action/issues/323

### DIFF
--- a/lychee-lib/Cargo.toml
+++ b/lychee-lib/Cargo.toml
@@ -40,13 +40,15 @@ path-clean = "1.0.1"
 percent-encoding = "2.3.1"
 pulldown-cmark = "0.13.0"
 regex = "1.12.2"
-# Use trust-dns to avoid lookup failures on high concurrency
-# https://github.com/seanmonstar/reqwest/issues/296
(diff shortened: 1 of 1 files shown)


lychee-lib/Cargo.toml
```diff
@@ -40,13 +40,15 @@ path-clean = "1.0.1"
...
-# https://github.com/seanmonstar/reqwest/issues/296
+# Disable default features to avoid rustls being enabled by default
 reqwest = { version = "0.13.1", default-features = false, features = [
     "gzip",
     "hickory-dns",
     "cookies",
     "json",
+    "charset",
+    "http2",
+    "system-proxy",
```
mre, line 51: These are the default features except `default-tls`, which causes problems.

See https://docs.rs/crate/reqwest/latest/features

thomas-zahner: Ah thanks, now I get it

thomas-zahner:
Looks good I think, definitely worth a try

Issue #1892, opened 2025-10-30
Speed up slow tests


thomas-zahner:
The following slow tests have been bugging me for a while now.

```
PASS [   7.266s] lychee::cli cli::test_lycheecache_exclude_custom_status_codes
PASS [   7.138s] lychee-lib client::tests::test_timeout
PASS [  12.799s] lychee::cli cli::test_compact_output_format_contains_status
SLOW [> 60.000s] lychee::cli cli::test_url_inputs_always_get_checked_no_matter_their_extension
PASS [ 117.764s] lychee::cli cli::test_url_inputs_always_get_checked_no_matter_their_extension
```

Speed up these tests without changing the functionality of the tests.
Without looking at the code, I assume these tests do network calls.
If this is the case, a solution is to set up local mock servers for testing with `wiremock::MockServer` to replace external network calls.
For inspiration see https://github.com/lycheeverse/lychee/blob/8096a569622a990a1c47c755440106a968c6436d/test-utils/src/lib.rs#L8

thomas-zahner:
Note the following test seems to be flaky. Would be nice to fix in the same PR.

```
        FAIL [   0.515s] lychee::cli cli::test_lycheecache_file
  stdout ───

    running 1 test
    Cache file contents: http://127.0.0.1:38935/,200,1765297468
    http://127.0.0.1:35441/,404,1765297468

    test cli::test_lycheecache_file ... FAILED

    failures:

    failures:
        cli::test_lycheecache_file

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 134 filtered out; finished in 0.50s

  stderr ───
    Error: Failed to remove cache file: "/home/thomas/Projects/lychee/fixtures/cache/.lycheecache", error: No such file or directory (os error 2)
```

PR #2019, opened 2026-02-01
Speed up slow tests

thomas-zahner:
Fix https://github.com/lycheeverse/lychee/issues/1892.

In CI we can see that tests now take [about 55 seconds](https://github.com/lycheeverse/lychee/actions/runs/21575222272/job/62161204666?pr=2019) instead of [70 seconds](https://github.com/lycheeverse/lychee/actions/runs/21585053842/job/62191042936). The CI action spends a lot of time in compiling lychee. This change is mainly an improvement for local development. E.g. on my machine the first stage of `make test` (`cargo nextest run --all-targets --all-features`) now takes about 5 seconds instead of 12 on average.

Apart from speeding up the tests I could also improve the reliability.

## Some background details

`test_lycheecache_exclude_custom_status_codes` requires no internet but was slow because of unnecessary retries.

`test_require_https` is using example.com. As it seems really unpractical to make localhost functional with HTTPS and it generally isn't slow I would prefer to keep this test unchanged.

Fix the flakiness of the cache tests. They were racing against each other writing and deleting the cache files of each other.

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1161,6 +1161,7 @@ The config file should contain every possible key for documentation purposes."
 
         cargo_bin_cmd!()
             .current_dir(test_path)
+            .arg("--insecure")
             .arg("TEST.md")
             .arg("--exclude-file")
             .arg(excludes_path)
@@ -1640,7 +1641,6 @@ The config file should contain every possible key for documentation purposes."
             .arg("--dump")
             .arg("example.com")
             .arg(&test_path)
-            .arg("https://example.org")
             .assert()
             .success();
     }


lychee-bin/tests/cli.rs
```diff
@@ -1440,58 +1442,6 @@ The config file should contain every possible key for documentation purposes."
         Ok(())
     }
 
-    /// Unknown status codes should be skipped and not cached by default
```
thomas-zahner, line 1456: This test is lying. 999 is cached as of today. Alternatively we could make the test pass similarly to the below code, but I think this might not make sense.

```patch
index e5368d4f..85cee011 100644
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1443,13 +1443,6 @@ The config file should contain every possible key for documentation purposes."
     /// The reason is that we don't know if they are valid or not
     /// and even if they are invalid, we don't know if they will be valid in the
     /// future.
-    ///
-    /// Since we cannot test this with our mock server (because hyper panics on
-    /// invalid status codes) we use LinkedIn as a test target.
-    ///
-    /// Unfortunately, LinkedIn does not always return 999, so this is a flaky
-    /// test. We only check that the cache file doesn't contain any invalid
-    /// status codes.
     #[tokio::test]
     async fn test_skip_cache_unknown_status_code() -> Result<()> {
         let base_path = fixtures_path!().join("cache");
@@ -1458,35 +1451,31 @@ The config file should contain every possible key for documentation purposes."
         // Unconditionally remove cache file if it exists
         let _ = fs::remove_file(&cache_file);
 
-        // https://linkedin.com returns 999 for unknown status codes
-        // use this as a test target
-        let unknown_url = "https://www.linkedin.com/company/corrode";
+        let mock_server = wiremock::MockServer::start().await;
+
+        Mock::given(wiremock::matchers::method("GET"))
+            .respond_with(ResponseTemplate::new(999))
+            .mount(&mock_server)
+            .await;
+
+        let url = mock_server.uri();
 
         // run first without cache to generate the cache file
         cargo_bin_cmd!()
             .current_dir(&base_path)
-            .write_stdin(unknown_url.to_string())
+            .write_stdin(url.to_string())
             .arg("--cache")
             .arg("--verbose")
             .arg("--no-progress")
-            .arg("--")
             .arg("-")
             .assert()
-            .success();
+            .failure();
 
         // If the status code was 999, the cache file should be empty
         // because we do not want to cache unknown status codes
         let buf = fs::read(&cache_file).unwrap();
-        if !buf.is_empty() {
-            let data = String::from_utf8(buf)?;
-            // The cache file should not contain any invalid status codes
-            // In that case, we expect a single entry with status code 200
-            assert!(!data.contains(",999,"));
-            assert!(data.contains(",200,"));
-        }
-
-        // clear the cache file
-        fs::remove_file(&cache_file)?;
+        let data = String::from_utf8(buf)?;
+        assert!(data.contains(",999,"));
 
         Ok(())
     }
```

thomas-zahner:
@katrinafyi FYI

lychee-bin/tests/cli.rs
katrinafyi: Why does it not make sense? It seems like a fine thing to test that 999 is cached. The test name will also have to be changed.

Also, if 999 is actually cached at the moment, why was the previous test passing?

lychee-bin/tests/cli.rs
```diff
@@ -2595,19 +2545,27 @@ The config file should contain every possible key for documentation purposes."
...
+            "https://example.com/2",
         ];
 
         let cmd = &mut cargo_bin_cmd!()
             .arg("--format")
             .arg("compact")
-            .arg(fixtures_path!().join(test_files[1]))
-            .arg(fixtures_path!().join(test_files[0]))
+            .arg(b_md.as_os_str())
+            .arg(a_md.as_os_str())
```
katrinafyi:
```suggestion
            .arg(b_md)
            .arg(a_md)
```

lychee-bin/tests/cli.rs
```diff
@@ -2617,8 +2575,7 @@ The config file should contain every possible key for documentation purposes."
 
         // Check that the input sources are sorted
         for file in test_files {
-            assert!(output.contains(file));
-
+            assert!(output.contains(file), "{file} not found");
```
katrinafyi: just to avoid confusion with overly generic file not found messages.
```suggestion
            assert!(output.contains(file), "{file} not found in lychee output");
```

lychee-bin/tests/cli.rs
```diff
@@ -2629,13 +2586,14 @@ The config file should contain every possible key for documentation purposes."
 
         // Check that the responses are sorted
         for url in test_urls {
-            assert!(output.contains(url));
-
+            assert!(output.contains(url), "{url} not found");
```
katrinafyi:
```suggestion
            assert!(output.contains(url), "{url} not found in lychee output");
```

lychee-bin/tests/cli.rs
```diff
@@ -2953,18 +2911,26 @@ The config file should contain every possible key for documentation purposes."
...
+                ResponseTemplate::new(200).set_body_string(format!("<a href=\"{url}\">hi</a>")),
+            )
+            .mount(&mock_server)
+            .await;
 
         cargo_bin_cmd!()
             .arg("--verbose")
             .arg("--dump")
-            .arg(url_input)
+            .arg(mock_server.uri())
```
katrinafyi: to match the purpose of the existing test's description, the CLI URL should be changed to end in .xml

```suggestion
            .arg(format!("{}/sitemap.xml", mock_server.uri()))
```

katrinafyi (commented):
looks good. it's a very good idea to disentangle the cache tests.

lychee-bin/tests/cli.rs
thomas-zahner: Oh, thanks for catching this 👍

lychee-bin/tests/cli.rs
thomas-zahner: True it does make sense. I thought that it's unnecessary to test such codes (999) because they are not treated any different as any other status code as of now. But it's surely good documentation if this behaviour changed over time. See 545168ea

It did pass because linkedin didn't return 999 but 200. This is because of the user agent.

`curl https://www.linkedin.com/company/corrode --user-agent lychee/0.21.0 -w "%{http_code}" -o /dev/null` returns `200` whereas `curl https://www.linkedin.com/company/corrode -w "%{http_code}" -o /dev/null` returns `999`.

I suspect that's the reason why the assertion was so weird. They might have started to return 200 for different user agents only after the test had been written and then the assertion was updated.

katrinafyi: It all makes sense. Thanks for looking into it and satisfying my curiosity :)

mre: > They might have started to return 200 for different user agents only after the test had been written and then the assertion was updated.

Funny that they accept lychee, but not curl... 😆

PR #2021, opened 2026-02-02
User hints

### LINKED ISSUE #2178: Range syntax for accepted status codes not working
I had workflow failures caused by 302 redirects failing since [updating to v2.8.0](https://github.com/Technetium1/BrowserExtensions/pull/21), but I didn't notice until recently. I changed my accept range to a list as shown in this commit, and no longer have the same errors: https://github.com/Technetium1/BrowserExtensions/commit/026b17cb50fc68fc08efe4f0539c9f3d3e1c3e26

I checked https://github.com/lycheeverse/lychee-action#passing-arguments-1 but I may have missed something.

Before: https://github.com/Technetium1/BrowserExtensions/actions/runs/24900371035
After: https://github.com/Technetium1/BrowserExtensions/actions/runs/24900764654

Please disregard the unrelated network errors in the latest run, I was manually triggering the workflow and got rate limited.

thomas-zahner:
Sorry if this isn't the right place to discuss this, but I always found the message `(this depends on your "accept" configuration)` on every line a bit noisy. Maybe we should put it at the end, as we do with the GitHub token hint.

```sh
[stdin]:
     [301] https://mock.httpstatus.io/301 | Rejected status code: 301 Moved Permanently: Redirects may have been limited by "max-redirects".
     [404] https://mock.httpstatus.io/404 | Rejected status code: 404 Not Found
     
Hint: You can configure accepted response codes with the "accept" option".
```

We could introduce a `Hint` struct that collects all of those hints (GitHub token, accept codes, redirects detected).

_Originally posted by @mre in https://github.com/lycheeverse/lychee/issues/2012#issuecomment-3816605045_

Closes #2178


lychee-bin/src/formatters/hints.rs
mre, line 1: I wonder if we should move it out of `formatters`. Maybe `src/hints.rs`?

lychee-bin/src/formatters/hints.rs
```diff
@@ -0,0 +1,67 @@
...
+
+use crate::{formatters::stats::ResponseStats, options::Config};
+
+const GITHUB_ERRORS: &str = "There were issues with GitHub URLs. You could try setting a GitHub token and running lychee again.";
+
+pub(crate) fn display_hints(stats: &ResponseStats, config: &Config) {
+    github_warning(stats, config);
+    redirect_warning(stats, config);
+    rejected_status_code_warning(stats);
+}
```
mre: This limits the design space. I was thinking of a generic Vec of hints, as in

```rust
type Hints = Vec<Hint>;
```

Then we could just "push" hints into that and impl a nice `Display` for it. Done. 😉

lychee-bin/src/formatters/hints.rs
```diff
@@ -0,0 +1,67 @@
...
+
+/// Display user-friendly message if there were any issues with GitHub URLs
+fn github_warning(stats: &ResponseStats, config: &Config) {
```
mre: I would move this function and the other functions below out of the hints module. So the logic becomes: "If we detect an issue throughout the link checking, we can easily add a hint."

lychee-lib/src/types/error.rs
```diff
@@ -152,7 +152,7 @@ pub enum ErrorKind {
 
     /// The given status code was not accepted (this depends on the `accept` configuration)
     #[error(
-        r#"Rejected status code: {code} {reason} (configurable with "accept" option)"#,
+        r#"Rejected status code: {code} {reason}"#,
```
mre: Thanks!

lychee-bin/src/formatters/hints.rs
thomas-zahner: Thank you for the feedback. For now I just kept the changes minimal and did the extraction, which is also why the PR is a draft.

Thinking about it I find it quite difficult to generalise all hints. It feels a bit like over-engineering, but maybe there is a good solution after all?

What would `Hint` be then?

Does it have state? Or is it just the rules and functions to know if the given hint should be triggered and if so, what the specific message was? Hints are only optionally triggered/shown and could take very different arguments.

The following wouldn't seem to work well.

```rust
type Hint = Box<dyn Display>;

type Hints = Vec<Hint>;
```

Instead a hint probably should define the rules to calculate if it should be triggered and what the message should be. 

```rust
trait Hint {
    fn show() -> Option<String>;
}
```

But the above would still not work, because each hint would have different arguments. This meant that the Hint itself would have to contain state, but that wouldn't make sense either, I think.

lychee-bin/src/formatters/hints.rs
thomas-zahner: So would you "generate" the hint message at the location where we encounter it? I currently feel like moving this logic into the hint module makes sense.

mre: Good point.  My idea was that we could provide a method like

```rust
fn add_hint<T: Into<Hint>>(hint: T);
```

And support converting strings to hints. So `impl From<String> for Hint`.
Not sure if that's a good idea, but that's what I had in mind to decouple the responsibilities.

lychee-bin/src/formatters/hints.rs
mre: > What would Hint be then?

Looking at the current usage pattern, `Hint` could just be a newtype wrapper around String?

```rust
#[derive(Debug)]
struct Hint(String);

impl Hint {
  fn show(&self) -> String {
    self.0
  }
}
```

Not sure if we need anything more than that? 🤔 
It would be great to check that a hint is non-empty on construction, hence the newtype; but apart from that, I'd keep it pretty lightweight.

It could be that I'm not accounting for additional requirements, though.

lychee-bin/src/hints.rs
```diff
@@ -0,0 +1,117 @@
...
+        })
+    });
+
+    if is_default && any_rejected_codes {
+        add_hint(MESSAGE.into());
+    }
+}
+
+fn unfollowed_redirects(stats: &ResponseStats, config: &Config) {
+    const MESSAGE: &str = "Rejected redirecional status codes. \
```
mre:
```suggestion
    const MESSAGE: &str = "Rejected redirectional status codes. \
```

lychee-bin/src/formatters/hints.rs
thomas-zahner: Thanks for your thoughts, I tried to incorporate them.

thomas-zahner:
@mre Thank you for your ideas. Initially I was thinking more of a fully functional approach were we wouldn't need to store state. But your idea also works.

Currently user hints is limited to lychee-bin which makes sense and keeps things simple. But this might also be limiting as there is no way to "trigger" a user hint in lychee-lib. For example with reporting bot protection as mentioned in https://github.com/lycheeverse/lychee/issues/2117 this might need to be updated.

Then I was also considering a possibility to hide these hints. Some users might not want to see them or maybe there are scenarios where we shouldn't display them? I was considering:

- Not showing hints in uncolored terminals
- Using env variables e.g. `NO_HINTS`
- Adding a flag e.g. `--no-hints`

Edit: ah maybe there's no need to any of this if instead we just use the logging level, so that they can be hidden with `-q` (see 78af898eb283e19a7dbc064d51310a7bb74691cd)

lychee-bin/src/hints.rs
thomas-zahner: Ah thank you :+1:

lychee-bin/src/hints.rs
```diff
@@ -0,0 +1,155 @@
...
+
+use std::{fmt::Display, sync::Mutex};
+
+use http::StatusCode;
+use lychee_lib::{ErrorKind, Status, StatusCodeSelector};
+
+use crate::{config::Config, formatters::stats::ResponseStats, verbosity::Verbosity};
+
+/// Hints are accumulated during a single program invocation.
+static HINTS: Mutex<Vec<Hint>> = Mutex::new(vec![]);
```
mre: Nit: Wondering if we should put them in a HashSet and sort them on output instead. This way, the output would be reproducible while we acknowledge the fact that there's no inherent ordering in the hints. So a set is more appropriate than a vec.

mre (approved):
I quite like it! Good work. The only remaining comment is optional.

mre:
> Edit: ah maybe there's no need to any of this if instead we just use the logging level, so that they can be hidden with -q (see https://github.com/lycheeverse/lychee/commit/78af898eb283e19a7dbc064d51310a7bb74691cd)

Yeah, that would be great. In general, I'd wait until someone asks for it; but that might just be a "me" thing, haha.

The functional approach also makes sense. While reading your comment, I thought about having a global registry of sorts, similar to how logging gets set up once at the beginning of a binary and then is configured for the rest of the runtime. Maybe you thought of something similar. I haven't looked into it, but I'm guessing the magic would be in a global constant and a function call which reads/writes from/into that.

TBH, I think the current implementation works well and what I like about it is that it's definitely an improvement and still simple enough to understand without obstructing the business logic. The fact that it works on stats is a really nice touch and that's what makes it so self-contained right now.

So I'd have no objections if we merged this.

lychee-bin/src/hints.rs
thomas-zahner: Currently they are reproducible already. But it might make sense nonetheless. But I feel like `Vec` is still simpler. See 5d65905ed3e7b2004ade3a9d3f4523423e8a5e56

thomas-zahner:
I've made some small adjustments, feel free to take another look.

> Yeah, that would be great. In general, I'd wait until someone asks for it; but that might just be a "me" thing, haha.

I like this approach. But I was already imagining the user asking for a way to hide the hints :)

> The functional approach also makes sense.

Agree in the current state it would maybe be even easier. But as soon as we need to have additional ways to add hints (e.g. https://github.com/lycheeverse/lychee/issues/2117) the current state is probably necessary.

mre (approved):
Looks really good. Ready to merge from my side.

thomas-zahner:
@mre Last updates:

* I've now moved hints to lychee-lib. This is necessary for us to also collect hints in lychee-lib, as is now done with the [accept hint](https://github.com/lycheeverse/lychee/pull/2169).
* I've added `hint!` for more convenience. This feels quite natural as it complements `info!`, `warn!`, etc. But it's not exactly the same as it doesn't immediately write something. The hints are only optionally (depending on verbosity) written before program termination.

lychee-bin/src/hints.rs
```diff
@@ -0,0 +1,111 @@
...
+        .find(|r| {
+            matches!(
+                r.status,
+                Status::Error(ErrorKind::RejectedStatusCode(StatusCode::TOO_MANY_REQUESTS))
+            )
+        })
+        .and_then(|b| b.uri.domain());
+
+    if default_host_config && let Some(domain) = first_rate_limited_domain {
+        hint!(
```
mre, line 30: The little exclamation marks make this looks so pretty. 😊

mre (approved):
Love it.

Issue #2034, opened 2026-02-08
CacheableResponse forces download of entire HTTP response


katrinafyi:
_Originally posted by @jcharaoui in [#27](https://github.com/lycheeverse/lychee/issues/27#issuecomment-3855438008)_

> I'm having a similar issue where a website I want to check has a lot of download links, but lychee is struggling to complete checking all the links and timing out on many of them.
> 
> Example URL: https://dist.torproject.org/torbrowser/16.0a2/tor-browser-linux-x86_64-16.0a2.tar.xz
> 
> When checking this link using the GET method, it downloads the whole file, while using the HEAD method the check completes in a matter of milliseconds.
> 
> The inconsistency observed is likely due to variations in the configurations of the various HTTP daemons out there. 

It looks like the entire response body is cached and this causes the entire body to be read, even when it's not needed (i.e., when fragment checking is disabled).

https://github.com/lycheeverse/lychee/blob/7d2fe388d31bcc2bafbb093e7a383eff47603cd4/lychee-lib/src/ratelimit/mod.rs#L40-L45

PR #2035, opened 2026-02-08
fix: CacheableResponse should only download file when needed

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/2034

This changes the CacheableResponse's text field to be an Option. By default, responses will be cached without the body text so they can be checked quickly even if the file is large. Cached responses can be "upgraded" to include the body if fragments need to be checked inside that URL. Once upgraded, the response with body will be cached for future fragment checks.

This is not a very intelligent fix to the issue, I don't love passing `needs_body: bool` everywhere. But it does work and it's not very invasive.

Note that this caching of body text is slightly redundant, because FragmentChecker already caches its own set of anchors within each URL. But it could be useful if the text is needed for other things in future (recursion???).

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -3334,7 +3334,7 @@ The config file should contain every possible key for documentation purposes."
             .stdout(contains("https://example.com"));
     }
 
-    /// URLs should NOT be downloaded fully, unless their link has a fragment.
+    /// URLs should NOT be downloaded fully, unless fragment checking is on and the link has a fragment.
     #[test]
     fn test_large_file_lazy_download() {
         cargo_bin_cmd!()
@@ -3343,10 +3343,24 @@ The config file should contain every possible key for documentation purposes."
             .arg("--timeout=5")
             .write_stdin(
                 "
-https://www.releases.ubuntu.com/noble/ubuntu-24.04.3-desktop-amd64.iso
-https://www.releases.ubuntu.com/noble/ubuntu-24.04.3-live-server-amd64.iso
-https://www.releases.ubuntu.com/noble/ubuntu-24.04.3-wsl-amd64.wsl
+https://proof.ovh.net/files/10Gb.dat
+https://proof.ovh.net/files/1Gb.dat
+https://proof.ovh.net/files/1Mb.dat
 https://lychee.cli.rs/guides/cli/#options
+            ",
+            )
+            .assert()
+            .success();
+
+        cargo_bin_cmd!()
+            .arg("-")
+            .arg("--timeout=5")
+            .write_stdin(
+                "
+https://proof.ovh.net/files/10Gb.dat#fragments-ignored
+https://proof.ovh.net/files/1Gb.dat#fragments-ignored
+https://proof.ovh.net/files/1Mb.dat#fragments-ignored
+https://lychee.cli.rs/guides/cli/#fragments-ignored
             ",
             )
             .assert()


lychee-bin/tests/cli.rs
```diff
@@ -3346,4 +3333,23 @@ The config file should contain every possible key for documentation purposes."
...
+        cargo_bin_cmd!()
+            .arg("-")
+            .arg("--include-fragments")
+            .arg("--timeout=5")
+            .write_stdin(
+                "
+https://www.releases.ubuntu.com/noble/ubuntu-24.04.3-desktop-amd64.iso
+https://www.releases.ubuntu.com/noble/ubuntu-24.04.3-live-server-amd64.iso
+https://www.releases.ubuntu.com/noble/ubuntu-24.04.3-wsl-amd64.wsl
+https://lychee.cli.rs/guides/cli/#options
```
thomas-zahner, line 3349: I feel like this test could be done in a more elegant way. It might be possible to do a more "low-level" unit/integration test, e.g. in website.rs but this might also be impractical.

Alternatively, I think it should be possible to replace the `www.releases.ubuntu.com` URLs with URLs to a mock server and then do assertions on the requests made to the mock server.

This would be more reliable; ubuntu.com is down, URL breaks as ISO is removed, I'm developing on lychee without internet connection, my PC or GitHub runner has incredibly fast internet so that the test passes on a regression even though it should fail but the ISOs just downloaded in less than 5 seconds.

thomas-zahner:
@katrinafyi Thank you for the PR! I see no issues with it apart from the test that could be made more precise.

> This is not a very intelligent fix to the issue, I don't love passing needs_body: bool everywhere. But it does work and it's not very invasive.

I think this approach is totally fine.

lychee-bin/tests/cli.rs
katrinafyi: Mm it would be nice to avoid the external URL. I also worried about Github actions being too fast. But this has some unusual requirements which make it harder. We need a URL which responds with 200 quickly, but then sends the rest of the large body slowly and preferably without holding it in memory.

It looks like [wiremock doesn't have](https://docs.rs/wiremock/latest/wiremock/struct.ResponseTemplate.html#method.set_body_bytes) this customisation. It only lets you supply bytes as the body, and you can't generate arbitrary "chunks" on demand. Its assertions also aren't granular enough to assert what we want: that the URL was fetched but the connection was closed without downloading the whole file.

So, I think it's easiest to use the Ubuntu URL which should be fairly long-term and reliable. I also acknowledge the other risks. The alternative would be to change to a different library which supports a generating function for the response body, e.g. [mockito](https://docs.rs/mockito/latest/mockito/struct.Mock.html#method.with_chunked_body). But I don't think it's worth the complexity for this one test.

Re a lower-level test, I think that the CLI test is ok here. The presenting issue is that big files shouldn't be downloaded in full, and it really doesn't matter how that's implemented. If a unit test is added, I'd still want to keep the CLI test.

jcharaoui: I think the concern about the specific URL used here is legitimate. If you look at the [home page](https://www.releases.ubuntu.com/), it doesn't list older Ubuntu releases, so it's fair to assume older releases are regularly removed from that page and hosted elsewhere such as https://old-releases.ubuntu.com/releases

It would probably be safer to use the URL of some ISO hosted at the above domain instead. Another alternative might be archive.org which also hosts ISO and large files of all sorts, with URLs that one hopes are intended to be stable long-term.

katrinafyi: Yeah fair call. I wouldn't want to use archive.org because they have constrained resources (and sometimes have downtime because of those limited resources).

Maybe one of the files from here https://proof.ovh.net/files/? I'm not _too_ worried about longevity of a particular URL, since it should be easy to find another when/if needed.

thomas-zahner: Okay, thanks for taking a closer look at wiremock, I would have assumed they have support for such a use case. I like the new URLs they look pretty stable and long-lived. So I'm fine this :+1:

PR #2036, opened 2026-02-08
fix: GITHUB_TOKEN should not cause rustls CryptoProvider panic

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/2024

Disables octocrab option which would lead to rustls/ring, so only rustls/aws-lc-rs is enabled.

Honestly, it looks like an octocrab bug because its default option set would lead to incompatible rustls features being enabled: https://docs.rs/crate/octocrab/latest/features

octocrab/default -> octocrab/rustls -> hyper-rustls/default -> rustls/aws-lc-rs
octocrab/default -> octocrab/rustls-ring -> hyper-rustls/ring -> rustls/ring

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -6,7 +6,7 @@ mod cli {
     use http::{Method, StatusCode};
     use lychee_lib::{InputSource, ResponseBody};
     use predicates::{
-        prelude::{PredicateBooleanExt, predicate},
+        prelude::PredicateBooleanExt,
         str::{contains, is_empty},
     };
     use pretty_assertions::assert_eq;
@@ -380,20 +380,6 @@ mod cli {
                 ..MockResponseStats::default()
             }
         )
-    }
-
-    #[test]
-    fn test_check_github_with_token() {
-        // https://github.com/lycheeverse/lychee/issues/2024
-        cargo_bin_cmd!()
-            .arg("-")
-            .write_stdin("https://github.com/lycheeverse/lychee")
-            .env("GITHUB_TOKEN", "dummy")
-            .assert()
-            .code(predicate::ne(101)) // assert not panicked
-            .stderr(contains("Could not automatically determine the process-level CryptoProvider from Rustls crate features.").not());
-        // we cannot assert for success or failure because the dummy token may
-        // or may not be used depending on the current github rate limiting state.
     }
 
     /// Test unsupported URI schemes
--- a/lychee-lib/Cargo.toml
+++ b/lychee-lib/Cargo.toml
@@ -41,6 +41,8 @@ octocrab = { version = "0.49.5", default-features = false, features = [
     "retry",
     "rustls",
     # "rustls-ring",
+    # Enabling `rustls-ring` leads to runtime panics.
+    # This is covered by `test_github_client_integration`.
     "timeout",
     "tracing",
 ] }
--- a/lychee-lib/src/checker/website.rs
+++ b/lychee-lib/src/checker/website.rs
@@ -326,3 +326,54 @@ impl Handler<Request, Status> for WebsiteChecker {
         ChainResult::Done(self.retry_request(input).await)
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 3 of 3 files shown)


thomas-zahner:
@katrinafyi Thank you very much for the fix! The merge commit should which includes additional changes and causes the merge conflict should not be necessary. See https://github.com/thomas-zahner/lychee/commits/rustls-fix/ for a rebased version. (I again wasn't allowed to push to this branch)

lychee-bin/tests/cli.rs
```diff
@@ -382,6 +382,20 @@ mod cli {
...
+    fn test_check_github_with_token() {
+        // https://github.com/lycheeverse/lychee/issues/2024
+        cargo_bin_cmd!()
+            .arg("-")
+            .write_stdin("https://github.com/lycheeverse/lychee")
+            .env("GITHUB_TOKEN", "dummy")
+            .assert()
+            .code(predicate::ne(101)) // assert not panicked
+            .stderr(contains("Could not automatically determine the process-level CryptoProvider from Rustls crate features.").not());
+        // we cannot assert for success or failure because the dummy token may
```
thomas-zahner: When does this happen?
I could not reproduce it (by running the single test nor with `make test`) and it seemed fine if I just checked for success instead.

```patch
index 3be86c39..18897377 100644
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -390,10 +390,8 @@ mod cli {
             .write_stdin("https://github.com/lycheeverse/lychee")
             .env("GITHUB_TOKEN", "dummy")
             .assert()
-            .code(predicate::ne(101)) // assert not panicked
+            .success()
             .stderr(contains("Could not automatically determine the process-level CryptoProvider from Rustls crate features.").not());
-        // we cannot assert for success or failure because the dummy token may
-        // or may not be used depending on the current github rate limiting state.
     }
 
     /// Test unsupported URI schemes
```

katrinafyi:
I'd sent you an invite to the repo. It should be in your email, or maybe there's a button at https://github.com/rina-forks/lychee.
 I found that there was no checkbox I could tick on the PR page

lychee-bin/tests/cli.rs
katrinafyi: The bad token would cause errors when the GitHub API is used, but it looks like the GitHub API is only used as a fallback when the basic website check fails (e.g., due to rate limiting). So the command could succeed when it's allowed through rate limiting but fail if the API is called when the rate limit is hit.

https://github.com/lycheeverse/lychee/blob/master/lychee-lib/src/checker/website.rs#L267-L273

thomas-zahner:
> I'd sent you an invite to the repo.

Ah thank you! I now got one step further. But due to branch rules/protections I'm unable to force push. Could you allow force pushes to all branches (or non-master) buy modifying the rule set or by deleting it?

```
remote: error: GH013: Repository rule violations found for refs/heads/rustls-fix.
remote: Review all repository rules at https://github.com/rina-forks/lychee/rules?ref=refs%2Fheads%2Frustls-fix
remote:
remote: - Cannot force-push to this branch
remote:
To github.com:rina-forks/lychee.git
 ! [remote rejected]   rustls-fix -> rustls-fix (push declined due to repository rule violations)
error: failed to push some refs to 'github.com:rina-forks/lychee.git'
```

katrinafyi:
Done~

thomas-zahner:
@katrinafyi Thank you. I think in that case it makes more sense to specifically test the function causing the panic instead of using such abstract cli tests. Do you agree with d3f3c60b?

katrinafyi:
Yeah I don't really mind either way. It is nice to be more specific.

thomas-zahner:
Thanks. I feel that being more specific with testing should be done when possible. In lychee we tend to have many CLI integration tests (which is great) but it might make sense to also have more "low-level" tests.

katrinafyi:
It's true. I definitely use CLI tests more often just because they're easier to write 😅

thomas-zahner:
Haha yes I also tend to do this, which might not always be ideal :smile:

Issue #1947, opened 2025-12-05
pre-commit doesn't maintain a version


max-sixty:
it seems like the pre-commit hook uses whichever version is installed, rather than locking the version to the pre-commit version, is that correct?

here's a full description, written with the help of Claude Code

thank you!

---

# Pre-commit hook doesn't pin lychee version

## Problem

The pre-commit hook uses `language: script`, so the `rev` field only pins the hook script—not the lychee binary version.

```yaml
# .pre-commit-hooks.yaml
- id: lychee
  entry: scripts/lychee_pre_commit.sh
  language: script  # <-- uses system lychee or downloads latest
```

With `rev: lychee-v0.18.0`, users expect lychee 0.18.0, but actually get whatever's installed locally or the latest download.

## Reproduction

```yaml
# .pre-commit-config.yaml
repos:
  - repo: https://github.com/lycheeverse/lychee
    rev: lychee-v0.18.0
    hooks:
      - id: lychee
```

```bash
$ lychee --version
lychee 0.22.0  # system version, not 0.18.0
$ pre-commit run lychee --all-files
# runs 0.22.0, not 0.18.0
```

## Suggestion

Use `language: rust` with the crate version matching the tag:

```yaml
- id: lychee
  entry: lychee
  language: rust
  additional_dependencies: []  # version derived from rev tag
```

This would install the correct version via cargo, matching user expectations.

## Workaround

Users can define a local hook with pinned version:

```yaml
- repo: local
  hooks:
    - id: lychee
      name: lychee
      language: rust
      entry: lychee
      types: [markdown]
      additional_dependencies: ["cli:lychee:0.22.0"]
```

mre:
I don't know much about pre-commit hooks, but here are some related issues, which might help:

- https://github.com/lycheeverse/lychee/pull/1829
- https://github.com/lycheeverse/lychee/issues/1635
- https://github.com/lycheeverse/lychee/issues/1601

As far as I remember, there were some issues with `language: rust`. Please refer to the links for more details.

If you like, you could attempt a fix by using the above information. Perhaps there's also a simple way to test the changes automatically.

katrinafyi:
I just quickly tried `language: rust` and it breaks because lychee uses workspaces, so the lychee-bin package is not at the root. pre-commit is not interested in supporting this setup https://github.com/pre-commit/pre-commit/issues/2931, which is quite frustrating. 

[...]

PR #2042, opened 2026-02-11
feat: pre-commit should use a fixed version according to yaml 'rev'

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/1947

this allows a `.pre-commit-config.yaml` like
```yaml
repos:
    rev: lychee-v0.99.0
    hooks:
      - id: lychee
```
and this will automatically download and run lychee 0.99.0 when the pre-commit hook is executed.

the pre-commit copy of lychee will be installed with cargo-binstall and it will be separate from the rest of the system cargo binaries, to avoid version conflicts. the pre-commit hook will never try to use a system lychee binary.

i've also chosen to only support stable versions in pre-commit, because it simplifies the code and surely nobody enforcing lychee on pre-commit would want to compile lychee from scratch.

pre-release usage
-----------------

for lychee, we have a pre-commit script which should be able to download and run any version of lychee (in theory).

however, rather inconveniently, the pre-commit program seems committed to tying the version of the invoked tool to the repository revision. this only makes sense if the pre-commit hook is expected to build the repo from source and certainly doesn't make sense with this new lychee_pre_commit.sh script.

this makes it hard if users want to use this new pre-commit as soon as it's merged, before a release is cut. to work around this, I made the script take a `LYCHEE_VERSION` argument which can manually set the version to something different. so, you could use something like this to fetch the script from master but run a stable version of lychee:
```yaml
repos:
    # temporarily... fetch script from master but run lychee 0.22.0
    rev: master
    hooks:
      - id: lychee
        args: ["LYCHEE_VERSION=0.22.0", "--no-progress"]
```
once a stable release is cut with this PR, then users could just do `rev: lychee-v0.23.0` as usual.

testing
---------------
in fact, this temporary usage is how you can test this PR - by pointing it to commit within the PR branch. 
for example, you can make a `.pre-commit-config.yaml` in the lychee repo:
```yaml
repos:
  - repo: https://github.com/rina-forks/lychee
    rev: 8c428e0
    hooks:
    - id: lychee
      verbose: true
      args: ['LYCHEE_VERSION=0.22.0', '--no-progress', '--root-dir=.']
```
then
```
pre-commit install
```
and make some changes to README.md, then commit them. it'll take 20 seconds ish and should succeed. you can make some changes again, and it should be much faster - basically only what's needed to run lychee.

alternatively, you could also (privately) tag a new lychee version and watch it work up until binstall.

tangent: honestly, this new script is very close to a "bootstrap" script and it would be nice to further streamline the process of going from zero to lychee with no rust and no building. it should be very possible since it's all static binaries. maybe something like https://github.com/zyedidia/eget/ or
https://github.com/houseabsolute/ubi

### DIFF
--- a/scripts/lychee_pre_commit.sh
+++ b/scripts/lychee_pre_commit.sh
@@ -16,20 +16,11 @@ pushd "$LYCHEE_DIR" >/dev/null
 
 # install tools into a subfolder of the pre-commit repo folder, so they can be cached
 # but still remain independent from other cargo tools.
-export CARGO_HOME="$LYCHEE_DIR/.cargo"
-export CARGO_BIN="$CARGO_HOME/bin"
+export LYCHEE_CARGO_HOME="$LYCHEE_DIR/.cargo"
+export CARGO_BIN="$LYCHEE_CARGO_HOME/bin"
 export PATH="$CARGO_BIN:$PATH"
 
-LYCHEE="$CARGO_BIN/lychee"
-
-# fast path if lychee has already been downloaded
-if [[ -x "$LYCHEE" ]]; then
-    popd >/dev/null
-    exec "$LYCHEE" "$@"
-fi
-
 if [[ -z "${LYCHEE_VERSION:-}" ]]; then
-
     # pre-commit doesn't fetch tags by default, so we might need to grab them.
     if [[ -z "$(git tag)" ]]; then
         git fetch --tags &>/dev/null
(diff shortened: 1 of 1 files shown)


scripts/lychee_pre_commit.sh
```diff
@@ -1,28 +1,87 @@
...
+
+# smoke test before we do destructive deletes of the local files
+[[ -x "$LYCHEE" ]] || {
+    echo 'lychee pre-commit binary not found after cargo-binstall.' \
+        "this shouldn't happen and is probably a bug in lychee!" >&2
+    exit 100
 }
 
+# clean up no-longer-needed files to save a little bit of space
+rm -rf .git assets target build "$CARGO_BIN/cargo-binstall"
```
mre: Hum, I wonder about deleting the .git folder.

If a user deletes the lychee binary but leaves the pre-commit cache directory intact, the script will run again. It will skip the "fast path" (line 26), proceed to version detection, fail to find any tags (because you deleted the real .git), and crash. The self-healing property of the cache is broken.

Keeping the original .git folder means that if the binary disappears, the script can simply re-read the tags and download it again.

scripts/lychee_pre_commit.sh
```diff
@@ -1,28 +1,87 @@
...
-    exit 1
+if [[ -n "$LYCHEE_VERSION" ]]; then
+
+    if ! command -v cargo-binstall &>/dev/null; then
+        echo "Installing cargo-binstall..." >&2
+        curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | BASH_XTRACEFD=7 bash 7>/dev/null
+    fi
+    # TODO: it's possible that a user has cargo on their path but not cargo-binstall,
+    # so `cargo binstall` would succeed but `cargo-binstall` wouldn't. it would be nice
+    # to support this case also.
```
mre: Yeah, that's a good idea. I think you could combine it with the lines above:

```bash
if command -v cargo-binstall &>/dev/null; then
    BINSTALL="cargo-binstall"
elif command -v cargo &>/dev/null && CARGO_HOME="" cargo binstall -V &>/dev/null; then
    BINSTALL="cargo binstall"
else
    echo "Installing cargo-binstall..." >&2
    curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | BASH_XTRACEFD=7 bash 7>/dev/null
    BINSTALL="$CARGO_BIN/cargo-binstall"
fi

echo 'Installing lychee by cargo-binstall...' >&2
$BINSTALL -y "lychee@$LYCHEE_VERSION"
```

katrinafyi:
Done in [`6ff9906` (this PR)](https://github.com/lycheeverse/lychee/pull/2042/commits/6ff99062152bb4d98bc0faa12840f56011d971e3)

mre:
Tested it and it works well! Thanks for the helpful test instructions. 

```sh
❯❯❯ git commit -m "foo"                                                                                                                                      
lychee...................................................................Passed
- hook id: lychee
- duration: 1.02s

🔍 116 Total (in 0s) ✅ 114 OK 🚫 0 Errors 🔀 2 Redirects

[testme-tmp 56761a1e] foo
 2 files changed, 13 insertions(+)
 create mode 100644 .pre-commit-config.yaml
 ```

Merging. 🎉

PR #2046, opened 2026-02-14
refactor: add WaitGroup for waiting for a dynamic set of tasks 

katrinafyi:
This is a small step towards recursion. It will enable waiting for a dynamic number of tasks which can spawn other tasks, which allows for correct and deadlock-free termination in the presence of (eventual) recursion.

I think recursion is not an astoundingly difficult feature but, because lychee is async and stream, it needs the combination of a few parts which might be tricky to get right. This is one of those parts. In fact, I had two different deadlock bugs while working on this PR before reaching the final implementation. If this was all done in one big recursion PR, then I think these bugs would be a lot harder to find and fix.

The old implementation in https://github.com/lycheeverse/lychee/pull/1603 used `Arc<AtomicUsize>` to implement this requirement. This new PR uses a RAII-based guard value to represent tasks, which should make it much easier to write correct code - it's all handled by the borrow checker and Drop, so it's impossible to forget to adjust the counter.

In terms of testing, changes the check.rs command to rely on WaitGroup for termination. You can read about this in the comments. This means that every run of lychee goes through the WaitGroup code and acts as a test for WaitGroup. There is a risk that if it is buggy, the impacts would be big. However, I think the unit tests are big enough that they're a decent exercise of the WaitGroup. My second deadlock bug was found by `make test`.

You can also see the docs at https://waitgroup.lychee-docs-katrinafyi.pages.dev/lychee_lib/waiter/

Related work: 
- https://docs.rs/waitgroup/latest/waitgroup/ - seems to work
- https://docs.rs/atomic-waitgroup/latest/atomic_waitgroup/index.html - doesn't support recursive.
- https://docs.rs/wgp/latest/wgp/ - seems to work, but uses unsafe and has [had correctness bug](https://www.github.com/Nugine/wgp/issues/1).
- https://docs.rs/crossbeam-utils/latest/crossbeam_utils/sync/struct.WaitGroup.html, https://docs.rs/omango/latest/omango/wg/struct.WaitGroup.html but this is for threading rather than async. 
- lots of things for waiting on a set of futures, but we can't use that here because we don't have individual features for each task https://docs.rs/tokio-util/latest/tokio_util/task/task_tracker/struct.TaskTracker.html, https://docs.rs/tokio/1.49.0/tokio/task/join_set/struct.JoinSet.html#method.join_all, https://docs.rs/futures/latest/futures/stream/struct.FuturesUnordered.html, https://docs.rs/spawn_groups/latest/spawn_groups/struct.SpawnGroup.html


lychee-bin/src/commands/check.rs
```diff
@@ -204,7 +232,7 @@ async fn request_channel_task(
             .await;
 
             send_resp
-                .send(response)
+                .send((guard, response))
```
mre, line 225: What are your thoughts on introducing a type for that? Seems like it's a central part of the design so from my perspective it merits its own name.

```rust
type Work<T> = (WaitGuard, T);
```

We could also call it "Envelope" or "Guarded" or so.

katrinafyi: Hmm I don't think it's needed. I think `(WaitGuard, ...)` already gives a lot meaning to the reader, and an alias like Work might even hide that a little bit.

I was wondering if I could make WaitGuard take some kind of parameter to indicate which WaitGroup it came from. Something like `WairGuard<"work">`. At the moment, all WaitGroups produce WaitGuards that look the same. Do you think that's worthwhile?

mre: Yeah makes sense. Let's keep it a tuple then.

lychee-lib/src/waiter.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// If needed, new guards should be created by cloning the returned [`WaitGuard`].
+    #[must_use]
+    pub fn new() -> (Self, WaitGuard) {
+        let (send, recv) = channel(1);
+        (Self { recv }, WaitGuard { _send: send })
+    }
+
+    /// Waits, asynchronously, until all the associated [`WaitGuard`]s have finished.
+    pub async fn wait(mut self) {
+        let None = self.recv.recv().await;
```
thomas-zahner, line 157: Huh, that's an interesting line. What does the `let None =` exactly mean?

katrinafyi: It means that the right hand side always returns None, same as `let () =`. It's guaranteed to be None because Some([Infallible](https://doc.rust-lang.org/std/convert/enum.Infallible.html)) is impossible, and the compiler can infer this.

Conceptually, this means that no real values will ever be received through the channel, so recv will only return when all senders are closed. When that happens, it returns with None.

It's a fun one :) (maybe too fun)

thomas-zahner: That's quite amazing. So instead of ignoring the return value we do a compile time type assertion. I definitely like it.

thomas-zahner:
Very cool PR! I don't see any problems with it.

@katrinafyi Could you add a test to waiter.rs to to test if recursion would really work that way? So that the unused part (`// unused for now, but will be used for recursion eventually.`) is covered and tested already. This would be great documentation and help me (and others) understand how this is used in a simpler scenario without lychee's additional complexity.

lychee-lib/src/waiter.rs
mre: Maybe I should add that to [this blog post](https://corrode.dev/blog/defensive-programming/). 🤔

katrinafyi:
@thomas-zahner That's a really interesting idea. I've added a (hopefully) simple example which you can see here: https://github.com/lycheeverse/lychee/compare/74046a9..b27a34f

lychee-lib/src/waiter.rs
thomas-zahner: Yeah, that would be very fitting :)

The minimal example is `let None: Option<Infallible> = None;` and `let None: Option<Never> = None;`

lychee-lib/src/waiter.rs
```diff
@@ -0,0 +1,184 @@
...
+                send.send((guard.clone(), n - 1)).expect("send 1");
+                send.send((guard.clone(), n - 2)).expect("send 2");
+            }
+        })
+        .await;
+}
+
+#[cfg(test)]
+mod tests {
+
```
thomas-zahner:
```suggestion
```

lychee-lib/src/waiter.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+/// Demonstrates use of the [`WaitGroup`] and [`WaitGuard`] to (very inefficiently)
+/// compute the Fibonacci number `F(n)` using recursive channels.
+///
+/// The given `waiter` will be used to detect when the work has finished and it will
+/// close the channels. Additionally, `waiter` can be omitted to show that without
+/// the [`WaitGroup`], the tasks would not terminate.
+#[allow(dead_code)]
```
thomas-zahner: Either move into `mod test` or annotate with `#[cfg(test)]`, this should resolve the `#[allow(dead_code)]`, right?

lychee-lib/src/waiter.rs
```diff
@@ -0,0 +1,184 @@
...
+/// An inefficient Fibonacci implementation. This computes `F(n)` by sending
+/// by `n-1` and `n-2` back into the channel. This shows how one work item can
+/// create multiple subsequent work items.
+///
+/// The result is determined by sending `()` into an increment channel and
+/// reading the number of increments.
+///
+/// This is wildly inefficient because it does not cache any results. Computing
+/// `F(n)` would generate `O(2^n)` channel items.
+#[allow(dead_code)]
```
thomas-zahner: Same here

lychee-lib/Cargo.toml
```diff
@@ -64,6 +64,7 @@ shellexpand = "3.1.1"
 strum = { version = "0.27.2", features = ["derive"] }
 thiserror = "2.0.18"
 tokio = { version = "1.49.0", features = ["full"] }
+tokio-stream = "0.1.18"
```
thomas-zahner: This is only depended on in the new test, right? If so, move it to the `[dev-dependencies]` group

thomas-zahner (commented):
@katrinafyi Thank you, that's very cool! This helps my understanding quite a bit.

lychee-lib/src/waiter.rs
katrinafyi: Hmm I put it outside of `mod tests` because I wanted it to be included in the docs. 

I tried a few things and it's a bit tricky. We need compiling to work (1) normally, (2) in docs, and (3) in tests.
- `#[cfg(any(test, doc))]` on the examples: dev-dependencies is not available in `doc`, so tokio-stream has to be in dependencies. but then, it's flagged as an unused crate in normal compilation.
- putting inside mod test and making test module `#[cfg(any(test, doc))]` has the same problems
- 
Do you have thoughts or other suggestions?

thomas-zahner: Hmm, I personally would really like to not have additional dependencies or code that is tests and documentation but possibly ends up in the final binary. (though in reality it should be optimised away)

I think [doc tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) would solve the problem. They are tested and great for documentation. It would be possible to highlight only the interesting bits of the code. It even seems possible to use dev dependencies in there!

katrinafyi: I just moved it into doctest and it looks great and works well, thanks for the suggestion! I should try and use doctests more often I think.

katrinafyi:
idk why that closed

lychee-lib/src/waiter.rs
thomas-zahner: Awesome :rocket: yeah they are quite powerful and useful. Thanks for addressing my concerns.

thomas-zahner: I have to admit that I'm confused by `#[cfg(all(test, not(doctest)))]`. Why is it `not(doctest)`? I see that it seems like the only working version where clippy doesn't complain. But do you understand why?

katrinafyi: I admit I didn't try other options very rigorously. But the thinking is that the unused crate warning appears when the crate is included but not used. Dev dependencies are enabled in test, and in doctest they get used.

So, test and not doctest is precisely the case where it's included but not used. That's the idea at least.

ETA: The not doctest is just adding a bit more precision. The line suppresses unused crate warning but in doctest, the crate is used so it should be unnecessary to suppress the warning.

thomas-zahner: Aah, yes I see that makes sense. Thanks for the great PR!

PR #2047, opened 2026-02-14
Add `make readme` to generate and update the readme's `--help` text

katrinafyi:
After https://github.com/lycheeverse/lychee/pull/2007, the options.rs help text would have `///` in front of each line making it harder to copy.

### DIFF
--- a/scripts/update_readme.py
+++ b/scripts/update_readme.py
@@ -17,20 +17,14 @@ def main():
     begin = '\n```help-message\n'
     end = '\n```\n'
 
-    with open('README.md', 'r+') as f:
+    with open('README.md', 'r') as f:
         text = f.read()
-        before, after = text.split(begin, 1)
-        _, after = after.split(end, 1)
+        before, rest = text.split(begin, 1)
+        _, after = rest.split(end, 1)
 
-        f.seek(0)
-
-        f.write(before)
-        f.write(begin)
-        f.write(new_help)
-        f.write(end)
-        f.write(after)
-
-        f.truncate()
+    with open('README.md', 'w') as f:
+        for part in [before, begin, new_help, end, after]:
+            f.write(part)
 
 if __name__ == "__main__":
     main()


scripts/update_readme.py
```diff
@@ -0,0 +1,36 @@
...
+
+import subprocess
+
+def main():
+    new_help = subprocess.check_output('cargo run -- --help'.split(), encoding='utf-8')
+    version = subprocess.check_output('cargo run -- --version'.split(), encoding='utf-8').split()[-1]
+    new_help = '\n'.join(
+        line.replace(f'lychee/{version}', 'lychee/x.y.z').rstrip()
+        for line in new_help.strip().split('\n')
+    )
```
mre: I find this a bit hard to read. Can we split it up into multiple steps?

```python
lines = help_output.strip().splitlines()

normalized = [
    line.rstrip().replace(f"lychee/{version}", "lychee/x.y.z")
    for line in lines
]

new_help = "\n".join(normalized)
```

mre:
Thanks, lgtm.

I want to mention that by merging this we would take on a dependency to Python. Might not be a big deal, given that every major OS ships it by default and it's also in GitHub actions, but the rest of the maintenance scripts are written in bash so far.

Not saying that we should rewrite it in bash. I actually tried and it looked awful. 😆 So, just an observation.

katrinafyi:
I've just pushed two commits to simplify it a bit, let me know what you think.

Re using Python, yeah it's true. But it's only a dev dependency and even then, it's only needed if the developer wants this convenience - you can always copy paste. 

For fun, I had a go at writing it in sed and it was also less readable :")
```sh
sed '
  /^```help-message$/,/^```$/ {
    /^```help-message$/ {
      r help.txt
      n
    }
    /^```$/ n
    d
  }
  ' README.md
```

mre (approved):
This looks great. Thanks for the simplifications.

Shall we add a test to see if it works? And I'm actually wondering if we should make that part of the release pipeline. 

But these are just ideas. We can also just merge it as is and come back to that later (or not at all)

katrinafyi:
Hmm, just thinking how it might interact with the existing unit test.  Maybe there could be a ci task that generates the readme and pushes it, then we don't need the unit test and the developer doesn't have to think about it at all. 

WDYT

mre:
Yeah, that would be cool. Maybe we should keep the changes minimal for now? We could do that in a separate PR?

katrinafyi:
Yeah a separate PR sounds good

mre:
Cool. I merged it. Great improvement. 😃

Issue #614, opened 2022-05-10
Support SARIF as an Output Format?


aj-stein-nist:
OK, this time I think I got the right repo. This is a little but more complicated than "just supporting a new output format" (see below) but it would give an increasing edge to make it perform like (ironically) the very specific static analysis tool it is.

[SARIF is a standard](https://docs.oasis-open.org/sarif/sarif/v2.0/sarif-v2.0.html) worth considering.

Why consider it? [For services like GitHub that we use now as I write and you type](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning), we can have link information overlay on Markdown or HTML within the repo. [One open source project has some screenshots](https://docs.kics.io/1.3.1/integrations_ghactions/) of how it would look.

Yes, I think it would require tracking lines and offsets, but that also seems more generally useful (if not currently possible, gotta read the Rust code 😅).

Thoughts?

aj-stein-nist:
> I'm also one of the maintainers of [analysis-tools.dev](https://github.com/analysis-tools-dev/static-analysis) and I've never heard of SARIF. 😅 But it looks great!

Dude, do you even sleep? lol

> Only problem is that we indeed don't support racking lines and offsets. We use html5gum, and [there once was a PR](https://github.com/untitaker/html5gum/pull/14) for adding span support (line numbers, basically), but it did not get merged. Not sure if that's still on the roadmap. @untitaker fyi.

Wow this comes full circle: we actually switched from `hyperlink` to `lychee` due to a small diff in feature set and `lychee` allows me to be lazy for non-local URLs. Hi untitaker! haha 

I didn't realize they're making this library core to `lychee`, what a small world.

> Apart from that I'm all for it.

untitaker:
hey @aj-stein-nist, if there are features missing from hyperlink that you'd like to see feel free to open feature requests in its repo. Although the scope of hyperlink is pretty narrow at the moment and is unlikely to expand to external links just like that, I'm always happy to discuss ideas.

yeah, html5gum currently doesn't support linenumbers. I believe it would not be too hard to add. though you can probably start building linenumber support for markdown files right now and leave HTML for later. honestly I believe linenumbers in HTML may be less interesting anyway, assuming most HTML is auto-generated.

[...]

PR #2056, opened 2026-02-19
Show line & column numbers

thomas-zahner:
Fixes https://github.com/lycheeverse/lychee/issues/1304

We call line & column numbers `span` and `RawUriSpan` in our code. Not sure if this is the best name, as it currently isn't really a span but a position, so the span is missing the end. But maybe we could update the span to include an end. (maybe not in this PR)

I've removed html5ever as an experiment in the second commit, to show that it makes it possible to unwrap the column. It seems like we tried to not rely too much on html5ever and the code comments discourage its usage. Maybe it's a good point to remove it then, but I'm not quite sure and don't now the background story. @mre What do you think?

Then, I'm also open to discuss how the output looks like. Do you have a better proposal? The current one seems simple and quite straight forward, basically minimal changes.

Note that @mihaigalos's [suggestion](https://github.com/lycheeverse/lychee/issues/1304#issue-2005161766) with `[./README.md:123:54]:` would make the current grouping by input source impossible or it would lead to repetition of the input source.

### DIFF
--- a/lychee-lib/src/types/request.rs
+++ b/lychee-lib/src/types/request.rs
@@ -4,8 +4,8 @@ use crate::{BasicAuthCredentials, ErrorKind, Uri, types::uri::raw::RawUriSpan};
 
 use super::ResolvedInputSource;
 
-/// A checkable element extracted from a document by lychee,
-/// containing a URI and its location within the source.
+/// A checkable element extracted from a [`ResolvedInputSource`],
+/// containing a [`Uri`] and additional metadata associated with it.
 #[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub struct Request {
     /// The extracted Uniform Resource Identifier
--- a/lychee-bin/src/formatters/stats/detailed.rs
+++ b/lychee-bin/src/formatters/stats/detailed.rs
@@ -20,9 +20,7 @@ const WIDTH: usize = 20;
 fn write_stat(f: &mut fmt::Formatter, title: &str, stat: usize, newline: bool) -> fmt::Result {
     f.write_str(title)?;
 
-    let stat = stat.to_string();
     let spacing = WIDTH.saturating_sub(title.chars().count());
-
     f.write_str(format!("{stat:.>spacing$}").as_str())?;
 
     if newline {
--- a/lychee-bin/src/formatters/stats/mod.rs
+++ b/lychee-bin/src/formatters/stats/mod.rs
@@ -81,6 +81,8 @@ where
 
 #[cfg(test)]
 fn get_dummy_stats() -> OutputStats {
+    use std::num::NonZeroUsize;
+
     use http::StatusCode;
     use lychee_lib::{RawUriSpan, Redirect, Redirects, ResponseBody, Status, ratelimit::HostStats};
     use url::Url;
@@ -96,8 +98,8 @@ fn get_dummy_stats() -> OutputStats {
                 .unwrap(),
             status: Status::Ok(StatusCode::NOT_FOUND),
             span: Some(RawUriSpan {
-                column: Some(1.try_into().unwrap()),
-                line: 1.try_into().unwrap(),
+                column: Some(NonZeroUsize::MIN),
+                line: NonZeroUsize::MIN,
             }),
         }]),
     )]);


lychee-bin/src/commands/check.rs
```diff
@@ -305,8 +312,13 @@ fn get_failed_urls(stats: &mut ResponseStats) -> Vec<(InputSource, Url)> {
...
         .iter()
         .flat_map(|(source, set)| {
-            set.iter()
-                .map(move |ResponseBody { uri, status: _ }| (source, uri))
+            set.iter().map(
+                move |ResponseBody {
+                          uri,
+                          status: _,
+                          span: _,
+                      }| (source, uri),
```
katrinafyi:
```suggestion
                move |body| (source, body.uri),
```

lychee-bin/tests/cli.rs
```diff
@@ -2244,14 +2213,20 @@ The config file should contain every possible key for documentation purposes."
...
             .arg("-vv")
             .arg(input)
             .assert()
             .failure();
 
+        let output = std::str::from_utf8(&result.get_output().stdout).unwrap();
+        let json: Value = serde_json::from_str(output).unwrap();
+
+        let actual_successes = extract_urls(&json["success_map"]);
+        let actual_errors = extract_urls(&json["error_map"]);
```
katrinafyi, line 2228: Thanks for cleaning up this test.

lychee-bin/tests/cli.rs
```diff
@@ -609,43 +613,6 @@ mod cli {
...
-            .env_clear()
-            .assert()
-            .success()
-            .stdout(contains("1 Total"))
-            .stdout(contains("1 OK"));
-    }
-
-    #[test]
-    // Test that two identical requests don't get executed twice.
-    fn test_caching_across_files() -> Result<()> {
```
katrinafyi, line 628: Why are the tests deleted?

edit: I see https://github.com/lycheeverse/lychee/pull/2056/commits/ab29984ab454fa602b139762992f231181a69461 and it makes sense.

katrinafyi (approved):
Looks reasonable and seems to be just wiring through the source locations into the output.

About the formatting, there are reasons to prefer `filename.md:4:1` if possible. Some editors or consoles can turn that into a clickable link. I'm not intrinsically opposed to repeating the filename (compilers and linters just do that, but often with fewer reports).

In any case, about the current format, I think it is a little hard to notice because it's two numbers with little context and placed after a possibly very long URL. 

Here is one very loose suggestion: It contextualises the numbers with the filename (but not the full path), and it moves them closer to the left. This makes them more noticeable and doesn't take up too much space since numbers are usually small.
```md
[a/x/b/a/a.md]:
   [ERROR] (a.md:39:1) https://a.com/#asd  | Error (cached)
     [404] (a.md:4:100) https://google.com/ajfidsajfidosa | Rejected status code: 404 Not Found (configurable with "accept" option)
```

Alternatively or additionally, we could also consider a new formatter for a more line-oriented output like grep:
```bash
grep -R struct lychee-lib --line-number
```
Image

lychee-bin/src/commands/check.rs
eread: @katrinafyi @thomas-zahner Keeping possible editor integration in mind is a great idea. Lychee could potentially act as an in-editor linter along with other Markdown linters. You can see, for example, what a SublimeLinter plugin would be looking for: https://www.sublimelinter.com/en/master/linter_plugin.html.

You could provide different line/column number formatting dependent on `--format`. An editor that was looking at a file at a time would only need incredibly simple output.

thomas-zahner: Thanks for your thoughts @eread 
Note that for any programmatic reuse of lychee's output there is `--format json`. So for e.g. building a sublime plugin one could use lychee's JSON output. This PR also adds the span information for JSON output. Alternatively, even better if possible, one could use lychee-lib

thomas-zahner: @katrinafyi Thanks for the suggestion, that's much better

thomas-zahner:
@katrinafyi Thank you for your thoughts and the output suggestion. I totally agree that we could consider adding a more grep like output format.

Thinking about it, it seems like there is a lot we could do. As said currently the "span" is only a location, as it's missing the an end point. Not sure how easy it is to update it to become a "real" span with an end point. If we had this we could also do the what grep does; to reprint the affected input line and highlight the broken URL. Even better is the more modern `rg` approach which groups the lines by file.

<details>
<summary>rg output</summary>

```
lychee-lib/src/types/status.rs
77:            s = serializer.serialize_struct("Status", 2)?;
81:            s = serializer.serialize_struct("Status", 2)?;
85:            s = serializer.serialize_struct("Status", 1)?;
```

</details>

But this is not yet possible because we're missing the endpoint of the span, but it would be great to have such an output format in the long run. We could even try to mimic the output of a compiler.

```
404: Rejected status code: 404 Not Found
  --> fixtures/TEST.md:71:9
   |
71 |         This is a [link](https://example.com).
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

For now in this PR I'd like to keep it simple and not implement a new format or the span endpoint. I'd also prefer to not changing the default output fundamentally.

> I think it is a little hard to notice because it's two numbers with little context and placed after a possibly very long URL

Totally agree.

What about the following? Is it still intuitively understandable? It's just that I think it might not be worth repeating the file name (as in `(a.md:39:1)`) because it will not make a clickable link as it's an incomplete path. (almost like generating broken links :thinking:)

```
[a/x/b/a/a.md]:
   [ERROR] (at 39:1) https://a.com/#asd  | Error (cached)
     [404] (at 4:100) https://google.com/ajfidsajfidosa | Rejected status code: 404 Not Found (configurable with "accept" option)
```

katrinafyi:
@thomas-zahner The `(at 39:1)` format looks good to me! I also agree that JSON format can be used for more specialised transformations.

About other output formats, yeah there's a lot we could do here and it will probably take longer to work out what's best. I do like the conciseness of rg, but we have a bit more information to display for each link. The compiler-like output gives us lots of space for extra details, but it might also be _too_ spacious. I also think that the `^` underline is less effective for lychee than compilers - the link is usually obvious and we won't be pointing to subtle details like a missing semicolon.

Lots to think about later :)

mre:
Great PR.

In regard to the formatting, the one thing I don't like is that the position info breaks URL alignment, making the output harder to scan.

I would like to discuss two additional options:

Option one puts the position after the URL:

```md
[a/x/b/a/a.md]:
   [ERROR] https://a.com/#asd (at 39:1) | Error (cached)
     [404] https://google.com/ajfidsajfidosa (at 4:100) | Rejected status code: 404
```

URLs stay front and center, which is arguably what matters most. Alignment is a non-issue since the position just floats at the end. The downside is that position info gets buried after a long URL and is easy to miss.

Option two pads the position prefix up to a threshold, so URLs still line up within each group:

```md
[a/x/b/a/a.md]:
   [ERROR] (at 39:1)  https://a.com/#asd | Error (cached)
     [404] (at 4:100) https://google.com/ajfidsajfidosa | Rejected status code: 404
```

Position is prominent and easy to scan down the left side. It only requires one pass per file group to compute the padding, not the whole output. The threshold cap (say, up to `999:99`) prevents weird wide gaps in edge cases, with the occasional misalignment beyond that being acceptable.

Any thoughts?

lychee-bin/src/commands/check.rs
```diff
@@ -304,10 +311,7 @@ fn get_failed_urls(stats: &mut ResponseStats) -> Vec<(InputSource, Url)> {
     stats
         .error_map
         .iter()
-        .flat_map(|(source, set)| {
-            set.iter()
-                .map(move |ResponseBody { uri, status: _ }| (source, uri))
-        })
+        .flat_map(|(source, set)| set.iter().map(move |body| (source, &body.uri)))
```
mre, line 314: Neat!

lychee-bin/src/main.rs
```diff
@@ -336,8 +336,6 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
         .include_verbatim(opts.config.include_verbatim)
         .headers(HeaderMap::from_header_pairs(&opts.config.headers())?)
         .excluded_paths(PathExcludes::new(opts.config.exclude_path.clone())?)
-        // File a bug if you rely on this envvar! It's going to go away eventually.
-        .use_html5ever(std::env::var("LYCHEE_USE_HTML5EVER").is_ok_and(|x| x == "1"))
```
mre, line 340: The nice thing about html5ever is that it has recently been picking up momentum again thanks to the Servo project. E.g. see the [recent commits](https://github.com/servo/html5ever/commits/main/). It's also well-documented and has a higher-level API, which makes it easier to work with. Supporting multiple engines reduces reliance on a single implementation, which is always a good thing.

On the other side, I understand that html5ever doesn't provide spans yet and the respective issue is open for a few years already, so we shouldn't hold our breath. Maybe it's indeed the best to simply remove html5ever to keep the code tidy. 🧹

lychee-lib/src/extract/css.rs
```diff
@@ -436,4 +433,9 @@ mod tests {
         assert_eq!(urls[0].element, Some("style".to_string()));
         assert_eq!(urls[0].attribute, Some("url".to_string()));
     }
+
+    /// Extract URLs from CSS content with default span
+    fn extract_css_with_default_span(input: &str) -> Vec<RawUri> {
+        extract_css(input, &SourceSpanProvider::from_input(input))
+    }
```
mre: Thanks for moving that into `#[cfg(test)]`.

lychee-lib/src/types/uri/raw.rs
```diff
@@ -127,14 +124,13 @@ impl SpanProvider for SourceSpanProvider<'_> {
         let column = self
             .input
             .get(line_offset..offset)
-            .or_else(|| self.input.get(line_offset..))
-            // columns are 1-based
-            .map(|v| ONE.saturating_add(v.chars().count()));
+            .unwrap_or_else(|| &self.input[line_offset..]); // As line_offset marks the beginning of a line it shouldn't panic.
```
mre: It's true that this shouldn't panic *right now*, but the tricky bit is if we accidentally break invariants during refactoring.

I was wondering if we should simply default to the end of the line instead of panicking? The change is pretty minimal.

```rust
fn span(&self, offset: usize) -> RawUriSpan {
        const ONE: NonZeroUsize = NonZeroUsize::MIN;
        let line = match self.line_starts.binary_search(&offset) {
            Ok(i) => i,
            Err(i) => i - 1,
        };

        // Since we get the index by the binary_search above and subtract `1` if it would be larger
        // than the length of the document, this shouldn't panic.
        let line_offset = self.line_starts[line];
        
        // Extract the slice of the line up to the offset, defaulting to the end of the line if necessary.
        let line_fragment = self.input.get(line_offset..offset)
            .unwrap_or_else(|| &self.input[line_offset..]);

        RawUriSpan {
            // lines are 1-based
            line: ONE.saturating_add(line),
            // columns are 1-based, counting Unicode characters rather than raw bytes
            column: ONE.saturating_add(line_fragment.chars().count()),
        }
    }
```

lychee-lib/src/types/request.rs
```diff
@@ -1,27 +1,31 @@
 use std::{borrow::Cow, convert::TryFrom, fmt::Display};
 
-use crate::{BasicAuthCredentials, ErrorKind, Uri};
+use crate::{BasicAuthCredentials, ErrorKind, Uri, types::uri::raw::RawUriSpan};
 
 use super::ResolvedInputSource;
 
-/// A request type that can be handle by lychee
+/// A URI which was extracted by lychee.
```
mre: Hm, it's more than just a URI, is it? The URI is just one field. Maybe:

> A checkable element extracted from a document, containing a URI and its location within the source.

lychee-lib/src/types/request.rs
```diff
@@ -1,27 +1,31 @@
...
     pub source: ResolvedInputSource,
 
-    /// Specifies how the URI was rendered inside a document
+    /// How the URI is rendered inside a document
     /// (for example `img`, `a`, `pre`, or `code`).
     /// In case of plaintext input the field is `None`.
     pub element: Option<String>,
 
-    /// Specifies the attribute (e.g. `href`) that contained the URI
+    /// What attribute (e.g. `href`) URI is contained in
```
mre:
```suggestion
    /// What attribute (e.g. `href`) the URI is contained in
```

lychee-lib/src/types/request.rs
```diff
@@ -63,6 +69,7 @@ impl TryFrom<Uri> for Request {
             None,
             None,
             None,
+            None,
```
mre: The list of "empty default values" is becoming concerning. 😬

mre: We could introduce our own "option" types to make this more declarative.
Here is how that would look like:

```rust
Request::new(
    uri.clone(),
    ResolvedInputSource::RemoteUrl(Box::new(uri.url)),
    Element::Unknown,
    Attribute::Unknown,
    Credentials::None,
    Span::Empty,
)
```

I'll let you be the judge if that is any cleaner.

mre: As a first step, we could also be more explicit about the field names. 
Maybe that's already good enough.

```rust
Request {
    uri: uri.clone(),
    source: ResolvedInputSource::RemoteUrl(Box::new(uri.url)),
    element: None,
    attribute: None,
    credentials: None,
    span: None,
)
```

lychee-bin/src/main.rs
thomas-zahner: Thank you for the details. I didn't really think this through it just looked nicer in my implementation to get rid of html5ever as I could unwrap the column field. But I definitely see how keeping the additional parser implementation would be nice, it also seems like a waste to just throw away all the good work, that's why I also commented it as an "experiment" in the commit message. So I got it back with 439a473df3b6d48d69f144a60cee57f6913946b4

> I understand that html5ever doesn't provide spans yet

Actually it's much better, it does provide spans! It's just it doesn't yet provide columns, only the line number. But it might even be possible to do that, not quite sure.

lychee-lib/src/extract/css.rs
thomas-zahner: Reverted with 439a473df3b6d48d69f144a60cee57f6913946b4 as it's used in html5ever

katrinafyi:
Another formatting suggestion, because I don't love the line number being put in the middle of the status information. This is a bit like ripgrep's output with line numbers at the beginning of lines.
```
[a/x/b/a/a.md]:
39:1: [ERROR] https://a.com/#asd | Error (cached)
4:100:  [404] https://google.com/ajfidsajfidosa | Rejected status code: 404
```
It might be complicated to do this left/right justification of the error and line number where they can take space from each other. But I think it looks a bit nicer than having gaps on both sides for alignment.

I would also be okay with @ mre's first suggestion:
```
   [ERROR] https://a.com/#asd (at 39:1) | Error (cached)
```

lychee-lib/src/types/uri/raw.rs
thomas-zahner: Well it's basically resolved by reverting (439a473d). The original implementation fell back on `None` if the column couldn't be calculated. So we use `None` if it cannot be calculated. Or do you want to change it still?

lychee-lib/src/types/request.rs
thomas-zahner: Thanks for the idea. Personally I don't think it's really cleaner to just introduce new types for `Option<String>`s in this case. So I prefer your second approach and took it a bit further and simplify `Request`'s usage a bit. See 1069ae96c44fbdc188cf7839e79cf1e81fdb40a2

lychee-lib/src/types/request.rs
```diff
@@ -1,50 +1,78 @@
...
+    /// Set [`Self::attribute`]
+    #[must_use]
+    pub fn with_attribute(mut self, attribute: String) -> Self {
+        self.attribute = Some(attribute);
+        self
+    }
+
+    /// Set [`Self::span`]
+    #[must_use]
+    pub const fn with_span(mut self, span: RawUriSpan) -> Self {
```
thomas-zahner, line 65: Interestingly we can mark this fn const but the other ones fail with [E0493](https://doc.rust-lang.org/error_codes/E0493.html#error-code-e0493)

thomas-zahner:
> Another formatting suggestion [...] This is a bit like ripgrep [...]

I really like the idea and tried it in https://github.com/thomas-zahner/lychee/commit/e329eb490c2d35833c83ca70ebfcb64841c28a3d. But as mentioned in the commit message it screws up a few things, like task formatting. Basically this concept doesn't really fit in the current format/mode concept. Instead we could maybe make a dedicated format for ripgrep's approach. But this would preferably be a separate PR.

> I would also be okay with @ mre's first suggestion:

Cool, that is much easier to implement. See a5e5cd18

thomas-zahner:
So I'm going to merge this. This allows me to tackle https://github.com/lycheeverse/lychee/issues/614. In summary I tried to keep the changes to the output formats minimal while also being a helpful improvement. In the future it would be pretty cool to try pursue the idea of ripgreps format I started in https://github.com/thomas-zahner/lychee/commit/e329eb490c2d35833c83ca70ebfcb64841c28a3d.

PR #1931, opened 2025-11-21
chore: add pypi release

kemingy:
- Add PyPI release, as listed in https://github.com/lycheeverse/lychee/issues/59

This is blocked by the rustls issue, tracked in https://github.com/lycheeverse/lychee/issues/1721

```sh
❯ cargo tree --no-default-features --features rustls-tls --features email-check -i openssl
openssl v0.10.73
└── native-tls v0.2.14
    ├── async-native-tls v0.4.0
    │   ├── async-smtp v0.6.0
    │   │   └── check-if-email-exists v0.9.1
    │   │       └── lychee-lib v0.21.0 (/home/user/GitHub/lychee/lychee-lib)
    │   │           ├── archive v0.1.0 (/home/user/GitHub/lychee/examples/archive)
    │   │           ├── benches v0.0.0 (/home/user/GitHub/lychee/benches)
    │   │           ├── builder v0.1.0 (/home/user/GitHub/lychee/examples/builder)
    │   │           ├── chain v0.1.0 (/home/user/GitHub/lychee/examples/chain)
    │   │           ├── client_pool v0.1.0 (/home/user/GitHub/lychee/examples/client_pool)
    │   │           ├── collect_links v0.1.0 (/home/user/GitHub/lychee/examples/collect_links)
    │   │           ├── extract v0.1.0 (/home/user/GitHub/lychee/examples/extract)
    │   │           ├── lychee v0.21.0 (/home/user/GitHub/lychee/lychee-bin)
    │   │           └── simple v0.1.0 (/home/user/GitHub/lychee/examples/simple)
    │   └── check-if-email-exists v0.9.1 (*)
    ├── hyper-tls v0.5.0
    │   └── reqwest v0.11.27
    │       ├── check-if-email-exists v0.9.1 (*)
    │       └── pwned v0.5.0
    │           └── check-if-email-exists v0.9.1 (*)
    ├── hyper-tls v0.6.0
    │   └── reqwest v0.12.24
    │       ├── builder v0.1.0 (/home/user/GitHub/lychee/examples/builder)
    │       ├── chain v0.1.0 (/home/user/GitHub/lychee/examples/chain)
    │       ├── collect_links v0.1.0 (/home/user/GitHub/lychee/examples/collect_links)
    │       ├── lychee v0.21.0 (/home/user/GitHub/lychee/lychee-bin)
    │       ├── lychee-lib v0.21.0 (/home/user/GitHub/lychee/lychee-lib) (*)
[...]

PR #2057, opened 2026-02-20
pre-commit hook: use pre-commit's built-in rust boostrapping

cconverse711:
When trying to use the `lychee` pre-commit hook at tag `lychee-v0.23.0`,  I get:

```
lychee...................................................................Failed
- hook id: lychee
- exit code: 127

  Installing lychee...
  Using cargo install (this may take a few minutes)...
  /root/.cache/prek/repos/2c023200f7174f5a/scripts/lychee_pre_commit.sh: line 18: cargo: command not found
```

because I don't have cargo installed.

Pre-commit supports [bootstrapping rust](https://pre-commit.com/#rust) so while this new hook that leverages `cargo install` under the hood may be slower upon first invocation than `cargo binstall`, it avoids all the hoops currently begin jumped through by `lychee_pre_commit.sh`, and ensures that the rust environment is fully set up and available.

Edit: I see that [this issue](https://github.com/pre-commit/pre-commit/issues/2931) has been causing problems and still remains a problem with my changes, however [prek](https://github.com/j178/prek) does seem to handle bootstrapping with workspaces as desired, so I still think my proposed changes are worthwhile. Not sure of the best way to indicate that this hook only works with the pre-commit replacement rather than the original, though...  

Alternatively, once #1931 is resolved, that may allow for a similar approach to how [`typos` manages it](https://github.com/crate-ci/typos)

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,6 +1,19 @@
 - id: lychee
   name: lychee
-  description: "Run 'lychee' for fast, async, stream-based link checking (auto-installs)"
+  description: "Run 'lychee' for fast, async, stream-based link checking (auto-installs from binary)"
+  entry: scripts/lychee_pre_commit.sh
+  language: script
+  # defaults to the version of the current repo checkout, i.e., the 'rev' field in
+  # .pre-commit-config.yaml. to use a different lychee version, pass "LYCHEE_VERSION=0.xx.0"
+  # as the first arg in 'args'.
+  args: ["--no-progress"]
+  types: [text]
+  pass_filenames: true
+  require_serial: true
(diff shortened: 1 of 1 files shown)


katrinafyi:
Have you seen this issue and in particular this comment https://github.com/lycheeverse/lychee/issues/1947#issuecomment-3920623368?

I touched this script very recently (after 0.23.0), and it should support classic pre-commit and it should also support downloading without cargo. Follow the instructions in the comment to test, setting rev to the latest commit and `LYCHEE_VERSION=0.23.0`.

Let me know if it works for you.

mre:
@cconverse711, any updates?

thomas-zahner:
I assume this has been fixed by now. Feel free to provide more details if this is not the case.

Issue #2053, opened 2026-02-18
Feature request: allow accepting/ignoring timeout errors


yegor256:
## Problem

Currently, the `--accept` flag only supports HTTP status codes. There is no way to ignore connection-level errors like timeouts.

When checking links across a large site, some external URLs may intermittently timeout due to network conditions or slow servers. These transient failures cause CI to fail even though the links are valid.

## Proposed Solution

Extend the `accept` configuration to support error types in addition to HTTP status codes. For example:

```toml
accept = [200, 429, "timeout"]
```

Or add a separate option:

```toml
skip_errors = ["timeout", "dns"]
```

## Use Case

I run lychee in CI to check links in a blog with hundreds of posts. Some external sites occasionally timeout, causing the entire CI run to fail. I would like to:

1. Report timeouts as warnings (not errors)
2. Or ignore them entirely

Currently, the only workarounds are:
- Increasing timeout and retries (doesn't fully solve the problem)
- Excluding specific URLs (requires maintenance as new timeout-prone URLs appear)
- Post-processing lychee output in CI scripts

## Alternatives Considered

- Excluding URLs via `.lycheeignore` — requires knowing which URLs will timeout in advance
- Increasing `timeout` and `max_retries` — helps but doesn't eliminate the issue
- Filtering output with grep in CI — fragile workaround

cristiklein:
In case it makes implementation easier, I could also imagine mapping various errors to HTTP Status Codes >=600. For example, TIMEOUT could be 601. [Rust documentation](https://docs.rs/http/latest/http/status/struct.StatusCode.html) seems to allow this:

> Values above 599 are unclassified but allowed for legacy compatibility, though their use is discouraged. Applications may interpret such values as protocol errors.

In that case, the user would simply set:

```
accept = [200, 429, 601]
```

thomas-zahner:
> For example, TIMEOUT could be 601.

We should really try to avoid such magic sentinel values. In fact lychee is intended to work with with non-standard/non-conformant status codes such as 601. A prominent example and use case is linkedin which returns the status code 999 (depending on the user-agent if I remember correctly). In this case the user can allow lychee to accept the code with e.g. `--accept 200..300,999`.

Only because it's neither recommended by the spec nor by the http crate doesn't mean it's invalid. But this is more of a HTTP spec problem.. So we specifically must be able to work with 100-999.

[...]

PR #2063, opened 2026-02-25
feat: allow accepting timeouts

cristiklein:
* Add `--accept-timeouts` which means timeouts are accepted as valid links and lychee exits with 0.
* This feature can also be used from `lychee.toml` by using `accept_timeouts = true`
* The output is unchanged, i.e., the user can still see the number of timeouts.
* Add test for this feature

Thanks to @yegor256, @mre and @thomas-zahner for the kind words and feedback.

### DIFF
--- a/lychee.example.toml
+++ b/lychee.example.toml
@@ -191,7 +191,7 @@ include_wikilinks = true
 offline = true
 
 # Accept timed out requests
-accept_timeouts = false
+accept_timeouts = true
 
 #############################  Other  #############################
 


cristiklein:
@yegor256 Wanna take this for a spin?

yegor256:
@cristiklein would it be possible to do this in the `lychee.toml`?

```
accept = [200, 202, 301, 302, 307, 401, timeout]
```

cristiklein:
> @cristiklein would it be possible to do this in the `lychee.toml`?
> 
> ```
> accept = [200, 202, 301, 302, 307, 401, timeout]
> ```

Just tested and it also works via `lychee.toml`:


```console
$ cat blah.md 
<http://127.0.0.1:8080>
$ cat lychee.toml 
accept = [200, 202, 301, 302, 307, 401, "timeout"]
timeout = 1
$ ./target/debug/lychee --config lychee.toml blah.md
1/1 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                       Issues found in 1 input. Find details below.

[blah.md]:
 [TIMEOUT] http://127.0.0.1:8080/ | Timeout

🔍 1 Total (in 11s) ✅ 0 OK 🚫 0 Errors ⏳ 1 Timeouts

$ echo $?
0
```

thomas-zahner:
Thank you for the PR. As mentioned [over on the issue](https://github.com/lycheeverse/lychee/issues/2053#issuecomment-3958468930) we should not use such magic/sentinel values.

Reusing `--accept` might not be ideal in this case. Tough it would be possible, if we wrapped the accept type in a newtype or struct/enum.

It might be simpler and more straight forward to introduce `--accept-timeouts` both for users and for the implementation. I'm not quite sure what all the implications of this change are. But they might be bigger than the PR in the current state.

For example all output formats still show an erroneous run.
`Issues found in 1 input. Find details below.` for `compact`,
`error_map` for `json`
`Errors in stdin` for `detailed`
`### Errors in stdin` for `markdown`

So the timeouts should maybe be marked as success as early as possible, instead of adding `is_success_ignoring_timeouts`.

mre:
Thanks for the initiative @cristiklein. It's nice to see someone taking on that task. 😊 

I agree with Thomas here. Those special values can cause hard-to-debug edge cases. I should know because we're still dealing with the aftermath of my ad-hoc decisions from the past. 😬 

So, yeah, let's make it explicit.

The lychee config would then look like

```toml
accept = [200, 202, 301, 302, 307, 401]

# Accept transient timeouts
accept_timeouts = false
```

cristiklein:
Thanks @mre and @thomas-zahner for the kind words and feedback.

I just force-pushed a new PR, which addresses your feedback.

Below is an example invocation:

```console
$ cat blah.md 
<http://127.0.0.1:8080>
$ cat lychee.toml 
accept = [200, 202, 301, 302, 307, 401]
accept_timeouts = true
timeout = 1
$ ./target/debug/lychee blah.md
1/1 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                          Issues found in 1 input. Find details below.

[blah.md]:
[TIMEOUT] http://127.0.0.1:8080/ (at 1:2) | Timeout

🔍 1 Total (in 11s) ✅ 0 OK 🚫 0 Errors ⏳ 1 Timeouts

$ echo $?
0
```

Let me know what you think.

yegor256:
@cristiklein maybe you can update the title and description of the PR, for clarity?

thomas-zahner:
@cristiklein Cool, that looks better. Also great to see the new test :+1:

Now we should think about how to display and output such accepted timeouts. Currently, they are still clearly marked as error. This might be unexpected, though maybe the current behaviour is also fine. But my thinking is that if timeouts are accepted timeouts should probably not be shown as errors, but as successes.

So for example the timeouts shouldn't show up in `error_map` for the json format. The same would be true for all other formats. Although, maybe this would make it harder to spot the timeouts? What are your thoughts @mre @katrinafyi?

katrinafyi:
I tried it out and I noticed that timeouts are still retried even if --accept-timeouts is given. This makes it take more time than necessary. Is this intentional?

Re output format, I agree that if timeouts are accepted, they shouldn't be printed amongst the errors. 

Looking at the related enums, I also think the current `enum Status` is not the best, because it mixes up two things which I think should be separated: (1) the actual outcome of the request, and (2) how that outcome should be interpreted. This makes it hard to represent cases like a Timeout that should be accepted and we also have to pass the accepted set around a lot. To implement special logic like this PR, it ends up being a lot of post-processing to "reinterpret" the status. If the two concerns were separated, it would be much easier to do this PR and other requests such as rejecting redirects.

But anyway, all that won't be fixed in this PR. In this PR, I guess you would change the lines around where it adds to the error map:

https://github.com/cristiklein/lychee/blob/6a8007ba3d1fb5de95c78c4e60126e824748340d/lychee-bin/src/formatters/stats/response.rs#L93-L106

cristiklein:
Hi @katrinafyi . Thanks for looking into the PR.

> I tried it out and I noticed that timeouts are still retried even if --accept-timeouts is given. This makes it take more time than necessary. Is this intentional?

Yes, my intention was for `--accept-timeout` to only change the exit code and nothing else. The user can run `lychee --accept-timeouts --max-retries=0 --timeout=1` if they want to minimize execution time.

> Re output format, I agree that if timeouts are accepted, they shouldn't be printed amongst the errors.

Just to make sure I understand correctly, do you argue for `lychee` to suppress outputting `[TIMEOUT]` and folding that into `OK`s?

My thinking is the following: This feature will most likely be used in a CI. In that case, the user might still want TIMEOUT in logs, but they want lychee to return 0, so their CI pipeline continues.
 
> Looking at the related enums, I also think the current `enum Status` is not the best, because it mixes up two things which I think should be separated: (1) the actual outcome of the request, and (2) how that outcome should be interpreted. This makes it hard to represent cases like a Timeout that should be accepted and we also have to pass the accepted set around a lot. To implement special logic like this PR, it ends up being a lot of post-processing to "reinterpret" the status. If the two concerns were separated, it would be much easier to do this PR and other requests such as rejecting redirects.

Just to make sure I understand correctly, are you arguing for moving accept logic to `ResponseStats`?

> But anyway, all that won't be fixed in this PR. In this PR, I guess you would change the lines around where it adds to the error map:
> 
> https://github.com/cristiklein/lychee/blob/6a8007ba3d1fb5de95c78c4e60126e824748340d/lychee-bin/src/formatters/stats/response.rs#L93-L106

Just to make sure, you propose timeouts to be added to `success_map` with `--accept-timeouts` and `error_map` without `--accept-timeouts`? Seems like timeouts are currently not added to any map.

As an alternative, would it make sense to add `timeout_map`? This would avoid having to pass `accept` to `ResponseStats.add_response_status()`.

katrinafyi:
> Just to make sure, you propose timeouts to be added to `success_map` with `--accept-timeouts` and `error_map` without `--accept-timeouts`? Seems like timeouts are currently not added to any map.

If you run
```bash
echo https://httpbin.org/delay/2 | cargo run --  - --timeout=1 --max-retries=0 --format json
```
the timeout are added to `error_map` (it's included in the `is_error()` case). I think it would be nice if `error_map` only contains the errors which would cause link checking to fail. Then, for example, `is_success` is simply `error_map.is_empty()` and we don't have to do the big sums.

However, at the moment, printing the timeouts (under the default verbosity) is tied to inclusion in the `error_map`. It would be nice if this were not the case and that _could_ be done by adding a `timeout_map` and a separate output section like redirects. But I also defer to @ thomas-zahner and @ mre for whether that is a good idea.

> Just to make sure I understand correctly, are you arguing for moving accept logic to ResponseStats?

Sorry, that paragraph was not necessarily directed to you. It's just an early idea and I'm not arguing for anything concrete.

cristiklein:
Thanks @katrinafyi for the feedback.

> > Just to make sure, you propose timeouts to be added to `success_map` with `--accept-timeouts` and `error_map` without `--accept-timeouts`? Seems like timeouts are currently not added to any map.
> 
> If you run
> 
> ```shell
> echo https://httpbin.org/delay/2 | cargo run --  - --timeout=1 --max-retries=0 --format json
> ```
> 
> the timeout are added to `error_map` (it's included in the `is_error()` case). I think it would be nice if `error_map` only contains the errors which would cause link checking to fail. Then, for example, `is_success` is simply `error_map.is_empty()` and we don't have to do the big sums.
> 
> However, at the moment, printing the timeouts (under the default verbosity) is tied to inclusion in the `error_map`. It would be nice if this were not the case and that _could_ be done by adding a `timeout_map` and a separate output section like redirects. But I also defer to @ thomas-zahner and @ mre for whether that is a good idea.

To make the alternative easier to reason about, I went ahead and moved timeouts into timeout_map as a concrete proposal. See https://github.com/lycheeverse/lychee/pull/2063/commits/3f9d495d10dd94a90685350b5f62a7b219ec4b87

I also simplified `is_success` and `is_success_ignoring_timeouts`. `is_success` now becomes a simple `error_map.is_empty() && timeout_map.is_empty()`. See https://github.com/lycheeverse/lychee/pull/2063/changes/84de85eda4279f47599a14fbc3f470f24ea004cc

Curious whether this direction makes sense conceptually, especially with regard to output verbosity and backward compatibility.

lychee-bin/src/formatters/stats/response.rs
```diff
@@ -41,6 +41,8 @@ pub(crate) struct ResponseStats {
     pub(crate) success_map: HashMap<InputSource, HashSet<ResponseBody>>,
     /// Store failed responses (if `detailed_stats` is enabled)
     pub(crate) error_map: HashMap<InputSource, HashSet<ResponseBody>>,
+    /// Store timed out responses (if `detailed_stats` is enabled)
```
thomas-zahner: The part in parentheses seems wrong. I noticed that this also is the case with `error_map` on master. I've fixed it in my junit PR, as I coincidentally touched the comments in this PR previously. See ec837a5563c4089505fb9377052d79fef0217471

```suggestion
    /// Store timed out responses
```

lychee-bin/src/formatters/stats/mod.rs
```diff
@@ -58,10 +58,36 @@ fn sort_stat_map<T>(stat_map: &HashMap<InputSource, HashSet<T>>) -> Vec<(&InputS
...
-        .map(|(source, responses)| {
-            let mut sorted_responses: Vec<&T> = responses.iter().collect();
+    merge_and_sort_stat_map(stat_map, None)
+}
+
+/// Merge two `ResponseStats` `HashMap` to a sorted Vec of key-value pairs
+/// The returned keys and values are both sorted in natural, case-insensitive order
+fn merge_and_sort_stat_map<'a, T>(
+    first: &'a HashMap<InputSource, HashSet<T>>,
+    second: Option<&'a HashMap<InputSource, HashSet<T>>>,
```
thomas-zahner: `first` and `second` seems quite arbitrary. It's better to make it a `Vec` and use `flat_map`

```
fn merge_and_sort_stat_maps<'a, T>(
    stat_maps: Vec<&'a HashMap<InputSource, HashSet<T>>>
) -> ...
```

Additionally it feels like this function now does two unrelated things, as indicated by the name. Can't we split it up into two separate functions?

cristiklein: Good point! I changed this to `sort_stat_maps` with the signature you suggested.

I agree that this function does a bit too much (merging and sorting). However, I'm not sure I found a satisfactory solution. Here are two options I explored:

1. Two `sort_stat_map`, one for HashMap-s containing owned values, the other one containing borrowed values. So the signatures would look like:

```
fn sort_stat_map<T>(stat_map: &HashMap<InputSource, HashSet<T>>) -> Vec<(&InputSource, Vec<&T>)>

fn merge_stat_maps<'a, T>(stat_maps: &'a Vec<&'a HashMap<InputSource, HashSet<T>>>) -> HashMap<&'a InputSource, HashSet<&'a T>>

fn sort_stat_map_ref<'a, T>(stat_map: &'a HashMap<&'a InputSource, HashSet<'a T>>) -> Vec<(&'a InputSource, Vec<&'a T>)>
```

I thought having two `sort_stat_map` is undesirable.

2. Let `merge_stat_map` return HashMap-s containing owned values:

```
fn merge_stat_maps<'a, T>(stat_maps: &Vec<&HashMap<InputSource, HashSet<T>>>) -> HashMap<InputSource, HashSet<T>>
```

but that would need a lot of `clone`-ing, which I also thought would be undesirable.

Is there any other option I am missing?

katrinafyi: One way of dealing with borrowed and owned hashmaps is to work with iterators instead. This easily generalises to multiple hashmaps because you can just chain their iterators. See https://github.com/rina-forks/lychee/commit/87c9fdf3d7840c5691a76ce927fc0a9497bd8711

cristiklein: Thanks @katrinafyi . I didn't think about iterators and I agree this is a really elegant solution.

I cherry-picked your branch, made clippy happy and added code to break ties.

lychee-bin/src/formatters/stats/json.rs
```diff
@@ -22,19 +22,12 @@ impl StatsFormatter for Json {
...
 
-    #[test]
-    fn test_json_formatter() {
-        let formatter = Json::new();
-        let result = formatter.format(get_dummy_stats()).unwrap();
-
-        assert_eq!(
-            result,
-            r#"{
+    const EXPECTED_JSON: &str = r#"{
```
thomas-zahner, line 25: Nitpick: I don't really like the idea of extracting a variable which is needed once in a specific function/test into the outer scope. Please keep it inside the test. You can still have it there as an extracted const variable.

lychee-bin/src/formatters/stats/mod.rs
```diff
@@ -54,29 +56,53 @@ pub(crate) fn output_statistics(stats: OutputStats, config: &Config) -> Result<(
...
+where
+    K: Display + Hash + Eq + 'a,
+    V: Display + 'a,
+    Entries: IntoIterator<Item = (&'a K, Vals)>,
+    Vals: IntoIterator<Item = &'a V>,
+{
+    let mut map: HashMap<&K, Vec<&V>> = HashMap::new();
+    for (k, vs) in it {
+        let vals = map.entry(k).or_default();
+        vals.extend(vs);
```
thomas-zahner:
```suggestion
        map.entry(k).or_default().extend(vs);
```

lychee-bin/src/formatters/stats/response.rs
```diff
@@ -44,6 +44,8 @@ pub(crate) struct ResponseStats {
     pub(crate) success_map: HashMap<InputSource, HashSet<ResponseBody>>,
     /// Failed responses
     pub(crate) error_map: HashMap<InputSource, HashSet<ResponseBody>>,
+    /// Store timed out responses
```
thomas-zahner: With merging my PR yesterday into master, I've updated these map comments to remove `Store ...`.

```suggestion
    /// Timed out responses
```

lychee-bin/src/options.rs
```diff
@@ -659,6 +659,11 @@ pub(crate) struct Config {
     #[arg(short, long, verbatim_doc_comment)]
     accept: Option<StatusCodeSelector>,
 
+    /// Accept timeouts as valid links
```
thomas-zahner:
```suggestion
    /// Accept timed out requests and return exit code 0
    /// when encountering timeouts but not any other errors.
```

lychee.example.toml
```diff
@@ -190,6 +190,9 @@ include_wikilinks = true
 # Only check local files
 offline = true
 
+# Accept timeouts as valid links
```
thomas-zahner:
```suggestion
# Accept timed out requests
```

thomas-zahner (changes requested):
I also suggest a few other improvements in 6dc06e9f9e6cef09e6040c5609dd314e33ad4f2f

lychee-bin/src/formatters/stats/json.rs
cristiklein: @thomas-zahner Me neither, but clippy was complaining that the function has too many lines (115, maximum 100). I see two options:

1. Extract the string outside the function.
2. Silence clippy.

Which one do you prefer?

thomas-zahner: Ah I see. Well let's keep it like this then. Currently it doesn't really matter that much as there is only one test.

cristiklein:
Thanks @thomas-zahner . I accepted all your suggestions and made the tests pass again. PTAL.

cristiklein:
> I also suggest a few other improvements in [6dc06e9](https://github.com/lycheeverse/lychee/commit/6dc06e9f9e6cef09e6040c5609dd314e33ad4f2f)

Sorry, I noticed this too late. I added those suggestions too and fixed the tests accordingly.

lychee.example.toml
```diff
@@ -190,6 +190,9 @@ include_wikilinks = true
 # Only check local files
 offline = true
 
+# Accept timed out requests
+accept_timeouts = false
```
thomas-zahner:
```suggestion
accept_timeouts = true
```

lychee-bin/src/formatters/stats/compact.rs
```diff
@@ -26,25 +26,27 @@ struct CompactResponseStats {
...
-            };
+        if issues > 0 {
+            let input = if issues == 1 { "input" } else { "inputs" };
 
             color!(
                 f,
                 BOLD_PINK,
-                "Issues found in {} {input}. Find details below.\n\n",
-                stats.error_map.len()
+                "Issues found in {issues} {input}. Find details below.\n\n",
```
thomas-zahner, line 41: Note that it might be better to make this a [user hint](https://github.com/lycheeverse/lychee/pull/2021) in the long run. (not in this PR) We could then take the value of `accept_timeouts` into account.

thomas-zahner:
@cristiklein Thank you for the work. I also really like the new `timeout_map` which emerged as part of this PR.

cristiklein:
> @cristiklein Thank you for the work. I also really like the new `timeout_map` which emerged as part of this PR.

And thank you for shepherding this PR. I really liked the infrastructure and tests in lychee, and the good discussions we had. It was fun!

PR #2066, opened 2026-02-26
Add JUnit

### LINKED ISSUE #614: Support SARIF as an Output Format?

thomas-zahner:
Fixes #614

I experimented both with the SARIF and JUnit XML report format. In the end I've settled for JUnit. I feel like the format is more fitting and it's simpler to integrate. In GitLab it's supported out of the box. In GitHub you would normally use an additional action such as [action-junit-report](https://github.com/mikepenz/action-junit-report). Using this additional action was easier IMO then setting up the code scanning in GitHub to get the SARIF format to properly work. It seems much more programmatic, where with SARIF you have to do some manual adjustments in the settings and the reported errors are considered "security" issues. With JUnit link check results are treated like tests which are passing, failing, or skipped.

I've made a little [demo for GitLab](https://gitlab.com/thomas-zahner/test-junit-report/-/pipelines/2351955460/test_report?job_name=lychee%20junit%20demo) and [one for GitHub](https://github.com/thomas-zahner/lychee/pull/10/changes).

This PR also introduces durations response body. See 2e0060cb

### DIFF
--- a/lychee-bin/src/formatters/stats/response.rs
+++ b/lychee-bin/src/formatters/stats/response.rs
@@ -17,9 +17,9 @@ use crate::formatters::suggestion::Suggestion;
 /// run. It also contains maps to store the responses for each status (success,
 /// error, excluded, etc.) and the sources of the responses.
 ///
-/// The `detailed_stats` field is used to enable or disable the storage of the
-/// responses in the maps for successful and excluded responses. If it's set to
-/// `false`, the maps will be empty and only the counters will be updated.
+/// The `detailed_stats` field indicates whether detailed statistics are shown.
+/// `success_map`, `redirect_map` and `excluded_map` remain empty if detailed
+/// statistics are disabled. This depends on the verbosity mode.
 #[derive(Default, Serialize, Debug)]
 pub(crate) struct ResponseStats {
     /// Total number of responses
@@ -42,7 +42,7 @@ pub(crate) struct ResponseStats {
     pub(crate) cached: usize,
     /// Successful responses (if `detailed_stats` is enabled)
     pub(crate) success_map: HashMap<InputSource, HashSet<ResponseBody>>,
-    /// Failed responses (if `detailed_stats` is enabled)
+    /// Failed responses
     pub(crate) error_map: HashMap<InputSource, HashSet<ResponseBody>>,
     /// Replacement suggestions for failed responses (if `--suggest` is enabled)
     pub(crate) suggestion_map: HashMap<InputSource, HashSet<Suggestion>>,


lychee-bin/src/formatters/stats/json.rs
```diff
@@ -83,12 +87,23 @@ mod tests {
...
+        },
+        "duration": {
+          "secs": 1,
+          "nanos": 0
         }
       }
     ]
   },
   "excluded_map": {},
-  "duration_secs": 0,
```
thomas-zahner, line 91: This is how `std::time::Duration` serialises by default. I felt that it's unnecessary to have additional logic only to make the output less precise. (the `duration_secs` approach)

lychee-lib/src/types/response.rs
```diff
@@ -92,6 +98,9 @@ pub struct ResponseBody {
     /// The location of the URI
     #[serde(skip_serializing_if = "Option::is_none")]
     pub span: Option<RawUriSpan>,
+    /// The time it took to perform the request and produce this [`ResponseBody`]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub duration: Option<Duration>,
```
mre, line 103: Hmmm... `ResponseBody` derives `Hash` and `PartialEq`, and now includes duration.

This means two responses to the same URL with the same status, but different measured durations won't deduplicate in the `HashSet`s used throughout the stats maps. This seems like a bug? 🤔 

On a more meta-level, I believe duration is a measurement artifact, not part of the identity of a response. Maybe we should keep them separate?

lychee-bin/tests/cli.rs
```diff
@@ -78,6 +78,24 @@ mod cli {
...
+        remove_non_deterministic_duration(&mut output_json["redirect_map"]);
+        return Ok(output_json);
+
+        fn remove_non_deterministic_duration(value: &mut Value) {
+            let map = value.as_object_mut().unwrap();
+            map.iter_mut().for_each(|(_, v)| {
+                v.as_array_mut().unwrap().iter_mut().for_each(|a| {
+                    a.as_object_mut().unwrap().remove("duration").unwrap();
+                });
+            });
```
mre: I know that these are just tests, but can we use `expect` in that case or add some assertions? Because I'm afraid that when this panics it will be harder to debug otherwise.

lychee-bin/src/formatters/stats/junit.rs
```diff
@@ -0,0 +1,165 @@
...
+pub(crate) struct Junit {}
+
+impl Junit {
+    pub(crate) const fn new() -> Self {
+        Self {}
+    }
+}
+
+impl StatsFormatter for Junit {
+    /// Format stats as JSON object
```
mre:
```suggestion
    /// Format stats as JUnit object
```

lychee-bin/src/formatters/stats/response.rs
```diff
@@ -37,18 +40,18 @@ pub(crate) struct ResponseStats {
...
-    /// Store redirected responses (if `detailed_stats` is enabled)
+    /// Redirected responses (if `detailed_stats` is enabled)
     pub(crate) redirect_map: HashMap<InputSource, HashSet<ResponseBody>>,
-    /// Store excluded responses (if `detailed_stats` is enabled)
+    /// Excluded responses (if `detailed_stats` is enabled)
     pub(crate) excluded_map: HashMap<InputSource, HashSet<ResponseBody>>,
-    /// Used to store the duration of the run in seconds.
-    pub(crate) duration_secs: u64,
+    /// The time it took to perform the full run
+    pub(crate) duration: Duration,
```
mre, line 54: This is cool and all, and we should keep it. I just want to point out that this is technically a breaking change in the JSON output. E.g. before it was  `"duration_secs": 0` and I believe now it would be `"duration": {"secs": 0, "nanos": 0}` or something. Maybe we should point that out in the release notes or in the PR description?

lychee-lib/src/types/response.rs
thomas-zahner: In my opinion this is no problem. Actually I've done the same thing with the spans. This makes it possible to understand where the URL comes from. Also it reflects the reality more closely. On a more abstract level we "perform" requests for each and every URL and we get different timings for each result and include the location/span. But of we should not perform real duplicate requests for duplicate URLs of course.

So the downside is that it might not be obvious that we actually only perform one real request for duplicate URLs. I've taken a closer look and noticed that we actually perform multiple requests for the same URL in some cases. This is what I'm addressing in #2067.

I think we might want to keep the new additions of span and duration. But it might make sense to show the duration only in verbose mode. What do you think?

The current master version does this:

```
{
  "total": 2,
  "successful": 2,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 0,
  "excludes": 0,
  "errors": 0,
  "cached": 0,
  "success_map": {
    "stdin": [
      {
        "url": "https://thomaszahner.ch/",
        "status": {
          "text": "200 OK",
          "code": 200
        },
        "span": {
          "line": 1,
          "column": 25
        }
      },
      {
        "url": "https://thomaszahner.ch/",
        "status": {
          "text": "200 OK",
          "code": 200
        },
        "span": {
          "line": 1,
          "column": 1
        }
      }
    ]
  },
  "error_map": {},
  "suggestion_map": {},
  "redirect_map": {},
  "excluded_map": {},
  "duration_secs": 0,
  "detailed_stats": true
}
```

Now this PR makes this:

```
{
  "total": 2,
  "successful": 2,
  "unknown": 0,
  "unsupported": 0,
  "timeouts": 0,
  "redirects": 0,
  "excludes": 0,
  "errors": 0,
  "cached": 0,
  "success_map": {
    "stdin": [
      {
        "url": "https://thomaszahner.ch/",
        "status": {
          "text": "200 OK",
          "code": 200
        },
        "span": {
          "line": 1,
          "column": 25
        },
        "duration": {
          "secs": 0,
          "nanos": 92491603
        }
      },
      {
        "url": "https://thomaszahner.ch/",
        "status": {
          "text": "200 OK",
          "code": 200
        },
        "span": {
          "line": 1,
          "column": 1
        },
        "duration": {
          "secs": 0,
          "nanos": 80457448
        }
      }
    ]
  },
  "error_map": {},
  "suggestion_map": {},
  "redirect_map": {},
  "excluded_map": {},
  "duration": {
    "secs": 0,
    "nanos": 98768175
  },
  "detailed_stats": true
}
```

lychee-bin/src/formatters/stats/response.rs
thomas-zahner: Yes that makes sense :+1:

lychee-lib/src/types/response.rs
mre: Ah yeah, that clears things up. Makes sense then. 👍

PR #2067, opened 2026-03-02
Prevent duplicate requests to the same URLs

thomas-zahner:
This is done by locking a Mutex for each Uri.
Previously duplicate Uris were sometimes checked, depending on the timing on when the other duplicates were cached.

### DIFF
--- a/lychee-lib/src/ratelimit/host/host.rs
+++ b/lychee-lib/src/ratelimit/host/host.rs
@@ -149,9 +149,7 @@ impl Host {
         let mut url = request.url().clone();
         url.set_fragment(None);
         let uri = Uri::from(url);
-
-        let uri_mutex = self.acquire_uri_mutex(&uri);
-        let _uri_guard = uri_mutex.lock().await;
+        let _uri_guard = self.lock_uri_mutex(uri.clone()).await;
 
         if let Some(cached) = self.get_cached_status(&uri, needs_body) {
             self.record_cache_hit();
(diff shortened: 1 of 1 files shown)


lychee-lib/src/ratelimit/host/host.rs
```diff
@@ -144,6 +151,8 @@ impl Host {
         let uri = Uri::from(url);
 
         let _permit = self.acquire_semaphore().await;
+        let uri_mutex = self.acquire_uri_mutex(&uri);
+        let _uri_guard = uri_mutex.lock().await;
 
         if let Some(cached) = self.get_cached_status(&uri, needs_body) {
```
katrinafyi: The first cache check could be done even earlier, before the semaphore acquire since it doesn't need to be limited by host concurrency.

Also, `acquire_uri_mutex` is named similar to `acquire_semaphore` but it does different things - the semaphore also gains the lock, but the mutex one doesn't. Could `acquire_uri_mutex` be changed to just return the lock guard?

thomas-zahner:
> Looks like it should work. I noticed this could happen while looking at this code for https://github.com/lycheeverse/lychee/pull/2035 but I thought that the benefit was small.

Ah cool. Yeah, you are totally right. But at the same time I think it's not that unlikely to happen. The chance that URL's are duplicated, potentially across multiple files, is quite big. Especially if you consider that the `Host` struct applies not for all URLs but only for the specific host/subdomain.

> To increase throughput, I think we'd need something higher up at the check.rs level.

Yeah, thanks for the idea. True, it does not increase throughput at all. But it should save resources might make it a bit faster, especially if encountering rate limiting.

lychee-lib/src/ratelimit/host/host.rs
thomas-zahner: > The first cache check could be done even earlier

Ah true, see 78634dd01e435fd1b940605b507e614f9176757a

> Could acquire_uri_mutex be changed to just return the lock guard?

Yeah, it would be nice to call the function in a single line, but I didn't get that to work unfortunately.

Oh right, the name might be weird. What about 0c6becec?

katrinafyi: Cache check looks good now. 

Is there a reason to prefer a function that returns the mutex rather than a function that returns the lock guard? (To make the lifetimes work, I think you'd need to use [`lock_owned`](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.lock_owned))

thomas-zahner: Oooh, that's awesome! So yeah, the reason for it was that I couldn't get it to work because it tried using `lock` instead of `lock_owned`. See 86c7f2ad778a0b5d8b470852ce74b7cf2e52d69f

thomas-zahner:
Thank you for reviewing and helping me with the `lock_owned` trick!

PR #2071, opened 2026-03-04
feat: Support sitemap.xml

### LINKED ISSUE #2062: Remote XML files are treated as HTML (including sitemap.xml)
> Hi, i ran into this same issue today, 
> It seems it is not only because of recursive lookup of sub-sitemap.xml links; it seems it can't extract links from sitemap.xml files accessed via http?
> 
> Apple's sitemap(https://www.apple.com/sitemap.xml) for example, does not have sub sitemap xmls - they had the actual links listed. 
> 
> ```
> <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
> <url><loc>https://www.apple.com/</loc></url>
> <url><loc>https://www.apple.com/accessibility/</loc></url>
> <url><loc>https://www.apple.com/accessibility/assistive-technologies/</loc></url>
> <url><loc>https://www.apple.com/accessibility/designed-for-students/</loc></url>
> ...
> </urlset>
> ```
> 
> Running lychee on this sitemap url does not yeild links
> ```
> lychee --version && lychee https://www.apple.com/sitemap.xml
> lychee 0.23.0
> 0/0 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                      🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors
> ```
> 
> But downloading the same sitemap xml and running lychee on the downloaded file is working fine
> ```
> lychee --version && lychee ./sitemap.xml 
> lychee 0.23.0
> 102/360 ━━━━━ ━━━━━━━━━━━━━━      [200] https://www.apple.com/feedback/calendar/
> ^C
> ```
>  

 _Originally posted by @mattr-rsolomon in [#1819](https://github.com/lycheeverse/lychee/issues/1819#issuecomment-3948874003)_

I did some quick investigation into https://www.apple.com/sitemap.xml, and it looks like the issue is that lychee is assuming that websites are HTML. The XML appears to have no links because the links are placed where ordinary non-link text would appear in HTML. 

I also note that downloading the file isn't a perfect workaround because it treats the file as plaintext. XML entities won't be interpreted. You can see this in `lychee --dump`.
```
https://www.apple.com/us/search/product-red?f=ipadpro_12_9_2&amp;fh=48e4&amp;sel=accessories
```

I think that we should add an XML file type which would help a lot with sitemaps. However, there's lots of different XML schemas so idk how we should handle them.

cristiklein:
Fixes #2062 

This PR adds support for extracting links from `<loc>` tags from `sitemap.xml`. The implementation is kept minimal, relying on a regex, similar to how links are extracted from CSS.

In future, the XML extractor could be extended with support for links in SVGs and perhaps a proper XML parser. However, those are out-of-scope for this PR.

### DIFF
--- a/lychee-lib/src/extract/xml.rs
+++ b/lychee-lib/src/extract/xml.rs
@@ -41,7 +41,8 @@ pub(crate) fn extract_xml<S: SpanProvider>(input: &str, span_provider: &S) -> Ve
                             let element = std::str::from_utf8(e.name().as_ref())
                                 .unwrap_or("")
                                 .to_string();
-                            let end_of_empty_tag: usize = reader.buffer_position().try_into().unwrap_or_default();
+                            let end_of_empty_tag: usize =
+                                reader.buffer_position().try_into().unwrap_or_default();
                             // Span is a bit imprecise, as it points to the end of the element. However, quick_xml does not provide the position of attributes, so this is the best we can do.
                             let span = span_provider.span(end_of_empty_tag);
 
@@ -73,7 +74,7 @@ pub(crate) fn extract_xml<S: SpanProvider>(input: &str, span_provider: &S) -> Ve
 
 #[cfg(test)]
 mod tests {
-    use crate::types::uri::raw::{ SourceSpanProvider, span };
+    use crate::types::uri::raw::{SourceSpanProvider, span};
 
     use super::*;
 
@@ -100,23 +101,26 @@ mod tests {
     </url>
 </urlset>"#;
 
-
-        let expected = vec![RawUri {
-            text: "https://elastisys.io/welkin/".to_string(),
-            element: Some("loc".to_string()),
-            attribute: None,
-            span: span(4, 15),
-        }, RawUri {
-            text: "https://elastisys.io/welkin/architecture/".to_string(),
-            element: Some("loc".to_string()),
-            attribute: None,
-            span: span(8, 15),
-        }, RawUri {
-            text: "https://elastisys.io/welkin/glossary/".to_string(),
-            element: Some("loc".to_string()),
-            attribute: None,
-            span: span(12, 15),
-        }];
+        let expected = vec![
+            RawUri {
+                text: "https://elastisys.io/welkin/".to_string(),
+                element: Some("loc".to_string()),
+                attribute: None,
+                span: span(4, 15),
+            },
+            RawUri {
+                text: "https://elastisys.io/welkin/architecture/".to_string(),
+                element: Some("loc".to_string()),
+                attribute: None,
+                span: span(8, 15),
+            },
+            RawUri {
+                text: "https://elastisys.io/welkin/glossary/".to_string(),
+                element: Some("loc".to_string()),
+                attribute: None,
+                span: span(12, 15),
+            },
+        ];
 
         let uris = extract(input);
 
@@ -140,17 +144,20 @@ mod tests {
     </channel>
 </rss>"#;
 
-        let expected = vec![RawUri {
-            text: "https://example.com".to_string(),
-            element: Some("link".to_string()),
-            attribute: None,
-            span: span(5, 15),
-        }, RawUri {
-            text: "https://example.com/item".to_string(),
-            element: Some("link".to_string()),
-            attribute: None,
-            span: span(9, 19),
-        }];
+        let expected = vec![
+            RawUri {
+                text: "https://example.com".to_string(),
+                element: Some("link".to_string()),
+                attribute: None,
+                span: span(5, 15),
+            },
+            RawUri {
+                text: "https://example.com/item".to_string(),
+                element: Some("link".to_string()),
+                attribute: None,
+                span: span(9, 19),
+            },
+        ];
 
         let uris = extract(input);
 
@@ -178,17 +185,20 @@ mod tests {
     </entry>
 </feed>"#;
 
-        let expected = vec![RawUri {
-            text: "https://example.com".to_string(),
-            element: Some("link".to_string()),
-            attribute: Some("href".to_string()),
-            span: span(4, 40),
-        }, RawUri {
-            text: "https://example.com/entry".to_string(),
-            element: Some("link".to_string()),
-            attribute: Some("href".to_string()),
-            span: span(12, 50),
-        }];
+        let expected = vec![
+            RawUri {
+                text: "https://example.com".to_string(),
+                element: Some("link".to_string()),
+                attribute: Some("href".to_string()),
+                span: span(4, 40),
+            },
+            RawUri {
+                text: "https://example.com/entry".to_string(),
+                element: Some("link".to_string()),
+                attribute: Some("href".to_string()),
+                span: span(12, 50),
+            },
+        ];
 
         let uris = extract(input);
 


lychee-lib/src/extract/xml.rs
```diff
@@ -0,0 +1,90 @@
...
+        <loc>(                      # Match '<loc>'
+        (?P<url>[^<]+)              # Capture the URL (anything until the next '<')
+        )</loc>                     # Match '</loc>'
+        ",
+    )
+    .expect("XML URL regex should be valid")
+});
+
+/// Extract unparsed URL strings from XML
+pub(crate) fn extract_xml<S: SpanProvider>(input: &str, span_provider: &S) -> Vec<RawUri> {
```
mre, line 9: Mh, not sure if we should call it extract_xml. It sounds a bit misleading, given that we can only handle a subset of XML files, namely sitemap files.

Maybe we could be explicit and call it `extract_sitemap_xml`?

lychee-lib/src/types/file.rs
```diff
@@ -149,6 +153,9 @@ impl FileType {
     /// All known plaintext extensions
     const PLAINTEXT_EXTENSIONS: &'static [&'static str] = &["txt"];
 
+    /// All known XML extensions
+    const XML_EXTENSIONS: &'static [&'static str] = &["xml"];
```
mre, line 157: Would be cool to support `xml.gz` as well.

lychee-lib/src/types/file.rs
```diff
@@ -157,6 +164,7 @@ impl FileType {
         extensions.extend(Self::html_extensions());
         extensions.extend(Self::css_extensions());
         extensions.extend(Self::plaintext_extensions());
+        extensions.extend(Self::xml_extensions());
```
mre, line 167: Adding XML to the list of default extensions could be confusing if all we support are `sitemap.xml` files.

E.g. imagine you see your XML files scrolling by in the output, then you might assume that the links got checked... but in fact everything that isn't a sitemap gets silently ignored.

I see two options:

1. We add a warning if we encounter an XML file that does not get checked
2. We add support for all XML files by using a proper XML parser.

Maybe 1 is enough for now?
I would like to hear people's thoughts on this.

lychee-lib/src/extract/xml.rs
cristiklein: I agree and added a commit to fix that.

lychee-lib/src/types/file.rs
cristiklein: I agree. I went for 1. See https://github.com/lycheeverse/lychee/pull/2071/commits/694f83bdd9d14d65098a1302797b1a4da492f683

lychee-lib/src/types/file.rs
cristiklein: I like the idea of supporting `xml.gz` as well. However:

1. I sample a few websites (e.g., `developers.google.com` and `apple.com`) and `sitemap.xml` seems to be more common than `sitemap.xml.gz`. I assume it's because compression is preferred in the TLS and/or HTTP/2 layer these days.
2. lychee currently dislikes binary files, to the point that there is [even a test for it](https://github.com/lycheeverse/lychee/blob/892cb9bee02a65a9070e1d31c018ed9e1b6df2ae/lychee-bin/tests/cli.rs#L2970).

These two combined make me think that the cost-to-benefit ratio for adding `xml.gz` is not very favorable. Happy to be challenged on this one.

lychee-lib/src/extract/xml.rs
```diff
@@ -0,0 +1,99 @@
...
+                span: span_provider.span(match_start),
+            })
+        })
+        .collect();
+
+    if uris.is_empty() {
+        warn!(
+            "No URLs found in sitemap XML input. Currently, lychee only supports extracting URLs from <loc> elements in sitemaps. If your XML contains links in a different format, please consider submitting a feature request or contributing support for additional XML formats."
+        );
+    }
```
mre, line 70: This warning can be confusing if the input file is not a sitemap.
Say I want to check an XML file that isn't a sitemap, then this warning would state that no URLs were found in the sitemap input.

Generally speaking, it feels like this check should happen earlier? Like, before we call the function?
I assume that sitemaps are all called sitemap.xml, so we could check the filename before the function call. E.g. in `lychee-lib/src/extract/mod.rs`.

lychee-lib/src/types/file.rs
mre: You're probably right. 👍

lychee-lib/src/extract/xml.rs
cristiklein: @mre I went for another option: What if we supported all common XML formats, like sitemaps, RSS and Atom? :smile: See https://github.com/lycheeverse/lychee/pull/2071/commits/9280d67bafe9b081a7d08e89854725d3fbcb6a1c

PTAL and let me know what you think.

Aside: Clippy complains, but I think it is unrelated to my PR.

mre: Yeah, that's better. There's probably a more generic way to look for links instead of searching for `loc` and `link`, but I like the direction. In the HTML extractor, we sometimes use the plaintext extractor. Maybe you can check out that part and see if you'd like to do something similar.

(Sorry for the short answer; not much time right now.)

thomas-zahner:
@cristiklein Regarding the clippy warnings/errors, by rebasing or merging master you should be able to resolve the issues. (edit: I did a rebase)

lychee-lib/src/extract/xml.rs
```diff
@@ -76,24 +99,85 @@ mod tests {
...
+        for link in &links {
+            assert_eq!(link.element, Some("link".to_string()));
+            assert_eq!(link.attribute, None);
+        }
+    }
+
+    #[test]
+    fn test_extract_atom_links() {
+        // Atom example
+        let input = r#"<?xml version="1.0" encoding="utf-8"?>
```
thomas-zahner, line 170: This doesn't seem to be entirely valid atom XML, according to [W3C](https://validator.w3.org/feed/#validate_by_input). Can we update the example to be fully valid?

lychee-lib/src/extract/xml.rs
```diff
@@ -76,24 +99,85 @@ mod tests {
...
+    </channel>
+</rss>"#;
+
+        let links: Vec<RawUri> = extract(input);
+        let expected_links = ["https://example.com", "https://example.com/item"];
+
+        assert_eq!(links.len(), 2, "Should extract all URLs");
+
+        for (expected_link, link) in expected_links.iter().zip(links.iter()) {
+            assert_eq!(expected_link, &link.text);
```
thomas-zahner: Can we assert the full struct, instead of accessing just some fields? This also forces us to test the `span` values. This is currently untested.

Ideally we have a single `assert_eq!(links, vec![ /* RawUri { ... }, ... */ ]);` instead of all the complicated assertions an the clever `.zip(links.iter())`. If the order would be non-deterministic we can sort the vector before the assertion.

lychee-lib/src/extract/xml.rs
```diff
@@ -76,24 +99,85 @@ mod tests {
...
+        <id>urn:uuid:1225c695-cfb8-4ebb-aaaa-80da344efa6a</id>
+        <updated>2026-03-04T12:00:00Z</updated>
+        <summary>Example Entry Summary</summary>
+    </entry>
+</feed>"#;
+
+        let links: Vec<RawUri> = extract(input);
+        let expected_links = ["https://example.com", "https://example.com/entry"];
+
+        assert_eq!(links.len(), 2, "Should extract all URLs");
```
thomas-zahner: Same here. Ideally it's one single `assert_eq!`.

lychee-lib/src/extract/xml.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2071/commits/167b2f9c49b361aff475173a870b837aed50378b

lychee-lib/src/extract/xml.rs
cristiklein: You're right, the span deserves testing too. Fixed in https://github.com/lycheeverse/lychee/pull/2071/changes/b945661889de36d732ed5c22bb819b58a6b7540e

lychee-lib/src/extract/xml.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2071/changes/b945661889de36d732ed5c22bb819b58a6b7540e

lychee-lib/src/extract/xml.rs
cristiklein: @mre Thanks for the kind words and tip. I checked how the HTML extractor is using the plaintext extractor, but I don't think we need it. Sitemaps, RSS and Atom are supposed to be "semantically cleaner" and the URL can be extracted directly from the XML attribute or text. I have been wrong before, though. :smile:

cristiklein:
@mre @thomas-zahner Thanks for the feedback on this PR. I believe I addressed all your comments. Otherwise, please let me know. :smile:

thomas-zahner:
@cristiklein Thank you for this cool addition to lychee! :rocket:

cristiklein:
> @cristiklein Thank you for this cool addition to lychee! 🚀

And thank you for the kind words and for shepherding this PR. :heart:

PR #2074, opened 2026-03-05
Report input source errors rather than tokio panics

katrinafyi:
for example:
```
cargo run -- https://google.com https://fdsjaifjdsaoi.invalid
```

before:
```
thread 'tokio-runtime-worker' (74220) panicked at lychee-bin/src/commands/check.rs:337:18:
response channel closed: SendError { .. }
Error: Task failed to execute to completion

Caused by:
    task 28 panicked with message "response channel closed: SendError { .. }"
```
note that the real error isn't printed at all, because it's meant to be returned to main and doesn't panic.

after:
```
Error: Network error: Connection failed. Check network connectivity and firewall settings (error sending request for url (https://fdsjaifjdsaoi.invalid/))

Caused by:
    0: error sending request for url (https://fdsjaifjdsaoi.invalid/)
    1: client error (Connect)
    2: dns error
    3: failed to lookup address information: Name or service not known
```
note also that this is not a panic

this was happening due to a subtlety in how errors get raised. the `collect_responses` task is the one which terminates first on errors, but it was being awaited after other tasks which would panic so its result was ignored. 

reporting the error does have to wait for the other tasks to finish or panic (they will panic the next time they try to send through a channel).

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -81,8 +81,7 @@ where
     let (stats_result, request_result, send_result) =
         futures::join!(stats_handle, request_handle, send_task);
 
-    // Fatal user errors are here so check it first. Unwraps two results, the
-    // first for JoinError and the second for ErrorKind.
+    // Fatal user errors are here so check it first.
     let mut stats: ResponseStats = stats_result??;
 
     let (cache, client) = request_result?;


lychee-bin/src/commands/check.rs
```diff
@@ -65,21 +65,29 @@ where
...
 
-    // Wait until all requests are sent
-    send_requests(params.requests, wait_guard, send_req, &progress).await?;
-    let (cache, client) = handle.await?;
+    // Send requests into the channel. Note that this will run within the main task
+    // and doesn't start until awaited.
+    let send_task = send_requests(params.requests, wait_guard, send_req, &progress);
+
+    // Waits for all futures to finish, either normally or due to panic.
+    let (stats, request_result, send_result) =
```
thomas-zahner: Maybe? But we can also leave it.

```suggestion
    let (stats_result, request_result, send_result) =
```

lychee-bin/src/commands/check.rs
```diff
@@ -154,22 +162,19 @@ async fn suggest_archived_links(
...
-{
+) -> Result<(), ()> {
     tokio::pin!(requests);
     while let Some(request) = requests.next().await {
         progress.inc_length(1);
         send_req
             .send((guard.clone(), request))
             .await
-            .expect("Cannot send request");
+            .map_err(|_| ())?;
```
thomas-zahner, line 176: Was the error variant, i.e. the value of `ErrorKind` unused?

thomas-zahner: Ah, never mind. The previous return type `Result<(), ErrorKind>` was confusing me. This is much better.

lychee-bin/src/commands/check.rs
```diff
@@ -65,21 +65,29 @@ where
...
+    // Send requests into the channel. Note that this will run within the main task
+    // and doesn't start until awaited.
+    let send_task = send_requests(params.requests, wait_guard, send_req, &progress);
+
+    // Waits for all futures to finish, either normally or due to panic.
+    let (stats, request_result, send_result) =
+        futures::join!(stats_handle, request_handle, send_task);
+
+    // Fatal user errors are reported in this Result, so check it first.
+    let stats: Result<ResponseStats, ErrorKind> = stats?;
```
katrinafyi:
```suggestion
    let stats: Result<ResponseStats, ErrorKind> = stats_result?;
```

lychee-bin/src/commands/check.rs
```diff
@@ -65,21 +65,29 @@ where
...
+    // and doesn't start until awaited.
+    let send_task = send_requests(params.requests, wait_guard, send_req, &progress);
+
+    // Waits for all futures to finish, either normally or due to panic.
+    let (stats_result, request_result, send_result) =
+        futures::join!(stats_handle, request_handle, send_task);
+
+    // Fatal user errors are reported in this Result, so check it first.
+    let stats: Result<ResponseStats, ErrorKind> = stats_result?;
+    let mut stats: ResponseStats = stats?;
```
thomas-zahner: Ah, actually this would even be possible :)

```suggestion
    let mut stats = stats_result??;
```

katrinafyi: It is! I didn't think it would work aha

thomas-zahner:
> note also that this is not a panic

If I run your provided command on the current version of the PR I get:

```
thread 'tokio-runtime-worker' (57434) panicked at lychee-bin/src/commands/check.rs:237:18:
cannot send response to queue: SendError { .. }
Error: Network error: Connection failed. Check network connectivity and firewall settings (error sending request for url (https://fdsjaifjdsaoi.invalid/))

Caused by:
    0: error sending request for url (https://fdsjaifjdsaoi.invalid/)
    1: client error (Connect)
    2: dns error
    3: failed to lookup address information: Name or service not known
```

This is fine and much better than before. But did you have a version in this PR where you didn't get the panic? Also, it might make sense to add a CLI test, to assure that we don't lose the detailed and useful stack track and details from the error message again in the future.

katrinafyi:
> This is fine and much better than before. But did you have a version in this PR where you didn't get the panic? 

Yeah it's a good question. Investigating more, the panic is *more likely* to be avoided if we race the futures using `select!`, but it's more complicated and the panic is still shown in some cases. I can still make it appear intermittently using `cargo  run -- a.md fdsjaifjdsaoi.com` and `a.md` has https links. I wrote a bit in the message of  730ed2a, let me know if you want to keep this or I can revert it.

> Also, it might make sense to add a CLI test

Agreed

thomas-zahner:
Oh wow, 730ed2a looks interesting but also really complicated :sweat_smile: 
Also it seems to introduce some deadlocks as the CI tests got stuck. I would much rather prefer the previous version which was simple in comparison. Panics on invalid user input aren't bad if we get such detailed and good error messages.

katrinafyi:
FYI the deadlocks were because I was awaiting a FusedFuture after it had already finished, and futures can only be awaited once so it never finished awaiting the second time. But yeah, I agree and I've reverted all the commits.

Idk why the lint is now failing with an error pointing to https://github.com/lycheeverse/lychee/pull/2075

lychee-bin/src/commands/check.rs
```diff
@@ -65,21 +65,29 @@ where
...
+    // Send requests into the channel. Note that this will run within the main task
+    // and doesn't start until awaited.
+    let send_task = send_requests(params.requests, wait_guard, send_req, &progress);
+
+    // Waits for all futures to finish, either normally or due to panic.
+    let (stats_result, request_result, send_result) =
+        futures::join!(stats_handle, request_handle, send_task);
+
+    // Fatal user errors are here so check it first. Unwraps two results, the
+    // first for JoinError and the second for ErrorKind.
```
thomas-zahner: I feel like the second part should be obvious and is prone to get outdated

```suggestion
    // Fatal user errors are here so check it first.
```

thomas-zahner (changes requested):
Looks good and seems to work.
Could you also add the CLI test?

katrinafyi:
Added the cli test in [3c61551](https://github.com/lycheeverse/lychee/pull/2074/commits/3c61551292852a19253505bcaa194783bd7e65a7) and checked that it fails before this PR :)

PR #2075, opened 2026-03-05
Remove deduplication of Request after source tracking

katrinafyi:
After source tracking, adding a Request to a HashSet is ineffective because it contains a unique location for each request.

As a side-effect, removing the deduplication means URLs will be sent to the stream in file order, which is nice for consistency.

### DIFF
--- a/lychee-lib/src/utils/request.rs
+++ b/lychee-lib/src/utils/request.rs
@@ -87,7 +87,7 @@ fn try_parse_into_uri(
 ///
 /// Requests are not deduplicated because repeated URLs may occur at different
 /// source locations. Caching and deduplication happens elsewhere (e.g., in the
-/// per-host HostCache and the top-level persistent cache).
+/// per-host `HostCache` and the top-level persistent cache).
 pub(crate) fn create(
     uris: Vec<RawUri>,
     source: &ResolvedInputSource,


thomas-zahner:
Thank you. This reduces unnecessary complexity for a bit.

Basically this duplication was useless as of now (it might have been useful in the past) because a `Request` contains data like `span` and `source`. The "real" deduplication happens at the lower level in host.rs, independently of `span`, `source` and other details, irrelevant to performing actual link checks efficiently.

lychee-lib/src/utils/request.rs
```diff
@@ -83,9 +82,9 @@ fn try_parse_into_uri(
     Ok(url.into())
 }
 
-/// Create requests out of the collected URLs.
-/// Returns a vector of valid URLs and errors. Valid URLs are deduplicated,
-/// request errors are not deduplicated.
+/// Create requests out of the collected URLs. Returns a vector of valid URLs
+/// and errors. URLs are not deduplicated because repeated URLs may occur at
+/// different source locations.
```
thomas-zahner: Maybe it would be worth mentioning that we don't do deduplication as it happens on a much lower level? (in host.rs) Also `Returns a vector of valid URLs and errors` is a bit confusing.

We could simplify it maybe something in the direction of:

```suggestion
/// Create requests out of the collected URLs.
/// [`Request`]s are not deduplicated.
/// Caching and deduplication happens on a lower level.
```

katrinafyi:
Let me know what you think after those commits.

thomas-zahner:
Looks great, thank you!

PR #2078, opened 2026-03-06
Add RelativeUri enum for types of relative links

katrinafyi:
This adds an enum:
```rust
pub enum RelativeUri<'a> {
    /// A root-relative link, e.g. `"/help"`. The contained string will
    /// start with `/` and not start with `//`.
    RootRel(&'a str),
    /// A scheme-relative link, e.g. `"//example.comhelp"`. The contained
    /// string will start with `//`.
    SchemeRel(&'a str),
    /// A locally-relative link, e.g. `"help"` or `"../home"`.
    LocalRel(&'a str),
}
```
This lets us cleans up some logic that depended on the relative link kind, and it lets us use a more explicit match in BaseInfo.

### DIFF
--- a/lychee-lib/src/types/uri/parsed.rs
+++ b/lychee-lib/src/types/uri/parsed.rs
@@ -26,7 +26,7 @@ impl<'a> ParsedUri<'a> {
         match Uri::try_from(text) {
             Ok(uri) => Ok(ParsedUri::Absolute(uri)),
             Err(ErrorKind::ParseUrl(ParseError::RelativeUrlWithoutBase, _)) => {
-                Ok(ParsedUri::Relative(RelativeUri::new(text)))
+                Ok(ParsedUri::Relative(RelativeUri::parse(text)))
             }
             Err(e) => Err(e),
         }
--- a/lychee-lib/src/types/uri/relative.rs
+++ b/lychee-lib/src/types/uri/relative.rs
@@ -33,12 +33,12 @@ pub enum RelativeUri<'a> {
 }
 
 impl RelativeUri<'_> {
-    /// Interprets the given text as a [`RelativeUri`] variant.
+    /// Parses the given text as a [`RelativeUri`].
     ///
     /// Determining between [`RelativeUri::Root`] and [`RelativeUri::Scheme`]
     /// is done based on how many initial slashes are in the text. If there
     /// are *no* initial slashes, the text is assumed to be a [`RelativeUri::Local`].
-    pub fn new(text: &str) -> RelativeUri<'_> {
+    pub fn parse(text: &str) -> RelativeUri<'_> {
         let text = text.trim_ascii_start();
 
         // important to check for scheme-rel before root-rel, as both of them


lychee-lib/src/types/uri/relative.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use crate::ErrorKind;
+use crate::Uri;
+
+/// A relative link text fragment. Leading whitespace is removed from the
+/// string reference.
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[expect(
+    clippy::enum_variant_names,
+    reason = "suffixing with Rel makes the names more sensible to read"
```
thomas-zahner: Maybe we could name this `Relativity` and then strip the `Rel` suffix?

And then

`pub use RelativeUri;`

instead of

`pub use RelativeUri::{LocalRel, RootRel, SchemeRel};`

But I'm also okay with the current approach.

lychee-lib/src/types/uri/relative.rs
```diff
@@ -0,0 +1,87 @@
...
+/// > Scheme-relative URL: `//developer.mozilla.org/en-US/docs/Learn_web_development` —
+/// > only the protocol is missing. The browser will use the same protocol as the one
+/// > used to load the document hosting that URL.
+///
+/// [MDN]: https://developer.mozilla.org/en-US/docs/Learn_web_development/Howto/Web_mechanics/What_is_a_URL#absolute_urls_vs._relative_urls
+pub(crate) fn is_scheme_relative_link(text: &str) -> bool {
+    text.trim_ascii_start().starts_with("//")
+}
+
+/// Attempts to parse the given text as a into a [`Uri`] or [`RelativeUri`].
```
thomas-zahner: Probably

```suggestion
/// Attempts to parse the given text into a [`Uri`] or [`RelativeUri`].
```

lychee-lib/src/types/uri/relative.rs
```diff
@@ -0,0 +1,87 @@
...
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[expect(
+    clippy::enum_variant_names,
+    reason = "suffixing with Rel makes the names more sensible to read"
+)]
+pub enum RelativeUri<'a> {
+    /// A root-relative link, e.g. `"/help"`. The contained string will
+    /// start with `/` and not start with `//`.
+    RootRel(&'a str),
+    /// A scheme-relative link, e.g. `"//example.comhelp"`. The contained
```
thomas-zahner: Maybe?

```suggestion
    /// A scheme-relative link, e.g. `"//example.com/help"`. The contained
```

lychee-lib/src/types/base_info.rs
```diff
@@ -272,41 +274,43 @@ impl BaseInfo {
...
-                    let locally_relative = format!(".{}", text.trim_ascii_start());
-                    origin.join(&locally_relative)
-                }
-
-                // Resolve all other relative links, including root-relative links
-                // of non-file bases.
-                Self::Full { origin, path } => origin.join(path).and_then(|x| x.join(text)),
+            (Self::None, _) => Err(ParseError::RelativeUrlWithoutBase),
+
+            (Self::NoRoot(base), LocalRel(text) | SchemeRel(text)) => base.join(text),
```
thomas-zahner: Hmm, what would the `(Self::NoRoot(base), SchemeRel(text))` case look like in practice? Is this tested?

lychee-lib/src/types/base_info.rs
```diff
@@ -272,41 +274,43 @@ impl BaseInfo {
...
+
+            (Self::Full { origin, .. }, RootRel(text)) if origin.scheme() == "file" => {
+                origin.join(&format!(".{text}"))
```
thomas-zahner: What does this mean? Why `format!(".{text}")`?

lychee-lib/src/types/uri/relative.rs
```diff
@@ -0,0 +1,87 @@
...
+    text.trim_ascii_start().starts_with("//")
+}
+
+/// Attempts to parse the given text as a into a [`Uri`] or [`RelativeUri`].
+///
+/// # Errors
+///
+/// Returns an error if the text cannot be parsed as a URL, and the parse error
+/// was not due to "relative link without base".
+pub fn parse_url_or_relative(text: &str) -> Result<Either<Uri, RelativeUri<'_>>, ErrorKind> {
```
thomas-zahner: I'm thinking that maybe we could try to avoid the new dependency on `Either`? But this would probably require a simple new enum?

Or I could imagine splitting the function into two parts, where they return `Result<Option<Uri>, ErrorKind>` and ``Result<Option<RelativeUri<'_>>, ErrorKind>`` respectively.

lychee-lib/src/types/base_info.rs
katrinafyi: For example, this would happen if the link `///home/x` appeared inside a local file `/tmp/index.html` without root-dir specified. It would resolve to the link `file:///home/x`, because it is relative to the `file://` scheme.

Worth noting, there is a triple slash in both the scheme-relative link and the resolved link. This is because the domain name is conventionally blank for file:// URLs, but it doesn't have to be. It *is* valid to have a URL like `file://example.com/file/on/example/dot/com`. According to the rustdocs, this won't do anything on Linux but it might do something on Windows: https://docs.rs/reqwest/0.12.23/reqwest/struct.Url.html#method.to_file_path. Maybe it accesses network shared files.

It's a bit goofy. But I think the case makes sense because a NoRoot base is mostly valid and usable aside from its root component, and a scheme-rel link does not need the root component.

It's probably not tested, I don't think there are many (or any) scheme-rel links in the suite.

lychee-lib/src/types/base_info.rs
katrinafyi: It previously had a comment
```
                // Resolve root-relative link with `file:` base by changing it to
                // a subpath of the origin.
```
does that help?

It's because a root-rel link starts with `/root`, and origin is a path like `file://a/b/`. Without the `.{text}`, the link `/root` would just go to `file:///root`. We need to turn it into a locally-relative link, by suppressing the leading `/`, so that it stays within the path of the root dir.

Maybe this would be much clearer if i make it a method on RelativeUri.

lychee-lib/src/types/uri/relative.rs
katrinafyi: It's not a new new dependency because it's a common crate and was already in the dependency tree, but I see your point. 

What would you make the enum? Would it be specific to the relative/absolute link case like `AbsoluteOrRelativeUri` (forgive my bad name), or more general?

I wouldn't want to make it two separate functions, because it's a bit inefficient. Because relative links are so general, the most reliable way to tell if a link is relative is to try and parse it. So, it makes sense to have one function that returns either a Uri or a RelativeUri. I wouldn't want to use regex to tell if a Uri is relative or not.

lychee-lib/src/types/base_info.rs
thomas-zahner: Oh, so it makes e.g. `/root` into `./root`? This is really not obvious to me :sweat_smile: 
Can we maybe comment this? Or extract the formatted string into a well named variable?

lychee-lib/src/types/base_info.rs
thomas-zahner: That's interesting. Can we add a little comment for that?

lychee-lib/src/types/uri/relative.rs
thomas-zahner: Hmm, yeah the enum would be really specific, I can't think of a great name either and I get that two methods aren't ideal. Let's keep it `Either` then.

lychee-lib/src/types/base_info.rs
katrinafyi: I've added this comment. Does this make sense or is it too much?

            // NoRoot and RelativeUri::Scheme happens when a link like `///a` occurs
            // within a local file without root-dir. note the triple slash because file
            // URLs typically don't have a hostname. however, file URLs with hostname
            // are also valid syntax, but they will be rejected by:
            // https://docs.rs/reqwest/0.12.23/reqwest/struct.Url.html#method.to_file_path

lychee-lib/src/types/uri/relative.rs
katrinafyi: I've removed the Rel suffix and kept the name as RelativeUri. Is that ok?

lychee-lib/src/types/base_info.rs
katrinafyi: Added this helper function. Let me know if it makes it clearer. It does mean we have to think about the Scheme case which is kind of nonsensical...

    /// Interprets the current relative link as a locally-relative link
    /// and returns the link text.
    ///
    /// For a root-relative link like `/rest`, this will return `./rest`.
    /// For a scheme-relative link like `//rest`, this will return
    /// `./rest`. A locally-relative link will be returned unchanged.
    ///
    /// This is occasionally useful - for example, to resolve root-relative
    /// links to a known local root directory.
    pub fn to_local_link_text(&self) -> String {
        match self {
            Self::Local(x) => x.to_string(),
            Self::Root(x) => format!(".{x}"),
            Self::Scheme(x) => {
                let x = &x[1..];
                format!(".{x}")
            }
        }
    }

katrinafyi (commented):
I think I've replied to all of the comments so far :)

lychee-lib/src/types/uri/relative.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::ErrorKind;
+use crate::Uri;
+
+/// A relative link text fragment. Leading whitespace is removed from the
+/// string reference.
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[expect(
+    clippy::enum_variant_names,
+    reason = "suffixing with Rel makes the names more sensible to read"
+)]
```
mre: Ooooh! That's fantastic. First time I see that new syntax. We should use it for other clippy warnings as well.

lychee-lib/src/types/uri/relative.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// `./rest`. A locally-relative link will be returned unchanged.
+    ///
+    /// This is occasionally useful - for example, to resolve root-relative
+    /// links to a known local root directory.
+    pub fn to_local_link_text(&self) -> String {
+        match self {
+            Self::Local(x) => x.to_string(),
+            Self::Root(x) => format!(".{x}"),
+            Self::Scheme(x) => {
+                let x = &x[1..];
```
mre: If Self::Scheme contains an empty string, `&x[1..]` will panic because the index is out of bounds. Even if it contains just "/", slicing strings in Rust can be dangerous if you aren't 100% sure where the UTF-8 character boundaries are (though / is safe as a single-byte ASCII character).

I *think* this should be equivalent?

```rust
pub fn to_local_link_text(&self) -> String {
    match self {
        Self::Local(x) => x.to_string(),
        Self::Root(x) => format!(".{x}"),
        Self::Scheme(x) => {
            // Safely remove the first '/' if it exists. 
            // If the string is empty, it just stays empty.
            let tail = x.strip_prefix('/').unwrap_or(x);
            format!(".{tail}")
        }
    }
}
```

mre: Come to think about this function a little more, what was the reason why we can't return a `Path` here? 🤔 If that's fine, I'd prefer that.

```rust
use std::path::{Path, PathBuf};

impl RelativeUri<'_> {
    pub fn resolve_to(&self, base: &Path) -> PathBuf {
        match self {
            Self::Local(x) => base.join(x),
            // strip_prefix("/") makes the path relative so .join() appends it 
            // instead of replacing the whole path.
            Self::Root(x) => base.join(x.strip_prefix('/').unwrap_or(x)),
            Self::Scheme(x) => base.join(x.strip_prefix("//").unwrap_or(x)),
        }
    }
}
```

Then, elsewhere we could avoid `origin.join(&rel.to_local_link_text())` and instead write `rel.resolve_to(origin)` or something.

mre:
I’d like to suggest a refactor for `parse_url_or_relative` following the ["parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) pattern.

By introducing a `Link` type, we can remove the dependency on `either` and replace `parse_url_text` with a `resolve_to` method. This consolidates the prefix logic into a single pass.

```rust
pub enum Link<'a> {
    Absolute(Uri),
    Relative(RelativeUri<'a>),
}

impl<'a> Link<'a> {
    pub fn parse(text: &'a str) -> Result<Self, ErrorKind> {
        let text = text.trim_ascii_start();
        match Uri::try_from(text) {
            Ok(uri) => Ok(Self::Absolute(uri)),
            Err(ErrorKind::ParseUrl(ParseError::RelativeUrlWithoutBase, _)) => {
                let rel = if text.starts_with("//") {
                    RelativeUri::Scheme(text)
                } else if text.starts_with('/') {
                    RelativeUri::Root(text)
                } else {
                    RelativeUri::Local(text)
                };
                Ok(Self::Relative(rel))
            }
            Err(e) => Err(e),
        }
    }

    pub fn resolve_to(self, base: &BaseInfo) -> Result<Url, ErrorKind> {
        match self {
            Self::Absolute(uri) => Ok(uri.url),
            Self::Relative(rel) => base.resolve_relative_link(&rel),
        }
    }
}
```

Not too happy with the term `Link`. We could also call it `InputUri` or something.
And `parse` could also be a `TryFrom`, but you get the idea.

Does this make sense? Sorry, not too deep into that topic, so there's a high chance I'm missing important context.

katrinafyi:
Parse don't validate (PDV) is about encoding properties that have been checked into the type itself and avoiding overly general types like String.

The refactor you suggested doesn't increase PDV, because it's the same *quantity* of information just in different names (but if you do want different names, we can do that). 

In fact, both my current code and your suggestion fall afoul of ["names are not type safety"](https://lexi-lambda.github.io/blog/2020/11/01/names-are-not-type-safety/) because we've just given a name to &str without constraining what's inside the string.

As you noted, the slicing `&x[1..]` could panic on general strings. But it was my intention (and I wrote in the doc comments) that SchemeRel should always start with //, so this should be safe. But this is an example of a validated, not parsed, property so it's not guaranteed in the types.

If we *were* to increase PDV, that would look like replacing the &str inside SchemeRel with a specific type that knows it always starts with //. Something like `SchemeRel(StrWithPrefix<'a, "//">)` - using const generic parameters. This would give us more safety and guarantees. 


This could be done using something like [refinement types](https://crates.io/crates/refined) - specifically, a [StartsWith predicate](https://docs.rs/refined/latest/refined/string/struct.StartsWith.html). Or, if that's too heavy, I can write a basic type myself. Or, we just use unwrap_or :)

lychee-lib/src/types/uri/relative.rs
katrinafyi: I replied to the first comment over here https://github.com/lycheeverse/lychee/pull/2078#issuecomment-4061763910.

About why we can't return a Path, it's because Path is a filesystem path and at this point, we're still dealing with URL-like things. Our relative links could have % encoding and they could have # fragment links, and going through Path would cause problems (and be OS-dependent).

lychee-lib/src/types/uri/relative.rs
thomas-zahner: Thanks, yeah you're right, that's actually better than my proposed name

lychee-lib/src/types/uri/relative.rs
thomas-zahner: Oh cool, I didn't even notice. That's better than `allow`, but now we don't need it at all :)

lychee-lib/src/types/base_info.rs
thomas-zahner: Makes sense :+1:

lychee-lib/src/types/uri/relative.rs
thomas-zahner:
```rust
            // Safely remove the first '/' if it exists. 
            // If the string is empty, it just stays empty.
            let tail = x.strip_prefix('/').unwrap_or(x);
            format!(".{tail}")
```

I like this version, as it's clear that nothing could go wrong.
But we can also keep the indexing, but in that case lets add a `// SAFETY: ` note, why it shouldn't panic. Personally I wouldn't create refinement types and add a new crate for this simple case.

lychee-lib/src/types/base_info.rs
thomas-zahner: The documentation is great and really helps!

lychee-lib/src/types/uri/relative.rs
mre: Yeah, probably `x.strip_prefix('/').unwrap_or(x);` is enough. 👍 

What do you think about replacing this with a `resolve_to` method that takes the base as an argument? It would make the intent clearer than keeping to_local_link_text and then calling `origin.join(&rel.to_local_link_text())`. AFAICT, that’s currently the only usage of `to_local_link_text`, so we could make that use case explicit.

mre:
You're right. Let's dismiss that idea. So the only open point from my side would be https://github.com/lycheeverse/lychee/pull/2078#discussion_r2939335735.

lychee-lib/src/types/uri/relative.rs
katrinafyi: I'm not sure what resovle_to would look like. Would it just be moving the `BaseInfo::resolve_relative_link` function into a method on RelativeUri? (If so, I prefer it in BaseInfo because BaseInfo is the more concrete concept, so BaseInfo should use RelativeUri and not the other way round)

 And would `to_local_link_text` be inlined? I'd extracted the function because @ Thomas said (and I agreed) that `format!(".{X}")` was too cryptic.

edit: added https://github.com/lycheeverse/lychee/commit/8e86a2e824bce9c1c3840852fc3ee31587d83611 for now to use strip_prefix + unwrap_or

mre: Sorry, yeah, I overlooked `resolve_relative_link`; that seems fine. I would personally inline `to_local_link_text` as it's only used in one specific case. Personally, I would move that method closer to where it gets used, but it's okay either way.

Speaking of which, what's the benefit of using `either`? 😅 I remember Either from Haskell, where I used it similar to Rust's Result type, but maybe we should come up with our own enum with two variants that are more aligned to lychee's domain?

katrinafyi: I've inlined to_local_link_text with a comment which should be clear (97fc444ac32ec05c6188ff52334fecbe963ea83f).

The main benefit of Either is we don't have to make a new type :") It's just a generic type with two possibilities, and no variant is more preferred than the other. In Haskell, Either serves dual purpose as the result type (Result in Rust) and as the generic two-variant type.

Either is also really clear at a glance. If it were an enum like `InputUri`, you'd have to click through to find out what an InputUri could be.

But to be honest, the biggest reason is that I can't think of a good name for a new enum. If we can come up with a good name, I would be fine either way. We also talked about it in https://github.com/lycheeverse/lychee/pull/2078#discussion_r2916854090.

mre: I see. How about this?

```rust
/// The result of parsing a string that could be either a full URL 
/// or a relative path/fragment.
pub enum ParsedUri<'a> {
    /// A fully-qualified, absolute URI.
    Absolute(Uri),
    /// A relative URI that requires a base for resolution.
    Relative(RelativeUri<'a>),
}
```

Even though I'm not in love with the naming, the reason why I like it more is that it's much easier to grep for `ParsedUri::Absolute` than to look for where `Either::Left` is being used in a URI context. If we add `either` as a dependency, it's very tempting to use it for more things in the future, so it could get more problematic with time.

katrinafyi: Done in 10cdae8f22ddf02639e0ab3120c6a83da675e7d4. I think it ended up pretty good.

thomas-zahner: I think that's great!

lychee-lib/src/types/uri/parsed.rs
```diff
@@ -0,0 +1,43 @@
...
+    ///
+    /// Returns an error if the text cannot be parsed as a URL, and the parse error
+    /// was not due to "relative link without base".
+    pub fn parse(text: &'a str) -> Result<Self, ErrorKind> {
+        let text = text.trim_ascii_start();
+
+        match Uri::try_from(text) {
+            Ok(uri) => Ok(ParsedUri::Absolute(uri)),
+
+            Err(ErrorKind::ParseUrl(ParseError::RelativeUrlWithoutBase, _)) => {
```
thomas-zahner, line 28:
```suggestion
            Ok(uri) => Ok(ParsedUri::Absolute(uri)),
            Err(ErrorKind::ParseUrl(ParseError::RelativeUrlWithoutBase, _)) => {
```

lychee-lib/src/types/uri/relative.rs
```diff
@@ -0,0 +1,72 @@
...
+            RelativeUri::Root(text)
+        } else {
+            RelativeUri::Local(text)
+        }
+    }
+
+    /// Returns whether the text represents a root-relative link. These look like
+    /// `/this` and are resolved relative to a base URL's origin.
+    fn is_root_relative_link(text: &str) -> bool {
+        !Self::is_scheme_relative_link(text) && text.trim_ascii_start().starts_with('/')
```
thomas-zahner: Nitpick: `trim_ascii_start` is performed twice, in the calling function and here. Same with `is_scheme_relative_link` but we can also keep it, if it is intended for other future use-cases

katrinafyi: It's not really intended for other uses, I was just being extra safe. honestly, we could just inline the two functions into RelativeUri::new. edit: inlined and removed the functions

thomas-zahner: Ah yes, that's better, also makes it clear that the order matters. I wouldn't have realised before

lychee-lib/src/types/uri/relative.rs
mre: Yeah! Reads nicely. :)

PR #2084, opened 2026-03-11
Fix/rustdoc warnings#2045

### LINKED ISSUE #2045: Fix rustdoc warnings
It would be good to fix these eventually. `cargo doc --no-deps` will report warnings.

It's a mix of unresolved links and non-escaped HTML. For example:
```rust
warning: unresolved link to `HostStats`
  --> lychee-lib/src/lib.rs:71:1
   |
13 | | //! }
   | |______^
...
71 | / /// Per-host rate limiting and concurrency control
   |
   = note: the link appears in this line:
           
           - [`HostStats`]: Statistics tracking for each host
              ^^^^^^^^^^^
   = note: no item named `HostStats` in scope
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`

warning: unclosed HTML tag `stdin`
   --> lychee-lib/src/types/input/input.rs:215:58
    |
215 |     /// - Stdin input is returned as the special string "<stdin>"
    |                                                          ^^^^^^^
```

In particular, the invalid HTML leads to broken formatting like this: https://docs.rs/lychee-lib/0.23.0/lychee_lib/struct.RawUri.html#structfield.element

krishnakurtakoti:
Fixes [issues/2045](https://github.com/lycheeverse/lychee/issues/2045)

### DIFF
--- a/lychee-lib/src/ratelimit/mod.rs
+++ b/lychee-lib/src/ratelimit/mod.rs
@@ -6,11 +6,11 @@
 //!
 //! # Architecture
 //!
-//! - [`host::HostKey`]: Represents a hostname/domain for rate limiting
-//! - [`host::Host`]: Manages rate limiting, concurrency, and caching for a specific host
-//! - [`host::HostPool`]: Coordinates multiple hosts and routes requests appropriately
-//! - [`host::HostConfig`]: Configuration for per-host behavior
-//! - [`host::HostStats`]: Statistics tracking for each host
+//! - [`crate::ratelimit::HostKey`]: Represents a hostname/domain for rate limiting
+//! - [`crate::ratelimit::Host`]: Manages rate limiting, concurrency, and caching for a specific host
+//! - [`crate::ratelimit::HostPool`]: Coordinates multiple hosts and routes requests appropriately
+//! - [`crate::ratelimit::HostConfig`]: Configuration for per-host behavior
+//! - [`crate::ratelimit::HostStats`]: Statistics tracking for each host
 
 mod config;
 mod host;
--- a/lychee-lib/src/ratelimit/mod.rs
+++ b/lychee-lib/src/ratelimit/mod.rs
@@ -6,11 +6,11 @@
 //!
 //! # Architecture
 //!
-//! - [`HostKey`]: Represents a hostname/domain for rate limiting
-//! - [`Host`]: Manages rate limiting, concurrency, and caching for a specific host
-//! - [`HostPool`]: Coordinates multiple hosts and routes requests appropriately
-//! - [`HostConfig`]: Configuration for per-host behavior
-//! - [`HostStats`]: Statistics tracking for each host
+//! - \[`HostKey`\]: Represents a hostname/domain for rate limiting
+//! - \[`Host`\]: Manages rate limiting, concurrency, and caching for a specific host
+//! - \[`HostPool`\]: Coordinates multiple hosts and routes requests appropriately
+//! - \[`HostConfig`\]: Configuration for per-host behavior
+//! - \[`HostStats`\]: Statistics tracking for each host
 
 mod config;
 mod host;
--- a/lychee-lib/src/ratelimit/mod.rs
+++ b/lychee-lib/src/ratelimit/mod.rs
@@ -6,11 +6,11 @@
 //!
 //! # Architecture
 //!
-//! - \[`HostKey`\]: Represents a hostname/domain for rate limiting
-//! - \[`Host`\]: Manages rate limiting, concurrency, and caching for a specific host
-//! - \[`HostPool`\]: Coordinates multiple hosts and routes requests appropriately
-//! - \[`HostConfig`\]: Configuration for per-host behavior
-//! - \[`HostStats`\]: Statistics tracking for each host
+//! - [`host::HostKey`]: Represents a hostname/domain for rate limiting
+//! - [`host::Host`]: Manages rate limiting, concurrency, and caching for a specific host
+//! - [`host::HostPool`]: Coordinates multiple hosts and routes requests appropriately
+//! - [`host::HostConfig`]: Configuration for per-host behavior
+//! - [`host::HostStats`]: Statistics tracking for each host
 
 mod config;
 mod host;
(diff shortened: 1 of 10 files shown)


lychee-bin/src/options.rs
```diff
@@ -409,7 +409,7 @@ pub(crate) struct Config {
 
     /// A list of status codes that will be ignored from the cache
     ///
-    /// The following exclude range syntax is supported: [start]..[[=]end]|code. Some valid
+    /// The following exclude range syntax is supported: `[start`]..`[`[=`]end`]|code. Some valid
```
thomas-zahner: You quoting doesn't make any sense. Let's just wrap the whole expression:

```suggestion
    /// The following exclude range syntax is supported: `[start]..[[=]end]|code`. Some valid
```

lychee-bin/src/options.rs
```diff
@@ -418,7 +418,7 @@ pub(crate) struct Config {
     /// - 500..=599 (excludes any status code from 500 to 599 inclusive)
     /// - 500..600 (excludes any status code from 500 to 600 excluding 600, same as 500..=599)
     ///
-    /// Use "lychee --cache-exclude-status '429, 500..502' <inputs>..." to provide a
+    /// Use "lychee --cache-exclude-status '429, 500..502' `<inputs>`..." to provide a
```
thomas-zahner: This makes no sense. `<inputs>` isn't supposed to represent an HTML element. It's a placeholder documenting CLI usage. Let's wrap the whole thing again.

```suggestion
    /// Use `lychee --cache-exclude-status '429, 500..502' <inputs>...` to provide a
```

lychee-bin/src/options.rs
```diff
@@ -642,7 +642,7 @@ pub(crate) struct Config {
 
     /// A List of accepted status codes for valid links
     ///
-    /// The following accept range syntax is supported: [start]..[[=]end]|code.
+    /// The following exclude range syntax is supported: `[start`]..`[`[=`]end`]|code. Some valid
```
thomas-zahner: `exclude` is wrong. We should do the same as above.

lychee-bin/src/options.rs
```diff
@@ -651,7 +651,7 @@ pub(crate) struct Config {
     /// - 200..=204 (accepts any status code from 200 to 204 inclusive)
     /// - 200..205 (accepts any status code from 200 to 205 excluding 205, same as 200..=204)
     ///
-    /// Use "lychee --accept '200..=204, 429, 500' <inputs>..." to provide a comma-
+    /// Use "lychee --accept '200..=204, 429, 500' `<inputs>`..." to provide a comma-
```
thomas-zahner: Same as in the other comment

lychee-lib/src/ratelimit/mod.rs
```diff
@@ -6,11 +6,11 @@
 //!
 //! # Architecture
 //!
-//! - [`HostKey`]: Represents a hostname/domain for rate limiting
-//! - [`Host`]: Manages rate limiting, concurrency, and caching for a specific host
-//! - [`HostPool`]: Coordinates multiple hosts and routes requests appropriately
-//! - [`HostConfig`]: Configuration for per-host behavior
-//! - [`HostStats`]: Statistics tracking for each host
+//! - \[`HostKey`\]: Represents a hostname/domain for rate limiting
```
thomas-zahner: Escaping makes no sense. Either we remove the links fully or use the fully qualified expression, so that the links properly resolve which is my preferred approach.

```suggestion
//! - [`host::HostKey`]: Represents a hostname/domain for rate limiting
```

lychee-lib/src/types/request_error.rs
```diff
@@ -7,7 +7,7 @@ use crate::{InputSource, ResolvedInputSource};
 
 static ERROR_URI: LazyLock<Uri> = LazyLock::new(|| Uri::try_from("error:").unwrap());
 
-/// An error which occurs while trying to construct a [`Request`] object.
+/// An error which occurs while trying to construct a \[`Request`\] object.
```
thomas-zahner: Same here, let's fully qualify the URL so that it resolves.

lychee-lib/src/types/input/input.rs
```diff
@@ -212,8 +212,8 @@ impl Input {
     /// - Remote URLs are returned as is, in their full form
     /// - Glob patterns are expanded and each matched entry is returned
     /// - Absolute or relative filepaths are returned as-is
-    /// - Stdin input is returned as the special string "<stdin>"
-    /// - A raw string input is returned as the special string "<raw string>"
+    /// - Stdin input is returned as the special string `"<stdin>"`
+    /// - A raw string input is returned as the special string `"<raw string>"`
```
thomas-zahner:
```suggestion
    /// - Stdin input is returned as the special string `<stdin>`
    /// - A raw string input is returned as the special string `<raw string>`
```

thomas-zahner (changes requested):

krishnakurtakoti:
Made suggested changes. Please check.

thomas-zahner:
Thank you for applying my suggestions, all changes look good now. However, in the current state the PR does not fix the mentioned issue so there are still things to do.

- [x] Fix all rustdoc warnings (`cargo doc --no-deps` still yields warnings)
- [x] Remove `continue-on-error: true` in linting step
- [x] CI doesn't pass, test failures need to be addressed

As a side-note I feel like the current PR description doesn't really help. It's really verbose and contains repeated content from my contents as well as confusing parts (e.g. `Rule of Thumb`) that seem to come from a LLM. Maybe you could shorten the description to the essential parts, or just leave the first line.

thomas-zahner:
@katrinafyi now addressed the comment I referenced previously with caa7d5c19540bc14c930271ab8ccc7cdb3f0091a. Consequently I've updated my above comment.

I recommend you to now rebase this PR on master, so that you can test the changes with this new commit included. You can then remove the line with `continue-on-error: true` and update the name of the step. Next you should address all warnings/errors which you can now see in the CI step.

mre:
@krishnakurtakoti, can you rebase this branch on top of `master`?

mre (approved):
Thanks for your contribution @krishnakurtakoti.
I went ahead and fixed the remaining issues.
Will merge this as the changes should be uncontroversial as they only fix lints.

PR #2085, opened 2026-03-11
Nlnet hex sticker

thomas-zahner:


mre:
Which reminds me, shall we add the "Sponsored by NlNet" banner somewhere?

thomas-zahner:
Oh yes, definitely! That's something I had in mind for quite some time :sweat_smile:

PR #2087, opened 2026-03-13
Increase value for --max-redirects

### LINKED ISSUE #1958: Increase value for --max-redirects

thomas-zahner:
Use a default value of 10 instead of 5. It was not entirely uncommon for websites to redirect more than five times. This fixes #1958.

### DIFF
--- a/README.md
+++ b/README.md
@@ -602,7 +602,7 @@ Options:
   -m, --max-redirects <MAX_REDIRECTS>
           Maximum number of allowed redirects
 
-          [default: 5]
+          [default: 10]
 
       --max-cache-age <MAX_CACHE_AGE>
           Discard all cached requests older than this duration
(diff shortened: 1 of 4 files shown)


thomas-zahner:
Thanks for the swift review :)
I was grooming a few issues and noticed these very low hanging fruits

PR #2088, opened 2026-03-13
Fix double count

### LINKED ISSUE #2072: Summary 'Total' count inflated after removal of cross-file deduplication
## Description

After upgrading from lychee v0.21.0 to v0.23.0 (via lychee-action v2.7.0 → v2.8.0), the "Total" count in the summary report jumped from ~3,800 to ~113,000 for the same documentation site (~489 HTML pages).

## Root cause

In v0.21.0, `collector.rs` used a `DashSet` to deduplicate sources across files:

```rust
let seen = Arc::new(DashSet::new());
// ...
.filter_map({
    move |source: Result<String>| {
        let seen = Arc::clone(&seen);
        async move {
            if let Ok(s) = &source {
                if !seen.insert(s.clone()) {
                    return None; // Skip already-seen
                }
            }
            Some(source)
        }
    }
})
```

This was removed between v0.21.0 and v0.23.0. As a result, the "Total" count now reflects **total link occurrences across all files** rather than **unique URLs to check**.

For a documentation site with ~489 HTML pages sharing common navigation, footer links, and cross-references, this means the same URLs are counted hundreds of times.

## Why this matters

- The summary becomes misleading — users see 113K "Total" and may think lychee is doing 113K network requests (it's not, thanks to the in-memory cache).
- It makes it harder to compare runs across versions.
- It inflates other metrics like "Successful" count.

## Suggestion

Consider showing both values in the summary:

```
| 🔍 Total (occurrences) | 113065 |
| 🔗 Unique URLs          | 3847   |
```

Or alternatively, restore deduplication for the "Total" count in the summary while still tracking per-file occurrences for error reporting.

## Environment

- lychee v0.21.0: Total = 3,847
- lychee v0.23.0: Total = 113,065
- Same site, same configuration, same number of pages

cristiklein:
Fixes #2072 

@mre Is this what we want? Notice that de-duplication happens over `Request.uri` so the `span`, `element` and `attribute` of URIs from different sources are lost.

### DIFF
--- a/lychee-bin/src/formatters/stats/response.rs
+++ b/lychee-bin/src/formatters/stats/response.rs
@@ -60,7 +60,7 @@ pub(crate) struct ResponseStats {
     pub(crate) detailed_stats: bool,
     /// Set for counting unique responses
     #[serde(skip)]
-    pub(crate) seen_responses: HashSet<Uri>,
+    pub(crate) seen_uris: HashSet<Uri>,
 }
 
 impl ResponseStats {
@@ -70,7 +70,7 @@ impl ResponseStats {
     pub(crate) fn extended() -> Self {
         Self {
             detailed_stats: true,
-            seen_responses: HashSet::new(),
+            seen_uris: HashSet::new(),
             ..Default::default()
         }
     }
@@ -119,7 +119,7 @@ impl ResponseStats {
 
     /// Update the stats with a new response
     pub(crate) fn add(&mut self, response: Response) {
-        if self.seen_responses.insert(response.body().uri.clone()) {
+        if self.seen_uris.insert(response.body().uri.clone()) {
             self.unique += 1;
         }
 
--- a/lychee-bin/src/formatters/stats/mod.rs
+++ b/lychee-bin/src/formatters/stats/mod.rs
@@ -190,7 +190,7 @@ fn get_dummy_stats() -> OutputStats {
         excluded_map: HashMap::default(),
         timeout_map,
         detailed_stats: true,
-        seen_responses: HashSet::new(),
+        seen_uris: HashSet::new(),
     };
 
     let host_stats = Some(HostStatsMap::from(HashMap::from([(


cristiklein:
@katrinafyi By reading your comment, I realized that I went completely in the wrong direction. The ask was not to reintroduce the DashSet, but to count unique URLs. Thanks for pointing me in the right direction.

I now changed the PR to do exactly that. See https://github.com/lycheeverse/lychee/pull/2088/changes/75d9f9f1b24773c7a2c3d3a0113d66e6bf90c1c9

So now we have:

```console
$ cargo run -- fixtures/double-count/
[...]
🔍 6 Total (in 4ms) 🔗 3 Unique ✅ 6 OK 🚫 0 Errors
```

@afalhambra-hivemq Would you have time to try out and check if this matches your expectations?

mre:
Ooh, the number of unique urls is nice.

katrinafyi:
I wonder if it's possible to use the Status::Cached enum variant to determine whether to increment or not. It would change the metric from "unique" to "non-cached".

I just think it would be nice to avoid another hashset because we already have caches at several layers.

cristiklein:
> I wonder if it's possible to use the Status::Cached enum variant to determine whether to increment or not. It would change the metric from "unique" to "non-cached".
> 
> I just think it would be nice to avoid another hashset because we already have caches at several layers.

I agree it would be nice to avoid another HashSet.

I tried displaying `stats.cached` in addition to `stats.unique` but the results look rather unexpected:

```console
$ cargo run -- fixtures/double-count/
   Compiling lychee v0.23.0 (/home/cklein/elastisys/lychee/lychee-bin)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.30s
     Running `target/debug/lychee fixtures/double-count/`
6/6 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                               
🔍 6 Total (in 4ms) 🔗 3 Unique ♻️ 0 Cached ✅ 6 OK 🚫 0 Errors

$ cargo run -- https://elastisys.io/ https://elastisys.io/welkin/
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/lychee 'https://elastisys.io/' 'https://elastisys.io/welkin/'`
216/216 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                             [WARN] lychee detected 3 redirects. You might want to consider replacing redirecting URLs with the resolved URLs. Run lychee in verbose mode (-v/--verbose) to see details about the redirections.
🔍 216 Total (in 8s 582ms) 🔗 180 Unique ♻️ 24 Cached ✅ 213 OK 🚫 0 Errors 🔀 3 Redirects
```

Notice that `216 (Total) - 24 (Cached) == 192 Non-Cached != 180 Unique`. I wonder if this indicates a bug in caching or if the number of non-cached requests and number of unique links are semantically different. :thinking: 

I see a few options moving forward:

1. Only display number of unique links, to avoid confusion, using the HashSet.
1. Display both non-cached and unique.
1. Try to figure out why non-cached != unique.
1. 2+3

I'm leaning towards 2 now and 3 later.

katrinafyi:
If the goal was to avoid the hash set, displaying both non-unique and cached counts still keeps the hash set around.

As a first guess, the different count could be due to multiple simultaneous requests to the same URL. Those get served by the second-layer cache (https://github.com/lycheeverse/lychee/pull/2067) rather than the top-level cache, and the second later cache doesn't return Status::Cached for its cache hits.

cristiklein:
> If the goal was to avoid the hash set, displaying both non-unique and cached counts still keeps the hash set around.
> 
> As a first guess, the different count could be due to multiple simultaneous requests to the same URL. Those get served by the second-layer cache (#2067) rather than the top-level cache, and the second later cache doesn't return Status::Cached for its cache hits.

Thanks for the explanation. I'm getting the feeling that "unique links" and "non-cached links" have different semantics. "unique links" measures the structure of the website and is "ground truth" which could also be measured by other tools. In contrast, "non-cached links" feels more like a metric measuring lychee's behavior.

As far as I understood, the ask in the original issue was about unique links, in which case I see the added HashSet as beneficial and necessary. 

I propose we add unique links with HashSet for semantic consistency.

What do you think?

katrinafyi:
I don't have strong feelings. It could be said that total links (including duplicates) is also a true value independent of lychee. Non-cached links could be useful as a measure of how much work lychee actually did. Unique links is probably more intuitive to understand. 

I'll leave it to you and others to decide, I don't have a stake here aside from not liking the hash set (but I can deal with it).

afalhambra-hivemq:
> @afalhambra-hivemq Would you have time to try out and check if this matches your expectations?

Just to clarify here. My ask was exactly that, to have a unique URL count. Otherwise, a total URL count jump from 4K to 113K, like we noticed, sounds very misleading.
Am also fine having a count regarding the non-cache URLs count as well. The more detail, the better. So we can configure and adjust Lychee setup accordingly based on this information.

If that's ok, I can give it a try and provide you with some feedback. Thanks for looking into this!

afalhambra-hivemq:
Tested locally against our documentation site (~489 HTML pages, same site referenced in #2072).

**Build:** `cargo install --git https://github.com/cristiklein/lychee.git --branch fix-double-count lychee`

**Command:**
```bash
lychee --insecure -c ./links-check.toml './build/site/**/latest/**/*.html' \
  './build/site/hivemq-edge/**/*.html' \
  './build/site/hivemq-platform-operator/**/*.html' \
  './build/site/hivemq-operator/**/*.html' \
  './build/site/hivemq-cloud/**/*.html' \
  './static/llms.txt' './static/llms-full.txt'
```

**Results:**
```
| Status         | Count  |
|----------------|--------|
| 🔍 Total       | 106097 |
| 🔗 Unique      | 4190   |
| ✅ Successful  | 105443 |
| ⏳ Timeouts    | 0      |
| 🔀 Redirected  | 102    |
| 👻 Excluded    | 212    |
| ❓ Unknown     | 0      |
| 🚫 Errors      | 340    |
```

The new `Unique` row is exactly what we needed — the 4,190 unique URLs aligns with the ~3,800 we saw on v0.21.0 (difference is expected from content growth). The errors are expected local file references (JavaDoc/REST API specs not generated in this build).

lychee-bin/src/formatters/stats/response.rs
```diff
@@ -113,6 +119,10 @@ impl ResponseStats {
 
     /// Update the stats with a new response
     pub(crate) fn add(&mut self, response: Response) {
+        if self.seen_responses.insert(response.body().uri.clone()) {
```
mre: This indicates that `seen_responses` doesn't contain respones, but rather URIs. Should we rename the variable? E.g. `seen_uris`?

mre:
For the final merge, you'd have to rebase one more time as well. :)

lychee-bin/src/formatters/stats/response.rs
cristiklein: I agree fixed.

cristiklein:
@mre Thanks for the feedback. I addressed your comment (yes, `seen_uris` makes a lot more sense). I rebased. CI looks green.

I'd say this is ready to push over the finish line. :checkered_flag:

mre:
Thanks for your contribution @cristiklein. Keep it up! 😊

cristiklein:
> Thanks for your contribution @cristiklein. Keep it up! 😊

And thank you for your mentoring and kind words @mre !

Issue #2061, opened 2026-02-24
https://example.com insecure


thomas-zahner:
On my machine https://example.com seems to encounter certificate issues. This also happens with curl. This has been the case since about 2 weeks. Because of this about 10-20 tests fail on my machine. This doesn't happen in the GitHub actions. It could be that it is simply an issue on my machine, but I have never messed with any certificates, so my setup should be pretty "standard". (NixOS)

<details>
<summary>curl & lychee demo</summary>

```
$ curl https://example.com
curl: (60) SSL certificate OpenSSL verify result: certificate rejected (28)
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the webpage mentioned above.

$ lychee https://example.com
Error: Network error: SSL certificate verification failed. Check certificate validity (error sending request for url (https://example.com/))

Caused by:
    0: error sending request for url (https://example.com/)
    1: client error (Connect)
    2: error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:2124: (certificate rejected)
    3: error:0A000086:SSL routines:tls_post_process_server_certificate:certificate verify failed:ssl/statem/statem_clnt.c:2124:
```

</details>

My questions:

* is anybody else observing this? (@katrinafyi @mre)
* has anybody got an idea why this could happen or how to properly debug? (I observe that the website loads without any issues in Firefox)
* would it make sense to make all of our tests independent of the website?

(related: https://github.com/openwisp/openwisp-controller/issues/1226 & https://github.com/libwww-perl/LWP-Protocol-https/issues/90 & https://github.com/certifi/python-certifi/issues/393 & https://github.com/libwww-perl/LWP-Protocol-https/pull/91)

katrinafyi:
I'm seeing the same issues on my home computer. I assumed it was my Arch packages being out of date 😆 It's good to know it's not just me. Unfortunately, TLS is not something I have good knowledge with.

I know that issuing a certificate needs a domain name, which might make it hard to make our tests independent. It'll always need to connect to _some_ domain, unless we install our own certificates (and that risks jeopardising the test validity)

thomas-zahner:
[...]

PR #2094, opened 2026-03-20
feat: show redirects & remaps

### LINKED ISSUE #1493: Preserve original URLs when `--remap` is used
I'm using the `--remap` feature to validate links pointing to private GitHub repositories containing documentation. Since there's no straightforward way to verify if a link to a private GitHub repository is broken, I'm mitigating this by using both the `--github-token` argument and setting the `--header "Authorization=token $GITHUB_TOKEN"`.

With this setup, the `--remap` feature substitutes links like:

```markdown
[`sre-design-docs/use-conventional-commit-messages.md`](https://github.com/<my org>/<my private repo>/blob/master/docs/design/docs/non-existint-document.md)
```

With their API equivalent:

```bash
✗ [404] https://api.github.com/repos/<my org>/<my private repo>/contents/docs/design/docs/non-existint-document.md?ref=master | Failed: Network error: Not Found
```

This approach works as expected, but I’m wondering if there's a way to convert the output link back to its original Markdown format. The goal is to prevent confusion when users search for the broken link in the document. Ideally, I’d like this validation to occur "in the background" to maintain a seamless user experience.

Thank you for considering this feature request. I hope others find this workaround helpful as well.

### LINKED ISSUE #2011: Redirects aren't reported for rejected status codes
```bash
echo '[a](http://github.com/lycheeverse/lychee/issues/fdsjakf)' > a.md
cargo run -- a.md -v --format markdown   
```
This redirects http to https but the output doesn't mention of redirects because the final status is 404.
```
* [404] <http://github.com/lycheeverse/lychee/issues/fdsjakf> | Rejected status code (this depends on your "accept" configuration): Not Found
```
You can see that there is actually a redirect happening:
```console
$ curl -I http://github.com/lycheeverse/issues/fdsjakf        
HTTP/1.1 301 Moved Permanently
Content-Length: 0
Location: https://github.com/lycheeverse/issues/fdsjakf
```

This is compared to a redirect leading to 200:
```
* [200] <http://github.com/lycheeverse/lychee/issues/2000> | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: http://github.com/lycheeverse/lychee/issues/2000 --[301]--> https://github.com/lycheeverse/lychee/issues/2000
```

It would be nice if the redirects would always be reported, independently of the final status code.

Kind of related to https://github.com/lycheeverse/lychee/issues/1958. If the redirects were reported, that issue would've been easier to debug.

thomas-zahner:
Fixes #1493
Fixes #2011

Please take a look a the commit messages for more details.

# Example

Running `echo 'http://github.com/lycheeverse/lychee' | cargo run - --remap 'github gitlab'` will now yield:

```

[stdin]:
   [403] http://gitlab.com/lycheeverse/lychee (at 1:1) | Rejected status code: 403 Forbidden (configurable with "accept" option) | Remapped: http://github.com/lycheeverse/lychee --> http://gitlab.com/lycheeverse/lychee) | Followed 2 redirects. Redirects: http://gitlab.com/lycheeverse/lychee --[301]--> https://gitlab.com/lycheeverse/lychee --[302]--> https://gitlab.com/users/sign_in

🔍 1 Total (in 325ms) ✅ 0 OK 🚫 1 Error 🔀 1 Redirects
```

# Notes

- As visible in the above example, with the new nesting "Total" no longer adds up to "OK" + "Errors" + "Redirects" ... But I'm not really sure if this was the case previously.
- Just noticed a dangling `)` in the details "Remapped: A --> B)", will adjust and write a test to cover this. It's currently uncovered.

### DIFF
--- a/lychee-bin/src/formatters/stats/response.rs
+++ b/lychee-bin/src/formatters/stats/response.rs
@@ -36,7 +36,7 @@ pub(crate) struct ResponseStats {
     pub(crate) timeouts: usize,
     /// Number of redirects encountered while checking links
     pub(crate) redirects: usize,
-    /// Number of Url remappings performed
+    /// Number of Url remaps performed
     pub(crate) remaps: usize,
     /// Number of links excluded from the run (e.g. due to the `--exclude` flag)
     pub(crate) excludes: usize,
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1924,7 +1924,7 @@ The config file should contain every possible key for documentation purposes."
             .assert()
             .failure()
             .stdout(contains(r#"
-[404] http://rust-lang.org/lycheeverse (at 1:1) | Rejected status code: 404 Not Found (configurable with "accept" option) | Followed 1 redirect. Redirects: http://rust-lang.org/lycheeverse --[301]--> https://rust-lang.org/lycheeverse | Remapped: http://github.com/lycheeverse --> http://rust-lang.org/lycheeverse
+[404] http://rust-lang.org/lycheeverse (at 1:1) | Rejected status code: 404 Not Found (configurable with "accept" option) | Followed 1 redirect. Redirects: http://rust-lang.org/lycheeverse --[301]--> https://rust-lang.org/lycheeverse | Remaps: http://github.com/lycheeverse --> http://rust-lang.org/lycheeverse
 "#))
         // It is debugged when URIs are remapped
         .stderr(contains("[DEBUG] Remapping http://github.com/lycheeverse --> http://rust-lang.org/lycheeverse"))
@@ -1943,7 +1943,7 @@ The config file should contain every possible key for documentation purposes."
             .assert()
             .failure()
             // The error message should be descriptive and helpful
-            .stderr(contains("Error checking URL https://example.com/: Invalid remapping pattern: the result `invalid/` is not a valid URL"))
+            .stderr(contains("Error checking URL https://example.com/: Invalid remap pattern: the result `invalid/` is not a valid URL"))
             // The original URI is shown as root cause in stdout
             .stdout(contains("The given URI is invalid, check URI syntax: https://example.com/"));
     }
@@ -1973,7 +1973,7 @@ The config file should contain every possible key for documentation purposes."
                 "url": "https://bbb.com/",
                 "status": {
                   "text": "Excluded",
-                  "details": "This is due to your 'exclude' values | Remapped: https://aaa.com/ --> https://bbb.com/"
+                  "details": "This is due to your 'exclude' values | Remaps: https://aaa.com/ --> https://bbb.com/"
                 },
                 "span": {
                   "line": 1,
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -13,7 +13,7 @@
     clippy::default_trait_access,
     clippy::used_underscore_binding
 )]
-use crate::remap::Remapping;
+use crate::remap::Remap;
 use std::{collections::HashSet, sync::Arc, time::Duration};
 
 use http::{
@@ -82,7 +82,7 @@ pub struct ClientBuilder {
     ///
     /// # Usage Notes
     ///
-    /// Use with caution because a large set of remapping rules may cause
+    /// Use with caution because a large set of remap rules may cause
     /// performance issues.
     ///
     /// Furthermore rules are executed sequentially and multiple mappings for
@@ -492,7 +492,7 @@ fn redirect_policy(redirect_history: RedirectHistory, max_redirects: usize) -> r
 /// options.
 #[derive(Debug, Clone)]
 pub struct Client {
-    /// Optional remapping rules for URIs matching pattern.
+    /// Optional remap rules for URIs matching pattern.
     remaps: Option<Remaps>,
 
     /// Rules to decide whether a given link should be checked or ignored.
@@ -541,7 +541,7 @@ impl Client {
         } = request.try_into()?;
 
         let start = std::time::Instant::now(); // Measure check time
-        let remapping = self.remap(&mut uri)?;
+        let remap = self.remap(&mut uri)?;
 
         let status = match uri.scheme() {
             _ if self.is_excluded(&uri) => Status::Excluded,
@@ -551,8 +551,8 @@ impl Client {
             _ => self.check_website(&uri, credentials).await?,
         };
 
-        let status = match remapping {
-            Some(remapping) => Status::Remapped(Box::new(status), remapping),
+        let status = match remap {
+            Some(remap) => Status::Remapped(Box::new(status), remap),
             None => status,
         };
 
@@ -570,13 +570,13 @@ impl Client {
         self.file_checker.check(uri).await
     }
 
-    /// Remap [`Uri`] as a side-effect, using the client-defined remapping rules.
-    /// Return `Some` only if a remapping was performed.
+    /// Remap [`Uri`] as a side-effect, using the client-defined remap rules.
+    /// Return `Some` only if a remap was performed.
     ///
     /// # Errors
     ///
     /// Returns an `Err` if the remapped `uri` is not a valid URI.
-    pub fn remap(&self, uri: &mut Uri) -> Result<Option<Remapping>> {
+    pub fn remap(&self, uri: &mut Uri) -> Result<Option<Remap>> {
         match self.remaps {
             Some(ref remaps) => {
                 let remapped = remaps.remap(uri)?;
@@ -663,7 +663,7 @@ mod tests {
     use crate::{
         ErrorKind, Redirect, Redirects, Request, Status, Uri,
         chain::{ChainResult, Handler, RequestChain},
-        remap::{Remapping, Remaps},
+        remap::{Remap, Remaps},
     };
 
     #[tokio::test]
@@ -1009,7 +1009,7 @@ mod tests {
                 Box::new(Status::Error(ErrorKind::InvalidFilePath(
                     format!("{mapped}/").try_into().unwrap(),
                 ))),
-                Remapping {
+                Remap {
                     original: input,
                     new: format!("{mapped}/").try_into().unwrap(),
                 },
--- a/lychee-lib/src/remap.rs
+++ b/lychee-lib/src/remap.rs
@@ -1,10 +1,10 @@
-//! Remapping rules which allow to map URLs matching a pattern to a different
+//! Remap rules which allow to map URLs matching a pattern to a different
 //! URL.
 //!
 //! # Notes
 //! Use in moderation as there are no sanity or performance guarantees.
 //!
-//! - There is no constraint on remapping rules upon instantiation or during
+//! - There is no constraint on remap rules upon instantiation or during
 //!   remapping. In particular, rules are checked sequentially so later rules
 //!   might contradict with earlier ones if they both match a URL.
 //! - A large rule set has a performance impact because the client needs to
@@ -30,14 +30,14 @@ use crate::{ErrorKind, Result, Uri};
 
 /// Records a single [`Uri`] remapping
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize)]
-pub struct Remapping {
+pub struct Remap {
     /// The original [`Uri`] before remapping
     pub original: Uri,
     /// The new [`Uri`] after applying [`Remaps`]
     pub new: Uri,
 }
 
-impl Display for Remapping {
+impl Display for Remap {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "{} --> {}", self.original, self.new)?;
         Ok(())
@@ -76,9 +76,9 @@ impl Remaps {
     ///
     /// # Errors
     ///
-    /// Returns an `Err` if the remapping rule produces an invalid URL.
+    /// Returns an `Err` if the remap rule produces an invalid URL.
     #[must_use = "Remapped URLs must be used"]
-    pub fn remap(&self, original: &Uri) -> Result<Option<Remapping>> {
+    pub fn remap(&self, original: &Uri) -> Result<Option<Remap>> {
         for (pattern, replacement) in self {
             if pattern.is_match(original.as_str()) {
                 let new = pattern.replace_all(original.as_str(), replacement);
@@ -86,25 +86,25 @@ impl Remaps {
                     ErrorKind::InvalidUrlRemap(format!("the result `{new}` is not a valid URL"))
                 })?;
 
-                let remapping = Remapping {
+                let remap = Remap {
                     original: original.clone(),
                     new: Uri { url: new },
                 };
-                debug!("Remapping {remapping}");
-                return Ok(Some(remapping));
+                debug!("Remapping {remap}");
+                return Ok(Some(remap));
             }
         }
 
         Ok(None)
     }
 
-    /// Returns `true` if there is no remapping rule defined.
+    /// Returns `true` if there is no remap rule defined.
     #[must_use]
     pub const fn is_empty(&self) -> bool {
         self.patterns.is_empty()
     }
 
-    /// Get the number of remapping rules.
+    /// Get the number of remap rules.
     #[must_use]
     pub const fn len(&self) -> usize {
         self.patterns.len()
@@ -122,7 +122,7 @@ impl Index<usize> for Remaps {
 impl TryFrom<&[String]> for Remaps {
     type Error = ErrorKind;
 
-    /// Try to convert a slice of `String`s to remapping rules.
+    /// Try to convert a slice of `String`s to remap rules.
     ///
     /// Each string should contain a Regex pattern and a URL, separated by
     /// whitespaces.
@@ -140,7 +140,7 @@ impl TryFrom<&[String]> for Remaps {
             let params: Vec<_> = remap.split_whitespace().collect();
             if params.len() != 2 {
                 return Err(ErrorKind::InvalidUrlRemap(format!(
-                    "Cannot parse into URI remapping, must be a Regex pattern and a URL separated by whitespaces: {remap}"
+                    "Cannot parse into URI remap rule, must be a Regex pattern and a URL separated by whitespaces: {remap}"
                 )));
             }
 
@@ -183,7 +183,7 @@ mod tests {
 
         assert_eq!(
             output,
-            Some(Remapping {
+            Some(Remap {
                 new: Uri::try_from(replacement).unwrap(),
                 original: input_url
             })
@@ -201,7 +201,7 @@ mod tests {
 
         assert_eq!(
             output,
-            Some(Remapping {
+            Some(Remap {
                 new: replacement,
                 original: input
             })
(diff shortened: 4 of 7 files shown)


lychee-lib/src/remap.rs
```diff
@@ -35,19 +39,29 @@ use crate::{ErrorKind, Result};
 /// # Notes
 /// See module level documentation of usage notes.
 #[derive(Debug, Clone)]
-pub struct Remaps(Vec<(Regex, String)>);
+pub struct Remaps {
+    patterns: Vec<(Regex, String)>,
+    history: RemapHistory,
+}
```
katrinafyi, line 58: Why does the remap history need to be centrally stored?

Remapping is only done in one place, at [lib client.rs](https://github.com/lycheeverse/lychee/blob/caa7d5c19540bc14c930271ab8ccc7cdb3f0091a/lychee-lib/src/client.rs#L545), and that function returns a Response which has a Status field. It seems like the Status::Redirected wrapping could just happen in that function and that would be much less work and memory, but maybe I'm missing something.

lychee-lib/src/checker/website.rs
```diff
@@ -214,9 +221,14 @@ impl WebsiteChecker {
...
 
         let status = self.check_website_inner(uri, &default_chain).await;
-        let status = self
+        let mut status = self
             .handle_insecure_url(uri, &default_chain, status)
             .await?;
+
+        if let Some(history) = &self.remap_history {
+            status = history.handle_remapped(&uri.url, status);
+        }
```
katrinafyi: Remapping can happen for all URL types not just websites (unlike redirects). See prev comment about moving the handling to client.rs.

It would be worth adding a test for remap history of file:// URLs.

lychee-lib/src/types/status.rs
```diff
@@ -94,9 +93,27 @@ impl Serialize for Status {
...
+        loop {
+            match current {
+                Status::Redirected(inner, _) | Status::Remapped(inner, _) => {
+                    current = inner;
+                }
+                other => break matches!(other, $pattern),
+            }
+        }
+    }};
+}
```
katrinafyi: This caused me to file https://github.com/lycheeverse/lychee/issues/2097 :') I think that the need for this macro means that the Status enum is not a good representation of the information we're interested in.

But it's not something isolated to this PR, the Status enum has been drifting and growing in scope for some time now. I wouldn't worry about any specific changes to this PR right now.

lychee-lib/src/remap.rs
thomas-zahner: Thank you for noticing. You are so right, there is no need! See d9d9320e. I did it probably because I was copying the redirect history approach which is really not necessary.

lychee-lib/src/checker/website.rs
thomas-zahner: Totally true! See d9d9320e with `test_remaps` as well as 22d9df0167f99673dfc8067536b5d9cad7907229

lychee-lib/src/types/status.rs
thomas-zahner: That really is great, thanks! I agree that it's better do tackle this separately in another PR.

lychee-lib/src/client.rs
```diff
@@ -556,13 +554,17 @@ impl Client {
 
         let start = std::time::Instant::now(); // Measure check time
 
-        let status = match uri.scheme() {
+        let mut status = match uri.scheme() {
             _ if uri.is_tel() => Status::Excluded, // We don't check tel: URIs
```
katrinafyi, line 548: If you want, you can consolidate the excluded case into this match to get remap history for excluded URLs. (This would also add elapsed time for excluded URLs - could be changed).

```suggestion
            _ if self.is_excluded(uri) => Status::Excluded,
            _ if uri.is_tel() => Status::Excluded, // We don't check tel: URIs
```

lychee-lib/src/client.rs
```diff
@@ -556,13 +554,17 @@ impl Client {
```
katrinafyi: ... with above
```suggestion
```

lychee-lib/src/client.rs
```diff
@@ -556,13 +554,17 @@ impl Client {
...
+        let mut status = match uri.scheme() {
             _ if uri.is_tel() => Status::Excluded, // We don't check tel: URIs
             _ if uri.is_file() => self.check_file(uri).await,
             _ if uri.is_mail() => self.check_mail(uri).await,
             _ => self.check_website(uri, credentials).await?,
         };
 
+        if let Some(remapping) = remapping {
+            status = Status::Remapped(Box::new(status), remapping);
+        }
```
katrinafyi: If you want, you can shadow status to avoid the `let mut`.

```suggestion
        let status = match remapping {
            Some(remapping) => Status::Remapped(Box::new(status), remapping),
            None => status,
        };
```

lychee-lib/src/client.rs
```diff
@@ -952,12 +963,12 @@ mod tests {
...
 
-        assert_eq!(
+        assert!(matches!(
             res.status(),
-            &Status::Error(ErrorKind::RejectedStatusCode(
-                StatusCode::PERMANENT_REDIRECT
-            ))
-        );
+            Status::Redirected(inner, redirects) if **inner == Status::Error(
+                ErrorKind::RejectedStatusCode(StatusCode::PERMANENT_REDIRECT)
```
katrinafyi, line 960: I was curious if there's a way to avoid the `**`.. maybe [in future](https://doc.rust-lang.org/beta/unstable-book/language-features/deref-patterns.html) :)

lychee-lib/src/client.rs
katrinafyi:
```suggestion
            mut uri,
```
I don't think it's specific to this PR, but if you change this to non-ref and uses to `&` / `& mut`, it can avoid cloning in the returned Response.

lychee-lib/src/remap.rs
```diff
@@ -19,12 +19,30 @@
...
+use crate::{ErrorKind, Result, Uri};
+
+/// Records a single [`Uri`] remapping
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize)]
+pub struct Remapping {
+    /// The original [`Uri`] before remapping
+    pub original: Uri,
+    /// The new [`Uri`] after applying [`Remaps`]
+    pub new: Uri,
+}
```
katrinafyi: I wonder if something like this would be good.
```suggestion
pub struct Remapping {
    /// The [`Uri`] after remapping is considered and, if applicable, applied.
    pub uri: Uri,
    /// The original [`Uri`], if remapping was applied.
    pub original: Option<Uri>,
}
```
It has some benefits:
- consolidates some code paths. you could add a method to `Remapping` that wraps in Status::Redirected if applicable, since Remapping now knows about remap status.
- avoids nested result+option `Result<Option<Remapping>>`
- `remap` functions could be changed to move `Uri` in, since it always gets returned back to the caller. this avoids `&mut Uri` and avoids one clone on the remapped code path, I think.

Some downsides:
- no longer fits into `Remapped(Box<Status>, Remapping)` which represents that a remap _has_ occurred. maybe this could be changed to `Remapped(Box<Status>, Uri, Uri)` if it's wanted.

But this is totally up to you, I'm just thinking.

lychee-lib/src/types/status.rs
katrinafyi: If you want, you could generalise the macro to just return the inner status. Then, the caller could use `matches!` or something else if they wanted. Very optional.

lychee-bin/tests/cli.rs
```diff
@@ -1914,6 +1912,42 @@ The config file should contain every possible key for documentation purposes."
...
+            .arg("-vv")
+            .arg("--remap")
+            .arg("github.com rust-lang.org")
+            .arg("-")
+            .write_stdin("http://github.com/lycheeverse\n")
+            .env_clear()
+            .assert()
+            .failure()
+            .stdout(contains(r#"
+[404] http://rust-lang.org/lycheeverse (at 1:1) | Rejected status code: 404 Not Found (configurable with "accept" option) | Followed 1 redirect. Redirects: http://rust-lang.org/lycheeverse --[301]--> https://rust-lang.org/lycheeverse | Remapped: http://github.com/lycheeverse --> http://rust-lang.org/lycheeverse
```
katrinafyi: Now this is a real nitpick, but it would be nice if the remaps could be printed *before* redirects. It would make all the remaps/redirects flow from left `-->` right.

Also could consider combining them which further emphasises the order of operations:
```
Followed remap and 1 redirect: http://github.com/lycheeverse --[remap]--> http://rust-lang.org/lycheeverse --[301]--> https://rust-lang.org/lycheeverse
```
But this might be difficult with the current Status::Redirected / Status::Remapped nesting, so it's definitely optional.

lychee-lib/src/types/status.rs
thomas-zahner: True! I've created `pub const fn innermost(&self) -> &Self` in 0f9411392fbbda0cd42e6de907b7ef4390a41192

lychee-lib/src/client.rs
thomas-zahner: Ah yes that would be great :)

lychee-lib/src/remap.rs
thomas-zahner: Hmm, I really like having `Remapping` in `Remapped(Box<Status>, Remapping)` which represents a valid remapping. I do see the benefit in the `remap` functions, but I think I'd like to keep `Remapped(Box<Status>, Remapping)`. Thanks for your thoughts.

lychee-bin/tests/cli.rs
thomas-zahner: Yeah, nesting makes it a bit difficult :sweat_smile: so I think this is out of scope for the PR but it's an awesome suggestion. See https://github.com/lycheeverse/lychee/issues/2102

thomas-zahner:
> In testing, I also noticed an interesting interaction with caching:

That's interesting. If the status is really loaded from the disk's cache I think it would be expected behaviour. But thinking about this specific case, I'm not sure this should happen at all. See https://github.com/lycheeverse/lychee/issues/2103

I also realised that the `redirect_map` became useless in the process, as it was no longer a "terminal" status code. See a2b72ecb

fixtures/TEST_HTTP.html
```diff
@@ -1 +1 @@
-<a href="http://example.com">Insecure HTTP link</a>
+<a href="http://rust-lang.org/">Insecure HTTP link</a>
```
mre: Not sure why this was changed? Is it related to the feature?

lychee-bin/src/formatters/stats/mod.rs
```diff
@@ -161,34 +161,19 @@ fn get_dummy_stats() -> OutputStats {
...
         redirects: 1,
-        cached: 0,
         suggestion_map,
         redirect_map,
-        success_map: HashMap::default(),
         error_map,
-        excluded_map: HashMap::default(),
         timeout_map,
         detailed_stats: true,
+        ..Default::default()
```
mre, line 177: Not sure about that one. In the future, we might add new fields and forget to explicitly check them here. Not a huge fan of `..Default::default()` anymore because of that, even though it reads nicer.

lychee-bin/src/formatters/stats/response.rs
```diff
@@ -32,8 +32,10 @@ pub(crate) struct ResponseStats {
     pub(crate) unsupported: usize,
     /// Number of timeouts
     pub(crate) timeouts: usize,
-    /// Redirects encountered while checking links
+    /// Number of redirects encountered while checking links
     pub(crate) redirects: usize,
+    /// Number of Url remappings performed
```
mre: Is it "remappings" or "remaps"? 🤔

lychee-bin/tests/cli.rs
```diff
@@ -1381,7 +1376,7 @@ The config file should contain every possible key for documentation purposes."
             .assert()
             .stderr(contains(format!("[200] {}/ (at 1:1)\n", mock_server_ok.uri())))
             .stderr(contains(format!(
-                "[204] {}/ (at 2:1) | 204 No Content: No Content\n",
```
mre, line 1384: Thanks for cleaning up those redundant bits.

lychee-bin/tests/cli.rs
```diff
@@ -1914,6 +1907,78 @@ The config file should contain every possible key for documentation purposes."
...
+            .arg("-vv")
+            .arg("--remap")
+            .arg("github.com rust-lang.org")
+            .arg("-")
+            .write_stdin("http://github.com/lycheeverse\n")
+            .env_clear()
+            .assert()
+            .failure()
+            .stdout(contains(r#"
+[404] http://rust-lang.org/lycheeverse (at 1:1) | Rejected status code: 404 Not Found (configurable with "accept" option) | Followed 1 redirect. Redirects: http://rust-lang.org/lycheeverse --[301]--> https://rust-lang.org/lycheeverse | Remapped: http://github.com/lycheeverse --> http://rust-lang.org/lycheeverse
```
mre: The term 'Redirects:' is a plural noun, whereas the term 'Remapped' is a past participle. I think we should align the naming and say either 'Redirected/Remapped' or 'Redirects/Remaps.'

lychee-bin/tests/cli.rs
```diff
@@ -1914,6 +1907,78 @@ The config file should contain every possible key for documentation purposes."
...
+        cargo_bin_cmd!()
+            .arg("--remap")
+            .arg("https://example.com invalid")
+            .arg("-")
+            .write_stdin("https://example.com")
+            .env_clear()
+            .assert()
+            .failure()
+            // The error message should be descriptive and helpful
+            .stderr(contains("Error checking URL https://example.com/: Cannot parse URL to URI: Error remapping URL: `The remapping pattern produced an invalid URL: invalid/`"))
```
mre: The comment says "descriptive and helpful." 😆  Not sure if a lot of users understand the error message. Was thinking of

```rust
.stderr(contains("Error checking URL https://example.com/: Invalid remapping pattern: the result `invalid/` is not a valid URL"))
```

But not sure how much impact we have on the error message format within the scope of this PR.

lychee-lib/src/types/error.rs
```diff
@@ -119,32 +122,30 @@ pub enum ErrorKind {
...
     /// The given glob pattern is not valid
-    #[error("UNIX glob pattern is invalid")]
+    #[error("Invalid glob pattern: {0}")]
     InvalidGlobPattern(#[from] glob::PatternError),
 
     /// The GitHub API could not be called because of a missing GitHub token.
-    #[error(
-        "GitHub token not specified. To check GitHub links reliably, use `--github-token` flag / `GITHUB_TOKEN` env var."
-    )]
+    #[error("GitHub token required")]
```
mre, line 129: Do we show the rest of the error message somewhere else? I always found the note about `GITHUB_TOKEN` quite helpful.

lychee-lib/src/types/error.rs
```diff
@@ -198,163 +199,120 @@ impl ErrorKind {
...
-                }
-                _ => Some(format!("Stdin read error: {e}")),
-            },
-            ErrorKind::ParseUrl(e, _url) => match e {
-                url::ParseError::RelativeUrlWithoutBase => Some(
-                    "This relative link was found inside an input source that has no base location"
-                        .to_string(),
                 ),
-                _ => None,
+                _ => format!("File read error for '{}': {}", path.display(), e),
```
mre: We *could* inline `{e}`.

lychee-lib/src/types/status.rs
```diff
@@ -31,7 +31,9 @@ pub enum Status {
     /// Request timed out
     Timeout(Option<StatusCode>),
     /// Got redirected to different resource
-    Redirected(StatusCode, Redirects),
+    Redirected(Box<Status>, Redirects),
+    /// The original [`Url`] was remapped in the link check process
+    Remapped(Box<Status>, Remapping),
```
mre: In comparison to the status code, the full Status might need quite some memory. Should be fine, though, since we only ever store a few concrete variants here.

lychee-lib/src/remap.rs
```diff
@@ -19,12 +19,30 @@
...
+
+impl Display for Remapping {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{} --> {}", self.original, self.new)?;
```
mre, line 42: Did we consider a "pretty" arrow instead? It's also Unicode: `→`.
We use arrows elsewhere, though, so it might not be consistent then, but at least this would also work: `── [ 200 ] ─→`.

fixtures/TEST_HTTP.html
thomas-zahner: Because it doesn't run on my machine as per https://github.com/lycheeverse/lychee/issues/2061 and I probably touched the tests with my changes :sweat_smile:

lychee-bin/src/formatters/stats/mod.rs
thomas-zahner: Well in this case I think it's okay, if we forget about and it's not a default value we get a failing test as a result. But if you prefer I can also revert that.

lychee-lib/src/types/error.rs
thomas-zahner: Yes exactly, with `details(&self)`, I've changed/cleaned up the logic quite a bit in how errors are formatted. This particular variant does the following in `details`:

```rust
            ErrorKind::MissingGitHubToken => {
                format!("{self}. Use --github-token flag or GITHUB_TOKEN environment variable")
            }
```

lychee-lib/src/types/status.rs
thomas-zahner: Yes I agree it might not be fully optimal but definitely okay. As the long-term approach we now have #2097

lychee-lib/src/remap.rs
thomas-zahner: Hmm yeah that might be a good idea. To not expand the scope of this PR further I will keep it ASCII.
I think this fits nicely into #2102, which I've updated accordingly.

fixtures/TEST_HTTP.html
mre: Ah yeah 😄

lychee-lib/src/remap.rs
mre: Thanks

lychee-bin/tests/cli.rs
thomas-zahner: True :laughing: done in e26e91ce6f0a5499368ff20d22d336c4eb3c4a7f

fixtures/TEST_HTTP.html
thomas-zahner: Undone in e26e91ce6f0a5499368ff20d22d336c4eb3c4a7f

lychee-bin/src/formatters/stats/response.rs
thomas-zahner: I've tried to replace all instances of "remapping" to "remap" where used as a gerund. So "remapping" is now only used where it's necessary, as in "the process of remapping". See a21c14a771dd902eb6d6324e5180d6985bf69dd7

lychee-bin/tests/cli.rs
thomas-zahner: See a21c14a771dd902eb6d6324e5180d6985bf69dd7, thank you for pointing out :+1:

Issue #1886, opened 2025-10-27
User agent not set when fetching command-line URLs


katrinafyi:
These two commands will behave differently. The second sets User-agent to lychee/0.21.0 (the default value), but the first does not set it at all.
```bash
cargo run -- https://webhook.site/a1895fe4-5d07-47e2-bbfe-28982b72d16f           
echo 'https://webhook.site/a1895fe4-5d07-47e2-bbfe-28982b72d16f' | cargo run -- -  
```
By going to [the view link](https://webhook.site/#!/view/a1895fe4-5d07-47e2-bbfe-28982b72d16f/dcd73421-e4c9-49eb-acf9-4c8f051e7c3a/1), you can see the two requests. (The link might expire - you can make your own at https://webhook.site).

This is related to the [`url_contents`](https://github.com/search?q=repo%3Alycheeverse%2Flychee+url_contents&type=code) function which implements the input fetching logic. This is a separate code path from the main link checking request logic which explains the discrepancy. (Imo, it would be good to unify the code paths - this would also help with doing recursion).

The omission of user-agent is the underlying cause of https://github.com/lycheeverse/lychee/issues/1767#issuecomment-3388580717, where a wikipedia page on the command line appears to contain no links at all.

This issue was suggested by @ thomas-zahner in https://github.com/lycheeverse/lychee/issues/1883#issuecomment-3450358700.

mre:
Fixed in `master` and soon to be released. 😊

PR #2099, opened 2026-03-25
set user-agent when fetching remote input URLs

mre:
This is a veeery conservative attempt at fixing https://github.com/lycheeverse/lychee/issues/1886.

Previously, the `reqwest::Client` used by the `UrlContentResolver` (which fetches CLI URL inputs) was built without a user-agent. This meant that passing a URL directly as a CLI argument resulted in a request with no user-agent header, unlike links discovered during checking which correctly use the configured user-agent.

This caused subtle bugs such as Wikipedia returning a 403 for CLI inputs (since it requires a user-agent), resulting in lychee finding zero links on pages that actually contain hundreds.

Fixed by storing the configured user-agent on the `Collector` and using it to build the resolver's `reqwest::Client`.

Let me say that I believe the `collector` is already doing way too much and that this is probably not the right approach going forward. The fact that the `collector` is responsible for building a `reqwest::Client` for the resolver is a sign of tight coupling and a violation of separation of concerns. The `collector` should ideally only be responsible for collecting links, while the resolver should be responsible for fetching content.

That said, I opened the PR to get some early feedback and perhaps to fix the immediate issue. I also wanted to avoid a larger refactor that would touch many parts of the codebase and potentially cause more disruption in the short term. (See below.)

As an alternative, I considered to unify the two code paths where we deal with "building request clients." Instead of the `Collector` using its own bare `reqwest::Client`, we can make it reuse the same fully-configured `lychee_lib::Client` (with rate limiting, retries, per-host config, TLS settings, etc.) that's used for link checking. The original issue already notes that unifying these paths would also help with recursion support. The downside is it's a much larger refactor touching the `Client` API.
But it's probably still worth it? I can create another PR for that and we can compare.
I'm a bit hesitant to do it because large PRs have caused some churn in the past and it took a while to get them merged. But I'm willing to do it if we think it's the right move. I just don't want to cause too much disruption in the short term.

Let me know what you think!

@katrinafyi  @thomas-zahner @cristiklein feedback welcome.

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -339,7 +339,8 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
         // File a bug if you rely on this envvar! It's going to go away eventually.
         .use_html5ever(std::env::var("LYCHEE_USE_HTML5EVER").is_ok_and(|x| x == "1"))
         .include_wikilinks(opts.config.include_wikilinks)
-        .preprocessor(opts.config.preprocess.clone());
+        .preprocessor(opts.config.preprocess.clone())
+        .user_agent(opts.config.user_agent());
 
     collector = if let Some(ref basic_auth) = opts.config.basic_auth {
         collector.basic_auth_extractor(BasicAuthExtractor::new(basic_auth)?)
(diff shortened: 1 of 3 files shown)


thomas-zahner:
I think we're all agreeing to go your less conservative approach proposed in #2100.

PR #2100, opened 2026-03-25
refactor: unify input URL fetching with the link-checker's HostPool

### LINKED ISSUE #1673: `--insecure` flag does not work with self-signed cert hosted in a Docker container

### LINKED ISSUE #1886: User agent not set when fetching command-line URLs

mre:
This is the follow-up to #2099, which took a conservative approach to fixing #1886.

Previously, the `reqwest::Client` used by `UrlContentResolver` (which fetches the body of remote CLI input URLs) was built without a user-agent, rate limiting, retries, TLS settings, or per-host configuration. This meant that passing a URL directly as a CLI argument silently diverged from how link checking works. For example, Wikipedia returns a `403` with no user-agent set, so `lychee https://en.wikipedia.org/wiki/...` would find zero links and report success.

The fix in #2099 was intentionally minimal: store the configured user-agent on the `Collector` and use it when building the resolver's `reqwest::Client`. It fixes the immediate issue but treats the two code paths separately, which isn't great.

This PR takes the approach I described in #2099 as the "alternative": instead of the `Collector` maintaining its own `reqwest::Client`, it now shares the same `Arc<HostPool>` that the link checker uses. The `lychee_lib::Client` is built before the `Collector` in `main.rs`, and its pool is handed to the `Collector` via the new `.host_pool()` builder method. Both input fetching and link checking now go through the same pool, so all configuration  (user-agent, custom headers, per-host headers, TLS, cookies, rate limiting, retries) is applied consistently to both paths.

As a side effect, fetching a remote input document now counts against the per-host rate limit bucket for that host. This is actually the correct behavior since we want lychee to be a good web citizen regardless of whether a request is for input fetching or link checking. =)

One tradeoff worth noting: `Collector::default()` and `Collector::new()` (which are used in tests without a full `ClientBuilder` setup) now fall back to `HostPool::default()` instead of `reqwest::Client::new()`. `HostPool::default()` is equally lightweight because it just wraps a default `reqwest::Client` with lazy host creation, so this should not be a big deal in practice, but it's worth mentioning.

I believe now that this is the superior approach to resolve the issue. wdyt?

Fixes #1886
Fixes #1673

@katrinafyi @thomas-zahner @cristiklein feedback welcome!

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -340,10 +340,6 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
         )
     })?;
 
-    // Build the lychee client first so we can share its HostPool with the
-    // collector. This ensures that fetching remote input documents uses the
-    // same user-agent, TLS settings, cookies, per-host rate limits and custom
-    // headers as regular link checks.
     let client = client::create(&opts.config, cookie_jar.as_deref())?;
 
     let mut collector = Collector::new(opts.config.root_dir.clone(), base.unwrap_or_default())?
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -2798,9 +2798,6 @@ The config file should contain every possible key for documentation purposes."
             .arg(server.uri())
             .assert()
             .success();
-
-        // wiremock will fail if the expected request was not received
-        server.verify().await;
     }
 
     #[tokio::test]
@@ -2825,8 +2822,9 @@ The config file should contain every possible key for documentation purposes."
         let user_agent = received_request
             .headers
             .get("user-agent")
-            .and_then(|v| v.to_str().ok())
-            .unwrap_or("");
+            .expect("User agent missing")
+            .to_str()
+            .unwrap();
 
         assert!(
             user_agent.starts_with("lychee/"),
--- a/lychee-lib/src/collector.rs
+++ b/lychee-lib/src/collector.rs
@@ -157,10 +157,12 @@ impl Collector {
     /// fetches share the same user-agent, TLS settings, cookies, per-host
     /// rate limits and headers as regular link checks:
     ///
-    /// ```ignore
-    /// let lychee_client = ClientBuilder::builder().client()?;
-    /// let collector = Collector::new(…)?
-    ///     .host_pool(lychee_client.host_pool());
+    /// ```
+    /// # use lychee_lib::{BaseInfo, ClientBuilder, Collector, ErrorKind};
+    /// let client = ClientBuilder::builder().build().client()?;
+    /// let collector = Collector::new(None, BaseInfo::none())?
+    ///     .host_pool(client.host_pool());
+    /// # Ok::<(), ErrorKind>(())
     /// ```
     #[must_use]
     pub fn host_pool(mut self, host_pool: Arc<HostPool>) -> Self {
@@ -621,13 +623,13 @@ mod tests {
         use wiremock::{Mock, MockServer, ResponseTemplate};
 
         let mock_server = MockServer::start().await;
+        let uri = Uri::try_from("https://example.com").unwrap();
 
         Mock::given(method("GET"))
             .and(path("/"))
             .and(header("user-agent", "test-agent/1.0"))
             .respond_with(
-                ResponseTemplate::new(200)
-                    .set_body_string(r#"<a href="https://example.com">Link</a>"#),
+                ResponseTemplate::new(200).set_body_string(format!(r#"<a href="{uri}">Link</a>"#)),
             )
             .expect(1)
             .mount(&mock_server)
@@ -653,8 +655,7 @@ mod tests {
             .collect::<std::collections::HashSet<_>>()
             .await;
 
-        assert!(links.iter().any(|u| u.url.as_str().contains("example.com")));
-        // wiremock will panic here if the expected request was not received
+        assert_eq!(links, HashSet::from([uri]));
     }
 
     #[tokio::test]
--- a/lychee-lib/src/types/resolver.rs
+++ b/lychee-lib/src/types/resolver.rs
@@ -51,7 +51,7 @@ impl UrlContentResolver {
         // Fetch with body since we need to extract links from the content.
         let response = self.host_pool.execute_request(request, true).await?;
 
-        // needs_body=true above guarantees text is populated on success.
+        // SAFETY: needs_body=true above guarantees text is populated on success.
         let content = response.text.unwrap_or_else(|| {
             unreachable!("execute_request with needs_body=true always returns text")
         });


cristiklein (commented):
Hi @mre . Thanks for involving me.

Overall, I like the idea of lychee having a single, shared HostPool which controls all host-related parameters and applies them uniformly to both fetching input URLs and collected URLs.

I have two comments:

1. It noticed that #2099 contains a few tests. Would be great to add them to this PR to show that something which was previously broken is now fixed.
2. I'm surprised by the need to go from `pub(create)` to `pub`. Is that really necessary?

mre:
> I'm surprised by the need to go from pub(create) to pub. Is that really necessary?

Oh, you're right. It *was* necessary, but thanks to some refactoring it's not necessary anymore. 😄 👍 Done.

> It noticed that https://github.com/lycheeverse/lychee/pull/2099 contains a few tests. Would be great to add them to this PR to show that something which was previously broken is now fixed.

Makes sense. Brought them over.

mre:
Cool. My vote goes to merging this and in turn closing https://github.com/lycheeverse/lychee/pull/2099. This will resolve a few issues and it's a relatively straightforward change. It shouldn't draw us into a corner when tackling the bigger, architectural issues.

lychee-bin/src/main.rs
```diff
@@ -328,6 +328,24 @@ async fn run(opts: &LycheeOptions) -> Result<i32> {
...
+                .cookie_jar
+                .as_ref()
+                .map_or_else(|| "<none>".to_string(), |p| p.display().to_string())
+        )
+    })?;
+
+    // Build the lychee client first so we can share its HostPool with the
+    // collector. This ensures that fetching remote input documents uses the
+    // same user-agent, TLS settings, cookies, per-host rate limits and custom
+    // headers as regular link checks.
```
thomas-zahner: IMO we could leave out this comment, as this is already well documented on `host_pool` of the `Collector` itself.

lychee-bin/tests/cli.rs
```diff
@@ -2776,6 +2776,64 @@ The config file should contain every possible key for documentation purposes."
...
+
+        let received_requests = server.received_requests().await.unwrap();
+        assert_eq!(received_requests.len(), 1);
+
+        let received_request = &received_requests[0];
+        let user_agent = received_request
+            .headers
+            .get("user-agent")
+            .and_then(|v| v.to_str().ok())
+            .unwrap_or("");
```
thomas-zahner: It's clearer to just `unwrap`. This would also make the error message clearer

```suggestion
        let user_agent = received_request
            .headers
            .get("user-agent")
            .expect("User agent missing")
            .to_str()
            .unwrap();
```

lychee-lib/src/collector.rs
```diff
@@ -145,10 +151,20 @@ impl Collector {
...
     }
 
-    /// Set client to use for checking input URLs
+    /// Set the [`HostPool`] to use when fetching remote input URLs.
+    ///
+    /// Pass the pool from a fully-configured [`crate::Client`] so that input
+    /// fetches share the same user-agent, TLS settings, cookies, per-host
+    /// rate limits and headers as regular link checks:
+    ///
+    /// ```ignore
```
thomas-zahner: Nitpick; would prefer the example to be either fully functional and tested or alternatively removed.

lychee-lib/src/collector.rs
```diff
@@ -599,6 +615,48 @@ mod tests {
...
+        let links = Collector::new(None, BaseInfo::none())
+            .unwrap()
+            .host_pool(client.host_pool())
+            .collect_links_from_file_types(inputs, crate::FileExtensions::default())
+            .map(|r| r.unwrap().uri)
+            .collect::<std::collections::HashSet<_>>()
+            .await;
+
+        assert!(links.iter().any(|u| u.url.as_str().contains("example.com")));
+        // wiremock will panic here if the expected request was not received
```
thomas-zahner: This comment should not be necessary. We already make use of this concept a dozen times across the tests and we never documented it. The comment feels out of place at the end, even though it does happen on drop.

lychee-bin/tests/cli.rs
```diff
@@ -2776,6 +2776,64 @@ The config file should contain every possible key for documentation purposes."
...
+            .await;
+
+        cargo_bin_cmd!()
+            .arg("--user-agent")
+            .arg("test-agent/1.0")
+            .arg(server.uri())
+            .assert()
+            .success();
+
+        // wiremock will fail if the expected request was not received
```
thomas-zahner: This is not necessary

thomas-zahner:
@mre I've created a54f81d5 to address my comments, you can of course amend/force push if you wouldn't agree with something.

mre:
Yes! Thanks for contributing @thomas-zahner. All great changes.

PR #2101, opened 2026-03-25
Respect HTTP error codes in CLI inputs

### LINKED ISSUE #1883: HTTP error codes in command-line inputs are ignored
If you type a URL on the command line which is a valid domain name but it returns a HTTP error code, then Lychee will ignore the error code and use the returned body anyway. In most cases, this returned body will be a meaningless error page. 

for example:
```console
$ cargo run --  http://example.com/this-link-does-not-exist   
  1/1 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links 
✅ 0 OK 🚫 0 Errors 🔀 1 Redirects

$ curl -I http://example.com/this-link-does-not-exist
HTTP/1.1 404 Not Found
Accept-Ranges: bytes
Content-Type: text/html
```

this can be very surprising if a web server switches its behaviour based on things like user agent.

for example, wikipedia returns a 403 when no user agent is set, and lychee doesn't set a user agent by default. this leads to things where the behaviour of lychee can _appear_ to diverge from other tools and there is no warning that this is happening.
```console
$ cargo run --  https://en.wikipedia.org/wiki/Tony_Hoare 
  0/0 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links
🔍 0 Total (in 0s) ✅ 0 OK 🚫 0 Errors

$ curl -I https://en.wikipedia.org/wiki/Tony_Hoare     
HTTP/2 200 
date: Fri, 24 Oct 2025 08:53:25 GMT
server: ATS/9.2.11
content-type: text/html; charset=UTF-8
content-length: 235766

$ curl https://en.wikipedia.org/wiki/Tony_Hoare | cargo run --  -  --offline
  169/169 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links
🔍 169 Total (in 0s) ✅ 0 OK 🚫 0 Errors 👻 169 Excluded
```
with some printing, we can see the input content that lychee is seeing that contains no links:
```
InputContent { source: RemoteUrl(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("en.wikipedia.org")), port: None, path: "/wiki/Tony_Hoare", query: None, fragment: None }), file_type: Html, content: "Please set a user-agent and respect our robot policy https://w.wiki/4wJS. See also T400119.\n" }
```
and finally, by setting an empty user agent to curl, we can see the same response:
```console
$ curl -i --user-agent '' https://en.wikipedia.org/wiki/Tony_Hoare
HTTP/2 403 
content-length: 92
content-type: text/plain
x-analytics: 
x-request-id: 26087a9d-3c9b-4ab3-9d7b-0f2d251eea50
server: HAProxy
x-cache: cp5020 int
x-cache-status: int-tls

Please set a user-agent and respect our robot policy https://w.wiki/4wJS. See also T400119.
```

what lychee can do (and maybe should do) is tell the user if one of their inputs receives a HTTP 4xx code. open questions:
- should it be a warning or an error?
- what if the user legitimately wants to check links on their 404 page? that should still be possible
- should it go through the status code `--accept` logic? what if you want different accept codes for link checking vs user input links?

mre:
Emit a ~warning~ hard error when a CLI input URL returns 4xx/5xx but keep processing other inputs. 
Fixes #1883.

### Before

```sh
$ lychee http://example.com/this-link-does-not-exist

1/1 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                    
[WARN] lychee detected 1 redirect. You might want to consider replacing redirecting URLs with the resolved URLs. Run lychee in verbose mode (-v/--verbose) to see details about the redirections.
🔍 1 Total (in 1s 10ms) ✅ 0 OK 🚫 0 Errors 🔀 1 Redirects
```

The error got silently ignored.

### After

```sh
$ lychee http://example.com/this-link-does-not-exist

0/0 ━━━━━━━━━━━━━━━━━━━━ Extracting links                                                             
[WARN] Input URL http://example.com/this-link-does-not-exist returned status code 404 Not Found
1/1 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                    
[WARN] lychee detected 1 redirect. You might want to consider replacing redirecting URLs with the resolved URLs. Run lychee in verbose mode (-v/--verbose) to see details about the redirections.
🔍 1 Total (in 783ms) ✅ 0 OK 🚫 0 Errors 🔀 1 Redirects
```

### Much After

Update! Converted to hard error (see discussion below):

```sh
$ lychee http://example.com/this-link-does-not-exist
Error: Cannot read input content from URL: status code 404 Not Found
```

Note that modifying `--accept` has no impact on the outcome since this is an input-level error.

### DIFF
--- a/lychee-lib/src/types/error.rs
+++ b/lychee-lib/src/types/error.rs
@@ -227,9 +227,14 @@ impl ErrorKind {
                 ),
                 _ => format!("File read error for '{}': {e}", path.display()),
             },
-            // TODO: This returns an empty string as a sentinel value because it's handled as a
-            // fatal application error rather than a link-level error. In the future, we should
-            // return an Option<String> or separate application errors from library errors.
+            // This `details()` method never gets called for incorrect CLI
+            // inputs, so whatever we put here, it won't be shown to the user.
+            //
+            // This returns an empty string as a sentinel value because it's handled as a
+            // fatal application error rather than a link-level error.
+            //
+            // TODO: In the future, we should return an Option<String> or separate
+            // application errors from library errors.
             ErrorKind::ReadInputUrlStatusCode(_) => String::new(),
             ErrorKind::ParseUrl(e, _url) => {
                 let detail = match e {


lychee-lib/src/types/resolver.rs
```diff
@@ -62,11 +62,16 @@ impl UrlContentResolver {
 }
 
 async fn get_request_body_text(client: &Client, request: Request) -> Result<String> {
-    client
+    let url = request.url().clone();
```
katrinafyi: This variable is only used in the error, right? Is there any way to avoid cloning it on every run?

lychee-lib/src/types/resolver.rs
```diff
@@ -62,11 +62,16 @@ impl UrlContentResolver {
...
         .execute(request)
         .await
-        .map_err(ErrorKind::NetworkRequest)?
-        .text()
-        .await
-        .map_err(ErrorKind::ReadResponseBody)
+        .map_err(ErrorKind::NetworkRequest)?;
+
+    let status = response.status();
+    if status.is_client_error() || status.is_server_error() {
```
thomas-zahner: I would be even more pedantic. Note that it still does not get triggered when using e.g. `http://endler.dev`, which redirects to `https`, as redirects get resolved as part of `client.execute(request)`. It would mean that we stopped following requests, which could happen when we reach the max-retries limit, which would be worth a warning, as the user is checking a not fully resolved/followed URL. Additionally it would include cases like Linkedin with `999`.

```suggestion
    if !status.is_success() {
```

lychee-bin/tests/cli.rs
```diff
@@ -3697,4 +3697,40 @@ https://lychee.cli.rs/guides/cli/#fragments-ignored
             .assert()
             .success();
     }
+    #[tokio::test]
+    async fn test_cli_input_url_status_warning() -> Result<()> {
+        let mock_server = wiremock::MockServer::start().await;
```
```patch
index 0b4c8b3a..a09baeea 100644
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -3697,8 +3697,9 @@ https://lychee.cli.rs/guides/cli/#fragments-ignored
             .assert()
             .success();
     }
+
     #[tokio::test]
-    async fn test_cli_input_url_status_warning() -> Result<()> {
+    async fn test_cli_input_url_status_warning() {
         let mock_server = wiremock::MockServer::start().await;
 
         wiremock::Mock::given(wiremock::matchers::method("GET"))
@@ -3730,7 +3731,5 @@ https://lychee.cli.rs/guides/cli/#fragments-ignored
         // We should see a warning for the error URL, but not for the success URL
         assert!(output.contains("returned status code 404"));
         assert!(!output.contains("returned status code 200"));
-
-        Ok(())
     }
 }
```

mre: Oh! Thanks. 😅

lychee-lib/src/types/resolver.rs
mre: Oh, right. Neat.

lychee-lib/src/types/resolver.rs
mre: Yesss. Done.

mre:
@katrinafyi 

> In a vacuum, I would lean towards making it an error with an option to bypass the error. I don't think it's too much disruption, and most people would probably appreciate the obvious error. But either way is much more helpful than silent with no message.

I was debating this myself! Initially, I tried to be conservative, but after thinking about it some more, let's be bold and the right thing. A hard error will definitely prevent users from accidentally checking an empty/error page and assuming everything is fine. Let me update this PR to return an error instead of a warning. What do you think about deferring the introduction the bypass-flag until someone explicitly asks for it? Too pragmatic?

lychee-lib/src/types/error.rs
```diff
@@ -202,6 +202,7 @@ impl ErrorKind {
     /// messages) and future changes are expected.
     #[must_use]
     #[allow(clippy::too_many_lines)]
+    #[allow(clippy::match_same_arms)]
```
mre: Added this because I think it's helpful if different error variants get their own match arm.

lychee-lib/src/types/error.rs
```diff
@@ -215,6 +220,7 @@ impl ErrorKind {
                     Some(e.to_string())
                 }
             }
+            ErrorKind::ReadInputUrlStatusCode(_) => None,
```
katrinafyi: Maybe add a message here suggesting that if the user does want to check the links inside an error page, they can download the error page and check it as a local file.

mre:
> I think I'm just very cautious when it comes to changing defaults and errors, after past experience 😅

Being cautious is generally the wise approach, yes, but in this case, my goal is to avoid increasing the number of CLI flags we have to maintain. There's a chance we will have to add another flag, but my hope is that people just treat it as an actual bugfix and accept it as the expected behavior when providing invalid inputs. We'll have to see. :^)

lychee-lib/src/types/error.rs
mre: Hm! Good idea. 👍

mre: So I added a note to the details() function, but it turns out we don't ever show that data for tool-level errors. ^^ I believe it is genuinely helpful for users to see that information, even in non-verbose mode. So I added a change to display those details.

```sh
> lychee  http://example.com/this-link-does-not-exist
 [ERROR] Cannot read input content from URL: status code 404 Not Found: To check links in error pages, download and check locally instead.
```

So the issue was:
- ✅ Link checking details already worked and were shown.
- ❌ Tool-level error details were missing (this is what I fixed).

@katrinafyi  @thomas-zahner do you both agree that this is a good change? I could create a separate PR, but it was small enough to add it here.

thomas-zahner: I'm definitely agreeing with both of you to add user-friendly details for a better UX. But I'm not sure with your change regarding error logging, because of the PR #2094, which I've merged just now. This PR, as a side-effect, changed to log details by default and it also made changes to the [`details()` function](https://github.com/lycheeverse/lychee/blob/4bff560d2b194294d7986f538a546a8220ea7988/lychee-lib/src/types/error.rs#L202). So it might also result in a merge conflict.

I think what you can do now the below change and probably remove any other changes to how logging is done.

```rust
            ErrorKind::ReadInputUrlStatusCode(_) => {
                format!("{self}. Some user friendly details here...")
            }
```

lychee-bin/src/main.rs
```diff
@@ -281,7 +282,17 @@ fn run_main() -> Result<i32> {
...
             exit(ExitCode::Success as i32);
         }
-        res => res,
+        Err(e) => {
+            // Check if this is a lychee error with details
+            if let Some(lychee_error) = e.downcast_ref::<LycheeErrorKind>()
+                && let Some(details) = lychee_error.details()
+            {
+                error!("{e}: {details}");
+                return Ok(ExitCode::UnexpectedFailure as i32);
```
thomas-zahner: I don't fully understand if this change was purely for a difference in logging or if the last line also affects lychee's return code. If it's only for logging it should be undone as part of resolving the merge conflicts, as mentioned in my other comment.

lychee-lib/src/types/error.rs
mre: Aha! That's cool, thanks. Will do.

lychee-bin/src/main.rs
mre: It was purely about logging. Thanks, I forgot about your recent changes. Cleaned up that part. :)

mre:
The actual change is pretty simple, but we went through a few iterations to find the right approach. I'm still not 100% satisfied, since the new match arm returns a [sentinel value](https://github.com/lycheeverse/lychee/blob/8816c4badbfc39f80a162b4ae79ee49ee3da0ddb/lychee-lib/src/types/error.rs#L230-L238). I've added a TODO to refactor that:

```rust
// This `details()` method never gets called for incorrect CLI
// inputs, so whatever we put here, it won't be shown to the user.
//
// This returns an empty string as a sentinel value because it's handled as a
// fatal application error rather than a link-level error.
//
// TODO: In the future, we should return an Option<String> or separate
// application errors from library errors.
ErrorKind::ReadInputUrlStatusCode(_) => String::new(),
```

lychee-bin/tests/cli.rs
```diff
@@ -3942,10 +3942,15 @@ exclude = ["cargo_exclude_test_str"]
...
+        std::fs::write(
+            dir.path().join("input.txt"),
+            "https://example.com/cargo_exclude_test_str",
+        )?;
+
         let mut cmd = cargo_bin_cmd!();
         let assert = cmd
             .current_dir(dir.path())
-            .arg("https://example.com/cargo_exclude_test_str")
+            .arg("input.txt")
```
mre, line 3953: Had to convert this to an input file since the original URL was invalid and that now causes an input error.

PR #2104, opened 2026-03-25
Support multiple config file formats and refactor config module

### LINKED ISSUE #1930: Support for `pyproject.toml` (and `cargo.toml`)
In addition to have lychee search for configuration in the `.lycheeignore` and `lychee.toml` file, I would love to have the possibility of adding lychee-specific configuration into a `pyproject.toml` file.

That way I can have all the configuration of the tools I use for my Python project nicely arranged in a single `pyproject.toml` file. For that to work, lychee-specific configuration needs to be put under its own section, like `[tool.lychee]`.

I know lychee is not a Python-specific project, and is not written in Python. But that does not prevent it to be nice to the Python ecosystem.

A good example is the [`typos`](https://github.com/crate-ci/typos) utility, which is written in Rust, but is supporting reading its configuration both from a `pyproject.toml` and a `cargo.toml` file:
- https://github.com/crate-ci/typos/issues/361
- https://github.com/crate-ci/typos/issues/361
- https://github.com/crate-ci/typos/issues/1060

mre:
This PR resolves #1930 by adding support for loading configuration from pyproject.toml and Cargo.toml. I've also decided to add package.json to the mix, as this is a very common way to configure tools in the JavaScript ecosystem.

While implementing this, I realized the config module had grown out of proportion, which caused me to split it up into logical submodules to keep the core high-level structs clean. (Sorry if that PR got a bit large due to that. I'm mostly just moving stuff around and adding the trait + impls.)

To handle the different config formats, I introduced a ConfigLoader trait. I went through a few iterations with this. Initially, I thought about a simple, imperative approach where we just attempt to strictly deserialize the entire file and return an option. However, because our config uses strict validation (denying unknown fields), a single typo by the user would cause the parsing to fail, and we would silently skip the file instead of telling them about the typo.

To fix this, I split the trait into is_match and load. The is_match method does a lightweight check just to see if the specific lychee section exists in the file. If it does, we pass it to load, which enforces strict schema deserialization and properly bubbles up validation errors so the user knows what went wrong.

Implementation-wise, I deliberately avoided heavyweight dependencies like the cargo_toml crate, which builds massive ASTs for the entire manifest. Instead, we use minimal, custom envelope structs with serde to extract only the lychee blocks. This keeps our binary size small and compilation fast. I also made sure that Cargo package metadata takes precedence over workspace metadata if both are present.

Finally, I added some tests for each loader to ensure invariants (valid blocks are mapped correctly, missing keys are ignored, precedence works as intended, and malformed schemas produce the expected errors).

### DIFF
--- a/lychee-bin/src/config/loaders/mod.rs
+++ b/lychee-bin/src/config/loaders/mod.rs
@@ -33,16 +33,13 @@ pub(crate) trait ConfigLoader {
 
 const LOADERS: [&dyn ConfigLoader; 4] = [
     &lychee_toml::LycheeTomlLoader,
-    &pyproject_toml::PyprojectTomlLoader,
     &cargo_toml::CargoTomlLoader,
+    &pyproject_toml::PyprojectTomlLoader,
     &package_json::PackageJsonLoader,
 ];
 
 /// Find the first matching default configuration file in the current directory
 /// and return the parsed configuration it contains.
-///
-/// This checks for files like `lychee.toml`, `pyproject.toml`, `Cargo.toml`,
-/// and `package.json` in a defined order of precedence.
 pub(crate) fn default_config_file() -> Result<Option<Config>> {
     for loader in LOADERS {
         let path = PathBuf::from(loader.filename());


lychee-bin/src/config/loaders/cargo_toml.rs
```diff
@@ -0,0 +1,166 @@
...
+            .is_some_and(|m| m.contains_key("lychee"));
+
+        let has_workspace = table
+            .get("workspace")
+            .and_then(|t| t.as_table())
+            .and_then(|t| t.get("metadata"))
+            .and_then(|m| m.as_table())
+            .is_some_and(|m| m.contains_key("lychee"));
+
+        has_package || has_workspace
```
thomas-zahner:
```suggestion
        table
            .get("package")
            .or(table.get("workspace"))
            .and_then(|t| t.as_table())
            .and_then(|t| t.get("metadata"))
            .and_then(|m| m.as_table())
            .is_some_and(|m| m.contains_key("lychee"))
```

lychee-bin/src/config/loaders/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// not recommended to call `load`, as it will likely fail with a parsing error.
+    fn load(&self, contents: &str) -> Result<Config>;
+}
+
+/// Find the first matching default configuration file in the current directory.
+///
+/// This checks for files like `lychee.toml`, `pyproject.toml`, `Cargo.toml`,
+/// and `package.json` in a defined order of precedence.
+pub(crate) fn find_default_config_file() -> Option<PathBuf> {
+    let loaders: [&dyn ConfigLoader; 4] = [
```
thomas-zahner: Extract it, to use the same loaders in `load_from_file` and make it const.

```suggestion
    const LOADERS: [&dyn ConfigLoader; 4] = [
```

lychee-bin/src/config/loaders/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    for loader in loaders {
+        let path = PathBuf::from(loader.filename());
+        if path.is_file() {
+            let contents = fs::read_to_string(&path).unwrap_or_default();
+            if loader.is_match(&contents) {
+                return Some(path);
+            }
+        }
+    }
+    None
```
thomas-zahner: Mabye? But I'm not sure if it's more readable..


```suggestion
    LOADERS
        .into_iter()
        .map(|loader| (loader, PathBuf::from(loader.filename())))
        .filter(|(_, path)| path.is_file())
        .find_map(|(loader, path)| {
            let contents = fs::read_to_string(&path).unwrap_or_default();
            loader.is_match(&contents).then_some(path)
        })
```

lychee-bin/src/config/loaders/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                        log::warn!("Failed to load config from {}: {e}", path.display());
+                        e
+                    });
+                }
+                return Ok(Config::default());
+            }
+        }
+    }
+
+    lychee_toml::LycheeTomlLoader.load(&contents).map_err(|e| {
```
thomas-zahner: Does this make sense? I think we might want to return an error here.

lychee-bin/src/config/loaders/package_json.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn filename(&self) -> &str {
+        PACKAGE_JSON_CONFIG_FILE
+    }
+
+    /// We use a generic JSON value to check for the presence of the
+    /// `lychee` configuration section. We don't want to strictly deserialize
+    /// into `Config` here, because if the user has a typo in their config
+    /// (e.g. `timeoutt = 10`), a strict deserialization would fail, and we
+    /// would incorrectly return `false`, causing lychee to silently ignore
+    /// the file instead of reporting the error.
```
thomas-zahner: I feel like this doc comment is not necessary. It's copy pasted in the other implementations. Let's remove it and only keep the comment in `ConfigLoader`. If some information is still missing in the doc comment of `ConfigLoader` which is mentioned here we can add it there.

lychee-bin/src/config/loaders/package_json.rs
```diff
@@ -0,0 +1,119 @@
...
+
+        value
+            .as_object()
+            .is_some_and(|obj| obj.contains_key("lychee"))
+    }
+
+    /// We strictly deserialize into our custom `PackageJson` envelope,
+    /// which contains our `Config` struct with `#[serde(deny_unknown_fields)]`.
+    /// Since we already know the section exists from `is_match`, any failure
+    /// here is a genuine configuration error that we want to bubble up.
```
thomas-zahner: The same is true for this doc comment. Let's keep the documentation in the trait instead of copy/pasting it into every implementation.

lychee-bin/src/config/loaders/cargo_toml.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        // Package metadata takes precedence over workspace metadata
+        // That's useful, because it allows users to define a
+        // workspace-wide default config and then override it in
+        // specific packages.
+        let config = [cargo.package, cargo.workspace]
+            .into_iter()
+            .flatten()
+            .find_map(|s| s.metadata.and_then(|m| m.lychee))
+            .unwrap_or_default();
```
thomas-zahner: What's the reason for this outermost `Option<_>` and `unwrap_or_default`?

lychee-bin/src/config/loaders/package_json.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn load(&self, contents: &str) -> Result<Config> {
+        let package_json = serde_json::from_str::<PackageJson>(contents)
+            .with_context(|| "Failed to parse lychee config from package.json")?;
+
+        Ok(package_json.lychee.unwrap_or_default())
```
thomas-zahner: Same here; what's the reason for the outermost `Option<_>` and the `unwrap_or_default`?

lychee-bin/src/config/loaders/pyproject_toml.rs
```diff
@@ -0,0 +1,107 @@
...
+pub(crate) const PYPROJECT_CONFIG_FILE: &str = "pyproject.toml";
+
+#[derive(Deserialize)]
+struct PyprojectToml {
+    tool: Option<Tool>,
+}
+
+#[derive(Deserialize)]
+struct Tool {
+    lychee: Option<Config>,
```
thomas-zahner, line 35: I think that this could be unwrapped, as with all other loaders:


```suggestion
    lychee: Config,
```

lychee-bin/src/main.rs
```diff
@@ -134,6 +133,16 @@ fn read_lines(file: &File) -> Result<Vec<String>> {
         .collect())
 }
 
+/// Find the default configuration file from the current directory.
```
thomas-zahner: This addition is very confusing and mostly wrong. It only uses the default configuration file if `opts.config_files.is_empty`, otherwise all specified config files are used. Also I find `A file is considered matching if it` too detailed and low-level, this is already explained in the new trait, and doesn't directly concern this function. IMO we could just remove this addition. Maybe optionally slightly update/improve the existing three lines.

lychee-bin/tests/cli.rs
```diff
@@ -3697,4 +3697,206 @@ https://lychee.cli.rs/guides/cli/#fragments-ignored
...
+            r#"
+[tool.lychee]
+exclude = ["pyproject_exclude_test_str"]
+"#,
+        )?;
+
+        let mut cmd = cargo_bin_cmd!();
+        let assert = cmd
+            .current_dir(dir.path())
+            .arg("https://example.com/pyproject_exclude_test_str")
```
thomas-zahner: Huh, this is confusing me.. Shouldn't we pipe a URL or write it to a file? For me the test fails locally because of the certificate issues with example.com (https://github.com/lycheeverse/lychee/issues/2061)

I would do this:

```rust
    #[tokio::test]
    async fn test_pyproject_toml() -> Result<()> {
        let dir = tempfile::tempdir()?;
        let pyproject = dir.path().join("pyproject.toml");
        std::fs::write(
            &pyproject,
            r#"
[tool.lychee]
exclude = ["exclude_test_str"]
"#,
        )?;

        let stdout = cargo_bin_cmd!()
            .current_dir(dir.path())
            .arg("-")
            .write_stdin("https://exclude/exclude_test_str")
            .arg("--format=json")
            .assert()
            .success()
            .get_output()
            .stdout
            .clone();

        assert_eq!(
            stdout_to_json(&stdout)?["excludes"],
            1,
            "Config must mark the URL as excluded"
        );
        Ok(())
    }
```

lychee-bin/tests/cli.rs
```diff
@@ -3697,4 +3697,206 @@ https://lychee.cli.rs/guides/cli/#fragments-ignored
...
+            .assert();
+        let output1 = String::from_utf8_lossy(&assert1.get_output().stdout);
+        assert!(!output1.contains("exclude_lychee.txt"));
+        assert!(output1.contains("exclude_pyproject.txt"));
+        assert!(output1.contains("exclude_cargo.txt"));
+
+        // Remove lychee.toml
+        std::fs::remove_file(dir.path().join("lychee.toml"))?;
+
+        // Test 2: pyproject.toml takes precedence over Cargo.toml
```
thomas-zahner: Is this important? IMO we don't need to test this. I only think it's important for lychee.toml to take precedence over any other formats. Everything else is no really important.

lychee-bin/src/config/loaders/mod.rs
mre: The jury is still out, but if you ask me, I find the simple for-loop more readable. But I change my mind on that every 6 months, sooo `¯\_(ツ)_/¯`

lychee-bin/src/config/loaders/cargo_toml.rs
mre: clever!

lychee-bin/src/config/loaders/cargo_toml.rs
mre: Because a single `Cargo.toml` file can contain *both* `[package.metadata]` and `[workspace.metadata]` blocks, we have to use `Option<Config>` in our struct definition:

```rust
#[derive(Deserialize)]
struct Metadata {
    lychee: Option<Config>,
}
```

If we made it strictly `lychee: Config` instead of an `Option`, the `serde` deserializer would throw an error if it encountered a `[package.metadata]` block that didn't have a `lychee` section, even if a perfectly valid `[workspace.metadata.lychee]` existed further down the file. So, it's a gnarly edge-case. Thought about adding a comment, but it might just be more confusing than the logic itself. wdyt?

lychee-bin/src/config/loaders/package_json.rs
mre: true. cleaned up.

lychee-bin/src/config/loaders/package_json.rs
mre: Yeah, this one can be cleaned up. Copy-paste error. 👍

lychee-bin/src/config/loaders/pyproject_toml.rs
mre: Nice.

lychee-bin/src/main.rs
mre: Ok, removed.

lychee-bin/tests/cli.rs
mre: Thanks. I copied your suggestion.

lychee-bin/tests/cli.rs
mre: Hm, you might be right. I cleaned it up.

lychee-bin/src/config/loaders/mod.rs
mre: done

lychee-bin/src/config/loaders/mod.rs
mre: @thomas-zahner can you check again?

I've slightly changed the logic by returning an error when a user specified a config file with an invalid lychee config. The tests still pass, so it should be fine?

mre:
Addressed all comments. (Thanks, Thomas!)

Also added a test which verifies that if the user points directly to a configuration file that doesn't contain a valid lychee section (like a basic `package.json`), the tool will output an error message `"No valid lychee configuration found in"` instead of silently accepting it and using default configuration. =)

Any thoughts?

lychee-bin/src/config/loaders/mod.rs
thomas-zahner: Haha yeah, I also feel like the imperative approach is slightly more readable

lychee-bin/src/config/loaders/mod.rs
thomas-zahner: Hmm, it doesn't look bad. But I feel like the error logging/handling is a bit messy. We use `bail`, `anyhow::Error::from`, `log::warn`.

It would be really nice if we could use `.context` everywhere and then do the logging outside at the call site `merge_file`.

Inside this function it would be something like this (didn't test this yet!)

```patch
diff --git a/lychee-bin/src/config/loaders/mod.rs b/lychee-bin/src/config/loaders/mod.rs
index 189ae8c7..91f32748 100644
--- a/lychee-bin/src/config/loaders/mod.rs
+++ b/lychee-bin/src/config/loaders/mod.rs
@@ -4,7 +4,7 @@ pub(crate) mod package_json;
 pub(crate) mod pyproject_toml;
 
 use super::Config;
-use anyhow::{Result, bail};
+use anyhow::{Context, Result, bail};
 use std::fs;
 use std::path::{Path, PathBuf};
 
@@ -64,27 +64,24 @@ pub(crate) fn find_default_config_file() -> Option<PathBuf> {
 /// using the corresponding loader. Otherwise, it falls back to the default
 /// TOML loader.
 pub(crate) fn load_from_file(path: &Path) -> Result<Config> {
-    let contents = fs::read_to_string(path).map_err(|e| {
-        log::warn!("Failed to read config file {}: {e}", path.display());
-        anyhow::Error::from(e)
-    })?;
+    let contents = fs::read_to_string(path)
+        .context(format!("Failed to read config file {}", path.display()))?;
 
     if let Some(filename) = path.file_name().and_then(|n| n.to_str()) {
         for loader in LOADERS {
             if loader.filename() == filename {
                 if loader.is_match(&contents) {
-                    return loader.load(&contents).map_err(|e| {
-                        log::warn!("Failed to load config from {}: {e}", path.display());
-                        e
-                    });
+                    return loader
+                        .load(&contents)
+                        .context(format!("Failed to load config from {}", path.display()));
                 }
+
                 bail!("No valid lychee configuration found in {}", path.display());
             }
         }
     }
 
-    lychee_toml::LycheeTomlLoader.load(&contents).map_err(|e| {
-        log::warn!("Failed to load config from {}: {e}", path.display());
-        e
-    })
+    lychee_toml::LycheeTomlLoader
+        .load(&contents)
+        .context(format!("Failed to load config from {}", path.display()))
 }
```

lychee-bin/src/config/loaders/cargo_toml.rs
thomas-zahner: Ah I see. So yes, let's keep the `Option` and I agree we don't need to add a comment. But I'm still not entirely sure about the `unwrap_or_default`. Doesn't this now behave differently from the other loaders, where if we specify the config explicitly, but the lychee config is empty we get an error? In this case we now get the default config, which might be unexpected. Maybe we should map it to an error instead of defaulting.

thomas-zahner:
> Also added a test which verifies that if the user points directly to a configuration file that doesn't contain a valid lychee section (like a basic package.json), the tool will output an error message "No valid lychee configuration found in" instead of silently accepting it and using default configuration. =)

That's great, thank you! Now I just think that pyproject still falls back to the default config and behaves differently, we could update that if possible and add that to the new test, which currently only covers package.json.

lychee-bin/src/config/loaders/cargo_toml.rs
```diff
@@ -0,0 +1,148 @@
...
+            .with_context(|| "Failed to parse lychee config from Cargo.toml")?;
+
+        // Package metadata takes precedence over workspace metadata
+        // That's useful, because it allows users to define a
+        // workspace-wide default config and then override it in
+        // specific packages.
+        let config = [cargo.package, cargo.workspace]
+            .into_iter()
+            .flatten()
+            .find_map(|s| s.metadata.and_then(|m| m.lychee))
```
katrinafyi: From the code here, I assume that "takes precedence" means that one is used and the other is ignored? Would it be better to merge them instead? I understand that you don't want to define a merge order between, e.g., pyproject.toml and package.json, but it makes sense between cargo package and workspace, I think.

lychee-bin/src/config/loaders/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// # Why separate `is_match` from `load`?
+    /// If we strictly deserialized the entire file directly into our `Config` struct
+    /// (which uses `#[serde(deny_unknown_fields)]`), we wouldn't be able to distinguish
+    /// between two scenarios:
+    /// 1. A file that simply doesn't contain any lychee configuration (which we should silently ignore).
+    /// 2. A file that *does* contain a lychee configuration section, but has a typo in one of the fields (which we should report as an error).
+    ///
+    /// `is_match` performs a loose structural check to safely handle the first scenario,
+    /// allowing `load` to be strict and bubble up actual configuration errors to the user.
+    fn is_match(&self, contents: &str) -> bool;
```
katrinafyi: I'm not sure I'm convinced by this rationale, and it leads to a lot of validation checks in is_match() followed by unwrap in load().

Is it possible to return a `Result<Option<Config>>` instead? Ok(None) means that lychee config is missing in an otherwise valid file, and the Err cases include would both invalid lychee config and invalid file syntax.

lychee-bin/src/config/loaders/cargo_toml.rs
mre: Hmm, how would we document this?

I'm a little worried about merge order behaving differently for cargo-related config than for other ecosystems. Keep in mind that lychee is independent of all of these ecosystems, including Rust. That said, one could argue that lychee config maps naturally to workspace-wide and package-wide configuration. Not sure if other languages have something similar.
My instinct is to keep it simple until someone explicitly asks for it, but I'm open to adding support.

katrinafyi: Hmm yeah, it could get complicated. Happy to leave this for later.

Maybe just touch up the comment to clarify that overriding is actually fully replacing.

lychee-bin/src/config/loaders/mod.rs
mre: Good point, I've merged the two methods back together.

One small addition: I introduced a `ConfigMatch` enum as a domain-specific alternative to `Option`. Semantically it's identical, but the named variants make the "no match" case explicit at the call site without needing to consult the docs to understand what None means in this context.

Here's the diff: https://github.com/lycheeverse/lychee/pull/2104/changes/c397b480ac62f1b1c956ba21a9e8af3f15ecac3f

lychee-bin/src/config/loaders/cargo_toml.rs
thomas-zahner: With c397b480ac62f1b1c956ba21a9e8af3f15ecac3f I my concerns are addressed via:

```rust
        match pyproject.tool.and_then(|t| t.lychee) {
            Some(config) => Ok(ConfigMatch::Found(Box::new(config))),
            None => Ok(ConfigMatch::NotFound),
        }
```

lychee-bin/src/config/loaders/mod.rs
thomas-zahner: Cool, think this is resolved :+1:

lychee-bin/src/config/loaders/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    match lychee_toml::LycheeTomlLoader
+        .load(&contents)
+        .with_context(|| format!("Failed to load config from {}", path.display()))?
+    {
+        ConfigMatch::Found(config) => Ok(*config),
+        ConfigMatch::NotFound => bail!("No valid lychee configuration found in {}", path.display()),
+    }
```
katrinafyi: it's splitting hairs, but i had a go at rewriting this to avoid the repeated error messages. i also saw your earlier thread about .iter() vs for loops, so it's fine to ignore this if you want.

```suggestion
    let filename = path.file_name().and_then(|n| n.to_str());
    let loader = if let Some(filename) = filename {
        LOADERS
            .iter()
            .find(|loader| filename == loader.filename())
            .copied()
            .unwrap_or(&lychee_toml::LycheeTomlLoader)
    } else {
        &lychee_toml::LycheeTomlLoader
    };

    match loader
        .load(&contents)
        .with_context(|| format!("Failed to load config from {}", path.display()))?
    {
        ConfigMatch::Found(config) => Ok(*config),
        ConfigMatch::NotFound => bail!("No valid lychee configuration found in {}", path.display()),
    }
```

lychee-bin/src/config/loaders/mod.rs
katrinafyi: thanks for the changes! i think the enum is good too.

lychee-bin/src/config/loaders/cargo_toml.rs
mre: Okay, I've updated the comment to reflect what the code does. 👍

lychee-bin/src/config/loaders/mod.rs
mre: Ah okay, that's a little shorter and more readable, so I've copied that over and committed it locally. Thanks for the refactor. Good use of `find()`. @thomas-zahner FYI. 😉

thomas-zahner: Oh that looks much nicer than the last time I glanced at it :)

mre:
This was fun! The code turned out much better than my initial attempt thanks to both of your feedback. Much appreciated. I will go ahead and merge this. :)

PR #2106, opened 2026-03-26
fix(cli): bump open files limit on macOS and Linux

### LINKED ISSUE #1248: Too many open files (os error 24) on macOS
When running `lychee README.md` against this repository https://github.com/dariubs/GoBooks, I am seeing the following error on macOS 13.5.1, however these links are considered fine by GitHub Actions with Ubuntu.

```
✗ [ERR] https://github.com/thewhitetulip/web-dev-golang-anti-textbook | Failed: Network error: dns error: proto error: io error: Too many open files (os error 24)
```

It's happening on quite a few links.

Running against commit 1c71e7329e369b510afb806f484aa4e415b9aeb2 of https://github.com/dariubs/GoBooks with lychee installed via Homebrew.

> ❯ lychee --version
> lychee 0.13.0

mre:
This automatically bumps the maximum number of open files (NOFILE) to the hard limit.
It prevents 'Too many open files (os error 24)' which happens easily on macOS due to its low default soft limit (256).

It addresses the issue originally mentioned in the comments, where users experience this during initial link extraction or link checking across multiple files.

Fixes #1248

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -134,34 +134,39 @@ fn read_lines(file: &File) -> Result<Vec<String>> {
         .collect())
 }
 
-/// Increase the maximum number of open files (if possible)
+/// Handle the system's `RLIMIT_NOFILE` limit to prevent
+/// opening too many file descriptors and encountering runtime errors.
+/// (see <https://github.com/lycheeverse/lychee/issues/1248>)
 ///
-/// This is helpful to respect system limits and because lychee opens many files
-/// concurrently during link extraction and checking. If the hard limit is low,
-/// we might need to reduce the `max_concurrency` to prevent lychee from
-/// crashing with a "Too many open files" error
-/// (<https://github.com/lycheeverse/lychee/issues/1248>).
-///
-/// The relation is roughly:
-/// total FDs ≈ `max_concurrency` + `baseline_overhead`
-fn set_fd_limits(opts: &mut LycheeOptions) {
-    /// Baseline overhead accounts for stdin, stdout, stderr, lychee.toml,
-    /// .lycheeignore, .lycheecache, and reqwest + DNS handles.
-    const BASELINE_OVERHEAD: u64 = 15;
+/// 1. We try to increase the soft limit.
+/// 2. If the soft limit is low we reduce lychee's `max_concurrency` accordingly
+fn handle_fd_limits(opts: &mut LycheeOptions) {
+    use rlimit::{Resource, getrlimit, increase_nofile_limit};
+    /// Baseline overhead estimate to account for file descriptors
+    /// which are opened before lychee actually does any link checking.
+    /// Estimate is based on observing `lsof -p $(pgrep lychee)`.
+    const BASELINE_OVERHEAD: u64 = 20;
 
-    if let Ok(limit) = rlimit::increase_nofile_limit(u64::MAX) {
-        let available_fds = limit.saturating_sub(BASELINE_OVERHEAD) / 2;
+    if let Ok(soft_limit) = increase_nofile_limit(u64::MAX)
+        .or_else(|_| getrlimit(Resource::NOFILE).map(|(soft, _)| soft))
+    {
+        // The relation between `concurrency` and `soft_limit` is roughly:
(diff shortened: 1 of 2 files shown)


lychee-bin/src/main.rs
```diff
@@ -115,6 +115,13 @@ enum ExitCode {
 const LYCHEEIGNORE_COMMENT_MARKER: &str = "#";
 
 fn main() -> Result<()> {
+    // Increase the maximum number of open files on macOS and Linux.
+    // This is helpful because lychee opens many files concurrently during
+    // link extraction and checking, which can exceed the default limit
+    // of 256 open files on macOS. See https://github.com/lycheeverse/lychee/issues/1248
+    #[cfg(any(target_os = "macos", target_os = "linux"))]
```
thomas-zahner: Peeking inside the function I see the guard `#[cfg(unix)] { ... } #[cfg(windows)] { Ok(lim) }`. So I would remove our config guard completely. It's less code and will handle all cases for us, including all Unixes, not just macOS and Linux.

katrinafyi:
I was reading [the manual](https://www.man7.org/linux/man-pages/man2/getrlimit.2.html) and this function increases the soft limit up to the hard limit of no. files, which is usually big enough. 

But is it possible that the hard limit is also a low number that would be too small for lychee? If that happens, we should probably lower max-concurrency to keep lychee functional, maybe with a warning. Do we know the relationship between max-concurrency and the number of used files?

lychee-bin/src/main.rs
mre: aha! that's great.

mre: removed.

mre:
> But is it possible that the hard limit is also a low number that would be too small for lychee? If that happens, we should probably lower max-concurrency to keep lychee functional, maybe with a warning. Do we know the relationship between max-concurrency and the number of used files?

Yes, it is possible to get a low number and be none the wiser. :) And also yes, we should clamp the max concurrency accordingly. But the tricky part is: clamp to what? I was roughly calculating `Number of file descriptors ≈ max_concurrency + baseline_overhead` where baseline_overhead is in the low two-digit range (stdin, stdout, stderr, lychee.toml, .lycheeignore, .lycheecache, reqwest + DNS handles), so maybe like 15?

Added a change for that.

mre:
Also added a regression test.

lychee-bin/src/main.rs
```diff
@@ -134,6 +134,38 @@ fn read_lines(file: &File) -> Result<Vec<String>> {
...
+/// (<https://github.com/lycheeverse/lychee/issues/1248>).
+///
+/// The relation is roughly:
+/// total FDs ≈ `max_concurrency` + `baseline_overhead`
+fn set_fd_limits(opts: &mut LycheeOptions) {
+    /// Baseline overhead accounts for stdin, stdout, stderr, lychee.toml,
+    /// .lycheeignore, .lycheecache, and reqwest + DNS handles.
+    const BASELINE_OVERHEAD: u64 = 15;
+
+    if let Ok(limit) = rlimit::increase_nofile_limit(u64::MAX) {
```
thomas-zahner: I might make sense to handle the case when setting the limit fails, but getting it would work.

thomas-zahner:
I've double checked the open file descriptors on my machine and it was 18 (in idle mode with `lychee -`), which is bigger than the advertised 15, so I've bumped it to 20. We could maybe be even more conservative.

I also tried to improve the docs and I now try to read the limit if setting it fails. See 12d4174e87a29c5dbd5bf01b68dace4f5fed2ec1.

Previously, you divided by 2. `let available_fds = limit.saturating_sub(BASELINE_OVERHEAD) / 2;` This division was not explained in your formula in the comment. Did you do it mainly to get rid of the clippy warning or to be more conservative? I've now removed it and am now expecting the clippy warning. If you did it for being more conservative I'd rather further increase the `BASELINE_OVERHEAD`.

mre:
Thanks for the review and for the improvements @thomas-zahner. :)

ngoldbaum:
Awesome, thanks for fixing this tricky issue!

mre:
Well, let's see if we fixed it. But thanks. :)

PR #2107, opened 2026-03-27
Cache for deduplicating top-level in-progress requests

katrinafyi:
This improves the top-level cache by teaching it to understand "in-progress" requests and abstracts it into a generic Cache struct. See the docs for the new types: https://in-flight-url-cache.lychee-docs-katrinafyi.pages.dev/lychee_lib/cache/

This means that when duplicate URLs are collected from sources (e.g., due to different positions in the file), there will be only one actual request to that URL. Additionally, this means that repeated requests can be sent to a side-queue to wait for the initial request to finish, freeing up the main worker tasks for real link checks. (This is an improvement compared to the Host cache, which I commented on here: https://github.com/lycheeverse/lychee/pull/2067#pullrequestreview-3875862783)

This is old work which I had started but never finished. I'm just posting it now to remind me to work on it and to gather opinions :) Comments are appreciated.

todo:
- [ ] fix prints and stuff
- [ ] clean up wording.
- [ ] think about storing a richer status in the in-memory cache, to avoid `OK (Cached)` messages without `--cache`.
- [ ] think about the role of the hostpool cache. i think it can be simplified a lot, because after this change the only duplicates it has to handle are requests to different fragments on the same URL.
- [ ] clean up handler tasks into separate functions
- [ ] change into_completed_entries to not panic

maybe later:
- [ ] think about applying the cache to more Arc<Mutex<Hashmap>> places, e.g. fragment cache
- [ ] think about caching based on checked http url, after remaps, rather than before remaps.

related work:
- maybe this instead of SetOnce: https://docs.rs/async-lazy/latest/async_lazy/struct.Lazy.html
- https://docs.rs/deduplicate/latest/deduplicate/
- or maybe the more feature-full moka, with eviction and bounds: https://docs.rs/moka/latest/moka/sync/struct.Cache.html


mre:
Haven't looked too closely at the code, but I believe this could be a missing puzzle piece for recursion support. For recursive crawling (where every page links to the same nav/footer), this would prevent concurrency starvation. The `lock_entry` / `CacheSetter` pattern could serve as a natural interception point for the recursion handler.

katrinafyi:
Closing in favour of:
- https://github.com/lycheeverse/lychee/pull/2188

which is a more up to date and smaller change.

I also agree with idea of the interception point for recursion. Although, we'll have to see how it shakes out if we want to do away with the top level cache.

mre:
#2188 got merged now. 🥳

Issue #1818, opened 2025-08-24
Cache fails to invalidate when changing `--remap` options


katrinafyi:
Using Lychee in the same directory with changing --remap options will fail to invalidate the cache when --remap is changed. I've run into this bug when testing remaps to try and write a correct remap expression.

I think a contributing factor to this bug is the .lycheecache entries are keyed by the original URL, before any remaps are applied.

These commands should demonstrate the problem.
```bash
echo 'https://google.com' > a.md
# run once. this succeeds
lychee a.md --cache -v
# introduce remap which should cause failure. still succeeds.
lychee a.md --cache -v --remap "https://google.com https://google.com/nonexistent"
cat .lycheecache

# try in other direction (with remap initially, then removing it)
rm .lycheecache
lychee a.md --cache -v --remap "https://google.com https://google.com/nonexistent"
# incorrectly fails
lychee a.md --cache -v
cat .lycheecache
```

An example log of the commands is below, with lychee 0.20.0.

```console
$ echo 'https://google.com' > a.md
$ lychee a.md --cache -v
     [200] https://google.com/

🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors
$ lychee a.md --cache -v --remap "https://google.com https://google.com/nonexistent"
   [INFO ] Cache is recent (age: 40s, max age: 1d 0h 0m 0s). Using.
     [200] https://google.com/ | OK (cached)

🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors
$ cat .lycheecache
https://google.com/,200,1756040828
$ rm .lycheecache
$ lychee a.md --cache -v --remap "https://google.com https://google.com/nonexistent"
     [404] https://google.com/nonexistent/ | Rejected status code (this depends on your "accept" configuration): Not Found

Issues found in 1 input. Find details below.

[a.md]:
     [404] https://google.com/nonexistent/ | Rejected status code (this depends on your "accept" configuration): Not Found

🔍 1 Total (in 0s) ✅ 0 OK 🚫 1 Error
$ lychee a.md --cache -v
   [INFO ] Cache is recent (age: 1m 26s, max age: 1d 0h 0m 0s). Using.
     [404] https://google.com/ | Error (cached)

Issues found in 1 input. Find details below.

[a.md]:
     [404] https://google.com/ | Error (cached)

🔍 1 Total (in 0s) ✅ 0 OK 🚫 1 Error
$ cat .lycheecache
[...]

PR #2109, opened 2026-03-29
Use the remapped URI for the cache key (#1818)

josrepo:
Fixes and adds a test for #1818.

### DIFF
--- a/lychee-bin/src/commands/dump.rs
+++ b/lychee-bin/src/commands/dump.rs
@@ -1,3 +1,4 @@
+use log::debug;
 use log::error;
 use log::warn;
 use lychee_lib::Request;
@@ -40,7 +41,10 @@ where
         };
 
         // Apply URI remaps (if any)
-        params.client.remap(&mut request.uri)?;
+        params
+            .client
+            .remap(&mut request.uri)?
+            .inspect(|r| debug!("Remapping {r}"));
 
         let excluded = params.client.is_excluded(&request.uri);
 
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -541,7 +541,7 @@ impl Client {
         } = request.try_into()?;
 
         let start = std::time::Instant::now(); // Measure check time
-        let remap = self.remap(&mut uri)?;
+        let remap = self.remap(&mut uri)?.inspect(|r| debug!("Remapping {r}"));
 
         let status = match uri.scheme() {
             _ if self.is_excluded(&uri) => Status::Excluded,
--- a/lychee-lib/src/remap.rs
+++ b/lychee-lib/src/remap.rs
@@ -21,7 +21,6 @@
 
 use std::{fmt::Display, ops::Index};
 
-use log::debug;
 use regex::Regex;
 use serde::Serialize;
 use url::Url;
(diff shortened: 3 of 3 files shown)


lychee-bin/src/commands/check.rs
```diff
@@ -284,7 +284,10 @@ async fn handle(
         Err(e) => return e.into_response(),
     };
 
-    let uri = request.uri.clone();
+    let mut uri = request.uri.clone();
```
katrinafyi: maybe rename this to `cache_key` to emphasise that it's only used for caching and not further link checking. i'd also wrap the .remap() call into a local block to restrict the scope of the mutable variable.

josrepo: Have made the suggested changes.

lychee-bin/src/commands/check.rs
```diff
@@ -284,26 +284,31 @@ async fn handle(
...
             // and they may have an impact on the interpretation of the status
             // code.
-            client.host_pool().record_persistent_cache_hit(&uri);
+            client.host_pool().record_persistent_cache_hit(&cache_key);
             Status::from_cache_status(v.value().status, &accept)
         };
 
         return Ok(Response::new(
-            uri.clone(),
+            cache_key.clone(),
```
katrinafyi, line 317: should work
```suggestion
            cache_key,
```

lychee-bin/src/commands/check.rs
```diff
@@ -284,26 +284,31 @@ async fn handle(
         Err(e) => return e.into_response(),
     };
 
-    let uri = request.uri.clone();
+    // Use the remapped URI for the cache key. See #1818.
```
katrinafyi: I think it's better to paraphrase the relevant details to make it immediately available.

wording is a suggestion only 
```suggestion
    // Cache key should be the actual HTTP URL which gets requested, i.e., after remaps.
```

josrepo:
@katrinafyi Thanks for the feedback!

I have removed the redundant `clone()` call and added a more descriptive comment.

> I don't love repeating the remap

I agree. Though I didn't see an easy way to move the remap further up without a larger refactor so I settled on the current approach.

I'm not sure why `cargo test` is failing either, it is passing when I run locally...

thomas-zahner:
Yes I agree that it's fine to remap twice.

> I'm not sure why cargo test is failing either, it is passing when I run locally...

`cargo test` doesn't cover all tests as features are skipped. You can use `make test` to get the same behaviour as in CI. I've fixed the issue in fb22a604.

Additionally, I tried to clean some things up, not directly related to this PR in 9fd55969.

josrepo:
> `cargo test` doesn't cover all tests as features are skipped. You can use `make test` to get the same behaviour as in CI.

I see, thanks for pointing out.

> I've fixed the issue in https://github.com/lycheeverse/lychee/commit/fb22a604b96216ed19dcb251a2914cda2768df77.

Many thanks!

katrinafyi:
@thomas-zahner Can we revert https://github.com/lycheeverse/lychee/commit/9fd55969f0d1afb24fa4e2ab06e1d9f334dfd617? It might be worthwhile, but I think it's too much to tack onto the tail of an otherwise small PR. 

Otherwise, lgtm. Thanks @josrepo!

thomas-zahner:
> @thomas-zahner Can we revert https://github.com/lycheeverse/lychee/commit/9fd55969f0d1afb24fa4e2ab06e1d9f334dfd617? It might be worthwhile, but I think it's too much to tack onto the tail of an otherwise small PR.

Yeah makes sense :+1: 
I've extracted the commit to https://github.com/lycheeverse/lychee/pull/2123

thomas-zahner:
Thank you @josrepo for the nice contribution! :rocket:

Issue #2113, opened 2026-03-29
When link highlights code like `#L23-L28` it fails


adamlui:
Like here https://github.com/adamlui/python-utils/blob/translate-messages-1.10.0/translate-messages/src/translate_messages/data/package_data.json#L23-L28

<img width="1065" height="333" alt="Image" src="https://github.com/user-attachments/assets/6b4f2e73-b138-48b0-9c2b-5d27ec180f2f" />

it did this https://github.com/adamlui/python-utils/actions/runs/23705745519/job/69057245856

<img width="1467" height="119" alt="Image" src="https://github.com/user-attachments/assets/7ba4ad24-ad86-4847-b2be-8e541cc027b5" />

Issue #1729, opened 2025-06-13
Regression on Markdown anchor links


fandreuz:
I'm facing the following error:
```
$ echo 'https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L119' | \
    docker run -i --rm lycheeverse/lychee:latest - --include-fragments
Issues found in 1 input. Find details below.

[stdin]:
[ERROR] https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L119 | Cannot find fragment

🔍 1 Total (in 1s) ✅ 0 OK 🚫 1 Error
   [WARN ] There were issues with GitHub URLs. You could try setting a GitHub token and running lychee again.
```

This used to work until last month:
```
$ echo 'https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L119' | \
    docker run -i --rm lycheeverse/lychee:sha-874b0f0-alpine - --include-fragments
🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors
```

I've bisected the first offending commit to be 0011896.

mre:
Getting the same error on Docker, but I just tried the latest binary release and it works without any issues:

```
❯❯❯ lychee --version                                                                                                                                                                                                                        
lychee 0.19.0
❯❯❯ echo 'https://github.com/googleapis/googleapis/blob/6a8c7914d1b79bd832b5157a09a9332e8cbd16d4/google/rpc/error_details.proto#L119' | lychee --include-fragments - 
🔍 1 Total (in 0s) ✅ 1 OK 🚫 0 Errors
```

That's strange because it should run the same code.

MichaIng:
It previously seems to have worked because the fragments were not actually checked: #1675

I also recognized that it fails for fragments in GitHub markdown files in some cases. E.g.:
- https://github.com/badaix/snapcast/blob/master/doc/configuration.md#airplay
  The fragment is functional, but GitHub's implementation for displaying markdown documents from repos does not actually add `airplay` as an HTML element ID.
- https://github.com/badaix/snapcast/blob/master/doc/configuration.md#user-content-airplay
[...]

PR #2116, opened 2026-03-30
quirk: remove line number fragments from github URLs

katrinafyi:
we have no hope of checking these and they're probably unlikely to break as long as the file itself continues to exist.

big remaining github fragment cases:
- `#readme` fragment going to implicit readme file - api
- fragments in implicit readme files (within a directory view) - api
- issue comment numbers - use api
- pr comment numbers - use api
- other branch tree views?

Might close https://github.com/lycheeverse/lychee/issues/2113 and also addresses https://github.com/lycheeverse/lychee/issues/1729 for this particular case.

### DIFF
--- a/lychee-lib/src/quirks/mod.rs
+++ b/lychee-lib/src/quirks/mod.rs
@@ -15,7 +15,7 @@ static CRATES_PATTERN: LazyLock<Regex> =
 static YOUTUBE_PATTERN: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^(https?://)?(www\.)?youtube(-nocookie)?\.com").unwrap());
 static YOUTUBE_SHORT_PATTERN: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r"^(https?://)?(www\.)?(youtu\.?be)").unwrap());
+    LazyLock::new(|| Regex::new(r"^(https?://)?(www\.)?(youtu\.be)").unwrap());
 static GITHUB_BLOB_MARKDOWN_FRAGMENT_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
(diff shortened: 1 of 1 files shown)


mre:
> #readme fragment going to implicit readme file

There's also an API endpoint to find the URL for a repo's readme. Is that helpful?
https://docs.github.com/en/rest/repos/contents?apiVersion=2026-03-10#get-a-repository-readme

katrinafyi:
oh, yes! that does help. we can add that in another pr

thomas-zahner:
Thanks for the quirk! I also like the new debugging logs.

> Might close https://github.com/lycheeverse/lychee/issues/2113

I've double-checked, it actually does resolve the issue :+1:

katrinafyi:
>  I also like the new debugging logs.

To be honest, they were kind of a lazy job just to see if it was working. Do you think it's worthwhile to add name strings to each quirk to avoid logging the big regex?

thomas-zahner:
> To be honest, they were kind of a lazy job just to see if it was working.

Haha :) but I definitely prefer that we now see when quirks are applied, could help a lot with troubleshooting

> Do you think it's worthwhile to add name strings to each quirk to avoid logging the big regex?

Yes that would be quite nice, I also considered that. Let's do it :+1:

katrinafyi:
I had to think about the precedence of the two github fragment quirks. I wrote about it in https://github.com/lycheeverse/lychee/commit/bd17ccbd7d0461c035e707cd4071190f257d1137, lmk wyt. I also wonder if it's time to start thinking about applying multiple quirks.

mre:
> I also wonder if it's time to start thinking about applying multiple quirks.

No objections. I leave it up to you to decide if you'd like to tackle this in the same PR or a separate one. :)

katrinafyi:
Added multiple quirk support in 93eef4ac0d979b8e8b5d4fca68950d0905966774. It isn't actually used yet, because the github line quirk deletes fragments so the github markdown fragment quirk won't apply after it. But it was a very simple change so I just did it here.

Issue #2097, opened 2026-03-23
Split up and refactor Status enum


katrinafyi:
I feel like the Status enum has grown beyond its original purpose and has become too big and complicated. 

The main problem is that it mixes up two things which I think should be separated: (1) the actual outcome of the request, and (2) how that outcome should be interpreted. Another problem is that the enum has grown data fields in the form of the Redirected variant. This history information should really be separate from the actual status and stored in a struct.

The presence of functions like `Status::is_success`, `is_error`, etc shows that the shape of the enum does not match the shape of the information you want it to store - ideally, these would all be match statements on a flat enum type. Functions like `is_success_ignoring_timeouts` are also desiring more flexibility than the current types can provide.

I've sketched out a rough idea for splitting it up into two parts:
- Outcome, the objective outcome of the check, and
- Verdict or Conclusion, the interpretation of the Outcome after taking into account --accept, etc.

(This naming is taken from [Github actions](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context))

```rust
/// Outcome of the checker, without any presumption of success or failure.
pub enum Outcome {
    Website {
        result: Result<StatusCode, reqwest::Error>,
        redirects: Redirects,
        fragment_check: Option<bool>,
    },
    Mail {
        result: mailify_lib::CheckResult,
    },
    File {
        result: Result<Metadata, io::Error>,
        applied_fallback_extension: Option<String>,
        applied_index_file: Option<String>,
        fragment_check: Option<bool>,
    },
    Excluded,
    RequestError(RequestError),
    Unsupported(ErrorKind),
    Cached(CacheStatus),
}

/// Response to a lychee Request. Bundles an Outcome along with metadata
/// (like source location) and pre-check transformations like remaps.
pub struct Response {
    outcome: Outcome,
    request: Request, // <-- remap_history is inside Request
    input_source: crate::InputSource,
    response_body: crate::ResponseBody,
}

/// An interpretation of an [`Outcome`], taking into account user-provided
/// configuration. Corresponds to counters in the CLI output.
pub enum Conclusion {
    Ok,
    Error,
    Timeout,
    Redirected,
    Cached,
    Excluded,
}

impl Conclusion {
    /// For example.
    pub fn from_response(
        response: Response,
        accept: StatusCodeSelector,
        accept_timeouts: bool,
        other_things: MoreOptions,
    ) -> Conclusion {
        Self::Ok
    }
}
```

[...]

Issue #2102, opened 2026-03-25
Combine remap & redirect info


thomas-zahner:
It would be nice if the remaps could be printed *before* redirects. It would make all the remaps/redirects flow from left `-->` right.

Additionally we should consider combining them which further emphasises the order of operations:
```
Followed remap and 1 redirect: http://github.com/lycheeverse --[remap]--> http://rust-lang.org/lycheeverse --[301]--> https://rust-lang.org/lycheeverse
```

This might be difficult with the current Status::Redirected / Status::Remapped nesting. But it should be easier once https://github.com/lycheeverse/lychee/issues/2097 is resolved.

Also consider using Unicode symbols as mentioned by @mre in https://github.com/lycheeverse/lychee/pull/2094#discussion_r2998670985.

_Originally posted by @katrinafyi in https://github.com/lycheeverse/lychee/pull/2094#discussion_r2984914922_

mre:
That would be great. On that note, I found the remap/redirect output hard to parse visually. The lines are very long.

Was wondering if we should only show a shorthand format in normal mode and the full version in verbose mode:

So, in normal mode:

```
http://github.com/lycheeverse --> https://rust-lang.org/lycheeverse
```

or maybe

```
http://github.com/lycheeverse --> https://rust-lang.org/lycheeverse (after remap and redirects)
```

In verbose mode:

```
Followed remap and 1 redirect: http://github.com/lycheeverse --[remap]--> http://rust-lang.org/lycheeverse --[301]--> https://rust-lang.org/lycheeverse
```

I also thought about a tree-like output:

```
http://github.com/lycheeverse
└─[remap]─▶ http://rust-lang.org/lycheeverse
└─[301]──▶ https://rust-lang.org/lycheeverse
```

(It might be fine since the output is meant for humans anyway. If users want machine-readable output, they can use `--format json`.)

Related: #1926

PR #2129, opened 2026-04-04
Remove the `Redirect` and `Remapped` wrapper variants from the `Status` enum.

mre:
The goal of this refactor is to make a step towards cleaning up the `Status` enum as proposed in #2097.

Previously, redirect and remap information was nested inside the `Status` enum, which forced every exhaustive match (~8 places) to either recurse through these wrappers or delegate via the `innermost()` helper. This made the enum harder to work with and obscured the actual status.

The proposal is to have redirect and remap data live as separate fields on `ResponseBody`.
This gets rid of innermost() and redirects() methods from Status and all is_success/is_error/etc. methods now match directly on self.

I am aware that the three `None`s in a row in `Response` are pretty bad. The constructor already had too many positional params before this change, and I made it worse.

However, this change will unlock the possibility to move redirects into `Outcome` in the future, e.g.

```rust
pub enum Outcome {
    Website {
        result: Result<StatusCode, reqwest::Error>,
        redirects: Redirects,
        fragment_check: Option<bool>,
    },
    // ...
}
```

This way, we can get rid of the `None` fields in `Response` again in the future when we introduce `Outcome`.

(I attempted that first, but it turned out to be a much bigger refactor, so I broke out a smaller chunk for now.)

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -255,7 +255,9 @@ async fn check_url(client: &Client, request: Request) -> Response {
         log::error!("Error checking URL {uri}: {e}");
         Response::new(
             uri.clone(),
-            Status::Error(ErrorKind::InvalidURI(uri.clone())),
+            Status::Error(ErrorKind::InvalidURI(uri)),
+            None,
+            None,
             source.into(),
             span,
             None,
@@ -316,6 +318,8 @@ async fn handle(
         return Ok(Response::new(
             cache_key.clone(),
             status,
+            None,
+            None,
             request.source.into(),
             request.span,
             None,
@@ -389,9 +393,9 @@ mod tests {
     async fn test_invalid_url() {
         let client = ClientBuilder::builder().build().client().unwrap();
         let uri = Uri::try_from("http://\"").unwrap();
-        let response = client.check_website(&uri, None).await.unwrap();
+        let (status, _redirects) = client.check_website(&uri, None).await.unwrap();
         assert!(matches!(
-            response,
+            status,
             Status::Unsupported(ErrorKind::BuildRequestClient(_))
         ));
     }
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1965,23 +1965,17 @@ The config file should contain every possible key for documentation purposes."
         let json = stdout_to_json(&stdout);
         assert_eq!(json["remaps"], 1);
         assert_eq!(json["excludes"], 1);
+        let excluded = &json["excluded_map"]["stdin"][0];
+        assert_eq!(excluded["url"], "https://bbb.com/");
+        assert_eq!(excluded["status"]["text"], "Excluded");
         assert_eq!(
-            json["excluded_map"],
-            json!({
-            "stdin": [
-              {
-                "url": "https://bbb.com/",
-                "status": {
-                  "text": "Excluded",
-                  "details": "This is due to your 'exclude' values | Remaps: https://aaa.com/ --> https://bbb.com/"
-                },
-                "span": {
-                  "line": 1,
-                  "column": 1
-                },
-              }
-            ]})
+            excluded["status"]["details"],
+            "This is due to your 'exclude' values"
         );
+        assert_eq!(excluded["remap"]["original"]["url"], "https://aaa.com/");
+        assert_eq!(excluded["remap"]["new"]["url"], "https://bbb.com/");
+        assert_eq!(excluded["span"]["line"], 1);
+        assert_eq!(excluded["span"]["column"], 1);
     }
 
     #[test]
@@ -2636,6 +2630,13 @@ The config file should contain every possible key for documentation purposes."
                 json["success_map"],
                 json!({
                 "stdin":[{
+                    "redirects": {
+                        "origin": redirect_url,
+                        "redirects": [{
+                            "code": 308,
+                            "url": ok_url,
+                        }]
+                    },
                     "span": {
                         "column": 1,
                         "line": 1,
@@ -2643,13 +2644,6 @@ The config file should contain every possible key for documentation purposes."
                     "status": {
                         "code": 200,
                         "text": "200 OK",
-                        "redirects": {
-                            "origin": redirect_url,
-                            "redirects": [{
-                                "code": 308,
-                                "url": ok_url,
-                            }]
-                        },
                     },
                     "url": redirect_url
                 }]})
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1961,17 +1961,31 @@ The config file should contain every possible key for documentation purposes."
         let json = stdout_to_json(&stdout);
         assert_eq!(json["remaps"], 1);
         assert_eq!(json["excludes"], 1);
-        let excluded = &json["excluded_map"]["stdin"][0];
-        assert_eq!(excluded["url"], "https://bbb.com/");
-        assert_eq!(excluded["status"]["text"], "Excluded");
         assert_eq!(
-            excluded["status"]["details"],
-            "This is due to your 'exclude' values"
+            json["excluded_map"],
+            json!({
+            "stdin": [
+              {
+                "url": "https://bbb.com/",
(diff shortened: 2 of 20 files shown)


lychee-bin/tests/cli.rs
```diff
@@ -2636,20 +2630,20 @@ The config file should contain every possible key for documentation purposes."
...
                     "status": {
                         "code": 200,
                         "text": "200 OK",
-                        "redirects": {
-                            "origin": redirect_url,
-                            "redirects": [{
-                                "code": 308,
-                                "url": ok_url,
-                            }]
-                        },
```
mre, line 2670: Note: redirects have moved out of `status`.

lychee-lib/src/types/response.rs
```diff
@@ -116,12 +143,27 @@ impl Display for ResponseBody {
...
+        if let Some(redirects) = &self.redirects {
+            let count = redirects.count();
+            let noun = if count == 1 { "redirect" } else { "redirects" };
+            write!(f, " | Followed {count} {noun}. Redirects: {redirects}")?;
+        }
+
+        if let Some(remap) = &self.remap {
+            write!(f, " | Remaps: {remap}")?;
+        }
+
```
katrinafyi: I'd print remaps first, because remaps occur before redirects.

Also, do you want to try tackle https://github.com/lycheeverse/lychee/issues/2102 at this time? it shouldn't be too hard to match on the case where both remaps and redirects are Some.

lychee-bin/tests/cli.rs
```diff
@@ -1965,23 +1965,17 @@ The config file should contain every possible key for documentation purposes."
...
-                  "details": "This is due to your 'exclude' values | Remaps: https://aaa.com/ --> https://bbb.com/"
-                },
-                "span": {
-                  "line": 1,
-                  "column": 1
-                },
-              }
-            ]})
+            excluded["status"]["details"],
+            "This is due to your 'exclude' values"
```
katrinafyi: do you prefer to move towards checking individual fields rather than a whole JSON object? i did like the json object because it shows all of the structure at once. not a big deal though.

katrinafyi (approved):
Very happy with the refactor and the resulting simplifications. Looks good!

It's not specific to this PR, but I was never clear on the separation between Response and ResponseBody.

In https://github.com/lycheeverse/lychee/issues/2097, it was also my intention to store Remap inside Request and then store Request inside Response. With that, we would avoid needing to move all the fields across into Response(Body), like span, source, etc. Nothing needs to change in this PR, but just checking we're on the same page :)

lychee-bin/tests/cli.rs
mre: Yeah good point. I don't mind keeping the JSON structure. Was experimenting with individual asserts to see how that feels, but I'll change it back. 👍

lychee-lib/src/types/response.rs
mre: Yeah, I can change the ordering. Will try to tackle #2102 in a separate commit. If it works out and doesn't get too big we can keep it. If not, I can move it to a separate PR. @thomas-zahner fyi.

mre:
> In https://github.com/lycheeverse/lychee/issues/2097, it was also my intention to store Remap inside Request and then store Request inside Response. With that, we would avoid needing to move all the fields across into Response(Body), like span, source, etc. Nothing needs to change in this PR, but just checking we're on the same page :)

Sounds like a plan. No objections right now. My only concern is that `Response` becomes quite a big type, so printing might get harder, but inspection should get easier. So it's a tradeoff.

I've never been a huge fan of the standalone `ResponseBody` anyway. Well, having a strong type instead of `Option<String>` would be nice, but the fact that `ResponseBody` was completely detached from `Response` caused some headaches in the past, like the field-copying/alignment that you mentioned.

lychee-bin/tests/cli.rs
mre: Right now I'm wondering if we should move redirects back into status. They sort of belong together (?). As a user, that's where I'd look for that info, I guess.
On the other hand, it would make the code messier, because we'd have to somehow merge redirects and that status vector. So, not sure what to do. Maybe it's fine? (Just wanted to mention that I'm still undecided.)

katrinafyi:
Yeah, the main use of ResponseBody is it appears in the response maps:
```rs
```
where ResponseBody is basically a Response minus the InputSource.
```rs
pub struct Response {
    input_source: InputSource,
    response_body: ResponseBody,
}
```
This makes some sense, as the hashmap already has the input source in the key. But yeah, just something to keep in mind for later.

lychee-bin/tests/cli.rs
katrinafyi: I think it's okay outside of status. For example, a browser (and lychee) makes a fresh request for each redirect, so it's not really a part of the final status. It's just a history of the path taken to get to the final status, so I think it makes sense outside of status and adjacent to remaps.

But I wouldn't be against a transformation to put it inside the `status` JSON, if you think that would be better (and it's less churn for users). The main refactor is in the Status enum, the JSON representation could be something else if desired.

lychee-bin/tests/cli.rs
thomas-zahner: Agree with @katrinafyi. By the way, that's a "pattern" I've observed quite a lot throughout the code base. We tend to often do assertions on individual fields, where asserting the full struct would have been easier and prevent potential regressions. I've changed this pattern quite a few times to make the assertions on the whole struct.

thomas-zahner (approved):
Thank you @mre, I have no objections. I agree that we can clean up the `None`s with a new PR as you explained.

Sorry for introducing this confusing approach in the first place. Additionally, the current nesting made it possible to introduce useless and confusing state such as containing `Redirected` or `Remapped` multiple times. This is something I didn't think about when I introduced it. So your refactor is definitely the way to go 👍

lychee-bin/tests/cli.rs
thomas-zahner: Personally I'm fine with this change form a user's perspective. From a code perspective this avoids additional logic, so I think this is the way to go.

lychee-bin/tests/cli.rs
mre: yeah, would be cool to have an `assert_json!`.

mre:
Cool! I will rebase the PR and apply the changes we agreed on when I get the chance. Will merge when everything is clean.

lychee-bin/src/commands/check.rs
```diff
@@ -255,7 +255,9 @@ async fn check_url(client: &Client, request: Request) -> Response {
         log::error!("Error checking URL {uri}: {e}");
         Response::new(
             uri.clone(),
-            Status::Error(ErrorKind::InvalidURI(uri.clone())),
+            Status::Error(ErrorKind::InvalidURI(uri)),
+            None,
```
thomas-zahner, line 259: I've just merged master into this branch to resolve conflicts, as we moved the cache functionality. This got me thinking that with these two `None` values we do lose some information. It would be pretty cool if we cached a whole `Response` instead of just the `Status`. But then `Status` also includes `span`, `source` and `Duration` which shouldn't be cached. But that's just a side note.

lychee-bin/tests/cli.rs
thomas-zahner: Not quite sure what you mean @mre but e97daccdcb21e6d21b113211eabdb0cec0b6f39c should revert the behaviour

lychee-lib/src/types/response.rs
thomas-zahner: Done in 41d072f345c6fe8587d5b063e8fe14f32be50b2b

mre: Thanks!

lychee-bin/src/commands/check.rs
mre: Okay, yeah, let's keep the `None`s for now then. 👍

mre:
Just some more lints to fix. 😅 @thomas-zahner, guess you're on it?

mre:
Thanks for pushing this over the finishing line, Thomas!

PR #2130, opened 2026-04-05
Fix matching logic for glob hidden files

### LINKED ISSUE #1937: Hidden files are always included by glob
I was thinking about this after https://github.com/lycheeverse/lychee/issues/1935, but idk if it's a bug or just imprecise documentation.

These commands
```bash
mkdir -p /tmp/hidden/asd/
echo 'https://a.com/' > /tmp/hidden/asd/.a.md
cargo run -q -- --dump  '/tmp/hidden/**/*' 
```
will print
```
https://a.com/
```

The view could be taken that the glob specifies that the file should be included. Alternatively, a view could be taken that shell globs would exclude hidden files by default and we should do the same.

Maybe this is not a big issue.

mre:
Hidden files were always included by glob patterns, even without `--hidden` flag. By default, shell globs exclude hidden files, but lychee behaved differently.

The trick is to pass `skip_hidden` to glob's`match_opts` to match shell behavior: wildcards like `*` and `**` should not match hidden files/dirs unless `--hidden` is passed.

Fixes: #1937

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -62,18 +62,12 @@ mod cli {
     /// Order of the lines is ignored.
     fn assert_lines_eq<S: AsRef<str> + Ord>(result: Assert, mut expected_lines: Vec<S>) {
         let output = &result.get_output().stdout;
-        let mut actual_lines: Vec<String> = output
-            .lines()
-            .map(|line| line.unwrap().to_string())
-            .collect();
+        let mut actual_lines: Vec<String> = output.lines().map(|line| line.unwrap()).collect();
 
         actual_lines.sort();
         expected_lines.sort();
 
-        let expected_lines: Vec<String> = expected_lines
-            .into_iter()
-            .map(|l| l.as_ref().to_owned())
-            .collect();
+        let expected_lines: Vec<&str> = expected_lines.iter().map(|l| l.as_ref()).collect();
 
         assert_eq!(actual_lines, expected_lines);
     }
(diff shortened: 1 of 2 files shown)


lychee-bin/tests/cli.rs
```diff
@@ -842,11 +842,15 @@ mod cli {
     #[tokio::test]
     async fn test_glob_recursive() -> Result<()> {
         let dir = tempfile::tempdir()?;
-        let subdir_level_1 = tempfile::tempdir_in(&dir)?;
-        let subdir_level_2 = tempfile::tempdir_in(&subdir_level_1)?;
+        // Use explicit non-hidden subdirectories instead of tempdir_in,
+        // which creates `.tmp*` dirs that are hidden and correctly skipped
+        // by glob when skip_hidden is true.
```
thomas-zahner, line 847: I found the comment quite confusing and had to reread it multiple times. Personally I think it makes sense to just remove it as it pretty much explains code which is removed

lychee-lib/src/types/input/resolver.rs
```diff
@@ -118,6 +118,9 @@ impl InputResolver {
                 let glob_expanded = tilde(pattern.as_str()).to_string();
                 let mut match_opts = glob::MatchOptions::new();
                 match_opts.case_sensitive = !ignore_case;
+                // Match shell behavior: wildcards like `*` and `**` should not
+                // match hidden files/dirs unless `--hidden` is passed.
```
thomas-zahner, line 122: Maybe? `--hidden` is very lychee-bin specific.

```suggestion
                // Match shell behavior: wildcards like `*` should not
                // match hidden files by default.
```

thomas-zahner (approved):
Really cool 👍

mre:
Thanks for the fixes @thomas-zahner. 🙂

PR #2135, opened 2026-04-08
Replace custom rate limit header parsing with rate-limits crate

### LINKED ISSUE #2027: Retry backoff mechanism kicks in unexpectedly
As one of the maintainers of https://support.torproject.org, I'm looking into options for incorporating dead link checking into our CI pipeline, and this project seems very promising so far.

I've hit a snag however related to crawling links specifically to `gitlab.torproject.org`, our self-hosted GitLab instance: it seems like when lychee hits that host it immediately goes into retry-backoff mode with a 1 minute delay between each request, which pretty much kills the performance of the whole process. If I exclude all `gitlab.torproject.org` links (and other links that redirect to that domain), it goes through something like 16 thousands links in a matter of seconds.

Given a bogus `test.html` file containing a small sample of such links:

~~~
<a href="https://gitlab.torproject.org/">foo</a>
<a href="https://gitlab.torproject.org/tpo/applications/vpn/-/issues/new">bar</a>
<a href="https://gitlab.torproject.org/tpo/applications/vpn-leak-test/-/blob/main/app/src/main/java/org/torproject/vpnleaktest/tests/DeviceIDsTest.kt#L235">baz</a>
~~~

Running lychee on this file takes several minutes:

~~~
$ podman run --init --rm -it -v $PWD:/app --workdir /app docker.io/lycheeverse/lychee:master test.html -vvv --max-concurrency 1
   [DEBUG] Following redirect to https://gitlab.torproject.org/tpo/team
    [WARN] Host gitlab.torproject.org sent an unexpectedly big rate limit backoff duration of 584542046years 1month 2days 15h 52m 15s 615ms. Capping the duration to 1m instead.
   [DEBUG] Host gitlab.torproject.org applying backoff delay of 60000ms due to previous rate limiting or errors
     [200] https://gitlab.torproject.org/ | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://gitlab.torproject.org/ --[302]--> https://gitlab.torproject.org/tpo/team
   [DEBUG] Following redirect to https://gitlab.torproject.org/users/sign_in
    [WARN] Host gitlab.torproject.org sent an unexpectedly big rate limit backoff duration of 584542046years 1month 2days 15h 52m 15s 615ms. Capping the duration to 1m instead.
   [DEBUG] Host gitlab.torproject.org applying backoff delay of 60000ms due to previous rate limiting or errors
     [200] https://gitlab.torproject.org/tpo/applications/vpn/-/issues/new | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://gitlab.torproject.org/tpo/applications/vpn/-/issues/new --[302]--> https://gitlab.torproject.org/users/sign_in
    [WARN] Host gitlab.torproject.org sent an unexpectedly big rate limit backoff duration of 584542046years 1month 2days 15h 52m 15s 615ms. Capping the duration to 1m instead.
     [200] https://gitlab.torproject.org/tpo/applications/vpn-leak-test/-/blob/main/app/src/main/java/org/torproject/vpnleaktest/tests/DeviceIDsTest.kt#L235
🔍 3 Total (in 2m 1s) ✅ 1 OK 🚫 0 Errors 🔀 2 Redirects
~~~

I'm not really sure what's happening here: where is this "584542046years 1month 2days 15h 52m 15s 615ms" duration coming from? From looking into the code a little but, it seems related to a `Retry-After` HTTP header, but as far as I can tell, no such header is actually sent from this host. It also mentions "previous rate limiting errors" but it's unclear where these errors were encountered, eg. is the 302 redirect being considered an error here?

I've tried running with `RUST_LOG=trace` hoping I could look at the actual HTTP headers sent and received by lynchee but I couldn't decipher those logs...

mre:
This migrates our rate-limit handling to the [rate-limits](https://github.com/mre/rate-limits) crate.

This work is based on my rewrite of the rate-limits crate in https://github.com/mre/rate-limits/pull/11. I rewrote it from scratch to be more efficient as well as use a state-machine under the hood for more reliable vendor detection.

I tested it locally as best as I could and so far I haven't seen any regressions.

Once all looks good, I will release a new version of the rate-limits crate based on this new approach and will remove the "draft" setting on this PR. We could also move the crate into the lycheeverse org, which I believe will make shared maintenance / bugfix releases easier in the future. wdyt?

Edit: while refactoring the code, I also decided to remove the 80% threshold for throttling request, which we had before. I don't think it's helpful. Instead, I now believe, we should go to the maximum number of allowed requests and get throttled. AFAICT, there is no penalty for that, and we could squeeze out some more performance this way. Once we get throttled, we just wait until we get unblocked. From what I understand, this is how to work with those headers correctly.

Closes #2027


lychee-lib/Cargo.toml
```diff
@@ -13,6 +13,7 @@ version.workspace = true
 rust-version = "1.88.0"
 
 [dependencies]
+rate-limits = { git = "https://github.com/mre/rate-limits", branch = "parser" }
```
mre, line 16: Will replace this with the version from crates.io, once https://github.com/mre/rate-limits/pull/11 got released.

thomas-zahner: Sounds good. Should I review the PR over there?

lychee-bin/tests/rate_limit.rs
```diff
@@ -0,0 +1,126 @@
...
+        .success();
+
+    let elapsed = start.elapsed();
+
+    assert!(
+        elapsed >= expected_min_duration,
+        "Rate limiting failed! Expected at least {expected}ms, got {actual}ms",
+        expected = expected_min_duration.as_millis(),
+        actual = elapsed.as_millis()
+    );
```
thomas-zahner, line 108: Maybe?

```suggestion
    assert!(
        elapsed >= expected_min_duration,
        "Rate limit headers were not respected! Expected minimum delay of {expected_min_duration}, but got {elapsed}"
    );
```

lychee-bin/tests/rate_limit.rs
```diff
@@ -0,0 +1,126 @@
...
+    run_rate_limit_test(
+        200,
+        &[
+            ("x-ratelimit-limit", "100"),
+            ("x-ratelimit-remaining", "0"),
+            ("x-ratelimit-reset", &reset_time.to_string()),
+        ],
+        &["/1", "/2"],
+        // Give 100ms leeway for scheduler flakiness in CI environments
+        Duration::from_millis(1900),
```
thomas-zahner: I don't understand these numbers here. `reset_time` indicates 3 seconds. And here we test with 2 seconds. It should be better documented if this is no mistake. Same is true for the other tests.

The 100ms could be moved as a constant into the test function, hidden away for better abstraction.

Note that there is a test called `test_retry_rate_limit_headers` in cli.rs. Did you see it? Maybe it can be incorporated?

lychee-bin/tests/rate_limit.rs
```diff
@@ -0,0 +1,126 @@
...
+        .arg("--max-concurrency")
+        .arg("1") // Ensure sequential execution so backoffs accumulate predictably
+        .write_stdin(inputs)
+        .assert()
+        .success();
+
+    let elapsed = start.elapsed();
+
+    assert!(
+        elapsed >= expected_min_duration,
```
thomas-zahner: I think we should also assert the other way (`<`) similar as in `test_retry_rate_limit_headers` and use some sort of tolerance. Currently we wouldn't notice if the duration is much bigger than expected unless the test gets really slow at some point.

lychee-lib/src/ratelimit/host/host.rs
```diff
@@ -272,51 +274,26 @@ impl Host {
...
-            let duration = match headers::parse_retry_after(retry_after_value) {
-                Ok(e) => e,
-                Err(e) => {
-                    warn!("Unable to parse Retry-After header as per RFC 7231: {e}");
-                    return;
+        if let Ok(rate_limit) = rate_limits::RateLimit::new(headers) {
+            if rate_limit.is_limited() {
+                let duration = rate_limit.reset().duration();
+                if !duration.is_zero() {
+                    self.increase_backoff(duration);
```
thomas-zahner, line 289: That's great 👍

thomas-zahner (changes requested):

lychee-lib/Cargo.toml
mre: Welll you can, but beware that it's a big diff. So it's optional, but highly appreciated. 😆

lychee-bin/tests/rate_limit.rs
mre: good idea

lychee-bin/tests/rate_limit.rs
mre: Remember that this includes the time to spawn lychee as a completely new process. If we set the reset time to 2s, the actual sleep duration inside `lychee` would be ~900ms and the test could randomly fail. 🤔 I wonder if I can isolate that part somehow to make the tests clearer...

lychee-bin/tests/rate_limit.rs
mre: done

lychee-bin/tests/rate_limit.rs
thomas-zahner: Hmm, I still don't really understand..
`reset_time` is in seconds and has `+ 3` and now we assert against `1900`ms. I don't understand the relation between the numbers. I think it should be explained somehow. Also it would be nice if the expected number (`1900`ms) is calculated instead of hardcoded if that is possible.

lychee-bin/tests/rate_limit.rs
```diff
@@ -0,0 +1,165 @@
...
+    server.verify().await;
+}
+
+#[tokio::test]
+async fn test_retry_after_seconds() {
+    run_rate_limit_test(
+        200,
+        &[("Retry-After", "2")],
+        &["/1", "/2"],
+        Duration::from_millis(1900),
```
katrinafyi: 1900 is a bit magical. Can the `run_rate_limit_test` take the predicted ideal duration then apply a threshold in its own code? edit:oh  @ thomas-zahner already got this in his comment.

katrinafyi (commented):
Code here looks reasonable. I also took a squiz at https://github.com/mre/rate-limits/pull/11.

lychee-bin/tests/rate_limit.rs
mre: The numbers are now derived from named constants (RATE_LIMIT_DELAY_SECONDS, RATE_LIMIT_RESET_OFFSET_SECONDS, TIMING_TOLERANCE, MAX_OVERHEAD) with explanatory doc comments, and the expected delay is computed rather than hardcoded. Hope that clears things up!

lychee-bin/tests/rate_limit.rs
mre: Done. The helper now takes an expected_delay and asserts elapsed < expected_delay + MAX_OVERHEAD.

lychee-lib/src/ratelimit/host/host.rs
mre: Thanks!

lychee-bin/tests/rate_limit.rs
mre: Done. run_rate_limit_test now takes expected_delay and applies TIMING_TOLERANCE / MAX_OVERHEAD internally, so no more magic numbers at the call sites.

lychee-bin/tests/rate_limit.rs
```diff
@@ -0,0 +1,332 @@
...
+/// measured server-side delay is `RATE_LIMIT_RESET_OFFSET_SECONDS` minus the
+/// time between `calculate_reset_time()` and lychee parsing the response
+/// headers. That includes process spawn, TLS setup, and the first HTTP
+/// round-trip, which can be well over 500ms on slow CI runners.
+const TIMING_TOLERANCE: Duration = Duration::from_millis(1000);
+
+/// Maximum additional time beyond the expected delay before the test fails.
+/// Intentionally generous to handle slow CI environments while still catching
+/// runaway delays (e.g. the 60-second `MAXIMUM_BACKOFF` cap being mistakenly applied).
+const MAX_OVERHEAD: Duration = Duration::from_secs(5);
```
thomas-zahner: I'm not sure if it makes sense to treat this "overhead" differently from the tolerance. Also I think a value of `5` is way too big. I would reduce it to one second in which case we could reuse "tolerance" which already sounds like it's intended to be used both ways.

```patch
diff --git a/lychee-bin/tests/rate_limit.rs b/lychee-bin/tests/rate_limit.rs
index 565caaef..a6cd921a 100644
--- a/lychee-bin/tests/rate_limit.rs
+++ b/lychee-bin/tests/rate_limit.rs
@@ -14,7 +14,7 @@ const RATE_LIMIT_DELAY_SECONDS: u64 = 1;
 /// backoff of at least `RATE_LIMIT_DELAY_SECONDS`.
 const RATE_LIMIT_RESET_OFFSET_SECONDS: u64 = RATE_LIMIT_DELAY_SECONDS + 1;
 
-/// How far below the expected delay is still considered a passing result,
+/// How far below or above the expected delay is still considered a passing result,
 /// accounting for timing imprecision in the test environment.
 ///
 /// For tests that encode an *absolute* Unix timestamp (GitHub, GitLab), the
@@ -24,11 +24,6 @@ const RATE_LIMIT_RESET_OFFSET_SECONDS: u64 = RATE_LIMIT_DELAY_SECONDS + 1;
 /// round-trip, which can be well over 500ms on slow CI runners.
 const TIMING_TOLERANCE: Duration = Duration::from_millis(1000);
 
-/// Maximum additional time beyond the expected delay before the test fails.
-/// Intentionally generous to handle slow CI environments while still catching
-/// runaway delays (e.g. the 60-second `MAXIMUM_BACKOFF` cap being mistakenly applied).
-const MAX_OVERHEAD: Duration = Duration::from_secs(5);
-
 /// Calculates an absolute Unix epoch timestamp `RATE_LIMIT_RESET_OFFSET_SECONDS` from now.
 fn calculate_reset_time() -> String {
     (SystemTime::now()
@@ -41,7 +36,7 @@ fn calculate_reset_time() -> String {
 
 /// Sets up a mock server with two paths, runs lychee sequentially over both,
 /// and asserts that the time between the two server-side request arrivals falls
-/// within `[expected_delay - TIMING_TOLERANCE, expected_delay + MAX_OVERHEAD]`.
+/// within `[expected_delay - TIMING_TOLERANCE, expected_delay + TIMING_TOLERANCE]`.
 ///
 /// Pass `Duration::from_secs(RATE_LIMIT_RESET_OFFSET_SECONDS)` for headers that
 /// encode an *absolute* Unix timestamp (GitHub, GitLab), and
@@ -103,7 +98,7 @@ async fn run_rate_limit_test(headers: &[(&str, &str)], expected_delay: Duration)
     };
 
     let expected_min = expected_delay.saturating_sub(TIMING_TOLERANCE);
-    let expected_max = expected_delay + MAX_OVERHEAD;
+    let expected_max = expected_delay.saturating_add(TIMING_TOLERANCE);
 
     assert!(
         elapsed >= expected_min,
@@ -232,7 +227,7 @@ async fn test_retry_rate_limit_headers() {
     let elapsed = second.duration_since(first);
 
     let expected_min = RETRY_DELAY.saturating_sub(TIMING_TOLERANCE);
-    let expected_max = RETRY_DELAY + MAX_OVERHEAD;
+    let expected_max = RETRY_DELAY.saturating_add(TIMING_TOLERANCE);
 
     assert!(
         elapsed >= expected_min,
@@ -317,7 +312,7 @@ async fn test_retry_after_seconds() {
 
     let expected_delay = Duration::from_secs(RATE_LIMIT_DELAY_SECONDS);
     let expected_min = expected_delay.saturating_sub(TIMING_TOLERANCE);
-    let expected_max = expected_delay + MAX_OVERHEAD;
+    let expected_max = expected_delay.saturating_add(TIMING_TOLERANCE);
 
     assert!(
         elapsed >= expected_min,
```

lychee-bin/tests/rate_limit.rs
```diff
@@ -0,0 +1,332 @@
...
+const RATE_LIMIT_RESET_OFFSET_SECONDS: u64 = RATE_LIMIT_DELAY_SECONDS + 1;
+
+/// How far below the expected delay is still considered a passing result,
+/// accounting for timing imprecision in the test environment.
+///
+/// For tests that encode an *absolute* Unix timestamp (GitHub, GitLab), the
+/// measured server-side delay is `RATE_LIMIT_RESET_OFFSET_SECONDS` minus the
+/// time between `calculate_reset_time()` and lychee parsing the response
+/// headers. That includes process spawn, TLS setup, and the first HTTP
+/// round-trip, which can be well over 500ms on slow CI runners.
```
thomas-zahner: Is this second paragraph the right place to document this? Not sure I really understand this. Sometimes I get test failures which might have to do with this comment?

Or maybe they are quite unrelated to this comment. Do we take into account that a Unix timestamp is heavily rounded? It may be off by almost one second (e.g. 999ms). This plus a little delay is more than a second.

```
running 3 tests
test test_github_rate_limit_exhausted ... FAILED
test test_gitlab_rate_limit_exhausted ... FAILED
test test_retry_rate_limit_headers ... ok

failures:

---- test_github_rate_limit_exhausted stdout ----

thread 'test_github_rate_limit_exhausted' (32706) panicked at lychee-bin/tests/rate_limit.rs:103:5:
Rate limit headers were not respected! Expected at least 1s, but got 944.096459ms

---- test_gitlab_rate_limit_exhausted stdout ----

thread 'test_gitlab_rate_limit_exhausted' (32707) panicked at lychee-bin/tests/rate_limit.rs:103:5:
Rate limit headers were not respected! Expected at least 1s, but got 944.096389ms


failures:
    test_github_rate_limit_exhausted
    test_gitlab_rate_limit_exhausted

test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 2 filtered out; finished in 1.09s

```

thomas-zahner (approved):
Thank you for all the work!

lychee-bin/tests/rate_limit.rs
mre: Much better. Applied.

lychee-bin/tests/rate_limit.rs
mre: True. I've updated the comment now.

thomas-zahner (approved):
Thank you! Thrilled to see this mess of rate-limit mechanisms distilled into a dedicated Rust crate and lychee making use of it :)

mre:
> Thank you! Thrilled to see this mess of rate-limit mechanisms distilled into a dedicated Rust crate and lychee making use of it :)

Oh, wow! Thanks so much for that comment. ☺️ And indeed, I'm also thrilled. The new mechanism is zero-copy and more robust than what we had earlier. I don't expect any surprises when we release this. It's been quite a while in the making, and I'm glad we can finally pull this in.

PR #2137, opened 2026-04-11
Allow merge bool v2

### LINKED ISSUE #2051: Unable to override bools in merge chain
When merging configs there is no way to override bools in the merge chain with `false`. If one item in the merge chain is `true` it will always stay `true`. This is a known limitation that we should ideally work around in the long term. The workaround is to convert all `bool`s of `Config` into `Option<bool>`. Then we can remove the `bool` case of the `merge!` macro and move the affected fields into the `option` case in the merge invocation.

Note that this is not a very pressing issue. It does not affect CLI options, as there is no way to set flags to false. (e.g. `lychee --dump=false` does not work)

Also [see the relevant discussion](https://github.com/lycheeverse/lychee/pull/2007#discussion_r2807428296).

TODO: add code snippet reference

cristiklein:
Fixes #2051 

Makes all boolean CLI flags mergeable by converting them to `Option<bool>`. For example, this allows `lychee.toml` to specify `accept_timeout = true` and calling `lychee --accept-timeout=false` with the combined effect being that timeouts are not accepted.

This PR uses the following overall patterns:

1. Boolean CLI flags are converted to `Option<bool>`.
2. Clap is configured with:
    ```
    #[arg(
        long,
        default_missing_value = "true", // `--flag` is equivalent to `--flag=true`
        num_args = 0..=1, // allow `--flag` with no value
        require_equals = true, // ensures `--flag parameter` is interpreted as `--flag=true parameter`
    )]
    ```
3. Field is made private and accessed via a getter to ensure all defaults are in `config/mod.rs`.

Decisions to be challenged:

1. The `#[arg(...)]` macro invokation looks tedious and repetative. I opted against factoring this into a macro, since the solution felt beyond my ability.
2. Should all boolean CLI flags be converted to `Option<bool>`, even e.g. `--dump`? I opted for this, since I felt this minimizes astonishment for the user.
3. Why not `--no-flag`, like ripgrep? This feels poorly supported in Clap. Furthermore, many CLI seems to move towards `--flag=false`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,8 +349,6 @@ Options:
       --accept-timeouts[=<false|true>]
           Accept timed out requests and return exit code 0 when encountering timeouts but not any other errors
 
-          [default: false]
-
       --archive <ARCHIVE>
           Web archive to use to provide suggestions for `--suggest`.
 
@@ -396,8 +394,6 @@ Options:
       --cache[=<false|true>]
           Use request cache stored on disk at `.lycheecache`
 
-          [default: false]
-
       --cache-exclude-status <CACHE_EXCLUDE_STATUS>
           A list of status codes that will be ignored from the cache
 
@@ -432,18 +428,12 @@ Options:
       --dump[=<false|true>]
           Don't perform any link checking. Instead, dump all the links extracted from inputs that would be checked
 
-          [default: false]
-
       --dump-inputs[=<false|true>]
           Don't perform any link extraction and checking. Instead, dump all input sources from which links would be collected
-
-          [default: false]
 
   -E, --exclude-all-private[=<false|true>]
           Exclude all private IPs from checking.
           Equivalent to `--exclude-private --exclude-link-local --exclude-loopback`
-
-          [default: false]
 
       --exclude <EXCLUDE>
           Exclude URLs and mail addresses from checking. The values are treated as regular expressions
@@ -454,20 +444,14 @@ Options:
       --exclude-link-local[=<false|true>]
           Exclude link-local IP address range from checking
 
-          [default: false]
-
       --exclude-loopback[=<false|true>]
           Exclude loopback IP address range and localhost from checking
-
-          [default: false]
 
       --exclude-path <EXCLUDE_PATH>
           Exclude paths from getting checked. The values are treated as regular expressions
 
       --exclude-private[=<false|true>]
           Exclude private IP address ranges from checking
-
-          [default: false]
 
       --extensions <EXTENSIONS>
           A list of file extensions. Files not matching the specified extensions are skipped.
@@ -530,8 +514,6 @@ Options:
       --glob-ignore-case[=<false|true>]
           Ignore case when expanding filesystem path glob inputs
 
-          [default: false]
-
   -h, --help
           Print help (see a summary with '-h')
 
@@ -547,8 +529,6 @@ Options:
 
       --hidden[=<false|true>]
           Do not skip hidden directories and files
-
-          [default: false]
 
       --host-concurrency <HOST_CONCURRENCY>
           Default maximum concurrent requests per host (default: 10)
@@ -577,12 +557,8 @@ Options:
       --host-stats[=<false|true>]
           Show per-host statistics at the end of the run
 
-          [default: false]
-
   -i, --insecure[=<false|true>]
           Proceed for server connections considered insecure (invalid TLS)
-
-          [default: false]
 
       --include <INCLUDE>
           URLs to check (supports regex). Has preference over all excludes
@@ -590,22 +566,14 @@ Options:
       --include-fragments[=<false|true>]
           Enable the checking of fragments in links
 
-          [default: false]
-
       --include-mail[=<false|true>]
           Also check email addresses
-
-          [default: false]
 
       --include-verbatim[=<false|true>]
           Find links in verbatim sections like `pre`- and `code` blocks
 
-          [default: false]
-
       --include-wikilinks[=<false|true>]
           Check WikiLinks in Markdown files, this requires specifying --base-url
-
-          [default: false]
 
       --index-files <INDEX_FILES>
           When checking locally, resolves directory links to a separate index file.
@@ -667,20 +635,14 @@ Options:
           Do not show progress bar.
           This is recommended for non-interactive shells (e.g. for continuous integration)
 
-          [default: false]
-
       --no-ignore[=<false|true>]
           Do not skip files that would otherwise be ignored by '.gitignore', '.ignore', or the global ignore file
-
-          [default: false]
 
   -o, --output <OUTPUT>
           Output file of status report
 
       --offline[=<false|true>]
           Only check local files and block network requests
-
-          [default: false]
 
   -p, --preprocess <COMMAND>
           Preprocess input files with the given command.
@@ -723,8 +685,6 @@ Options:
       --require-https[=<false|true>]
           When HTTPS is available, treat HTTP links as errors
 
-          [default: false]
-
       --root-dir <ROOT_DIR>
           Root directory to use when checking absolute links in local files. This option is
           required if absolute links appear in local files, otherwise those links will be
@@ -747,12 +707,8 @@ Options:
       --skip-missing[=<false|true>]
           Skip missing input files (default is to error if they don't exist)
 
-          [default: false]
-
       --suggest[=<false|true>]
           Suggest link replacements for broken links, using a web archive. The web archive can be specified with `--archive`
-
-          [default: false]
 
   -t, --timeout <TIMEOUT>
           Website timeout in seconds from connect to response finished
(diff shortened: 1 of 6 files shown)


thomas-zahner:
1. I agree, it's very verbose and introduces boilerplate code with the getters. Creating a attribute macro like `#[bool_arg]` unfortunately would require us to create a new crate as shown in [the docs](https://doc.rust-lang.org/reference/procedural-macros.html#the-proc_macro_attribute-attribute), I think this is a current language limitation. Clap could also be configured programmatically instead of using the macros but it probably makes no sense to combine the current attribute approach with a programmatic approach. (`LycheeOptions::command().arg(clap::Arg::new("some-bool-arg"))`) So it seems like there is no simple elegant solution.
2. I agree that we should treat all bool flags the same.
3. Hmm, from a user-perspective I actually like this PR. I think it's easier to have a single arg which can be set to true/false instead of having different args which affect the same boolean value.

@katrinafyi @mre What are your thoughts?

lychee-bin/src/config/mod.rs
```diff
@@ -819,28 +1049,11 @@ impl Config {
...
-                include_verbatim,
-                include_wikilinks,
-                insecure,
-                no_ignore,
-                no_progress,
-                offline,
-                require_https,
-                skip_missing,
-                suggest,
+                /*
```
thomas-zahner: I prefer to remove `bool` here and from the macro and remove this comment.

README.md
```diff
@@ -631,19 +663,25 @@ Options:
 
           [possible values: plain, color, emoji, task]
 
-  -n, --no-progress
+  -n, --no-progress[=<NO_PROGRESS>]
```
katrinafyi: I will point out that there are some arguments that already have "no" in front of them. This leads to a double negation when doing `--no-progress=false`, for example.

This is a little weird to read but I think it's perfectly fine to leave for now. Although it's weird, the behaviour is not surprising.

katrinafyi (approved):
Yeah I agree with @thomas-zahner on the discussion points. 

The PR looks reasonable, it's basically exactly what I expected for this change. Somewhat large, but fairly rote.

Thanks for tackling it!

lychee-bin/src/config/mod.rs
cristiklein: Fixed https://github.com/lycheeverse/lychee/pull/2137/commits/efb234fa2ad95b5551954d09ec136cf3e89803c9

README.md
cristiklein: :smile: Indeed, `--no-progress=false` is odd, but still understandable. Thanks for noticing.

cristiklein:
Thanks @katrinafyi and @thomas-zahner for your feedback. I believe I addressed all your comments. PTAL.

README.md
mre, line 1: The diff looks good. Just one question from my side: it's not possible to hide those `[=<SOMETHING>]` directives with clap, or? Because they are slightly noisy and don't add much value. As a user, I'd be wondering if the flags also work without the setting and what would be the default value if not set. (Before the change it's clear that setting the flag would enable the option, but now it's not so clear.)

Just my 2ct. Sorry if that was already discussed before. Flyby comment. 😅

mre (approved):

katrinafyi:
Re the big `#[arg]` macro attribute, I asked the question [in the Rust discord](https://discord.com/channels/273534239310479360/1120175689124036669/1493563356337148004) and got this answer (among other, much more complicated macro answers):

> I think you could define an extension trait on Arg and call that with arg() syntax

> I have no idea if it actually does but can't imagine why it wouldn't

If you want, it's worth a try @cristiklein. The idea is that the `#[arg]` macro just transforms itself into the builder API. For example, `#[arg(num_args = 1]` is transformed into [the builder syntax](https://docs.rs/clap/latest/clap/struct.Arg.html#method.num_args) as:
```
arg.num_args(1)
```
So, if you add an extension method on `Arg`, it might be able to be invoked the macro.

```rs
pub trait ArgExt  {
    fn optional_with_bool_value(self) -> Arg {
        // apply the needed options using the builder API
    }
}

// then later,
#[arg(optional_with_bool_value)]
```

README.md
thomas-zahner: Glancing at clap's source code I don't see a possibility to change this behaviour. Personally I think it's okay. This "help syntax" seems to be quite standard practice in help messages and man pages. The brackets indicate that the following argument is optional. Without it people wouldn't be able to tell that the value can be set with `=true`/`=false`. Without it this new "feature" remains undocumented.

cristiklein:
Great find @katrinafyi and works like a charm. See https://github.com/lycheeverse/lychee/pull/2137/commits/2521d071ad98007d00464924a99eaf04c331e3a5. Help message didn't change at all during refactoring.

@mre I found a few ways to beautify the help message. For example:

```
            .value_name("false|true") // Set the value name for help messages
            .hide_possible_values(true)
            .default_value("false")
```

Leads to:

```
      --suggest[=<false|true>]
          Suggest link replacements for broken links, using a web archive. The web archive can be specified with `--archive`
          
          [default: false]
```

See https://github.com/lycheeverse/lychee/pull/2137/commits/593100483aa6b00fc8cd1a23056232d6ebb140b5

cristiklein:
Arrrgggh, sorry, I pressed ENTER too fast. `.default_value("false")` has the unintended effect that missing flags become `false` and not `None`, which defies the whole point of this PR.

So the best I could obtain was:

```
            .value_name("false|true") // Set the value name for help messages
            .hide_possible_values(true)
```

Which leads to:

```
      --suggest[=<false|true>]
          Suggest link replacements for broken links, using a web archive. The web archive can be specified with `--archive`
```

katrinafyi:
> Great find @katrinafyi and works like a charm.

Incredible! Honestly I had no idea if it would work. Thanks for doing that, I think it's a bit cleaner :)

thomas-zahner:
@cristiklein @katrinafyi Thank you, that's quite a lot less boilerplate code and quite an improvement!

lychee-bin/src/config/mod.rs
```diff
@@ -55,6 +55,24 @@ occurrences take precedence over previous occurrences.
...
+trait ArgBoolOptionalExt {
+    fn optional_bool_flag(self) -> Self;
+}
+
+impl ArgBoolOptionalExt for Arg {
+    fn optional_bool_flag(self) -> Self {
+        self.default_missing_value("true") // Ensure `--flag` is treated as `--flag=true`
+            .num_args(0..=1) // Allow the flag to be specified with no value
+            .require_equals(true) // Ensure that `--flag value` is not misinterpreted as `--flag=true value`
+            .value_name("false|true") // Set the value name for help messages
```
thomas-zahner, line 71: The value name is originally probably not intended for "value" purposes. So it could be debated that the previous version was more consistent and that `hide_possible_values` should not be set to `true`.

But we also have `--header <HEADER:VALUE>` which already does exactly that. So the inconsistency was there already and I actually do like how this version is more concise. So I think it's fine.

(feel free to ignore this comment, just thinking out loud)

README.md
thomas-zahner: `--flag[=<false|true>]` while hiding ` [possible values: true, false]` (two additional lines) in help message and man pages as done in 593100483aa6b00fc8cd1a23056232d6ebb140b5 is a nice middle ground I think. It's probably the best we can do. [Lack of consistency could be debated](https://github.com/lycheeverse/lychee/pull/2137/changes/593100483aa6b00fc8cd1a23056232d6ebb140b5#r3080445446) but that's really nitpicky.

In general I feel like we're hitting capability limits of clap, as noticeable with this change, but also things like [documenting default values](https://github.com/lycheeverse/lychee/pull/2007). It's probably a reason why e.g. ripgrep switched from clap to their custom cli-arg parser implementation. As a result we now have some boilerplate code and conventions (e.g. `test_default_values`, getter/setters with private fields) but it's still well manageable.

Sorry for the digression..

lychee-bin/tests/cli.rs
```diff
@@ -4123,6 +3899,22 @@ https://lychee.cli.rs/guides/cli/#fragments-ignored
             .stdout(contains("resource-1.md").count(2));
     }
 
+    #[tokio::test]
+    async fn test_set_bool_to_false() {
```
thomas-zahner: One last thing. I feel like this test is overly complicated and slow (1 second) for what it's supposed to test and it's not very obvious that it's only about testing the `--flag=false` syntax.

thomas-zahner: @cristiklein Are you okay with c9bbc4ff83ccfa36aea1bfb597f68660794b2182?

Alternatively I'm okay with reverting the commit and fully removing the `test_set_bool_to_false` test. I think both tests are not hugely important.

cristiklein:
Love it! Thanks for catching that the test was slow.

lychee-bin/tests/cli.rs
cristiklein: Love it! Thanks for catching that the test was slow.

README.md
mre: Thanks for the thorough explanation. I agree.

PR #1600, opened 2024-12-25
Add support for Text fragment feature (#1545)

thiru-appitap:
Text Fragment feature implementation pull request. The feature follows the published URL Fragment Text Directives specification (https://wicg.github.io/scroll-to-text-fragment/). 

(base) lychee % lychee -vv --include-text-fragments  https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments
...
[DEBUG] tdirective: "text=From%20the%20foregoing%20remarks%20we%20may%20gather%20an%20idea%20of%20the%20importance
[DEBUG] status: Completed
[DEBUG] result: "From the foregoing remarks we may gather an idea of the importance"
[200] https://mdn.github.io/css-examples/target-text/index.html#:~:text=From%20the%20foregoing%20remarks%20we%20may%20gather%20an%20idea%20of%20the%20importance
[DEBUG] tdirective: "text=linked%20URL,-'s%20format"
[DEBUG] status: Completed
[DEBUG] result: "linked URL"
[DEBUG] tdirective: "text=Deprecated-,attributes,attribute"
[DEBUG] status: Completed
[DEBUG] result: "attributes     charset Deprecated   Hinted at the character encoding of the linked URL.   Note:This attribute"
[200] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#:~:text=linked%20URL,-'s%20format&text=Deprecated-,attributes,attribute
[DEBUG] tdirective: "text=downgrade:-,The%20Referer,be%20sent,-to%20origins"
[DEBUG] status: Completed
[DEBUG] result: "The Referer header will not be sent"
[200] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#:~:text=downgrade:-,The%20Referer,be%20sent,-to%20origins
[DEBUG] tdirective: "text=linked%20URL,defining%20a%20value"
[DEBUG] status: Completed
[DEBUG] result: "linked URL as a download. Can be used with or without a filename value:    Without a value, the browser will suggest a filename/extension, generated from various sources:   The Content-Disposition HTTP header  The final segment in the URL path  The media type (from the Content-Type header, the start of a data: URL, or Blob.type for a blob: URL)     filename: defining a value"
[200] https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#:~:text=linked%20URL,defining%20a%20value
...     

If the fragment directive is not found, a TextDirectiveNotFound error will be returned.

Below changes are completed:
[...]

PR #2138, opened 2026-04-12
Check text fragments

### LINKED ISSUE #1545: feature request: check text fragments
Text fragments are a [standardized way to link to webpage text without relying on HTML ID attributes](https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments#:~:text=without%20relying%20on%20the%20presence%20of%20IDs). They are [supported by all modern browsers](https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments#browser_compatibility). I'd like to use lychee to make sure my text fragment links are valid!

Example:
```html
<!-- file.html -->
<html>
  <body>
    <div>
      <a href="https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments#:~:text=without%20relying%20on%20the%20presence%20of%20IDs">
        valid
      </a>
      <a href="https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments#:~:text=asdfawefaweafa">
        invalid
      </a>
    </div>
  </body>
</html>
```
```sh
; lychee --version
# lychee 0.16.1
; lychee -vv --include-fragments ./file.html
#     [200] https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments#:~:text=without%20relying%20on%20the%20presence%20of%20IDs
#     [200] https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments#:~:text=asdfawefaweafa
```

Desired behavior: if lychee is configured to check fragments and is passed a `#:~:text=...` fragment, lychee should throw an error if the text is not found within the linked document.

cristiklein:
Fixes #1545 

This PR adds `--include-text-fragments` which checks [text fragments](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments). `--include-fragments`'s behavior is changed to only check the anchor part.

Hence, for a fragment like `#anchor:~:text=some-text`, the user may choose between three behaviors:

1. `--include-fragments` checks only if `#anchor` is a valid anchor.
2. `--include-text-fragments` checks only if `some-text` is a valid text.
3. `--include-fragments --include-text-fragments` checks both.

Decisions to be challenged:

1. The text fragment checker uses only `html5ever`. I found no good reason to also include an implementation based on `html5gum`.
2. The text fragment checker extracts all visible text, then checks the text fragment. This is rather wasteful on memory compared to a state-machine-based approach. I opted for the more memory-intense approach, both due to simplicity and because we might want to support [multiple text fragments](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments#urls_with_multiple_text_fragments) in the future. (If that ever gets requested.)
3. Extracting visible text assume all tags contain visible text, except `head`, `script`, `style` and `template`. I believe this is a good start, in particular `head` also filters out `title` and `meta`. However, I see this more like a starting point, than an academically correct response. I'm hoping for more user feedback here.


cristiklein:
@SKalt wanna take this for a spin?

thomas-zahner:
Wow, thank you for this nice PR! :rocket: 
It looks great already. Just last week I've started to rework #1600 via https://github.com/thomas-zahner/lychee/tree/text-fragment-v2 but now with this great PR it probably makes sense to continue here. (I've also considered to do a full rewrite instead of using #1600 as a starting point)

1. Ideally we should respect the `use_html5ever` option/flag to align with the other parsing behaviour.
2. This indeed sounds quite wasteful. I will take a closer look and will think about how this can be improved.
3. This sounds sensible.

One more thing that could be debated is if `--include-text-fragments` makes sense. In my rework on #1600 I've integrated the feature into `--include-fragments`, after all text fragments are fragments. Do you see a use-case in checking text fragments but not checking any other fragments? I'm currently in favour of a single flag.

katrinafyi:
to make it faster, something like an inverted index: https://swtch.com/~rsc/regexp/regexp4.html. then, to search for a string, you can just go through each word inside the search string, in turn, and look at its indices and check they're contiguous.

but I don't know if we'll see text fragments often enough for preprocessing to make sense. if they're rare, matching on the fly as you go through the html cursor could be fine. 

I also agree that the two include fragment options should be folded together.

cristiklein:
@thomas-zahner and @katrinafyi Thank you for your feedback!

> Ideally we should respect the use_html5ever option/flag to align with the other parsing behaviour.

Do I understand correctly that you want to also see an implementation based on `html5gun` before merging this PR?

> but I don't know if we'll see text fragments often enough for preprocessing to make sense. if they're rare, matching on the fly as you go through the html cursor could be fine.

This sounds like a fascinating idea, but I agree that text anchor might be too rare for this to make sense.

> I also agree that the two include fragment options should be folded together.

Can I challenge this a bit? In my experience, text fragments are less stable than anchor fragments. I could imagine users wanting to perform anchor fragment checking (as they are more stable and can be seen as an "API" of the webpage), but not text fragment checking (as they might be less stable and are definitely not an "API" of the webpage).

Not sure this is a good enough reason to separate the two checks, but I felt a need to point it out. :smile: I'm fine with whatever is decided.

katrinafyi:
> In my experience, text fragments are less stable than anchor fragments.

True, that's a really good point, thanks for bringing it up. To be honest, I've never used text fragments myself. 

How about we combine it with --include-fragments but make it take an argument string that's none, anchor, text, or all? I think that's possible with clap and I think joining them together makes sense and makes the features more discoverable. You should also be able to write --include-fragments without a value and it would default to.. "anchor", probably.

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+    if !status.is_success() || file_type != FileType::Html {
+        return status;
+    }
+
+    let directives = parse_text_directives(url);
+    if directives.is_empty() {
+        return status;
+    }
+
+    let document = normalize_whitespace(&extract_visible_text(content));
```
katrinafyi: it seems unnecessarily slow to split then join the entire document. can we just avoid emitting redundant whitespace in the callback?

alternatively, the document can emit into a Vec of whitespace-separated words and you just have to split the fragment in the same way. this would avoid all the normalising atm

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+
+        let suffix = if parts.last().is_some_and(|part| part.starts_with('-')) && parts.len() >= 2 {
+            let suffix = parts.pop().expect("checked length above");
+            Some(suffix[1..].to_owned())
+        } else {
+            None
+        };
+
+        let [start] = parts.as_slice() else {
+            let [start, end] = parts.as_slice() else {
```
katrinafyi: chained let/else can just be a match statement, I believe.

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+        .collect()
+}
+
+fn normalize_whitespace(input: &str) -> String {
+    input.split_whitespace().collect::<Vec<_>>().join(" ")
+}
+
+fn extract_visible_text(input: &str) -> String {
+    #[derive(Default)]
+    struct TextExtractor {
```
katrinafyi: I think the inner structure is too big. also, I think using default for an initial value is a misuse, just use pub fn new() :)

this feature is kinda interesting outside of lychee, so if possible it would be nice to clean up the API a little bit (even if we don't export it)

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+            &mut self,
+            event: CallbackEvent<'_>,
+            _span: html5gum::Span<usize>,
+        ) -> Option<()> {
+            match event {
+                CallbackEvent::OpenStartTag { name } => {
+                    let tag = String::from_utf8_lossy(name).into_owned();
+                    if TextExtractor::is_hidden_tag(&tag) {
+                        self.hidden_stack.push(tag);
+                    }
```
katrinafyi: I was looking at [the spec](https://wicg.github.io/scroll-to-text-fragment/#example-73638554) and [a JS implementation](https://github.com/GoogleChromeLabs/text-fragments-polyfill/blob/abc6ed408b3f20e91d9cbda9977748459f5e3877/src/text-fragment-utils.js#L30) and a "block" tag should terminate (prevent) any matches across its boundary. this is kinda tricky, so i'm fine with the lychee mplementation accepting too much. I just thought it was worth mentioning and maybe deserves a comment in the code.

in general, Id like to see some refrences to the spec in the comments :) is the matching algorithm based on some reference implementation?

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    let mut extractor = TextExtractor::default();
+    let emitter = CallbackEmitter::new(&mut extractor);
+    let _: Result<Vec<_>, _> = Tokenizer::new_with_emitter(input, emitter).collect();
+    extractor.text
+}
+
+#[cfg(test)]
+mod tests {
```
katrinafyi, line 152: you seem to support it in your code, so it would be nice to have a test where the start and end could be overlapping:

start=aa, end=aa should match in aaaa but not aaa or aa

also, where the start is part of a word

start=aa, end=b should match in aab, aaab.

actually, does the spec say anything about whether a fragment match can start in the middle of a word?

thomas-zahner:
> In my experience, text fragments are less stable than anchor fragments.

Text fragments are also [supposed to get/become an API](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments#api) at some point. But it's true that it's a newer feature and maybe still not fully mature.

> How about we combine it with --include-fragments but make it take an argument string that's none, anchor, text, or all?

Yes, that would be a middle ground. Something like the following?

```rust
enum FragmentMode {
  /// Don't do any fragment checking
  #[default]
  None,
  /// Check text anchors but not text fragments
  Simple,
  /// Check both text anchors and text fragments
  Full,
}
```

I'm not really sure/happy with the above variant names, but `AnchorsWithTextFragments` & `AnchorsWithoutTextFragments` is not really any better. Admittedly, I still feel like the better approach is to make text fragment checking part of `include_fragments`/`--include-fragments` while keeping this a boolean type. I keeps UX simpler & doesn't require additional logic. Also we don't really have to fear introducing this as part of include_fragments in terms of stability/correctness as currently lychee reports an error on any text fragment so we would greatly improve correctness even if the implementation might not be fully conformant yet.

lychee-lib/src/checker/website.rs
```diff
@@ -61,6 +73,33 @@ pub(crate) struct WebsiteChecker {
...
 impl WebsiteChecker {
+    fn parse_fragment(url: &Url) -> ParsedFragment<'_> {
+        let Some(fragment) = url.fragment() else {
+            return ParsedFragment {
+                anchor_fragment: None,
+                text_directive: None,
+            };
+        };
+
+        let Some((element, text_directive)) = fragment.split_once(":~:text=") else {
```
katrinafyi: duplicated with code in text_fragment.rs?

also I'd use "anchor" instead of element, i makes me think of html elements

lychee-lib/src/checker/website.rs
```diff
@@ -47,6 +54,11 @@ pub(crate) struct WebsiteChecker {
     /// Will be disabled if the request method is `HEAD`.
     include_fragments: bool,
 
+    /// Whether to check the existence of text fragments in the response HTML files.
+    ///
+    /// Will be disabled if the request method is `HEAD`.
+    include_text_fragments: bool,
```
katrinafyi: this is only in website checker ATM. as far as I can tell, it should be in FileChecker too. is there any way of doing this nicely?

honestly, fragment_checker should be be an Option field rather than carrying the adjacent include_fragments flag, and maybe the text fragments can be hooked into FragmentChecker somehow

cristiklein:
> How about we combine it with --include-fragments but make it take an argument string that's none, anchor, text, or all? I think that's possible with clap and I think joining them together makes sense and makes the features more discoverable. You should also be able to write --include-fragments without a value and it would default to.. "anchor", probably.

I love this! I'll produce a PR as soon as I have time.

> Text fragments are also [supposed to get/become an API](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments#api) at some point. But it's true that it's a newer feature and maybe still not fully mature.

Sorry for not being clearer. What I meant with the "API" analogy is the following: As a website author, if I want to encourage people to link to specific parts of a page, I'd ask them to use anchors. I'd do my best to keep anchor unchanged, even if the section title has changed. So website authors are very likely to include anchors in their stable interface promise.

In contrast, as a website author, I would definitely not want people to expect me to keep text stable, so that their text fragments don't break.

lychee-lib/src/checker/website.rs
cristiklein: > duplicated with code in text_fragment.rs?

Not really. This code only splits anchor from text fragment. The code in `text_fragment.rs` actually parses the text fragment into prefix, start, stop and suffix. I thought to keep this separation, as fully parsing the text directive at this point is not needed.

> also I'd use "anchor" instead of element, i makes me think of html elements

Fully agree, nice catch!

mre:
> Do I understand correctly that you want to also see an implementation based on html5gun before merging this PR?

don't know how hard it is to add html5gum support, but I would personally very much appreciate that. The reason is that we'd like to maintain feature parity between both engines. They each have their advantages and disadvantages. html5gum is faster and allows more direct/raw HTML access, while html5ever is very mature and provides some helpful high-level functionality.

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+            });
+        };
+
+        Some(Self {
+            prefix: prefix.map(|s| normalize_whitespace(&s)),
+            start: normalize_whitespace(start),
+            end: None,
+            suffix: suffix.map(|s| normalize_whitespace(&s)),
+        })
+    }
```
mre: I personally found this a bit harder to read than it could be.

Tried my hand on a refactoring and added a few comments. Also turned the separator and markers into constants. I feel like some examples are also helpful. 

```rust
/// Parses a text directive value (the part after `text=`) into its components.
///
/// The format is: `[prefix-,] start [,end] [,-suffix]`
///
/// # Examples
///
/// ```
/// // Exact match
/// let d = TextDirective::parse("an example").unwrap();
/// assert_eq!(d.start, "an example");
/// assert!(d.prefix.is_none() && d.end.is_none() && d.suffix.is_none());
///
/// // Range match
/// let d = TextDirective::parse("an example,text fragment").unwrap();
/// assert_eq!(d.start, "an example");
/// assert_eq!(d.end.unwrap(), "text fragment");
///
/// // Prefix and suffix context terms
/// let d = TextDirective::parse("this is-,an example,-text fragment").unwrap();
/// assert_eq!(d.prefix.unwrap(), "this is");
/// assert_eq!(d.start, "an example");
/// assert_eq!(d.suffix.unwrap(), "text fragment");
///
/// // Invalid: no start term
/// assert!(TextDirective::parse("").is_none());
/// ```
fn parse(input: &str) -> Option<Self> {
    // The parameter separator splits the directive into its components:
    // [prefix-,] start [,end] [,-suffix]
    const PARAMETER_SEPARATOR: char = ',';

    // The context term marker distinguishes prefix/suffix from start/end terms.
    // e.g. "prefix-,start,-suffix"
    const CONTEXT_TERM_MARKER: char = '-';

    // Literal commas in the text are percent-encoded upstream, so splitting on ',' is safe.
    let mut parts: Vec<&str> = input.split(PARAMETER_SEPARATOR).collect();
    if parts.is_empty() {
        return None;
    }

    // A prefix is indicated by a trailing context term marker on the first token
    let prefix = if parts.first().is_some_and(|part| part.ends_with(CONTEXT_TERM_MARKER)) && parts.len() >= 2 {
        let prefix = parts.remove(0);
        Some(prefix.strip_suffix(CONTEXT_TERM_MARKER).unwrap_or(prefix).to_owned())
    } else {
        None
    };

    // A suffix is indicated by a leading context term marker on the last token
    let suffix = if parts.last().is_some_and(|part| part.starts_with(CONTEXT_TERM_MARKER)) && parts.len() >= 2 {
        let suffix = parts.pop()?;
        Some(suffix.strip_prefix(CONTEXT_TERM_MARKER).unwrap_or(suffix).to_owned())
    } else {
        None
    };

    // After removing prefix/suffix, exactly 1 or 2 tokens must remain (start [,end])
    let (start, end) = match parts.as_slice() {
        [start] => (start, None),
        [start, end] => (start, Some(normalize_whitespace(end))),
        _ => return None,
    };

    Some(Self {
        prefix: prefix.map(|s| normalize_whitespace(&s)),
        start: normalize_whitespace(start),
        end,
        suffix: suffix.map(|s| normalize_whitespace(&s)),
    })
}
```

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+            if prefix_matches && suffix_matches {
+                return true;
+            }
+
+            start_offset = match_start + 1;
+        }
+
+        false
+    }
+}
```
mre, line 71: I think there's a bug. When `end` is not found, it immediately returns `false`, but it should `continue` to the next candidate match of `start` instead, since a later occurrence of `start` might have a valid `end` after it.

```rust
fn matches(&self, document: &str) -> bool {
    let mut start_offset = 0;

    // Find each occurrence of `start` and check whether the surrounding
    // context (prefix, end, suffix) also matches, per the spec's algorithm.
    while let Some(relative_start) = document[start_offset..].find(&self.start) {
        let match_start = start_offset + relative_start;
        let mut match_end = match_start + self.start.len();

        // For range matches, find the first occurrence of `end` after `start`
        if let Some(end) = &self.end {
            let Some(relative_end) = document[match_end..].find(end.as_str()) else {
                // No `end` found after this `start`; try the next occurrence of `start`
                start_offset = match_start + 1;
                continue;
            };
            match_end += relative_end + end.len();
        }

        let prefix_matches = self
            .prefix
            .as_ref()
            .is_none_or(|prefix| document[..match_start].trim_end().ends_with(prefix.as_str()));
        let suffix_matches = self
            .suffix
            .as_ref()
            .is_none_or(|suffix| document[match_end..].trim_start().starts_with(suffix.as_str()));

        if prefix_matches && suffix_matches {
            return true;
        }

        start_offset = match_start + 1;
    }

    false
}
```

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+    };
+    let Some((_, directive)) = fragment.split_once(":~:") else {
+        return Vec::new();
+    };
+
+    url::form_urlencoded::parse(directive.as_bytes())
+        .filter(|(key, _)| key == "text")
+        .filter_map(|(_, value)| TextDirective::parse(value.as_ref()))
+        .collect()
+}
```
mre: Can we make `:~:` a constant? Maybe also the text directive type?

```rust
fn parse_text_directives(url: &Url) -> Vec<TextDirective> {
    // The fragment directive delimiter separates the element fragment from the directive.
    // e.g. "https://example.com#section:~:text=foo"
    const FRAGMENT_DIRECTIVE_DELIMITER: &str = ":~:";

    // The only directive type defined by the spec
    const TEXT_DIRECTIVE_PREFIX: &str = "text";

    let Some(fragment) = url.fragment() else {
        return Vec::new();
    };
    let Some((_, directive)) = fragment.split_once(FRAGMENT_DIRECTIVE_DELIMITER) else {
        return Vec::new();
    };

    url::form_urlencoded::parse(directive.as_bytes())
        .filter(|(key, _)| key == TEXT_DIRECTIVE_PREFIX)
        .filter_map(|(_, value)| TextDirective::parse(value.as_ref()))
        .collect()
}
```

Of course, we could move all constants to the beginning of the file to group them together, but I find reading code with meaningful constants easier in general.

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,298 @@
...
+
+fn normalize_whitespace(input: &str) -> String {
+    input.split_whitespace().collect::<Vec<_>>().join(" ")
```
mre: [Soon](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.intersperse) 😉 (This would avoid extra allocations if only it was stable already.)

lychee-lib/src/checker/website.rs
```diff
@@ -61,6 +73,33 @@ pub(crate) struct WebsiteChecker {
 }
 
 impl WebsiteChecker {
+    fn parse_fragment(url: &Url) -> ParsedFragment<'_> {
+        let Some(fragment) = url.fragment() else {
+            return ParsedFragment {
+                anchor_fragment: None,
+                text_directive: None,
+            };
```
mre: Could be slightly cleaner if ParsedFragment implemented `Default`:

```rust
ParsedFragment::default()
```

lychee-lib/src/checker/website.rs
mre: I like constants. 😉 

```rust
pub(crate) const TEXT_DIRECTIVE_SEPARATOR: &str = ":~:text=";
```

lychee-lib/src/checker/website.rs
```diff
@@ -120,22 +161,25 @@ impl WebsiteChecker {
     async fn check_default(&self, request: Request) -> Status {
         let method = request.method().clone();
         let request_url = request.url().clone();
+        let parsed_fragment = Self::parse_fragment(&request_url);
```
mre: This should technically be `parsed_fragments`. I was confused for a moment while reading this. Thought it returned a single value and then you did `parsed_fragment.anchor_fragment`. The plural would make it clearer what's going on.

lychee-lib/src/checker/website.rs
```diff
@@ -120,22 +161,25 @@ impl WebsiteChecker {
...
+            Self::with_element_fragment(&request_url, parsed_fragment.anchor_fragment);
 
-        let check_request_fragments = self.include_fragments
+        let check_anchor_fragments = self.include_fragments
             && method == Method::GET
-            && request_url.fragment().is_some_and(|x| !x.is_empty());
+            && parsed_fragment.anchor_fragment.is_some();
+        let check_text_fragments = self.include_text_fragments
+            && method == Method::GET
+            && parsed_fragment.text_directive.is_some();
```
mre: The wording is a little off. Here you use `text_directive` and two lines above it's `include_text_fragments`. Can we align that naming? E.g. maybe `text_fragment`?
But I could also see how the naming is just off because of the spec and lychee's cli params. Just wanted to raise that concern.

lychee-lib/src/checker/website.rs
```diff
@@ -120,22 +161,25 @@ impl WebsiteChecker {
...
+        let needs_body = check_anchor_fragments || check_text_fragments;
```
mre: How about?

```suggestion
        let fetch_body = check_anchor_fragments || check_text_fragments;
```

cristiklein:
@mre and @katrinafyi Thanks tons for reviewing this PR.

I extracted the following high-level comments:

- [x] Make lychee accept `--include-fragments[=<none|anchor-only|text-only|all>`]
- [x] Add an `html5gum` implementation
- [x] Check text fragments in `FileChecker` too.
- [x] Somehow avoid splitting and recombining the whole document (or documenting why it's really necessary)
- [x] Add comments to link implementation to spec
- [ ] Add tests for overlapping start and end
- [x] Revisit the matching code, make it more readable by adding constants

I'll keep you posted. :smile:

lychee-lib/src/checker/website.rs
```diff
@@ -152,25 +120,21 @@ impl WebsiteChecker {
...
-
-        let check_anchor_fragments = self.fragment_checker_options.check_anchor_fragments
+        let check_request_fragments = self.fragment_checker_options.any_enabled()
             && method == Method::GET
-            && parsed_fragment.anchor_fragment.is_some();
-        let check_text_fragments = self.fragment_checker_options.check_text_fragments
-            && method == Method::GET
-            && parsed_fragment.text_directive.is_some();
-        let fetch_body = check_anchor_fragments || check_text_fragments;
+            && request_url.fragment().is_some_and(|x| !x.is_empty());
```
mre, line 126: Do we currently allow empty fragments? That's confusing. I might misunderstand, but I would expect `Some("")` to instead rather be a `None`. Maybe we can add that check when creating the fragment.

(Don't think that was a consequence of the refactor, but it could be a thing we could fix while we're at it.)

lychee-lib/src/utils/fragment_checker/mod.rs
```diff
@@ -34,6 +36,59 @@ impl FragmentInput<'_> {
...
+struct ParsedFragment<'a> {
+    element_id: Option<&'a str>,
+    text_directive_value: Option<String>,
+}
+
+const FRAGMENT_DIRECTIVE_DELIMITER: &str = ":~:";
+const TEXT_DIRECTIVE_KEY: &str = "text";
+
+impl<'a> ParsedFragment<'a> {
+    /// This method does top-level parsing of the fragment, separating the element id (if any) from the text directive (if any).
```
mre: Can you wrap that comment?

lychee-lib/src/checker/website.rs
cristiklein: @mre If I understand this correctly:

- `http://localhost:8080/` -> `fragment()` is `None`, so from lychee's PoV the fragment is valid.
- `http://localhost:8080/#` -> `fragment()` is `Some("")`, which means "scroll to top", so from lychee's PoV the fragment is valid.

I believe the intention of the code was to act like a short-circuit for "scroll to top" fragments and is tested [here](https://github.com/lycheeverse/lychee/blob/master/fixtures/fragments/file1.md#to-top-fragments).

mre: Yeah, probably. Shall we add a comment that this is expected behavior?

mre:
Some more lints to fix (which are unrelated to this specific change, it seems). Looks good from my side.

cristiklein:
Hi folks,

Please find below a short status update on the PR:

Status | Description | Explanation
-- | -- | --
✅ | Make lychee accept --include-fragments[=<none\|anchor-only\|text-only\|all>] | Done
✅ | Add an html5gum implementation | I confused html5ever with html5gum. My initial implementation was already with html5gum. 😄 FragmentChecker only has an html5gum implementation, so the text fragment part follows this pattern.
✅ | Check text fragments in FileChecker too | Done as a side-effect of moving text fragment checking into FragmentChecker. Also added a test, just to make sure.
🟡 | Somehow avoid splitting and recombining the whole document (or documenting why it's really necessary) | I believe this can be optimized to avoid the splitting and recombining, but I didn't have time to do that yet.
✅ | Add comments to link implementation to spec | Done.
⛔ | Add tests for overlapping start and end | As far as I could test, Chrome does not generate such text fragments.
✅ | Revisit the matching code, make it more readable by adding constants | Done.

To sum up, the PR is in a mergeable state and ready to receive user feedback. Let me know if you want the optimization to be part of this PR or of a future PR. At any rate, I believe the wiring and tests are good enough to make this a very localized change.

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,213 @@
...
+const TEXT_DIRECTIVE_KEY: &str = "text";
+
+impl<'a> ParsedFragment<'a> {
+    /// This method parses the fragment, separating the element id (if any) from the text directives (if any).
+    pub(super) fn parse(url: &'a Url) -> Self {
+        let Some(fragment) = url.fragment() else {
+            return Self {
+                element_id: None,
+                text_directives: Vec::new(),
+            };
```
mre: Nit: Still think we should implement `Default` for `ParsedFragment` and `return Self::default()` here. 😉 
Especially since we use the same code again a bit further down. This way, we can keep the "default return value" paths in sync.

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,213 @@
...
+        let Some((element_id, fragment_directive)) =
+            fragment.split_once(FRAGMENT_DIRECTIVE_DELIMITER)
+        else {
+            return Self {
+                element_id: Some(fragment),
+                text_directives: Vec::new(),
+            };
+        };
+
+        let element_id = (!element_id.is_empty()).then_some(element_id);
```
mre, line 50: This line is hard to understand. (Might just be me, though.) Can you add a comment how it works and why it's needed?

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        Self {
+            element_id,
+            text_directives,
+        }
+    }
+}
+
+impl TextDirective {
+    fn parse(input: &str) -> Option<Self> {
```
mre, line 104: I know that it is an internal function, but can we add a doc-comment?

```suggestion
    /// Parse a [text directive] from a string.
    ///
    /// Text directives follow the format:
    ///
    /// ```text
    /// prefix-,start,end,-suffix
    /// ```
    ///
    /// where `prefix-` and `-suffix` are optional context anchors and `end` is an
    /// optional range end. Only `start` is required.
    ///
    /// Returns `None` if the input is empty or contains more than two parts
    /// after stripping the optional prefix and suffix anchors.
    ///
    /// [text directive]: https://wicg.github.io/scroll-to-text-fragment/#the-text-directive
    fn parse(input: &str) -> Option<Self> {
```

Writing the comment made me notice a subtle little wrinkle in the code. (See next comment.)

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,213 @@
...
+        let mut parts: Vec<&str> = input.split(',').collect();
+        if parts.is_empty() {
```
mre: This isn't quite right. [split(',') on an empty string yields [""], not []](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=dcc77bd784deb7ec4478e439145554d1), so parts.is_empty() is actually never true. The real failure condition is having 0 or 3+ parts after stripping prefix/suffix, not an empty input.

So this is dead code: the parts.is_empty() guard is unreachable and should probably be removed or replaced with a check on the trimmed/split result.

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,213 @@
...
+        } else {
+            None
+        };
+
+        let suffix = if parts.last().is_some_and(|part| part.starts_with('-')) && parts.len() >= 2 {
+            let suffix = parts.pop().expect("checked length above");
+            Some(normalize_whitespace(&suffix[1..]))
+        } else {
+            None
+        };
```
mre: Just an idea: how about we move that to separate helper functions?

```rust
fn strip_prefix_anchor(parts: &mut Vec<&str>) -> Option<String> {
    if parts.len() >= 2 && parts.first().is_some_and(|p| p.ends_with('-')) {
        let part = parts.remove(0);
        Some(normalize_whitespace(&part[..part.len() - 1]))
    } else {
        None
    }
}

fn strip_suffix_anchor(parts: &mut Vec<&str>) -> Option<String> {
    if parts.len() >= 2 && parts.last().is_some_and(|p| p.starts_with('-')) {
        let part = parts.pop().expect("checked length above");
        Some(normalize_whitespace(&part[1..]))
    } else {
        None
    }
}
```

It looks a bit repetitive, but then this function just becomes

```rust
// ...
let mut parts: Vec<&str> = input.split(',').collect();
let prefix = strip_prefix_anchor(&mut parts);
let suffix = strip_suffix_anchor(&mut parts);
// ...

And also, the prefix/suffix functions can be tested individually with, say `rstest`.

File: lychee-lib/src/utils/fragment_checker/parsed_fragment.rs [hunk 23]
```diff
@@ -0,0 +1,213 @@
...
+            });
+        };
+
+        Some(Self {
+            prefix,
+            start: normalize_whitespace(start),
+            end: None,
+            suffix,
+        })
+    }
```
mre: How about

```rust
let (start, end) = match parts.as_slice() {
    [start] => (start, None),
    [start, end] => (start, Some(normalize_whitespace(end))),
    _ => return None,
};

Some(Self {
    prefix,
    start: normalize_whitespace(start),
    end,
    suffix,
})
```

I find this cleaner. ;)

Alternatively

```rust
let (start, end) = match parts.as_slice() {
    [start] => (normalize_whitespace(start), None),
    [start, end] => (normalize_whitespace(start), Some(normalize_whitespace(end))),
    _ => return None,
};

Some(Self {
    prefix,
    start,
    end,
    suffix,
})
```

mre:
Thanks for the overview. Very nicely done.
I've added a few more thoughts, but it is already in a great shape, so feel free to ignore. There are also some open comments by @katrinafyi as well.

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/commits/1fbdfd4f954518caefc3366ee69a9b34c29475f1

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: Fixed  in https://github.com/lycheeverse/lychee/pull/2138/commits/ef333e4479b6e69cd363c1f22b5f0e71bb795c6c

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: I'm not sure to fully understand your comment.

~Could you explain why `new()` is better than `default()` in this case?~ Nevermind, clippy explained it to me. :smile: Fixed in https://github.com/lycheeverse/lychee/pull/2138/commits/d68c43eb0819ea0f9c5a68b4069d8d786fc73f3e

Also, to me `extract_visible_text` feels like a clean API (with TextExtractor being an internal implementation detail). Did you mean to clean up the API so that in the future TextExtractor could be used outside of lychee?

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: Fixed in many places, e.g., https://github.com/lycheeverse/lychee/pull/2138/commits/0f21d09a7731be786a80f4690ba59f500d0e8e9e

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: I checked and I couldn't get Chrome to generate `start=aa,end=aa` because that would be rather ambiguous. It seems to prefer something like `prefix-,aaaa,-suffix`.

Regarding word boundaries, seems like the proposal wants that: https://wicg.github.io/scroll-to-text-fragment/#word-boundaries

What exactly constitutes a word boundary seems to be ... complicated (see link above). I need to think about what lychee could/should do to check if the text fragment is aligned with word boundaries.

lychee-lib/src/checker/website.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/changes/83531d9ce68552b887dc28ceee0c8ba54d6d4930

lychee-lib/src/checker/website.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/changes/239b0ace01d8f03e345190920f19b01071b3184b

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/commits/1fbdfd4f954518caefc3366ee69a9b34c29475f1

lychee-lib/src/checker/website.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/commits/fb2ecb9c62e151f57581f0891b748d905a8f782e

lychee-lib/src/checker/website.rs
cristiklein: Fixed

lychee-lib/src/checker/website.rs
cristiklein: Outdated, I moved fragment parsing into FragmentChecker.

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed https://github.com/lycheeverse/lychee/commit/fb2ecb9c62e151f57581f0891b748d905a8f782e

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/commits/d5d5e2624f4e84708beee640064770f48408da00

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/changes/ef333e4479b6e69cd363c1f22b5f0e71bb795c6c

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/changes/ef333e4479b6e69cd363c1f22b5f0e71bb795c6c

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/changes/ef333e4479b6e69cd363c1f22b5f0e71bb795c6c

cristiklein:
@thomas-zahner @mre @katrinafyi Thanks a lot for the time you put into reviewing this PR. This ended up being more complex than I initially expected. :smile:

I believe I addressed all your comments (but I was wrong before :smile:).

I'm unsure why CI / test fails. It works alright on my machine:

```console
$ git log -1
commit 961515d50cf8f2c4d3f0aa7c33ca4360aabed942 (HEAD -> check-text-fragments, cristiklein/check-text-fragments)
Author: Cristian Klein <cristian.klein@elastisys.com>
Date:   Fri Apr 17 22:32:05 2026 +0200

    Force CI re-run
$ make test
cargo nextest run --all-targets --all-features
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.17s
────────────
 Nextest run ID 7ba43ba1-257c-42bf-b196-3b5677913228 with nextest profile: default
    Starting 699 tests across 16 binaries (3 tests skipped)
[...]
     Summary [  20.305s] 699 tests run: 698 passed, 1 failed, 3 skipped
        FAIL [  20.112s] lychee::cli cli::test_email
```

(Note that cli::test_email does not work at my place, due to SMTP being blocked by my ISP.)

Let me know if there is anything I missed.

lychee-lib/src/checker/website.rs
mre: Oh, is there a reason why the Default trait can't be used?

https://doc.rust-lang.org/std/default/trait.Default.html

mre:
Don't worry about the failing test. The wayback machine test is flaky. We have to tackle that at some point.

lychee-lib/src/utils/fragment_checker/text.rs
katrinafyi: Yeah interesting, I'm happy to leave this for now. Thanks for looking into it!

lychee-bin/src/config/mod.rs
```diff
@@ -797,8 +822,22 @@ impl Config {
...
+            Some(FragmentMode::TextOnly) => FragmentCheckerOptions {
+                check_anchor_fragments: false,
+                check_text_fragments: true,
+            },
+            Some(FragmentMode::Full) => FragmentCheckerOptions {
+                check_anchor_fragments: true,
+                check_text_fragments: true,
+            },
+            Some(FragmentMode::None) | None => FragmentCheckerOptions::default(),
+        }
```
katrinafyi: Can the bulk of this logic be moved into into a `From<FragmentMode>` for `FragmentCheckerOptions`? This function would just need to handle the `None` case, which lines up with the pattern for other default values.

lychee-bin/src/config/mod.rs
```diff
@@ -479,9 +490,23 @@ pub(crate) struct Config {
     accept_timeouts: Option<bool>,
 
     /// Enable the checking of fragments in links.
-    #[arg(long, optional_bool_flag())]
+    ///
+    /// Use `none` to disable fragment checks,
+    /// `anchor-only` for anchor fragments like `#section`,
+    /// `text-only` for text fragments like `#:~:text=example`,
+    /// or `full` to check both.
```
katrinafyi: a paragraph should flow to the end of the line. or, if you prefer to use line breaks, you can turn it into a proper bullet list.
```suggestion
    /// Use `none` to disable fragment checks, `anchor-only` for anchor fragments
    /// like `#section`, `text-only` for text fragments like `#:~:text=example`,
    /// or `full` to check both.
```

lychee-lib/src/utils/fragment_checker/text.rs
katrinafyi: I'd just move TextExtractor outside of the function. I'm always surprised by inner structs and its impl is fairly big. You can keep it private and it won't be accessible outside the file. 

Thanks for adding the `new()`!

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,283 @@
...
+
+/// Helper function to strip a suffix from the end of the text directive parts, if present and correctly formatted.
+fn strip_suffix(parts: &mut Vec<&str>) -> Option<String> {
+    if parts.len() >= 2 && parts.last().is_some_and(|p| p.starts_with('-')) {
+        let part = parts.pop().expect("checked length above");
+        Some(percentage_decode(&part[1..]))
+    } else {
+        None
+    }
+}
```
katrinafyi: These functions have names which are too generic. I'd move them inside `impl TextDirective` or rename them to be more obviously associated with text directives.

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    if directives.is_empty() {
+        return true;
+    }
+
+    // The algorithm to find a range in a document likely requires a full implementation of a browser.
+    // See https://wicg.github.io/scroll-to-text-fragment/#finding-ranges-in-a-document
+    // Here, we try to approximate it by extracting visible text.
+    // This ensures that `Hell<i>o</i> <strong>world</strong>` is matched.
+    let document = extract_visible_text(content);
```
katrinafyi, line 29: I'm not familiar with text fragments, but I know that HTML itself has a fair bit of whitespace normalisation. It seems that by doing _no_ normalisation, it will fail to find links like
```
http://127.0.0.1:8000/a.html#:~:text=a-,b%20cd
```
in
```html
<html>
  <body>
    <p>
      a
      b
      cd
    </p>
  </body>
</html>
```
Using "copy link to highlight" in Firefox gave me that link from this HTML. 

Is this limitation okay? Is this something which would "likely requires a full implementation of a browser"? I'm not a stakeholder here, so I'm happy to leave this decision to you or others.

lychee-lib/src/utils/fragment_checker/text.rs
```diff
@@ -0,0 +1,220 @@
...
+                | CallbackEvent::Doctype { .. }
+                | CallbackEvent::Error(_) => {}
+            }
+            None
+        }
+    }
+
+    let mut extractor = TextExtractor::new();
+    let emitter = CallbackEmitter::new(&mut extractor);
+    let _: Result<Vec<_>, _> = Tokenizer::new_with_emitter(input, emitter).collect();
```
katrinafyi: save a tiny bit of memory by avoiding the `Vec<()>`
```suggestion
    let _: Result<(), _> = Tokenizer::new_with_emitter(input, emitter).collect();
```

lychee-bin/src/config/mod.rs
```diff
@@ -74,6 +74,17 @@ impl ArgBoolOptionalExt for Arg {
     }
 }
 
+/// Values for `--include-fragments` option, which controls how URL fragments are checked.
+#[derive(Debug, Clone, Deserialize, EnumString)]
```
katrinafyi: if you want
```suggestion
#[derive(Debug, Clone, Copy, Deserialize, EnumString)]
```

katrinafyi (approved):
Looks really good, just a few code things which should be easy fixes and one question about whitespace normalisation. Apart from that, I think it's very nearly ready!

lychee-bin/src/config/mod.rs
cristiklein: Indeed, looks a lot cleaner. Fixed in https://github.com/lycheeverse/lychee/pull/2138/commits/5aa7d9346e18fb637b2d391e7aa946e36ba45b01

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed in https://github.com/lycheeverse/lychee/pull/2138/commits/dd533cf8366685fb45ed47f7c067ff3ba2141ab9

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: I'm a stakeholder and I definitely want reasonable cases of "odd whitespace" (like the example you gave above) to work.

I added your example to the tests and re-added whitespace normalization: https://github.com/lycheeverse/lychee/pull/2138/commits/4bbc203d860e313e49c97d582d81da7b11fcf54b

I kind of dislike how the code re-splits and re-join the string, but I'm unsure how to otherwise consistently normalize whitespace, both within tag as well as between tags, without too much complexity.

Any ideas?

katrinafyi:
Oh one more thing (sorry!), is there a reason you use the lower level CallbackEmitter rather than the higher level default Tokeniser? It seems to do everything we need: https://docs.rs/html5gum/latest/html5gum/index.html#html5gum

lychee-lib/src/utils/fragment_checker/text.rs
katrinafyi: It's something like: in the String event handler, split by whitespace and then append the split words to the output. You have to stitch adjacent text together across elements, but only if not separated by whitespace. Matching would be splitting the search string by whitespace, then [windows](https://doc.rust-lang.org/std/primitive.slice.html#method.windows) to match the fragment words amongst the document words.

But thinking about it, this sounds rather slow and it will prevent running more optimised string search algorithms. Alternatively, you could compile the fragment pattern into a regex with `\s+`. This could be even better because it could consider the prefix and suffix as well. 

However, the Vec of words approach has an advantage over regex in that it could be more space efficient by only storing references. The regex would need the full text to be put into a string.

If you want, I can give it a try tomorrow.

cristiklein:
> Oh one more thing (sorry!), is there a reason you use the lower level CallbackEmitter rather than the higher level default Tokeniser? It seems to do everything we need: https://docs.rs/html5gum/latest/html5gum/index.html#html5gum

Indeed, that looks so much nicer (https://github.com/lycheeverse/lychee/pull/2138/commits/0ea1412d73e7782f695e43ec5f5478f8b7d45060). Thanks for pointing me in the right direction.

lychee-lib/src/utils/fragment_checker/text.rs
cristiklein: Would be great if you gave it a try. Check out what my agentic friend came up with:

```diff
@@ -78,6 +78,7 @@ fn extract_visible_text(input: &str) -> String {
 
     let mut text = String::new();
     let mut hidden_stack: Vec<String> = Vec::new();
+    let mut last_was_whitespace = true;
 
     for Ok(token) in Tokenizer::new(input) {
         match token {
@@ -94,17 +95,23 @@ fn extract_visible_text(input: &str) -> String {
                 }
             }
             Token::String(value) if hidden_stack.is_empty() => {
-                text.push_str(&String::from_utf8_lossy(&value));
+                for ch in String::from_utf8_lossy(&value).chars() {
+                    if ch.is_whitespace() {
+                        if !last_was_whitespace {
+                            text.push(' ');
+                            last_was_whitespace = true;
+                        }
+                    } else {
+                        text.push(ch);
+                        last_was_whitespace = false;
+                    }
+                }
             }
-
             _ => { /* Ignore other token types */ }
         }
     }
 
-    // It's kind of wasteful to split and re-join the text. However, given that extra whitespace may appear both inside
-    // a tag and between tags, this is a simple way to ensure that the extracted text is normalized in a way that matches
-    // how browsers treat whitespace for text fragments.
-    text.split_whitespace().collect::<Vec<_>>().join(" ")
+    text.trim_end().to_string()
 }
 ```

It basically pushes one character at a time and only adds whitespace if the last character wasn't a whitespace. Is this the kind of solution we want to go for? I'm split. :thinking:

File: lychee-lib/src/utils/fragment_checker/text.rs (hunk 9 above)
Comment (katrinafyi) [reply lychee-lib/src/utils/fragment_checker/text.rs]: I think that this is not a bug, since if there's no `end` starting at an earlier position, then there will still be no `end` when starting from a later position.

However, there was a very similar bug where it was failing to try later occurrences of `end` which might have a valid `suffix` after them.

**Review by katrinafyi — COMMENTED** [review]
@cristiklein I had a look through this in https://github.com/cristiklein/lychee/pull/1 and I think that whitespace normalisation by split is necessary and okay for now.

I also found a bug when the `end` string occurs multiple times. Previously, it only checked the earliest end string for each start position. This would fail to match if the first `end` had no suffix, even if there was a later `end` which had the right suffix. This is similar to mre's comment here: https://github.com/lycheeverse/lychee/pull/2138#discussion_r3079443966. To fix this, I just changed the loop into a regex match which is less work and possibly faster. Handling this in the manual loop code would've needed an inner loop over all possible end positions for each start position.

I also looked at the behaviour around alternative whitespaces (using `&nbsp;` as an example) and fixed a bug there.

And a few more test cases. Hopefully that is okay :)

File: lychee-lib/src/utils/fragment_checker/text.rs (hunk 27 above)
Comment (katrinafyi) [reply lychee-lib/src/utils/fragment_checker/text.rs:29]: Yeah, it ended up something like that in https://github.com/cristiklein/lychee/pull/1. I just used iterators instead of one character at a time because it's (probably) faster.

**Comment by cristiklein** [comment]
Really love @katrinafyi regex-based matching. So much easier to understand.

I'd say this is as good as it gets. Ready to merge from my end.

File: lychee-lib/src/utils/fragment_checker/text.rs [hunk 30]
```diff
@@ -0,0 +1,352 @@
...
+/// considered visible text, except known hidden text (according to [`is_hidden_tag`]).
+fn extract_visible_text(input: &str) -> String {
+    /// Pushes a space if not already ending in a space.
+    fn push_space(text: &mut String) {
+        if !text.ends_with(char::is_whitespace) {
+            text.push(' ');
+        }
+    }
+
+    let mut text: String = String::new();
```
katrinafyi: you're right, not needed :)
```suggestion
    let mut text = String::new();
```

lychee-bin/tests/cli.rs
```diff
@@ -52,6 +52,41 @@ mod cli {
...
+            let body = fs::read(fixtures_dir.join(name))?;
+            Ok(ResponseTemplate::new(200).set_body_raw(body, "text/html"))
+        };
+
+        Mock::given(method("GET"))
+            .and(path("/"))
+            .respond_with(html_response("index.html")?)
+            .mount(&mock_server)
+            .await;
+
```
thomas-zahner: Nitpick; that's currently unused

```patch
index 2d649908..f70a8e66 100644
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -60,12 +60,6 @@ mod cli {
             Ok(ResponseTemplate::new(200).set_body_raw(body, "text/html"))
         };
 
-        Mock::given(method("GET"))
-            .and(path("/"))
-            .respond_with(html_response("index.html")?)
-            .mount(&mock_server)
-            .await;
-
         Mock::given(method("GET"))
             .and(path("/index.html"))
             .respond_with(html_response("index.html")?)
```

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,299 @@
...
+    pub(super) end: Option<String>,
+    pub(super) suffix: Option<String>,
+}
+
+const FRAGMENT_DIRECTIVE_DELIMITER: &str = ":~:";
+const TEXT_DIRECTIVE_KEY: &str = "text";
+
+impl<'a> ParsedFragment<'a> {
+    /// By default, there is no fragment.
+    pub(super) const fn default() -> Self {
```
thomas-zahner: This should be more idiomatic:

```patch
diff --git a/lychee-lib/src/utils/fragment_checker/parsed_fragment.rs b/lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
index 99739332..b876f1eb 100644
--- a/lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
+++ b/lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
@@ -19,15 +19,17 @@ pub(super) struct TextDirective {
 const FRAGMENT_DIRECTIVE_DELIMITER: &str = ":~:";
 const TEXT_DIRECTIVE_KEY: &str = "text";
 
-impl<'a> ParsedFragment<'a> {
+impl Default for ParsedFragment<'_> {
     /// By default, there is no fragment.
-    pub(super) const fn default() -> Self {
+    fn default() -> Self {
         Self {
             element_id: None,
             text_directives: Vec::new(),
         }
     }
+}
 
+impl<'a> ParsedFragment<'a> {
     /// This method parses the fragment, separating the element id (if any) from the text directives (if any).
     pub(super) fn parse(url: &'a Url) -> Self {
         let Some(fragment) = url.fragment() else {
```

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,299 @@
...
+        Self {
+            element_id: None,
+            text_directives: Vec::new(),
+        }
+    }
+
+    /// This method parses the fragment, separating the element id (if any) from the text directives (if any).
+    pub(super) fn parse(url: &'a Url) -> Self {
```
thomas-zahner, line 34: Hmm first time seeing `pub(super)` personally but also in lychee. How is it different to `pub(crate)` in this context? Is it less visible than `pub(crate)`?

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
```diff
@@ -0,0 +1,299 @@
...
+    #[test]
+    fn parses_text_directive_with_encoded_utf8() {
+        let url = Url::parse("http://127.0.0.1:8000/a.html#:~:text=b%C2%A0cd").unwrap();
+        let parsed = ParsedFragment::parse(&url);
+
+        assert_eq!(
+            parsed.text_directives,
+            vec![TextDirective {
+                prefix: None,
+                start: "b\u{a0}cd".to_string(),
```
thomas-zahner: I was wondering what `\u{a0}` is and only saw in @katrinafyi's PR that it's a non-breaking space. Would be worth a comment and especially why we do this and why we test it. I'm currently not sure I understand this. We basically replace any consecutive whitespaces with a single ` ` space, right? (edit: I think I get it now with the space-interspersing, but still worth a comment)

thomas-zahner (commented):
@cristiklein @katrinafyi Thank you for your efforts! 🚀 Remarking just a few tiny things I noticed. I'd like to include this in the next release I'm intending to do the next few days.

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
katrinafyi: Yeah, it is related to whitespace splitting/normalisation and the main test for this is over at `check_text_fragments_alternative_whitespaces`. This test just makes sure that %-encoded nbsp inside a URL fragment get decoded.

It would be good to add a comment or an assertion message. @cristiklein I leave it up to you :)

lychee-bin/tests/cli.rs
cristiklein: Fixed in 1e1b45100d5475560a48cd1a0f38b2e632c6deb7. (I'm really impressed you caught this. :open_mouth: )

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed in 118028b843c20a7174c7f4e994c5df8e244ac2ed

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Indeed, `pub(super)` is less visible than `pub(crate)`. I think it expresses the intention that this method is internal only a bit better, but we can also stick to `pub(crate)`. It doesn't really have an impact in this case. Let me know what you prefer.

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
cristiklein: Fixed in 7c2707e4bac8005a63ecdd0033f153d8dac1007f. I added a comment and converted the magic numbers into constants.

lychee-lib/src/utils/fragment_checker/parsed_fragment.rs
thomas-zahner: Thank you, let's leave it

thomas-zahner:
@cristiklein @katrinafyi The implementation really looks solid and it's a great feature. Thank you for your hard work! :rocket:

thomas-zahner:
One last question, have we decided that an additional `html5gum` implementation is not worth the effort? I don't think it's a top priority and it leads to additional code but theoretically it would be pretty nice if `use_html5ever` was respected in this new feature as well. Maybe I have missed a discussion.

cristiklein:
Thanks all! This was fun!

Regarding `use_html5ever`, we opted to only implement text fragment checking using `html5gum`, because that's what the anchor fragment checker does.

If it's valuable, I can take a shot at an `html5ever` implementation.

thomas-zahner:
It's definitely not urgent which is why I've merged and released it. @mre @katrinafyi What are your thoughts on the dual parser implementation?

PR #2140, opened 2026-04-15
refactor: make website checker return Status always, rather than Result

katrinafyi:
this means that the Err case of `Client::check` will only happen due to remap failures. this means the error cases we have to handle are much clearer.

this commit re-lands this refactor which was previously part of https://github.com/lycheeverse/lychee/pull/2124. there's a bit more context in that PR.


thomas-zahner:
Resolved the conflicts with ef5652c0 due to the updated signatures and test additions.

thomas-zahner:
Thank you for this nice and simple PR! I don't see any problems.

PR #2144, opened 2026-04-16
fix(formatters/markdown): stop leaking <error:> placeholder into output

### LINKED ISSUE #2143: `<error:>` leaks into markdown format output for unparseable URLs.
```sh
> lychee --format markdown fixtures/TEST_HTML5_MALFORMED_LINKS.html

...

* [ERROR] <error:> (at 6:14) | Cannot parse 'https://example]org/malformed_two' into a URL: invalid international domain name
```

tmchow:
## Summary

When lychee encounters a link whose URL can't be parsed, the markdown stats formatter was leaking `<error:>` into the report because the placeholder Uri constructed by `lychee_lib::types::request_error` has scheme `"error"` and the formatter always wraps the Uri in `<…>`. Addresses #2143.

## Why

`RequestError::into_response` (in `lychee-lib/src/types/request_error.rs`) constructs a synthetic `Response` for reporting purposes whenever a raw link can't be turned into a valid URL. It uses a single `static ERROR_URI: LazyLock<Uri> = LazyLock::new(|| Uri::try_from("error:").unwrap());` as the placeholder, backed by the `error:` scheme.

The existing markdown formatter treated every response identically:

```rust
let mut formatted = format!(
    "* [{}] <{}>",
    response.status.code_as_string(),
    response.uri,
);
```

For the placeholder, `response.uri` renders as `error:`, so the output became:

```
* [ERROR] <error:> (at 6:14) | Cannot parse 'https://example]org/malformed_two' into a URL: invalid international domain name
```

That's bad for two reasons: `<error:>` looks like a clickable anchor pointing at an unresolvable custom scheme, and it visually overshadows the actual failing URL which now only lives in the `| details` tail.

## Changes

- `lychee-bin/src/formatters/stats/markdown.rs`:
  - New private constant `UNPARSEABLE_URI_SCHEME = "error"`, kept in lockstep with `lychee-lib`'s placeholder (the comment points back at `request_error.rs` so it doesn't drift silently).
  - `markdown_response` now branches: when `response.uri.scheme() == UNPARSEABLE_URI_SCHEME`, it omits the `<uri>` segment and emits just `* [STATUS]`; the subsequent `(at span)` and `| details` tail continue to render normally.
  - No behavioural change for any non-placeholder URI -- the old code path remains default.

## Testing

- New `test_markdown_response_unparseable_url_omits_uri_segment` directly exercises the placeholder path:
  - Constructs a `ResponseBody` with `Uri::try_from("error:")` (the same value `RequestError::into_response` emits).
  - Asserts `<error:>` is not present in the output.
  - Asserts the prefix stays `* [ERROR]` so downstream consumers still see the status code.
  - Asserts the span `(at 1:1)` and the underlying malformed URL text survive in the details tail.
- `cargo test --package lychee --bin lychee` -> 78 passed, 0 failed.

Fixes #2143

This contribution was developed with AI assistance (Claude Code).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/lychee-bin/src/formatters/stats/markdown.rs
+++ b/lychee-bin/src/formatters/stats/markdown.rs
@@ -73,22 +73,24 @@ fn stats_table(stats: &ResponseStats) -> String {
         .to_string()
 }
 
-/// Placeholder scheme used in [`lychee_lib::types::request_error`] when a
-/// malformed link can't be parsed into a URL. Treat it as "no URL".
-const UNPARSEABLE_URI_SCHEME: &str = "error";
+/// Placeholder Uri used in [`lychee_lib::types::request_error`] when a
+/// malformed link can't be parsed into a URL. Matched exactly (not by scheme
+/// alone) so a real input like `error:foo` still renders its URI.
+const UNPARSEABLE_URI_SENTINEL: &str = "error:";
 
 /// Helper function to format single response body as markdown
 ///
 /// Optional details get added if available.
 fn markdown_response(response: &ResponseBody) -> Result<String> {
     // For links that could not be parsed as a URL, the underlying
-    // `RequestError::CreateRequestItem` path emits a placeholder `Uri`
-    // whose scheme is `error` (see `lychee-lib/src/types/request_error.rs`).
-    // Rendering `<{uri}>` on that placeholder leaks `<error:>` into the
-    // markdown output and obscures the actual failure (see #2143). Skip the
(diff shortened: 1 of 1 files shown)


mre:
Thanks for your pull request @tmchow. We decided to try and fix the issue on a more fundamental level by removing the sentinel value entirely. Thanks for honestly pointing out the AI assistance, too.

We'll continue collecting thoughts on how to fix the issue in #2143.

PR #2145, opened 2026-04-16
fix: prevent 0 value for max-concurrency and threads

### LINKED ISSUE #2141: Panic on `--max-concurrency=0`
Currently, we panic on the following call:

```sh
lychee --max-concurrency=0  https://en.wikipedia.org/wiki/Tony_Hoare
thread 'main' (32032398) panicked at lychee-bin/src/commands/check.rs:90:60:
mpsc bounded channel requires buffer > 0
```

I think we should show a more user-friendly message.

Discovered in https://github.com/lycheeverse/lychee/pull/2136#issuecomment-4259400512

katrinafyi:
Using NonZeroUsize means we get nice error messages automatically from clap.

    error: invalid value '0' for '--max-concurrency <MAX_CONCURRENCY>': number would be zero for non-zero type

    For more information, try '--help'.

Unfortunately, the error is more cryptic when using a zero in the toml:

    [ERROR] Error while loading config: Cannot load configuration file: lychee.toml
    See: https://github.com/lycheeverse/lychee/blob/lychee-v0.23.0/lychee.example.toml

Fixes https://github.com/lycheeverse/lychee/issues/2141

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -31,7 +31,7 @@ where
     S: futures::Stream<Item = Result<Request, RequestError>>,
 {
     // Setup
-    let max_concurrency = params.cfg.max_concurrency();
+    let max_concurrency = params.cfg.max_concurrency().get();
     let (send_req, recv_req) = mpsc::channel(max_concurrency);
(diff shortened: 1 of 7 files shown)


mre:
Love it! Thanks for the fix!

PR #2148, opened 2026-04-16
Fix output path not validated before the run

### LINKED ISSUE #2147: Output path not validated before the run
It's frustrating to see all checks run, only to find out that the final output file can't be written to.

```sh
~ ❯❯❯ lychee --output /does/not/exist https://endler.dev -v                                                                                                                                                               
[EXCLUDED] https://twitter.com/matthiasendler (at 9:12262)
   [200] https://endler.dev/about (at 9:9705) | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://endler.dev/about --[308]--> https://endler.dev/about/
   [200] https://endler.dev/about (at 9:9983) | Redirect: Followed 2 redirects resolving to the final status of: OK. Redirects: https://endler.dev/about --[308]--> https://endler.dev/about/ --[308]--> https://endler.dev/about/
   [200] https://endler.dev/about (at 9:10305) | Redirect: Followed 2 redirects resolving to the final status of: OK. Redirects: https://endler.dev/about --[308]--> https://endler.dev/about/ --[308]--> https://endler.dev/about/
   [200] https://endler.dev/archive (at 9:10395) | Redirect: Followed 1 redirect resolving to the final status of: OK. Redirects: https://endler.dev/archive --[308]--> https://endler.dev/archive/
Error: Cannot write status output to file

Caused by:
    No such file or directory (os error 2)
~ ❯❯❯
```

 I think we should show an error message immediately if we detect that.

KonstantinosVasilopoulos:
- Added check before execution, ensuring that the parent of the given output path exists
- Added test for the new logic

The already existing behavior is kept, the only difference being that the checks do not run if the parent directory of the output path does not exist.

`lychee --output tmp.txt https://example.com` succeeds even if tmp.txt does not exist (just like before).
`lychee --output does/not/exist/tmp.txt https://example.com` now fails before checks, since directory `does/not/exist/` does not exists.

Note: also affects --dump.

Closes #2147

### DIFF
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -301,7 +301,9 @@ fn run_main() -> Result<i32> {
     // Exit if output path parent directory does not exist
     if let Some(output) = &opts.config.output {
         let parent = output.parent().filter(|p| !p.as_os_str().is_empty());
-        if let Some(parent) = parent && !parent.exists() {
+        if let Some(parent) = parent
+            && !parent.exists()
+        {
             error!(
                 "Output path `{}` is not writable: parent directory `{}` does not exist",
                 output.display(),
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -300,6 +300,7 @@ fn run_main() -> Result<i32> {
 
     // Exit if output path parent directory does not exist
     if let Some(output) = &opts.config.output {
+        // parent() returns Some("") for paths with no directory component
         let parent = output.parent().filter(|p| !p.as_os_str().is_empty());
         if let Some(parent) = parent
             && !parent.exists()
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -298,6 +298,19 @@ fn run_main() -> Result<i32> {
         }
     };
 
+    // Exit if output path parent directory does not exist
+    if let Some(output) = &opts.config.output {
+        let parent = output.parent().filter(|p| !p.as_os_str().is_empty());
+        if let Some(parent) = parent && !parent.exists() {
+            error!(
+                "Output path `{}` is not writable: parent directory `{}` does not exist",
+                output.display(),
+                parent.display()
+            );
+            exit(ExitCode::UnexpectedFailure as i32);
+        }
+    }
+
     if let Some(mode) = opts.config.generate {
         print!("{}", generate(&mode)?);
         exit(ExitCode::Success as i32);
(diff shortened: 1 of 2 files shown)


lychee-bin/tests/cli.rs
```diff
@@ -4405,3 +4405,16 @@ fn test_file_limit_low_concurrency() {
...
+#[test]
+fn test_output_invalid_path() {
+    let mut cmd = assert_cmd::Command::cargo_bin("lychee").unwrap();
+    cmd.arg("--output")
+        .arg("does/not/exist")
+        .arg("https://example.com");
+    cmd.assert().stderr(predicates::str::contains(
+        "Output path `does/not/exist` is not writable: parent directory `does/not` does not exist",
+    ));
+}
```
mre, line 4420: Thanks for adding a test! 👍 


The test doesn't assert the exit code.
It only checks stderr output but doesn't verify the process actually failed.
Can we add something like

```rust
cmd.assert()
    .failure() // <--
    .stderr(predicates::str::contains(...));
```

lychee-bin/src/main.rs
```diff
@@ -299,6 +298,21 @@ fn run_main() -> Result<i32> {
         }
     };
 
+    // Exit if output path parent directory does not exist
+    if let Some(output) = &opts.config.output {
+        let parent = output.parent().filter(|p| !p.as_os_str().is_empty());
```
mre, line 304: Hmmm `filter(|p| !p.as_os_str().is_empty())` is quite subtle. 

I *believe* this guards against output being a bare filename like report.txt with no parent, where `.parent()` returns `Some("")`? 🤔 

Worth a short comment explaining why:

```rust
// parent() returns Some("") for paths with no directory component
let parent = output.parent().filter(|p| !p.as_os_str().is_empty());
```

mre:
Thanks for the PR, Kostas. Two minor remarks from my side.

lychee-bin/tests/cli.rs
KonstantinosVasilopoulos: Thanks for your input!

I've updated the test to assert the exit code with `.failure()` as suggested.

lychee-bin/src/main.rs
KonstantinosVasilopoulos: Exactly. `parent()` returns `Some("")` given only a filename without a parent directory. There is no need for a path validation in that case, hence empty parent strings are converted to `None`.

I added your comment to better document that case.

KonstantinosVasilopoulos:
Thanks for your input @mre ! I have added the changes you proposed.

mre (approved):

mre:
Thanks a lot for the nice contribution. 😃

PR #2149, opened 2026-04-17
Always show full config parse errors with full details in output

### LINKED ISSUE #2146: TOML error messages regressed
It seems that after d5d22bc2cf4e485f4a8beeaa66c42a794b0607e9, the error messages from TOML parsing no longer show the specific error message or the source location.

With this `lychee.toml`:
```
threads = 'a'
```

Old, good (e.g., 1a903e421775f1a2e5a0eced6768491ddcb7a9a7):
```
 [ERROR] Error while loading config: Cannot load configuration file `lychee.toml`: Failed to parse configuration file

Caused by:
    TOML parse error at line 1, column 11
      |
    1 | threads = 'a'
      |           ^^^
    invalid type: string "a", expected usize
    
See: https://github.com/lycheeverse/lychee/blob/lychee-v0.23.0/lychee.example.toml
```

After d5d22bc2cf4e485f4a8beeaa66c42a794b0607e9:
```
 [ERROR] Error while loading config: Cannot load configuration file: lychee.toml
See: https://github.com/lycheeverse/lychee/blob/lychee-v0.23.0/lychee.example.toml
```

mre:
The root cause is that the error was formatted with `{}` (Display), which for `anyhow` only shows the outermost message in the error chain. 

Since https://github.com/lycheeverse/lychee/commit/d5d22bc2cf4e485f4a8beeaa66c42a794b0607e9 restructured the error handling to use `e.context(msg)`, which wraps the TOML details *under* the context message, the Display format only shows the context message and loses the TOML parse error details (line number, column, offending value, etc.).

Changing `{}` to `{:?}` uses `anyhow`'s Debug format, which prints the full error chain:

Fixes #2146

### DIFF
--- a/fixtures/configs/invalid-type.toml
+++ b/fixtures/configs/invalid-type.toml
@@ -1,0 +1,1 @@
+threads = 'a'
--- a/lychee-bin/src/main.rs
+++ b/lychee-bin/src/main.rs
@@ -290,7 +290,7 @@ fn run_main() -> Result<i32> {
         Ok(opts) => opts,
         Err(e) => {
             error!(
-                "Error while loading config: {}\n\
+                "Error while loading config: {:?}\n\
                 See: https://github.com/lycheeverse/lychee/blob/lychee-v{}/lychee.example.toml",
                 e,
                 crate_version!()
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -1018,6 +1018,28 @@ mod cli {
             .stderr(contains("TOML parse error"));
     }
 
+    /// Make sure that TOML type errors include all error details (line/column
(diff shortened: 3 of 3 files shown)


katrinafyi (approved):
It's a quick and easy fix, awesome!

I was curious and this is what the error looks like for package.json after this change.
```
 [ERROR] Error while loading config: Cannot load configuration file: package.json

Caused by:
    0: Failed to parse lychee config from package.json
    1: invalid type: string "a", expected a nonzero usize at line 3 column 18
See: https://github.com/lycheeverse/lychee/blob/lychee-v0.23.0/lychee.example.toml
```

mre:
That looks really good!

mre:
Thanks for the review and the check. 😀

PR #2151, opened 2026-04-20
Prevent autolink duplication

### LINKED ISSUE #2150: Bracketed autolink `<...>` in markdown is detected twice
For example, `a.md`:
```
<http://127.0.0.1:8000/>
```

leads to
```console
$ cargo run -- a.md -vvv 
   [200] http://127.0.0.1:8000/ (at 1:2)
   [200] http://127.0.0.1:8000/ (at 1:2)
🔍 2 Total (in 9ms) 🔗 1 Unique ✅ 2 OK 🚫 0 Errors
```

Maybe this is being returned by both the markdown extractor and the plaintext linkify extractor?

thomas-zahner:
Fixes #2150

I'm not yet sure if this covers all cases.

### DIFF
--- a/lychee-lib/src/extract/markdown.rs
+++ b/lychee-lib/src/extract/markdown.rs
@@ -31,11 +31,9 @@ pub(crate) fn extract_markdown(
     include_verbatim: bool,
     include_wikilinks: bool,
 ) -> Vec<RawUri> {
-    // In some cases it is undesirable to extract links from within code blocks,
-    // which is why we keep track of entries and exits while traversing the input.
     let mut inside_code_block = false;
-    let mut inside_link_block = false;
-    let mut skip_following_link_text = false;
+    let mut inside_link_label = false; // encountering `X` in `[X]()`
+    let mut inside_extracted_link = false; // prevent double extraction when encountering `Text(X)` in `<X>` or `[[X]]`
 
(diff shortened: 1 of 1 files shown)


lychee-lib/src/extract/markdown.rs
```diff
@@ -35,7 +35,7 @@ pub(crate) fn extract_markdown(
     // which is why we keep track of entries and exits while traversing the input.
     let mut inside_code_block = false;
     let mut inside_link_block = false;
-    let mut inside_wikilink_block = false;
+    let mut skip_following_link_text = false;
```
katrinafyi: optionally, you can rename this because it's specifically about inner text which repeats the URL.
```suggestion
    let mut inside_link_url = false;
```
or maybe `inside_extracted_link` - something which emphasises the link has already been handled.

The other "inside link" variable can also be renamed to `inside_link_label` because that is only for the X in `[X]()`

katrinafyi (approved):
Makes sense! Thanks for the fast fix and investigation. I'm not too worried about other cases, I think the issue was fairly isolated.

lychee-lib/src/extract/markdown.rs
thomas-zahner: Thank you :+1: I've used `inside_extracted_link`. Also added short explanations.

Issue #2112, opened 2026-03-29
Valid fragments w/ URL-encoded emoji false positive


adamlui:
like https://github.com/adamlui/js-utils/actions/runs/23705290073/job/69056053612

<img width="1274" height="891" alt="Image" src="https://github.com/user-attachments/assets/12ebfb06-4497-4723-9747-2ed1ad318d18" />

katrinafyi:
Investigating and there seems to be a problem with the "🏛️ MIT 许可证" header link at https://github.com/adamlui/js-utils/blob/main/minify.js/node.js/docs/zh-cn/README.md#%EF%B8%8F-mit-%E8%AE%B8%E5%8F%AF%E8%AF%81. It works in browser but not in lychee.

```
[a](https://github.com/adamlui/js-utils/blob/main/minify.js/node.js/docs/zh-cn/README.md#%EF%B8%8F-mit-%E8%AE%B8%E5%8F%AF%E8%AF%81)
```
```console
$ cargo run b.md --include-fragments -vvvv     
 [DEBUG] Applied quirk 'fetch raw GitHub Markdown files' to https://github.com/adamlui/js-utils/blob/main/minify.js/node.js/docs/zh-cn/README.md#%EF%B8%8F-mit-%E8%AE%B8%E5%8F%AF%E8%AF%81
 [ERROR] https://github.com/adamlui/js-utils/blob/main/minify.js/node.js/docs/zh-cn/README.md#%EF%B8%8F-mit-%E8%AE%B8%E5%8F%AF%E8%AF%81 (at 2:1) | Cannot find fragment
  [WARN] There were issues with GitHub URLs. You could try setting a GitHub token and running lychee again.
Issues found in 1 input. Find details below.

[b.md]:
 [ERROR] https://github.com/adamlui/js-utils/blob/main/minify.js/node.js/docs/zh-cn/README.md#%EF%B8%8F-mit-%E8%AE%B8%E5%8F%AF%E8%AF%81 (at 2:1) | Cannot find fragment
```

The percent encoded string is really odd:
```python
from urllib.parse import *
unquote('%EF%B8%8F-mit-%E8%AE%B8%E5%8F%AF%E8%AF%81').encode('unicode_escape')
# b'\\ufe0f-mit-\\u8bb8\\u53ef\\u8bc1'
```
[...]

PR #2153, opened 2026-04-21
feat: match github's markdown fragment generation

katrinafyi:
This copies the approach of [Flet/github-slugger#56](https://www.github.com/Flet/github-slugger/issues/56) for deriving fragment identifiers from markdown headings.

Basically, it deletes characters from certain Unicode categories, then it lowercases and replaces spaces with `-`. and finally applies a disambiguation with numbers. 

To check the test cases, you can see [this gist](https://gist.github.com/katrinafyi/c5f767d2952ae2ac2a3059b267325334) or create your own markdown file on github.

Fixes https://github.com/lycheeverse/lychee/issues/2112. This can be verified with
```console
$ cargo run -- https://raw.githubusercontent.com/adamlui/js-utils/refs/heads/main/minify.js/node.js/docs/zh-cn/README.md  -vvvvv --include-fragments

...

   [200] https://raw.githubusercontent.com/adamlui/js-utils/refs/heads/main/minify.js/node.js/docs/zh-cn/README.md#%EF%B8%8F-mit-%E8%AE%B8%E5%8F%AF%E8%AF%81 (at 28:10)
```


lychee-lib/src/extract/markdown.rs
```diff
@@ -277,7 +285,7 @@ pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
     let mut in_heading = false;
     let mut heading_text = String::new();
     let mut heading_id: Option<CowStr<'_>> = None;
-    let mut id_generator = HeadingIdGenerator::default();
+    let mut id_generator = GithubSlugify::new();
```
mre, line 288: Kinda liked the old name, but I could see how we'd like to disambiguate different vendors. As a compromise, maybe we could namespace it? `github::HeadingIdGenerator`?

katrinafyi: Yeah that's fair. I don't love the namespace github because I think it's more fragment/slug-related rather than Github related. What about `fragments::GithubHeadingIdGenerator`? Also, do you want to change away from "slugging" terminology everywhere else too?

mre: Sorry for the bikeshedding. 😅 I like your idea! I believe "slug" is indeed a bit misleading because it's also commonly used for blog post URLs. It's about fragment ID generation, so you're right that we should use something along those lines.

katrinafyi: No worries! Should be done in e9f6275d927c7ea87cc44474cd85622eae22986e

mre: Yay! Love it.

lychee-lib/src/extract/fragments.rs
```diff
@@ -0,0 +1,198 @@
...
+            "Underscores foo_bar_, dots . and numbers 1.7e-3",
+            "underscores-foo_bar_-dots--and-numbers-17e-3",
+        );
+        check("Many          spaces", "many----------spaces");
+    }
+
+    #[test]
+    fn test_github_generate_with_repeats() {
+        let headings = ["foo 1", "foo", "foo", "foo", "foo 1", "FOO 1"];
+        let expected = vec!["foo-1", "foo", "foo-2", "foo-3", "foo-1-1", "foo-1-2"];
```
mre, line 239: I appreciate the thorough testing. One remark though: this could be using `rstest`. It would get a little more verbose, but the advantage would be that we'd see exactly which case failed pretty easily. Same for `test_generate_kebab_case`, but that's admittedly more subjective.

But overall, I don't care too much as long as we've got tests. :)

mre (commented):
Just one more fly-by comment. ;)

lychee-lib/src/extract/fragments.rs
katrinafyi: I changed the other two tests to rstest and merged them. This one with the vector is testing the numeric suffix when generating a sequence of identifiers so it shouldn't be an rstest.

lychee-lib/src/extract/fragments.rs
```diff
@@ -0,0 +1,222 @@
...
+                })
+                .unwrap_or(/* in case of overflow only */ usize::MAX)
+        });
+
+        let next_suffix = ONE.saturating_add(this_suffix.unwrap_or(0));
+        self.next_suffixes.insert(base_id, next_suffix);
+
+        if this_suffix.is_some() {
+            self.next_suffixes.insert(candidate.clone(), ONE);
+        }
```
mre: Can you help me understand this logic?

If base_id is "foo" and it's seen for the second time, candidate becomes "foo-1". You then insert "foo-1" -> 1 into the map. If the very next heading in the document is actually titled "# foo 1", the generator will:

1. Generate base_id = "foo-1".
2. See that "foo-1" is already in the map.
3. Increment it and produce "foo-1-1".

Doesn't this create a "jump" in numbering if headings naturally collide with generated suffixes? Would that be an issue?

lychee-lib/src/extract/fragments.rs
```diff
@@ -0,0 +1,222 @@
...
+//! [markdownlint]: https://github.com/DavidAnson/markdownlint/blob/v0.40.0/doc/md051.md
+//! [committed]: https://github.com/DavidAnson/markdownlint/commit/30353cc733561af72bf5d226105429c07b43a666
+//! [Flet/github-slugger]: https://github.com/Flet/github-slugger
+use std::{collections::HashMap, num::NonZeroUsize, sync::LazyLock};
+
+use regex::Regex;
+
+/// From <https://github.com/gjtorikian/html-pipeline/blob/f13a1534cb650ba17af400d1acd3a22c28004c09/lib/html/pipeline/toc_filter.rb#L30>
+static REGEX_TO_REMOVE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"[^\w -]").expect("fragment regex failed"));
```
mre, line 24: So I'm not an expert on regexes, but doesn't Rust match unicode characters with `\w` while Ruby only matches ASCII [a-zA-Z0-9_]?
If our goal is to perfectly mimic the Ruby html-pipeline, we should test that. Maybe we should use (?-u:[^\w -]) or specifically [^a-zA-Z0-9 _-] to ensure you aren't accidentally keeping Unicode characters that the Ruby version would have stripped.

test-utils/src/lib.rs
```diff
@@ -133,7 +133,7 @@ macro_rules! fixtures_path {
 #[macro_export]
 macro_rules! load_fixture {
     ($filename:expr) => {{
-        let path = fixtures_path!().join($filename);
+        let path = test_utils::fixtures_path!().join($filename);
```
mre: This change assumes that the crate using the macro has test_utils in its dependency tree under that specific name. It's not a big deal right now but if a sub-crate imports the macro but doesn't have test_utils as a direct dependency (or renames it), this will fail to compile. Using $crate::fixtures_path!() is usually safer for macros intended to be exported.

I haven't tested it, but this *should* compile and work:
 
```rust
#[macro_export]
macro_rules! load_fixture {
    ($filename:expr) => {{
        // $crate ensures that we always point to the fixtures_path 
        // defined inside this specific crate.
        let path = $crate::fixtures_path!().join($filename);
        std::fs::read_to_string(path).unwrap()
    }};
}
```

Same for the other macro below.

mre (approved):
We can merge this. Added a few more comments, but they are quite nitpicky tbh. No blockers.

lychee-lib/src/extract/fragments.rs
katrinafyi: You're right about the languages having different \w. However, the Ruby code actually goes through the code path that uses `\p{Word}`. This matches unicode-aware word chats, just like Rust's default \w.

lychee-lib/src/extract/fragments.rs
katrinafyi: I'm not sure what you mean by jump. Do you mean because the foo-1-1 ID has two hyphens? Or, are you worried about seeing a "foo" next and skipping some numbers?

 If it's the second case, the next "foo" will resume the sequence from 2 and no numbers will be skipped. The incrementing numbers are always appended onto the original base_id.

Edit: I should add that avoiding conflicts with generated suffixes is the most complicated part of this code. There's two ways to write this code, one that has complicated after-generarion logic (this code) and one that has complicated collision detection. I can try changing to the complicated collision detection version which avoids this conditional insert (but introduces conditional queries).

Edit 2: the behaviour of headings that conflict with generated suffixes can also be seen in this test case https://github.com/rina-forks/lychee/blob/41362802c150490473c06155d0e11d2ccc3a2c6e/lychee-lib/src/extract/fragments.rs#L212

lychee-lib/src/extract/fragments.rs
mre: Aaah, cool. Glad that's the case.

lychee-lib/src/extract/fragments.rs
katrinafyi: See if https://github.com/lycheeverse/lychee/commit/41612674c2353b922efbc178934d8d937e2d1c67 is easier to follow.

lychee-lib/src/extract/markdown.rs
thomas-zahner: This link is broken..
I would assume that the specs have been removed in the pulldown-cmark library as it's supposed to be 100% compliant with CommonMark which is specified here: https://spec.commonmark.org/

lychee-lib/src/extract/fragments.rs
```diff
@@ -0,0 +1,244 @@
...
+    // but GitHub's algorithm does not, presumably because it is implemented
+    // in Ruby: https://ruby-doc.org/3.4.1/case_mapping_rdoc.html#label-Default+Case+Mapping
+    REGEX_TO_REMOVE
+        .split(text)
+        .flat_map(str::chars)
+        .map(|c| {
+            match c {
+                ' ' => '-',
+                '\u{0130}' => 'i', // U+0130 LATIN CAPITAL LETTER I WITH DOT ABOVE
+                'Σ' => 'σ',        // U+03A3 GREEK CAPITAL LETTER SIGMA
```
thomas-zahner: What's up with these two characters? 😅
Does the above comment refer to this?

lychee-lib/src/extract/fragments.rs
```diff
@@ -0,0 +1,244 @@
...
+        "underscores-foo_bar_-dots--and-numbers-17e-3"
+    )]
+    #[case("Many          spaces", "many----------spaces")]
+    #[case("À, Á, Â, Ã, Ä, Å or à, á, â, ã, ä, å", "à-á-â-ã-ä-å-or-à-á-â-ã-ä-å")]
+    // Regression tests for https://github.com/lycheeverse/lychee/issues/2112
+    #[case::emoji_variation_selector_kept("#️⃣ b", unpercent("%EF%B8%8F⃣-b"))]
+    #[case::emoji_variation_selector_kept("☔️ c", unpercent("%EF%B8%8F-c"))]
+    #[case::alphabetic_emoji_kept("🅰️ d", unpercent("🅰%EF%B8%8F-d"))]
+    // Should NOT apply Unicode's special casing rules: https://www.unicode.org/Public/3.2-Update/SpecialCasing-3.2.0.txt
+    #[case::capital_dotted_i("aİb", "aib")]
```
thomas-zahner: So does this really only affect exactly this `İ` and `Σ` character?

typos.toml
```diff
@@ -7,3 +7,9 @@ extend-exclude = [
 
 [default.extend-words]
 excluder = "excluder"
+
+[default]
+extend-ignore-re = [
+  # Ignore the line after `// spellchecker:ignore-next-line`:
+  "(#|//)\\s*spellchecker\\s*:\\s*ignore-next-line\\n.*",
```
thomas-zahner: Maybe?
```suggestion
  "(#|//|///|//!)\\s*spellchecker:ignore-next-line\\n.*",
```

lychee-lib/src/extract/markdown.rs
katrinafyi: Fixed the link, seems to have been a master/main difference.

lychee-lib/src/extract/fragments.rs
katrinafyi: It does. I've moved the lowercasing into its own function and added a comment closer to this code to try and clarify it.

lychee-lib/src/extract/fragments.rs
katrinafyi: Hm in my rough testing that was the case. As we are doing lowercasing, a lot of the rules don't affect us. For instance, there's a special rule to uppercase ß to SS but lowercasing SS simply goes to ss without needing special handling.

For completeness, I should try to check all of the special case rules. This will take some time. https://gist.github.com/katrinafyi/b616773773588ce6fb3d488903a9048b

katrinafyi (commented):
Thanks for the review, I'll follow up on the special case rules.

lychee-lib/src/extract/fragments.rs
thomas-zahner: Oh thank you, that's awesome! Maybe it's possible to add this gist as a test to lychee?

katrinafyi: I've added the language-independent rules of special casing into the repository as special-casing.md and its current Github output as special-casing.html and there's now a test checking that they match. This is powered by a bit of a messy python file that translates the SpecialCasing.txt into markdown headings. Let me know if you have any ideas to organise this better.

mre (approved):
This all looks good to me!

I don't think there are any open questions, other than whether we'd like to add the Unicode special-casing gist as a test. I think that's a great idea, but it's also not a blocker, and we can add it later. (@thomas-zahner, hope that's fine.) 

I'd like to merge this in to mark the work as done.
Great improvement! Thank you, @katrinafyi. 👍

lychee-lib/src/extract/fragments.rs
thomas-zahner: @katrinafyi This looks like very thorough testing, thanks!

Just a side-comment when looking through this PR.
I'm somewhat confused by how and why the files special-casing.html are produced. The special-casing.html file was manually created via GitHub gist, right? This could maybe be documented or even automated. And what exactly is the purpose of putting the contents in a Markdown code block in make-special-casing.py?

katrinafyi: Oh yeah I should improve that process. I thought I'd written a comment but it must've gotten lost.

The HTML is obtained [using an API](https://docs.github.com/en/rest/repos/contents?apiVersion=2026-03-10#get-repository-content) which is really nice. It renders a MD file in a repo into HTML and doesn't need to create a gist. I did think about fully automating this by calling the API directly in the test. But it has the downside that running the test would need an API token or risk rate-limiting.

What do you think? I could fully automate it in the test (and maybe ignore rate limit errors), or I could make a shell script or just write a comment.

The non-heading text in the txt file is put into the Markdown code blocks. This is just to give the context for each special casing rule and doesn't affect the test itself.

katrinafyi (commented):
Re special casing test case...

lychee-lib/src/extract/fragments.rs
thomas-zahner: Thanks for the details. It would be really cool if the external python script could be converted into a Rust test. This test could then also write the md/html files and fail, if the contents have changed. This feels much more integrated than a separate python script which we forget about in the future.

If this is possible without API token and without being flaky this could be worth the change. Otherwise I'm fine with the current approach and an additional comment to clear up my confusion would totally suffice :)

PR #2158, opened 2026-04-24
check HTTP status in archive snapshots and ignore 503 in test

katrinafyi:
this should fix the flakiness while still detecting an error if the API ever changes and it starts 404ing or something.


thomas-zahner (approved):
Ah thank you!
It really is unfortunate how unstable their API is. At some days it feels like there's a 50% chance the requests fail, while other days the API seems quite stable.

mre:
Ah, this is such a great way to fix it. I learned something. Thanks for fixing.

PR #2164, opened 2026-04-24
 refactor: move basic auth credentials out of Request 

katrinafyi:
the basic auth extractor is now held by WebsiteChecker which makes sense because it's only applicable to website checks.

previously, the basic auth credentials were attached to the Request during the collecting phase, which is unnecessarily early i think. this change also avoids the need to repeatedly pass the "basic auth extractor" into helper functions.

i should add that i find the basic auth extractor and basic auth selector names to be very confusing. to me, "basic auth extractor" sounds like it parses basic auth strings from URLs when it's actually just a known database of basic auth credentials. "basic auth selector" sounds like it should do what basic auth extract actually does - select from a list of known basic auths. i haven't changed it in this PR but just something to think about.

One goal of this refactor is to make it more palatable to store Request inside Response (for https://github.com/lycheeverse/lychee/issues/2097), since it no longer has sensitive information.

### DIFF
--- a/lychee-bin/src/client.rs
+++ b/lychee-bin/src/client.rs
@@ -2,6 +2,7 @@ use crate::config::{Config, HeaderMapExt};
 use crate::parse::parse_remaps;
 use anyhow::{Context, Result};
 use http::{HeaderMap, StatusCode};
+use lychee_lib::BasicAuthExtractor;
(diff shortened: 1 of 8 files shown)


thomas-zahner (approved):
Looks like a nice little improvement!

@mre any objections?

mre (approved):
Looks fine. If we like, we can rename it, but we can do that in a follow-up PR. ✌️

Issue #2160, opened 2026-04-24
Changed release asset names break lychee-action


Donnerbart:
Release [0.24](https://github.com/lycheeverse/lychee/releases/tag/lychee-v0.24.0) changed the release asset names. For example, `lychee-aarch64-unknown-linux-gnu.tar.gz` is now `lychee-lychee-v0.24.0-aarch64-unknown-linux-gnu.tar.gtz`.

This breaks the filename determination in [lychee-action](https://github.com/lycheeverse/lychee-action/blob/master/action.yml#L74).

I think this breaks all setups that use `latest`:
```
Run lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411
Run echo "$HOME/.local/bin" >> "$GITHUB_PATH"
Run # Remove any existing lychee binary to prevent conflicts
Run # Create a temporary directory for downloads and extraction
Downloading from: https://github.com/lycheeverse/lychee/releases/latest/download/lychee-x86_64-unknown-linux-gnu.tar.gz
Error: Process completed with exit code 22.
```

thomas-zahner:
Thank you for reporting! I think we should retain the old naming scheme. There's not much value in `lychee-lychee-v0.24.0-`.

WIP PR: https://github.com/lycheeverse/lychee/pull/2161

@mre FYI

thomas-zahner:
I've now restored the old release file names, see https://github.com/lycheeverse/lychee/releases/tag/nightly

Next issue is that the contents differ, which is why the docker step [now consistently fails](https://github.com/lycheeverse/lychee/actions/runs/24889954050/job/72884969343) with `tar: lychee: Not found in archive`.

Previously it was:

```
lychee
docs/
docs/README.md
docs/TROUBLESHOOTING.md
docs/lychee.1
docs/PRE_COMMIT.md
```

Now it is:

```
lychee-x86_64-unknown-linux-gnu/
lychee-x86_64-unknown-linux-gnu/README.md
lychee-x86_64-unknown-linux-gnu/docs/
lychee-x86_64-unknown-linux-gnu/docs/lychee.1
lychee-x86_64-unknown-linux-gnu/docs/TROUBLESHOOTING.md
lychee-x86_64-unknown-linux-gnu/docs/PRE_COMMIT.md
lychee-x86_64-unknown-linux-gnu/complete/
lychee-x86_64-unknown-linux-gnu/complete/_lychee.ps1
[...]

PR #2165, opened 2026-04-24
fix: update binstall metadata after 0.24.0 restructure

katrinafyi:
just remove bin-dir and pkg-fmt because they can be auto-detected and this is more robust against future changes and different operating systems: https://github.com/cargo-bins/cargo-binstall/blob/main/SUPPORT.md#defaults

this is the error without these changes:
```console
$ cargo binstall --no-cleanup lychee@0.24.1  --strategies crate-meta-data
 INFO resolve: Resolving package: 'lychee@=0.24.1'
ERROR resolve: When resolving lychee bin lychee is not found. This binary is not optional so it must be included in the archive, please contact with upstream to fix this issue.
 WARN resolve: Error while downloading and extracting from fetcher github.com: failed to find or install binaries: bin file /home/rina/.cargo/cargo-binstall4XM5HP/bin-lychee-x86_64-unknown-linux-gnu-GhCrateMeta/lychee not found
ERROR resolve: When resolving lychee bin lychee is not found. This binary is not optional so it must be included in the archive, please contact with upstream to fix this issue.
 WARN resolve: Error while downloading and extracting from fetcher github.com: failed to find or install binaries: bin file /home/rina/.cargo/cargo-binstall4XM5HP/bin-lychee-x86_64-unknown-linux-musl-GhCrateMeta/lychee not found
ERROR Fatal error:
  \x1b[31m×\x1b[0m For crate lychee: Fallback to cargo-install is disabled
\x1b[31m  ╰─▶ \x1b[0mFallback to cargo-install is disabled
```

related to https://github.com/lycheeverse/lychee/issues/2160. the failure of binstall breakas pre-commit (and probably other things)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,13 @@ jobs:
       - name: Verify the MSRV
         run: make verify
 
+  binstall-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: cargo-bins/cargo-binstall@main
+      - run: cargo binstall --manifest-path . --strategies crate-meta-data lychee --no-confirm
+
   publish-check:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


mre:
Ouch, yeah, that's on me. Thanks for fixing.

PR #2167, opened 2026-04-24
Replace Wayback Availability API with direct snapshot URLs

mre:
I came across [this internal Wikipedia page](https://en.wikipedia.org/wiki/Help:Using_the_Wayback_Machine) and learned about Wayback Machine URL formats. Based on this, I believe we can simplify our code.

We can replace the call to https://archive.org/wayback/available with a client-side construction of https://web.archive.org/web/0/<url>.

~The '0' timestamp tells Wayback to redirect to the newest available snapshot, which is far more reliable than the Availability JSON API (which is heavily rate-limited and is a frequent source of frustration due to flakiness. It also sometimes returns empty archived_snapshots for pages that are clearly archived).~

Edit: turns out "0" means "give me the first snapshot". What I was looking for was https://web.archive.org/web/<url> without the `/0/`, which returns the **latest** snapshot. Thanks for the correction.

### DIFF
--- a/lychee-lib/src/archive/wayback/mod.rs
+++ b/lychee-lib/src/archive/wayback/mod.rs
@@ -1,181 +1,151 @@
+//! Wayback Machine integration for suggesting archived versions of dead pages.
+//!
+//! Builds URLs of the form:
+//!
+//! ```text
+//! https://web.archive.org/web/<url>
+//! ```
+//!
+//! Omitting the timestamp tells Wayback "give me the latest available
+//! snapshot". The server's response tells us whether a snapshot exists:
+//!
+//! - **`302 Found`** with a `Location:` header pointing at the actual
+//!   timestamped snapshot (e.g. `/web/20020120142510/http://example.com/`).
+//!   We read that header directly, without following the redirect, so users
+//!   see the capture date in the suggested link without paying to download
+//!   the archived page.
+//! - **`404 Not Found`** when no snapshot exists for that URL. We return
+//!   `None` so lychee won't suggest a dead-end "page not archived" link.
+//!
+//! This is far more reliable than the Availability JSON API
+//! (`https://archive.org/wayback/available`), which is heavily rate-limited,
+//! flaky, and frequently returns empty `archived_snapshots` for pages that are
+//! clearly archived.
+//!
+//! See <https://en.wikipedia.org/wiki/Help:Using_the_Wayback_Machine>.
+
 use std::sync::LazyLock;
 use std::time::Duration;
 
-use serde::de::Error as SerdeError;
-use serde::{Deserialize, Deserializer};
-
 use http::StatusCode;
+use reqwest::header::LOCATION;
+use reqwest::redirect::Policy;
 use reqwest::{Client, Error, Url};
 
-static WAYBACK_URL: LazyLock<Url> =
-    LazyLock::new(|| Url::parse("https://archive.org/wayback/available").unwrap());
+/// Per-request timeout for Wayback lookups.
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(20);
 
-pub(crate) async fn get_archive_snapshot(
-    url: &Url,
-    timeout: Duration,
-) -> Result<Option<Url>, Error> {
-    get_archive_snapshot_internal(url, timeout, WAYBACK_URL.clone()).await
-}
+/// Shared HTTP client for all Wayback suggestion lookups.
+///
+/// The suggestion path may issue dozens of lookups in quick succession
+/// (one per failed URL), all aimed at the same host. Sharing a client
+/// lets them reuse the connection pool, TLS session cache, and DNS
+/// resolver instead of paying those costs per request.
+///
+/// Redirects are intentionally *not* followed: a Wayback hit answers with a
+/// `302` whose `Location` already holds the snapshot URL, so following it
+/// would just download the archived page for no reason.
+static CLIENT: LazyLock<Client> = LazyLock::new(|| {
+    Client::builder()
+        .timeout(REQUEST_TIMEOUT)
+        .redirect(Policy::none())
+        .build()
+        .expect("Wayback HTTP client should always build with default config")
+});
 
-async fn get_archive_snapshot_internal(
-    url: &Url,
-    timeout: Duration,
-    mut api: Url,
-) -> Result<Option<Url>, Error> {
-    let url = url.to_string();
+/// Construct a Wayback Machine URL pointing at the latest available snapshot
+/// of `url`, or `None` if Wayback has no snapshot of the page.
+///
+/// Performs a single GET request against `https://web.archive.org/web/<url>`
+/// and inspects the response:
+///
+/// - a redirect (`302`) carries the timestamped snapshot in its `Location`
+///   header, which we return without following, and
+/// - a `404` means there is no snapshot, so we return `None`.
+pub(crate) async fn get_archive_snapshot(url: &Url) -> Result<Option<Url>, Error> {
+    let wayback = format!("https://web.archive.org/web/{url}");
+    // `Url::parse` cannot fail here for any well-formed input `url`, but
+    // if it ever did we'd rather report "no suggestion" than panic.
+    let Ok(snapshot_url) = Url::parse(&wayback) else {
+        log::debug!("failed to construct Wayback URL for {url}");
+        return Ok(None);
+    };
 
-    // The Wayback API doesn't return any snapshots for URLs with trailing slashes
-    let stripped = url.strip_suffix("/").unwrap_or(&url);
-    api.set_query(Some(&format!("url={stripped}")));
+    let response = CLIENT.get(snapshot_url).send().await?;
+    let status = response.status();
 
-    let response = Client::builder()
-        .timeout(timeout)
-        .build()?
-        .get(api)
-        .send()
-        .await?
-        .error_for_status()?
-        .json::<InternetArchiveResponse>()
-        .await?;
+    if status == StatusCode::NOT_FOUND {
+        return Ok(None);
+    }
 
-    Ok(response
-        .archived_snapshots
-        .closest
-        .map(|closest| closest.url))
-}
+    if status.is_redirection() {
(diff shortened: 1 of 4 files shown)


lychee-lib/src/archive/wayback/mod.rs
```diff
@@ -1,181 +1,125 @@
...
+
+/// Shared HTTP client for all Wayback suggestion lookups.
+///
+/// The suggestion path may issue dozens of lookups in quick succession
+/// (one per failed URL), all aimed at the same host. Sharing a client
+/// lets them reuse the connection pool, TLS session cache, and DNS
+/// resolver instead of paying those costs per request.
+static CLIENT: LazyLock<Client> = LazyLock::new(|| {
+    Client::builder()
+        .timeout(REQUEST_TIMEOUT)
```
mre: Decided to add this `LazyLock` around the client as a small performance improvement. The price we'd have to pay for it is that we can no longer use the configured request timeout, since `reqwest::Client`'s timeout is fixed at build time on the `ClientBuilder` and `LazyLock` only initializes the client once.

There is a per-request escape hatch (`RequestBuilder::timeout` overrides the client default), so in principle we could keep threading the user's `--timeout` through if it's worth it. Happy to flip it if we'd rather honor `--timeout` here.

thomas-zahner:
> The '0' timestamp tells Wayback to redirect to the newest available snapshot, which is far more reliable than the Availability JSON API (which is heavily rate-limited and is a frequent source of frustration due to flakiness. It also sometimes returns empty archived_snapshots for pages that are clearly archived).

Hmm are you really sure? On my machine this does not seem to be the case. `curl 'https://web.archive.org/web/0/http://example.com/' -v` yields a snapshot from 2002 with accordingly dated website contents.

Ah the `/0/` in the path seems to mean "oldest" version, i.e. snapshot number 0. Getting rid of it should work as intended.

Sidenote; Wikipedia also links a CLI tool called waybackpy which can be used to obtain latest snapshots. This tool [seems to use a different API](https://github.com/akamhy/waybackpy/blob/3b3e78d901a600bb22943202c6a8981ca04a5e48/waybackpy/cdx_api.py#L322) called [Wayback CDX Server API](https://archive.org/help/wayback_api.php). This could also be investigated. But I assume using the simplest approach is the way to go. Especially since I wouldn't expect this CDX Server API to be more reliable than the direct approach.

lychee-lib/src/archive/wayback/mod.rs
thomas-zahner: Hmm, yes I'm okay with not reusing lychee-lib's configured timeout. It is quite a different use case compared to what `--timeout` is normally used for by users. So lets keep this!

lychee-lib/src/archive/wayback/mod.rs
```diff
@@ -1,181 +1,125 @@
...
-    D: Deserializer<'d>,
-{
-    let value: &str = Deserialize::deserialize(deserializer)?;
-    let result = value
-        .parse::<u16>()
-        .map_err(|e| D::Error::custom(e.to_string()))?;
-    StatusCode::from_u16(result).map_err(|e| D::Error::custom(e.to_string()))
+    // After redirect-following, `response.url()` is the resolved
+    // timestamped snapshot URL (e.g. `/web/20020120142510/http://...`).
+    Ok(Some(response.url().clone()))
```
thomas-zahner: We should immediately return the value of the location header without following and fully loading the URL.

This is basically

```
time curl 'https://web.archive.org/web/http://google.com' -L
curl 'https://web.archive.org/web/http://google.com' -L  0.03s user 0.01s system 1% cpu 2.148 total
```

vs.

```
time curl 'https://web.archive.org/web/http://google.com'
curl 'https://web.archive.org/web/http://google.com'  0.05s user 0.01s system 10% cpu 0.676 total
```

Following the URL is about 2-5 times slower. I can take a shot at that if you agree.

katrinafyi: +1. This should just be a `Policy::none()` then checking for 302

lychee-lib/src/archive/wayback/mod.rs
```diff
@@ -1,181 +1,125 @@
...
-                            "timestamp": "20130919044612",
-                            "status": "200"
-                        }
-                    }
-                }
-                "#,
-            "application/json",
-        );
+    use super::*;
+    use std::error::Error as StdError;
```
katrinafyi: nit: it's a bit nicer to avoid renaming by using explicit imports. there's very few imports anyway.
```suggestion
    use http::StatusCode;
    use std::error::Error;
    use url::Url;

    use super::get_archive_snapshot;
```

katrinafyi (commented):
Mostly good! Hopefully it's more reliable.

I also observe the same thing as Thomas - to link to the latest snapshot, the URL should omit the `/0/`. This seems to match what the wikipedia help page says: https://en.wikipedia.org/wiki/Help:Using_the_Wayback_Machine#Latest_archive_copy

mre:
I've rebased and fixed the bug. :) Using the latest snapshot now, not the oldest. So basically dropped the `/0/`.

lychee-lib/src/archive/wayback/mod.rs
mre: Great idea. Fixed.

lychee-lib/src/archive/wayback/mod.rs
mre: True. Updated.

lychee-lib/src/archive/wayback/mod.rs
```diff
@@ -1,186 +1,156 @@
...
-    pub(crate) url: Url,
-    pub(crate) archived_snapshots: ArchivedSnapshots,
-}
-
-#[derive(Debug, Deserialize, Eq, PartialEq)]
-pub(crate) struct ArchivedSnapshots {
-    pub(crate) closest: Option<Closest>,
-}
+/// Per-request timeout for Wayback lookups.
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(20);
```
Souradip121: The timeout was previously passed in from params.cfg.timeout(), meaning users could control it via the --timeout CLI flag. Now it's hardcoded to 20 seconds. Is this intentional? 

Users who set a shorter global timeout (e.g. --timeout 10) might be surprised that Wayback lookups still wait 20 seconds.

lychee-lib/src/archive/wayback/mod.rs
```diff
@@ -1,186 +1,156 @@
...
-        It is flaky because the API does not reliably return snapshots,
-        i.e. the `archived_snapshots` field is unreliable.
-        That's why the test is ignored. For development and documentation this test is still useful."]
+    /// Both cases run in a single test (and therefore a single Tokio runtime)
+    /// on purpose: `get_archive_snapshot` uses a process-wide static
+    /// `reqwest::Client` whose connection pool is bound to the runtime that
+    /// first initializes it. Splitting these into two `#[tokio::test]`s would
+    /// give each its own runtime and let one reuse a pooled connection whose
+    /// runtime has already been torn down, failing with `DispatchGone`.
     #[tokio::test]
```
Souradip121, line 129: The #[ignore] flag is missing wont this would break the CI pipeline if wayback machine is down by any chance. 
Was there any specific reason of not putting down the flag here? @mre

lychee-lib/src/archive/wayback/mod.rs
```diff
@@ -1,186 +1,156 @@
...
-    let value: &str = Deserialize::deserialize(deserializer)?;
-    let result = value
-        .parse::<u16>()
-        .map_err(|e| D::Error::custom(e.to_string()))?;
-    StatusCode::from_u16(result).map_err(|e| D::Error::custom(e.to_string()))
+    // Any other non-success (rate limiting, 5xx, ...) bubbles up so the
+    // caller can decide; the suggestion path currently swallows errors.
+    // Anything else (an unexpected `2xx`) means no usable snapshot.
+    response.error_for_status()?;
+    Ok(None)
```
Souradip121, line 106: The comment says rate limit and other errors "bubble up so the caller can decide", but the caller in check.rs silently swallows them with if let Ok(Some(suggestion)). Should there be a log::warn! so users know why suggestions are missing?

Souradip121 (commented):
@mre Added my review, kindly check and let me know what do you think.

lychee-lib/src/archive/wayback/mod.rs
mre: Oh yeah, great point! I've added it back.

lychee-lib/src/archive/wayback/mod.rs
mre: True! Changed.

mre:
Hey @Souradip121, thanks for your review comments. I've applied the changes now. 👍

lychee-lib/src/archive/wayback/mod.rs
mre: Yeah, I've added the `#[ignore]`. Good call.

mre (commented):
So this turned out to not be much shorter than the previous version, but I'd argue it's still an improvement because we've got less moving parts, and we make link suggestion quicker, because we don't download the full page anymore. (We read the location from the response headers.) Besides, many changes were just comments, so it's fine.

katrinafyi (commented):
Without at least one test for the "real" API, we wouldn't know if Wayback machine makes breaking changes that break this feature. I'd hoped that with this new approach, combined with the status code filtering, the test would be much more reliable and less flaky so we could leave it enabled.

It seems a bit hasty to preemptively ignore the test without a trial period to see if it's really flaky. That said, I think Wayback machine, by its nature, is very unlikely to make breaking changes so I'm not tooo worried. The ignoring just seems like a change made in haste.

mre:
Good point. I'll bring the test back in a new PR.

mre:
This is fixed now in https://github.com/lycheeverse/lychee/pull/2226

thomas-zahner:
Awesome :rocket: 
I do think (and hope) that this endpoint we're using now is more stable than their API endpoint we've previously used

PR #2169, opened 2026-04-25
feat: warn when exclusive '..' syntax is used in accept ranges

### LINKED ISSUE #2178: Range syntax for accepted status codes not working

mre:
Closes lycheeverse/lychee#2178.

The exclusive `..` range syntax follows Rust semantics and excludes the upper bound, which is a common source of confusion (e.g. `301..302` only accepts 301). Emit a warning to make this visible.

This could be promoted to an error or downgraded to a hint once #2021 lands.

### DIFF
--- a/lychee-lib/src/types/accept/range.rs
+++ b/lychee-lib/src/types/accept/range.rs
@@ -76,11 +76,12 @@ impl FromStr for StatusRange {
             if inclusive {
                 Self::new(start, end)
             } else {
-                warn!(
-                    "Accept range '{s}' uses exclusive '..' syntax, which excludes the upper bound ({end}). \
-                     This matches Rust range semantics but is a common source of confusion. \
-                     If you intended to include {end}, use '{start}..={end}' or list codes explicitly (e.g. '{start}, {end}')."
-                );
+                if end == start + 1 {
+                    warn!(
+                        "Accept range '{s}' only matches status code {start}. \
+                         Did you mean '{start}..={end}' or '{start}, {end}'?"
+                    );
+                }
                 Self::new(start, end - 1)
             }
         }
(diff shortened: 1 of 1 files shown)


thomas-zahner:
Hmm actually this is not getting logged :/

The reason being that the logger is initialised only after the options are parsed.

```rust
    let mut opts = LycheeOptions::parse();
    init_logging(&opts.config.verbose(), &opts.config.mode());
```

This would be another reason to add user-hints to lychee lib. I'm not sure how easy it is

mre:
Wellll.... I should have added a cli test for that. Nice catch.

thomas-zahner:
I've now cherry picked it into #2021. The user hints are a nice workaround for this issue. It's nice that these two PRs coincided. I think other workaround wouldn't be so nice. We would have to initialise the logger with some dummy verbosity which would be quite a hack.

I'm closing this in favour of #2021.

mre:
Yes, makes a lot of sense. Than for picking that up.

PR #2170, opened 2026-04-25
ci: fix flaky binstall-check job

mre:
The binstall-check job from #2165 fails for two reasons:

- Unauthenticated api.github.com requests get 403 under contention. Fix by passing GITHUB_TOKEN.
- Release-plz PRs bump the in-tree version to one that is not published yet, so binstall (with --manifest-path .) cannot find a matching release. Drop the flag so binstall resolves the latest published version, and skip the job on release-plz branches as a safety net.

Should unblock #2169 and future release-plz PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,10 +70,20 @@ jobs:
 
   binstall-check:
     runs-on: ubuntu-latest
+    # Skip on release-plz PRs: those bump the version to one that has not been
+    # published yet, so binstall cannot find a matching release. We still want
+    # PR coverage on regular PRs to catch regressions in the binstall metadata.
+    if: ${{ !startsWith(github.head_ref, 'release-plz-') }}
+    env:
+      # Authenticate against api.github.com to avoid 403 rate-limit failures
+      # when binstall queries GitHub releases.
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v6
       - uses: cargo-bins/cargo-binstall@main
-      - run: cargo binstall --manifest-path . --strategies crate-meta-data lychee --no-confirm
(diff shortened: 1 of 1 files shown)


katrinafyi (approved):
Thanks for fixing :) 

Yeah, manifest path is needed but it's still an imperfect test. It's installing the latest release archive (which is likely older than master) using the binstall metadata from master.

I don't know if there's a good way around this.. maybe we can try to install a nightly release archive? It would still be 24 hours behind, but it would let us test with the upcoming release binaries.

thomas-zahner:
> Yeah, manifest path is needed but it's still an imperfect test. It's installing the latest release archive (which is likely older than master) using the binstall metadata from master.

I think this is fine. We basically want to lint/test if binstall can parse our metadata. The archive contents are not really that relevant. I found out that it's possible to specify `--git "file:///$(pwd)"` but that also doesn't seem to be really helpful either. 

In other words, I don't think we would detect any more errors if we test against nightly instead of latest release with `--manifest-path .` as the archive contents should contain a valid binary anyways. But maybe I'm missing something.

Going to merge now. If there actually still is an improvement to be made it can be done in a fallow up PR.

katrinafyi:
With testing nightly, the purpose would be to guard against changes in the release archive structure and to catch this before the official release is pushed. But it's not super urgent and there's probably not many more release changes coming up.

thomas-zahner:
With v0.24.2 [the job fails](https://github.com/lycheeverse/lychee/actions/runs/25220870253/job/73952477307) :/ but maybe rightfully so?

Edit: huh, seems to be fixed after [rerunning the job](https://github.com/lycheeverse/lychee/actions/runs/25220870253/job/73968920039) :shrug:

katrinafyi:
It might be using the version from Cargo.toml trying to install the new version before the binaries are uploaded. Just a guess though.

PR #2171, opened 2026-04-25
readme: add text fragments to features table, other tweaks

katrinafyi:
I wanted to add the new text fragments feature because I think it's a real distinguishing feature (thanks @cristiklein!).

While doing this, I also made some changes which should promote the information which people want to see (and better show off lychee).

- Move Github actions section much earlier, since it's probably a very common use case.
- Hide feature flag information in a details block.
- Mention cargo binstall in pre-build binaries section.
- Add subheadings in the feature table.
- Reorder and group feature table using these subheadings.
- Hyperlinks some :heavy_check_mark: marks to lychee CLI arguments. I remember there was an issue that said they thought fragment checking was enabled by default because of the :heavy_check_mark:.

Uhh let me know what you think. I can slice and subtract things from this PR if you think they're too much.

See here for the rendered readme: https://github.com/rina-forks/lychee/tree/readme-text-frags#features

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,8 @@ outdated information.
 | Basic Auth           | ![yes]  | ![no]         | ![no]    | ![yes]                | ![no]        | ![yes]               | ![no]                 | ![no]  |
 | Custom user agent    | ![yes]  | ![no]         | ![no]    | ![yes]                | ![no]        | ![yes]               | ![no]                 | ![no]  |
 | Relative URLs        | ![yes]  | ![yes]        | ![no]    | ![yes]                | ![yes]       | ![yes]               | ![yes]                | ![yes] |
-| Anchors/Fragments    | ![yes]  | ![no]         | ![no]    | ![no]                 | ![no]        | ![yes]               | ![yes]                | ![no]  |
+| [URL anchor fragments]| ![yes] | ![no]         | ![no]    | ![no]                 | ![no]        | ![yes]               | ![yes]                | ![no]  |
+| [URL text fragments] | ![yes]  | ![no]         | ![no]    | ![no]                 | ![no]        | ![no]                | ![no]                 | ![no]  |
 | Include patterns     | ![yes]️  | ![yes]        | ![no]    | ![yes]                | ![no]        | ![no]                | ![no]                 | ![no]  |
 | Exclude patterns     | ![yes]  | ![no]         | ![yes]   | ![yes]                | ![yes]       | ![yes]               | ![yes]                | ![yes] |
 | Handle redirects     | ![yes]  | ![yes]        | ![yes]   | ![yes]                | ![yes]       | ![yes]               | ![yes]                | ![yes] |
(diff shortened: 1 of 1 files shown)


cristiklein (approved):
I'm obviously biased, but I like this change. I assume that most people bumping into the README file are users and not developers, so getting quicker to usage is positive.

Nit-pick: The latest documentation for [include-fragments](https://lychee.cli.rs/guides/cli/#--include-fragments) is not deployed yet. Until then, clicking on the checkmarks might be a bit confusing.

README.md
```diff
@@ -233,15 +254,28 @@ outdated information.
...
+[7]: https://lychee.cli.rs/guides/cli/#-e---exclude-all-private
+
 [yes]: ./assets/yes.svg
 [no]: ./assets/no.svg
 [maybe]: ./assets/maybe.svg
+
 [custom headers]: https://github.com/rust-lang/crates.io/issues/788
 [filter status code]: https://github.com/tcort/markdown-link-check/issues/94
 [skip private domains]: https://github.com/appscodelabs/liche/blob/a5102b0bf90203b467a4f3b4597d22cd83d94f99/url_checker.go
 [use as library]: https://github.com/raviqqe/liche/issues/13
```
thomas-zahner: Noticed that the link currently doesn't seem to make much sense. Do you agree with my suggestion?

```suggestion
[use as library]: https://docs.rs/lychee-lib/latest/lychee_lib/
```

README.md
```diff
@@ -233,15 +254,28 @@ outdated information.
...
+[6]: https://lychee.cli.rs/guides/cli/#--include-mail
+[7]: https://lychee.cli.rs/guides/cli/#-e---exclude-all-private
+
 [yes]: ./assets/yes.svg
 [no]: ./assets/no.svg
 [maybe]: ./assets/maybe.svg
+
 [custom headers]: https://github.com/rust-lang/crates.io/issues/788
 [filter status code]: https://github.com/tcort/markdown-link-check/issues/94
 [skip private domains]: https://github.com/appscodelabs/liche/blob/a5102b0bf90203b467a4f3b4597d22cd83d94f99/url_checker.go
```
thomas-zahner: Seems to be unused.

```suggestion
```

README.md
```diff
@@ -233,15 +254,28 @@ outdated information.
...
+[5]: https://lychee.cli.rs/guides/cli/#-x---method
+[6]: https://lychee.cli.rs/guides/cli/#--include-mail
+[7]: https://lychee.cli.rs/guides/cli/#-e---exclude-all-private
+
 [yes]: ./assets/yes.svg
 [no]: ./assets/no.svg
 [maybe]: ./assets/maybe.svg
+
 [custom headers]: https://github.com/rust-lang/crates.io/issues/788
 [filter status code]: https://github.com/tcort/markdown-link-check/issues/94
```
thomas-zahner: These two are rather unhelpful / confusing I think. Could we update them to point to our docs and move them to the lychee column as you did with the other 7 flags?

thomas-zahner (changes requested):
Thank you 🚀
I love the `* May need configuration` info :)

I agree with all your changes but noticed some inconsistencies with old links in the feature table. Could you adjust these?

README.md
katrinafyi: Indeed, I've updated them to point to the lychee docs website. For these ones, I linked to the guide pages rather than the CLI options since the guide pages might be a bit friendlier. What do you think about this or should it be CLI options for consistency?

thomas-zahner: No that's great. It does seem more informative and useful to link to the dedicated documentation parts.

thomas-zahner (approved):

katrinafyi:
Thanks for the review @thomas-zahner ! We should also merge https://github.com/lycheeverse/lycheeverse.github.io/pull/152 to get the new `--include-fragments` docs.

thomas-zahner:
Oh right, thanks for reminding :)

PR #2172, opened 2026-04-25
Treat .qmd and .Rmd files as Markdown

### LINKED ISSUE #2157: Treat `.qmd`/ `.Rmd` as markdown
[Quarto](https://quarto.org/docs/get-started/hello/positron.html) and [Rmarkdown](https://rmarkdown.rstudio.com/lesson-2.html) are markdown variants popular with data scientists.

Lychee currently parses `.qmd` (Quarto) and `.Rmd` (RMarkdown) files with the plaintext parser, which leads to some surprising link checking failures if one converts a file from `.md` to `.qmd`, such as:

sample.qmd:

````markdown
---
title: Sample
---

Real link: <https://example.com>.

```{.bash}
curl -O https://cdn.posit.co/r/rhel-10/pkgs/R-${R_VERSION}-1-1.$(arch).rpm
```
````

```bash
$ lychee --dump --verbose sample.qmd
https://example.com/ (sample.qmd) [excluded]
https://cdn.posit.co/r/rhel-10/pkgs/R-$%7BR_VERSION%7D-1-1.$(arch).rpm (sample.qmd)

```

Would you be open to a PR adding `.qmd` and `.Rmd` to the list of [markdown extensions?](https://github.com/lycheeverse/lychee/blob/master/lychee-lib/src/types/file.rs#L144)

Thanks for considering, and for all of your work.

mre:
Quarto (.qmd) and R Markdown (.Rmd) are popular Markdown variants. Lychee was parsing them as plaintext, so URLs inside fenced code blocks were extracted as links, including ones with shell substitutions like $(arch) that fail link checks.

This adds qmd and rmd to the list of Markdown extensions. Case-insensitive lookup already handles .Rmd.

Includes fixtures reproducing the example from the issue and CLI tests confirming that real links are extracted while code-block URLs are skipped by default.

Closes #2157

### DIFF
--- a/lychee-lib/src/types/file.rs
+++ b/lychee-lib/src/types/file.rs
@@ -141,7 +141,7 @@ impl std::fmt::Display for FileType {
 impl FileType {
     /// All known Markdown extensions
     const MARKDOWN_EXTENSIONS: &'static [&'static str] = &[
-        "markdown", "mkdown", "mkdn", "mdwn", "mdown", "mdx", "mkd", "md",
+        "markdown", "mkdown", "mkdn", "mdwn", "mdown", "mdx", "mkd", "md", "qmd", "rmd",
     ];
 
     /// All known HTML extensions
@@ -266,6 +266,9 @@ mod tests {
         assert_eq!(FileType::from("foo.md"), FileType::Markdown);
         assert_eq!(FileType::from("foo.MD"), FileType::Markdown);
         assert_eq!(FileType::from("foo.mdx"), FileType::Markdown);
+        assert_eq!(FileType::from("foo.qmd"), FileType::Markdown);
+        assert_eq!(FileType::from("foo.Rmd"), FileType::Markdown);
+        assert_eq!(FileType::from("foo.rmd"), FileType::Markdown);
 
         // Test that a file without an extension is considered plaintext
         assert_eq!(FileType::from("README"), FileType::Plaintext);
--- a/lychee-lib/src/types/file.rs
+++ b/lychee-lib/src/types/file.rs
@@ -140,8 +140,10 @@ impl std::fmt::Display for FileType {
 
 impl FileType {
     /// All known Markdown extensions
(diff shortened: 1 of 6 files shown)


lychee-lib/src/types/file.rs
```diff
@@ -140,8 +140,10 @@ impl std::fmt::Display for FileType {
 
 impl FileType {
     /// All known Markdown extensions
+    // NOTE: Stored in reverse-popularity order because the `Iterator` impl for
+    // `FileExtensions` pops from the end
```
katrinafyi: This comment really makes me raise an eyebrow at `impl Iterator for FileExtensions`. Conceptually, I do not think FileExtensions should be a stateful iterator that gets mutated. 

Is iterating over FileExtensions even necessary? If iteration is needed, it should delegate to Vec's iterator, or there can just be a function that returns a slice of strings.

katrinafyi (commented):
The functional change set is fine but this comment and its implications are eyebrow raising. It doesn't need to be fixed in this PR, but also might as well.

lychee-lib/src/types/file.rs
mre: Yeah, true, that crossed my mind as well. Didn't want to change too much at once, but I could change that in this PR.

katrinafyi:
I've made a commit at https://github.com/lycheeverse/lychee/compare/treat-qmd-rmd-as-markdown...rina-forks:lychee:treat-qmd-rmd-as-markdown to fix up the iterator.

mre:
Thanks, @katrinafyi, I've cherry-picked your commit into the branch. Looks fine to me. 😊
I also dropped the obsolete "stored in reverse-popularity order because the iterator pops from the end" comment.
One more thing that I noticed was that the README still contained the old ordering. I updated that.

katrinafyi (approved):
Looks good!

mre:
Thanks for the support.

PR #2182, opened 2026-05-02
ci: move binstall-check to on release and use `/nightly/` URL

katrinafyi:
running in the release-binary workflow and after the build-release step means that the release binaries will always be available when the binstall check runs.

for nightly releases, we also change the URL to `/nightly/` which lets us preview upcoming archive structure changes in the next version.

should fix https://github.com/lycheeverse/lychee/pull/2170#issuecomment-4360450732, but i can't test it.

moving the check into release-binary might make a failing check less visible, but maybe that's okay. we should keep an eye on https://github.com/lycheeverse/lychee/actions?query=event%3Aschedule+event%3Arelease or add the badges somewhere visible, maybe in the release PR text.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,24 +68,6 @@ jobs:
       - name: Verify the MSRV
         run: make verify
 
-  binstall-check:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


mre:
Thanks! Let's try it.

PR #2183, opened 2026-05-03
chore: generate artifact attestations for release assets

scop:
https://docs.github.com/en/actions/concepts/security/artifact-attestations

This makes use of workflow artifacts to accomplish generating a single combined attestation for all release assets. Generating one per asset is kind of an antipattern, could at present generate one for successful matrix entries even when some others in it fail, as well as require id-token and attestations permission for the build process. Doing it this way with workflow artifacts is a bit clunky, but it currently serves as a widely used workaround for inability to nicely set and consume multiple outputs from matrix builds.

Python wheel builds already have the attestations, per the maturin generated workflow.

Caveat: untested, but I suppose it could work. Anyway as the asset job is a separate one after release builds, a failure in it should not prevent GitHub release from succeeding.

### DIFF
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -339,6 +339,12 @@ jobs:
           # uploads idempotent across reruns or overlapping triggers.
           gh release upload "$version" ${{ env.ASSET }} ${{ env.ASSET_SUM }} --clobber
 
+      - name: Upload release asset digest
+        uses: actions/upload-artifact@v7
+        with:
+          path: ${{ env.ASSET_SUM }}
+          archive: false
+
   binstall-check:
     needs: [build-release]
     runs-on: ubuntu-latest
@@ -356,3 +362,26 @@ jobs:
       # Use `--manifest-path .` so binstall validates the in-tree binstall metadata.
       - run: cargo binstall --manifest-path . --strategies crate-meta-data lychee --no-confirm --verbose
 
+  attest:
+    name: attest
+    needs: [build-release]
+    runs-on: ubuntu-latest
+    permissions:
+      attestations: write # to write attestations
+      id-token: write # to sign attestations
+    steps:
+      - name: Download release asset digests
+        uses: actions/download-artifact@v8
+        with:
+          path: release-asset-digests/
+          merge-multiple: true
+
+      - name: Combine release asset digests
+        shell: bash
+        run: |
+          cat release-asset-digests/*.sha256 >release-asset-digests.sha256
+
+      - name: Generate release asset artifact attestations
+        uses: actions/attest@v4
+        with:
+          subject-checksums: release-asset-digests.sha256
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -378,7 +378,9 @@ jobs:
     permissions:
       attestations: write # to write attestations
       id-token: write # to sign attestations
-      artifact-metadata: write # required by actions/attest@v4
+      # artifact-metadata: write is documented as required by actions/attest@v4.
+      # However, it does not actually seem to be required in our use of the action,
+      # and the documentation around it is quite terse, so omitting for now.
     steps:
       - name: Download release asset digests
         uses: actions/download-artifact@v8
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -294,7 +294,11 @@ jobs:
         if: startsWith(matrix.os, 'windows')
         run: |
           7z a "$ARCHIVE.zip" "$ARCHIVE"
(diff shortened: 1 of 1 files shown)


mre:
Sorry for leaving this hanging for so long.

Might need a rebase.

> Caveat: untested, but I suppose it could work. Anyway as the asset job is a separate one after release builds, a failure in it should not prevent GitHub release from succeeding.

Question: how can this be tested?

scop:
Rebased.

Re testing, at least just pushing a tag to my private fork doesn't seem to yield anything useful, there are a bunch of unrelated things failing. I guess it could be acceptable to just go with this and see how it fares on next release, the attestation job is the last in the release binary workflow, so even if it fails, it does not prevent publishing of the binaries. I'm not familiar offhand with the whole lychee release pipeline nor have I checked if a failing release-binary could affect some other parts of the release pipeline adversely.

mre:
Thanks for the rebase and for your contribution.

Yeah, that makes sense. Let's give it a shot. We can always make modifications in case it doesn't work as expected. 😊

mre:
Hold on, before I merge, I wondered about a few things.

1. What about Windows support?

Windows `.sha256` files aren't in `shasum` format, or?
At least, the `attest` README is explicit about `subject-checksums`:

> The file referenced by the `subject-checksums` input must conform to the same format used by the shasum tools. Each subject should be listed on a separate line including the hex-encoded digest (either SHA256 or SHA512), a space, a single character flag indicating either binary (`*`) or text (` `) input mode, and the filename.

For Unix jobs, `shasum -a 256` works. But our Windows job uses:

```bash
certutil -hashfile "$ARCHIVE.zip" SHA256 > "$ARCHIVE.zip.sha256"
```

`certutil` emits a three-line block with a header/footer and no filename:

```
SHA256 hash of lychee-x86_64-pc-windows-msvc.zip:
e3b0c44298fc1c149afbf4c8996fb924...
CertUtil: -hashfile command completed successfully.
```

When the attest job does `cat release-asset-digests/*.sha256`, the Windows entry won't parse. So the Windows `.zip` ends up unattested (and the malformed lines may error the step). I believe the fix is to normalize the Windows checksum file to `sha256sum` format (e.g. by changing it to `sha256sum` as well).

2. The aarch64-macos job uploads two space-joined files with `archive: false`

For the Apple-Silicon build, `ASSET_SUM` is set to two values:

```bash
echo "ASSET_SUM=${ASSET_SUM} $ARCHIVE.dmg.sha256" >> $GITHUB_ENV
```

That results in 
```
"lychee-aarch64-apple-darwin.tar.gz.sha256 lychee-aarch64-apple-darwin.dmg.sha256"
```

The new upload step uses `path: ${{ env.ASSET_SUM }}`, which renders as a single line with a space in it. But `archive: false` supports only a single file, and `path` treats that line as one glob pattern (filenames have no spaces, so it matches nothing, which runs into `if-no-files-found: warn`, which means nothing uploaded). The net effect is that the macOS DMG and its tar.gz digests for that target silently won't be attested (from what I can tell). If this is true, this needs either a separate upload per file, `archive: true` with a unique `name:`, or splitting the paths onto newlines.

3. What about permissions?

I checked the `attest@v4` README and it lists three permissions:

```yaml
id-token: write
attestations: write
artifact-metadata: write
```

The PR sets only the first two. Was that on purpose?

`artifact-metadata: write` is documented as needed for storage records, which are only created when `push-to-registry: true` (not the case here), so it's *probably* not strictly required for `subject-checksums`.

mre:
I took a stab at it in https://github.com/scop/lychee/pull/1. If you merge that one, it will update your PR to lychee.

Please review carefully as I'm not an expert in GitHub action workflows and also don't know much about checksums or Windows internals. 😅 There's a chance that you thought about all of these things already, and I'm just overcomplicating things. If so, please let me know.

scop:
Yes, omitting `artifact-metadata` is intentional. I've never seen it needed in practice in basic uses like this, and the documentation around it is a bit lacking.

Good catch re non-linux digest formats, I did not do my homework on them. I'll have a look.

scop:
Heh, I was a bit hasty and jumped to fixing this up myself, missing the PR you had filed. I came up with pretty much exactly the same changes as you did, apart from the `artifact-metadata` addition. So with https://github.com/scop/lychee/pull/1#discussion_r3368936570 I'm good with this, thanks for having a look!

mre:
Cool. No worries. 😊 So if I understand correctly, you would my branch into yours now, right? I've applied your suggestion in my branch, so you can merge it if you like.

scop:
Yeah, that's how I gather it goes. Merged your PR into mine now. This is the first time I've done things this way, but it works actually fine I think :)

mre (approved):

mre:
Sorry for the long delay; this slipped through the cracks. It looks good, so merging right now.

PR #2184, opened 2026-05-04
chore: strip the release debug info

kemingy:
As we discussed in https://github.com/lycheeverse/lychee/pull/1931#issuecomment-4365754245

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,3 @@ resolver = "2"
 
 [workspace.package]
 version = "0.24.2"
-
-[profile.release]
-debug = true


mre (approved):
Nice. Thanks!

thomas-zahner:
Guess this makes sense as this simply uses Cargo's default behaviour.

@mre What was the reason for enabling the debug profile in release mode? Do we get additional debug information in the stack trace when lychee panics?

mre:
Yeah, I wondered about that, too. Apparently, I introduced it when I refactored the extractor from a pure function into its own struct. Not sure why I enabled the debug symbols in general. Probably because I benchmarked during that refactor to make sure there were no regressions. I believe they need to be enabled so that cargo-flamegraph can show the function names in the chart. But not 100% sure. Either way, it's wasteful to enable that for release builds without a good reason.

thomas-zahner:
Ah that makes sense :+1:

PR #2188, opened 2026-05-07
refactor: add generic Cache to deduplicate in-progress tasks

katrinafyi:
This introduces a new generic Cache which understands and deduplicates "in-progress" tasks according to a given key. See the docs for the new types: https://cache-upgrade.lychee-docs-katrinafyi.pages.dev/lychee_lib/async_lib/cache/

When a key is requested concurrently by different tasks, exactly one task is designated the leader and is responsible for setting the value. Other tasks wanting the same key are called followers and should await until the leader computes the value. 

This supersedes and closes https://github.com/lycheeverse/lychee/pull/2107. Compared to the earlier PR, this chooses the fragment checker cache as the first use-case for this new cache, rather than the more complicated top-level request cache.

I'd like feedback on naming and the API design. Should we change this Cache struct's name, should we rename other cache types? Should the cache provide a method for checking whether a key is in the cache but without locking the entry?


lychee-lib/src/utils/fragment_checker/mod.rs
```diff
@@ -107,14 +103,14 @@ impl FragmentBuilder {
 /// a `HashSet` of fragments as the value.
 #[derive(Default, Clone, Debug)]
 pub(crate) struct FragmentChecker {
-    cache: Arc<Mutex<HashMap<String, HashSet<String>>>>,
+    cache: Cache<String, HashSet<String>>,
```
mre, line 106: This looks cleaner, thanks.

Do we even need to make the cache generic, though? We only use it here. Alternatively, shall we create a type alias for that specific cache? Then again, I'm not a huge fan of aliases. Perhaps a newtype wrapper around the cache value is enough...

Just thinking out loud here. The goal is to make this even nicer, although it's already an improvement.

lychee-lib/src/cache.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn set(mut self, value: T) -> CacheFollower<T> {
+        let arc = std::mem::take(&mut self.0);
+        let _ = arc.set(value);
+        CacheFollower(arc)
+    }
+
+    /// Returns a new dissociated [`CacheLeader`]. That is, a setter which is
+    /// not backed by any value within the cache. This can be useful to let
+    /// uncacheable entities use the same cache-handling logic.
+    pub fn dissociated() -> Self {
```
mre: Is dissociated a technical term? I'm not a native speaker and all that I could find were references to psychology. Maybe "detached" or "deferred" could be alternatives worth considering?

lychee-lib/src/cache.rs
```diff
@@ -0,0 +1,187 @@
...
+        Self(Arc::default(), None)
```
mre: That looks a lot like Self's default value.

lychee-lib/src/cache.rs
```diff
@@ -0,0 +1,187 @@
...
+        let arc = std::mem::take(&mut self.0);
+        self.1 = None;
+        CacheLeader(arc, Some(default))
+    }
+
+    /// Writes the given value into the cache, consuming this [`CacheLeader`] and
+    /// returning a [`CacheFollower`] referencing the stored value.
+    pub fn set(mut self, value: T) -> CacheFollower<T> {
+        let arc = std::mem::take(&mut self.0);
+        let _ = arc.set(value);
```
mre: Can we propagate the `SetOnceError` to the call-site? I understand that setting it more than once should never happen but if it does, we have a data race. Better to surface that somehow. If returning an error is too viral, I'd be okay with an explicit panic instead.

lychee-lib/src/cache.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn dissociated() -> Self {
+        Self(Arc::default(), None)
+    }
+}
+
+/// Drop implementation that calls the stored [`CacheLeader::with_fallback`]
+/// closure, if it is configured and no value has been manually stored.
+impl<T, Fn: FnOnce() -> T> Drop for CacheLeader<T, Fn> {
+    fn drop(&mut self) {
+        if let Some(f) = self.1.take()
```
mre: Can we rename f to `fallback_fn` or so to make this easier to understand?

mre: ...and sorry if that's a stupid question, but why do we set the value on drop?

lychee-lib/src/cache.rs
```diff
@@ -0,0 +1,187 @@
+//! Facility for cached asynchronous operations, with operations keyed by a
+//! key type and ensuring mutual exclusion of operations with the same key.
```
mre: Should we also explain why we have this cache given that we also have a per-host cache? I know that it's generic, but at the moment we use it only for fragments, right?

mre (commented):
That API looks pretty ergonomic. Good job!

mre:
Since I saw your recent rename commit, I wanted to propose CacheSetter/CacheGetter as well. Leader/Follower sounds a bit strange to me. Or maybe Publisher/Subscriber similar to Redis PubSub? 🙃

lychee-lib/src/cache.rs
katrinafyi: Logically, the SetOnceError is impossible because the DashMap will only let one task see a Vacant entry and insert it, but I don't think this is something which Rust can understand in its type system. I'll see what I can do to improve the error logic.

lychee-lib/src/cache.rs
katrinafyi: It's possible that the user drops the CacheLeader/CacheSetter without calling `set`. It's rare, but possible in case of user error or if a panic happens. In that case, we want to set a value so that the CacheFollower/CacheGetters won't deadlock forever. It's commented here: https://github.com/rina-forks/lychee/blob/894d3fe3489dda8b8f9b57c2480c0ef3a5d7b7e4/lychee-lib/src/cache.rs#L43-L46

To be honest, this is my least liked part of this code. I'll have a think about what else we can do.

lychee-lib/src/cache.rs
katrinafyi: At the moment, yeah, but the idea with the new Cache type is that it would be used in all cases where we need a cache that prevents duplicated in-progress work.

For instance, it should be used to replace the `HostCache` + `active_requests` logic in the host struct. I planned to do this in a later PR to keep this PR small, but I can do it here if you think it makes sense to do it together.

lychee-lib/src/utils/fragment_checker/mod.rs
katrinafyi: See other comment.

katrinafyi (commented):
Thanks!

lychee-lib/src/cache.rs
mre: Ah, cool! Yeah, replacing `HostCache` would be great. No need to do it here; can be a separate PR. Thanks for the explanation.

lychee-lib/src/cache.rs
katrinafyi: Big changes to the error handling in https://github.com/lycheeverse/lychee/compare/894d3fe...c3ede4d , as well as other review comments. The error handling for a dropped setter is now propagated as an error on the getter side, which I think is much nicer.

eread:
Out of curiosity, @katrinafyi, would this result in a performance boost for Lychee?

katrinafyi:
As it is, this PR shouldn't have much effect on performance, it's mostly a code refactor. It's also comparable to the pre-existing cache from https://github.com/lycheeverse/lychee/pull/2067 (I think that PR would've improved performance when it was merged).

The main benefit kf this PR is being versatile enough to use in different places. The addition of a generic in-progress cache could lead to performance increases when applied to things which aren't currently cached. 

Off the top of my head, I think there's currently no deduplication of in-progress local file checks, so it could have a benefit there. But it's hard to say *how much* of an impact it will have, since it only has an effect when identical URLs appear amongst the links to be checked *and* when these identical URLs are racing.

lychee-lib/src/async_lib/cache/iter.rs
```diff
@@ -0,0 +1,99 @@
+//! Types which implement the iterators for [`super::Cache`].
```
katrinafyi: Looking back at this PR, I now think these iterator implementations are kinda overkill.

We could probably just use a `Box<dyn Iterator>`. It's a touch slower but iterating a cache should be very rare anyway. WDYT?

mre: If it makes the code easier, sure.

katrinafyi: Done in https://github.com/lycheeverse/lychee/pull/2188/commits/776c24676da06e7de2f6649c28b1622a2fd5089c

mre (approved):

mre:
This is in good shape now. Shall we merge it?

mre:
I'll go ahead and merge this. 👍 It's still clean on top of `master` and as far as I can tell, there are no open discussion points. Thanks, @katrinafyi. ✌️

PR #2194, opened 2026-05-12
fix: Use GitHub API for certain fragment checks

gittihub-jpg:
Closes #2128

## Summary
Routes specific GitHub fragment checks through the REST API when a `GITHUB_TOKEN` is set, resolving implicit README and directory-view fragments without downloading full pages.

## Special notes
- `lychee-lib/src/quirks/mod.rs`: Intercepts matching GitHub URLs and rewrites them to the corresponding REST API endpoints.
- `lychee-lib/src/client.rs`: Attaches the `GITHUB_TOKEN` as a Bearer header for requests targeting `api.github.com`.
- `lychee-lib/src/utils/fragment_checker/mod.rs`: Parses JSON responses from these API calls and maps them to standard HTTP status codes.
- `lychee-bin/tests/cli.rs`: Integration tests covering all four supported fragment types.

- [x] Backwards compatibility

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2345,6 +2345,7 @@ version = "0.24.2"
 dependencies = [
  "async-stream",
  "async-trait",
+ "base64",
  "cookie_store",
  "dashmap 6.1.0",
  "doc-comment",
--- a/lychee-lib/Cargo.toml
+++ b/lychee-lib/Cargo.toml
@@ -14,6 +14,7 @@ rust-version = "1.88.0"
 
 [dependencies]
 async-stream = "0.3.6"
+base64 = "0.22"
 async-trait = "0.1.88"
 cookie_store = "0.22.1"
 dashmap = { version = "6.1.0" }
@@ -57,6 +58,7 @@ reqwest_cookie_store = { version = "0.10.0", features = ["serde"] }
 # https://github.com/Homebrew/homebrew-core/pull/70216
 ring = "0.17.14"
 secrecy = "0.10.3"
+serde_json = "1.0.149"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_with = "3.19.0"
 shellexpand = "3.1.2"
--- a/lychee-lib/src/checker/website.rs
+++ b/lychee-lib/src/checker/website.rs
@@ -313,6 +313,12 @@ impl WebsiteChecker {
             return status;
         }
 
+        // If the fragment checker already determined the fragment is invalid,
+        // trust that result — don't override it with a generic repo-exists check.
+        if matches!(status, Status::Error(ErrorKind::InvalidFragment(_))) {
+            return status;
+        }
+
         if let Ok(github_uri) = GithubUri::try_from(uri) {
             let status = self.check_github(github_uri).await;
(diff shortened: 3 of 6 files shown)


katrinafyi:
I see it's draft so lmk if you're still working on it, but I don't like the idea of using a magic parameter to indicate the URLs. I think it's quite a hack. What're your thoughts about this?

gittihub-jpg:
Hi, I wanted to fix the CI failures before marking it for review. Also you are right, the magic parameter could be confusing later. I used it to avoid changing the existing interfaces, but I'm open to refactoring. Would you prefer a typed wrapper struct, or an enum variant passed alongside the URL? Of course possible to redesign before marking this ready.

katrinafyi:
From my initial task scoping, I don't think any enums or wrapper structs should be needed. Have you found that they would be needed? The API URL can just be treated as a normal URL. Attaching the token should also not be duplicated, it can be done through a RequestChain item. 

I'm also interested to learn, do you think there's a need to break out all the quirk functions into helper functions? Also, is it ok that `needless_pass_by_value` is allowed in all those functions?

gittihub-jpg:
I initially added the magic query params to pass extra context (fragment, branch) through the call stack without changing function signatures — but you're right, it's a hack. Treating the API URL as a plain URL and routing purely by pattern matching is cleaner, I'll go that route. 
For the token, using a RequestChain item makes much more sense than duplicating the attachment logic — I'll refactor that. On the helper functions: I split them out because each pattern hits a different API endpoint with different response parsing, and lumping it together felt harder to follow. Happy to keep them if that reads better to you too, or collapse if you'd prefer. 
The needless_pass_by_value allows: the clippy lint fires because the quirk functions take owned String arguments that could be &str. I allowed it to avoid touching the existing function signatures, but the cleaner fix is just switching to &str parameters — I'll do that instead of silencing the lint.

After your response Ill refactorre whats been discussed and fix the CI failures.

katrinafyi (changes requested):
Yeah I'd prefer you fix all of the issues I've mentioned. After that, also fix the CI issues?

gittihub-jpg:
Hi sorry for the delay, im pretty busy currently so it might take some time but ill get it done this week.

katrinafyi:
No worries! I should mention that, if this is a feature you want to use, it might be ready sooner if one of the maintainers works on it - it's a somewhat complicated task. But if you just want to learn, you're welcome to keep working on it and I'll review.

gittihub-jpg:
My goal is to learn here so I will try myself, if it is too much Ill request help.

katrinafyi:
I must admit... I'm not thrilled about the state of this PR. It needs a lot of work, but I'm not going to repeatedly prompt an LLM through GitHub comments. I can't tell if you genuinely understand anything I've said, and I don't think you have a mental understanding of the problem, the solution, or the problems with your solution. 

It's my opinion that using an LLM is a very bad way to learn. If you are genuinely trying to learn, I recommend choosing a smaller task, doing it *yourself*, and building up. I'm still generally happy to help—but absolutely not if you're going to paste me into the LLM.

PR #2197, opened 2026-05-18
Use hints for root-relative links error

### LINKED ISSUE #2195: UX improvement for path-relocatable sites?
I maintain `testers.lycheeLinkCheck` in Nixpkgs. Recently I've had some trouble with mdbook sites and I noticed that if `--root-dir` is not passed, lychee will report errors like:

```
Cannot resolve root-relative link '/': To resolve root-relative links in local files, provide a root dir
```

That's correct, but not great for sites that need to actively avoid root-relative links.

Would you be interested in a PR that allows this intent to be declared in the options?

```
 --relocatable
    Declare that the site must be mountable onto any path.
    While this is the de facto default behavior, this option makes the error message explicit about the requirement.
    This is the opposite of specifying `--root-dir`; they're mutually exclusive.
```

### New UX

1. Instead of failing, ask the user to elaborate:
```
Don't know how to resolve root-relative link '/'! If your site must be relocatable to any path, pass `--relocatable` or `relocatable = true` in the configuration. Otherwise, pass `--root-dir`.
```

2. Either success, or, if `relocatable`:
```
Root-relative links (starting with '/') are not allowed because this site should be relocatable to any path.
```

### Context

I've prototyped this idea in https://github.com/NixOS/nixpkgs/pull/520792 and that PR would solve the immediate problem for me, but I'd much prefer a built-in solution of course.

### So

Does this match your design philosophy for lychee?

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/2195 by moving the suggestion into a user hint and adding a sentence about possible impacts on relocatable sites.

New error looks like:
```

[a.md]:
 [ERROR] error: (at 4:1) | Cannot resolve root-relative link '/a'

🔍 3 Total (in 60ms) 🔗 3 Unique ✅ 0 OK 🚫 3 Errors

Hint: Unresolved root-relative links in local files. To resolve root-relative links to a local directory, provide a root dir. Alternatively, if the site should be relocatable, consider using locally-relative links.
```

Happy to tweak any part of the phrasing.

I've kept it as "locally-relative" (rather than @ roberth's suggestion of page-/document-relative), because I think it's more straightforward when it refers to files on disk. Page and document both need the user to think about web pages before connecting it a file path. MDN calls them [current directory relative](https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references#current_directory_relative). That's a bit long but I also wouldn't mind "directory-relative".

### DIFF
--- a/lychee-bin/src/hints.rs
+++ b/lychee-bin/src/hints.rs
@@ -123,7 +123,7 @@ fn root_relative_links(stats: &ResponseStats) {
         hint!(
             "Unresolved root-relative links in local files. \
             To resolve these links to a local directory, provide a root dir with `--root-dir`. \
-            Alternatively, if the site should be relocatable, consider using locally-relative links."
+            Alternatively, if the site should be relocatable, consider using directory-relative links instead."
         );
     }
 }
--- a/lychee-bin/src/hints.rs
+++ b/lychee-bin/src/hints.rs
@@ -122,7 +122,7 @@ fn root_relative_links(stats: &ResponseStats, _config: &Config) {
     if any_root_relative_links {
         hint!(
             "Unresolved root-relative links in local files. \
-            To resolve these links to a local directory, provide a root dir. \
+            To resolve these links to a local directory, provide a root dir with `--root-dir`. \
             Alternatively, if the site should be relocatable, consider using locally-relative links."
         );
     }
--- a/lychee-bin/src/hints.rs
+++ b/lychee-bin/src/hints.rs
@@ -1,5 +1,5 @@
 use http::StatusCode;
-use lychee_lib::{ErrorKind, Status, StatusCodeSelector, hint};
+use lychee_lib::{ErrorKind, RequestError, Status, StatusCodeSelector, hint};
 
 use crate::{config::Config, formatters::stats::ResponseStats};
 
@@ -10,6 +10,7 @@ pub(crate) fn handle_stats(stats: &ResponseStats, config: &Config) {
     any_redirects(stats, config);
     rejected_status_codes(stats, config);
     unfollowed_redirects(stats, config);
+    root_relative_links(stats, config);
 }
 
 fn rate_limit(stats: &ResponseStats, config: &Config) {
@@ -106,6 +107,23 @@ fn unfollowed_redirects(stats: &ResponseStats, config: &Config) {
             "Rejected redirectional status codes. \
              This means some redirects were not followed. \
              You might want to increase the limit for `-m`/`--max-redirects`."
+        );
+    }
+}
+
+fn root_relative_links(stats: &ResponseStats, _config: &Config) {
+    let any_root_relative_links = stats.error_map.values().flatten().any(|r| match &r.status {
+        Status::RequestError(RequestError::CreateRequestItem(_, _, e)) => {
+            matches!(**e, ErrorKind::RootRelativeLinkWithoutRoot(_))
+        }
+        _ => false,
+    });
+
+    if any_root_relative_links {
+        hint!(
+            "Unresolved root-relative links in local files. \
+            To resolve these links to a local directory, provide a root dir. \
+            Alternatively, if the site should be relocatable, consider using locally-relative links."
         );
     }
 }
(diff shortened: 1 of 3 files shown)


lychee-bin/src/hints.rs
```diff
@@ -109,3 +110,20 @@ fn unfollowed_redirects(stats: &ResponseStats, config: &Config) {
         );
     }
 }
+
+fn root_relative_links(stats: &ResponseStats, _config: &Config) {
```
thomas-zahner: I'd simply remove `_config`. We currently don't have a trait enforcing it to be specified. If we did in the future we could still add it again.

lychee-bin/src/hints.rs
```diff
@@ -109,3 +110,20 @@ fn unfollowed_redirects(stats: &ResponseStats, config: &Config) {
...
+        Status::RequestError(RequestError::CreateRequestItem(_, _, e)) => {
+            matches!(**e, ErrorKind::RootRelativeLinkWithoutRoot(_))
+        }
+        _ => false,
+    });
+
+    if any_root_relative_links {
+        hint!(
+            "Unresolved root-relative links in local files. \
+            To resolve these links to a local directory, provide a root dir. \
```
thomas-zahner: We are in lychee-bin, I feel that it makes a lot of sense to mention the specific flag

```suggestion
            To resolve these links to a local directory, provide a root dir with `--root-dir`. \
```

lychee-bin/src/hints.rs
```diff
@@ -109,3 +110,20 @@ fn unfollowed_redirects(stats: &ResponseStats, config: &Config) {
...
+            Alternatively, if the site should be relocatable, consider using locally-relative links."
```
thomas-zahner: Maybe?

```suggestion
            Alternatively, if the site should be relocatable, consider updating the link to be directory-relative."
```

Personally, I find "locally-relative link" quite confusing. The term "directory-relative" you cited from MDN is much more obvious to me. When thinking about URLs "directory" makes sense to me (oh it has to do with the path and slashes as in `/a/b/c.txt`), while I don't associate "locally" with anything URL related. But we could further debate this if you disagree.

Then I'm suggesting to add `updating the link` or something similar. It wasn't immediately clear to me that this is about changing/updating the URL itself when reading the hint for the first time.

Or maybe we could even add an example?
We even know what the non-resolvable root-relative URL is at this point.
Would it be possible to do the same thing as in `fn rate_limit` where I'm showing ```You might be able to work around this by adding `[hosts.\"{domain}\"]` to the TOML config```?

So if we failed on `/a` we could even explain, that `--root-dir https://example.com` would make it resolve to `https://example.com/a`. And we could also add an example to what locally-relative / directory-relative is with the specific error value. -> ```Alternatively, if the site should be relocatable, consider updating the link to be directory-relative. (e.g. ../a instead of just `/a`)```

katrinafyi: All good ideas. What about "Alternatively, if the site should be relocatable, consider using directory-relative links instead."? I think that "updating the link" is a bit wordy.

I'm not sure we can add a specific example, since we do not know what the correct suggestion would be (indeed, that's the whole cause of the relative link error). We could add a generic example, but I think that the message risks becoming too long.

thomas-zahner: I see your point of being quite generic when adding the example. I do think it's okay to be a bit more verbose with user hints in comparison to "normal" logging (error, warn, info) as it's intended to be a practical assistance for users and can be disabled with e.g. `-q`. But of course being overly generic doesn't help. Let's go with your suggestion then :+1:

katrinafyi: True, a longer message is fine in this context. I guess it comes down to whether we think that a user would understand "directory-relative" without the clarification. I would assume they could work out the meaning, especially when it is juxtaposed with "root-relative". Or, there is always google, so it's good that we use MDN's terminology so it will show up :).

I've made the changes from your review comments, thanks!

thomas-zahner: I agree. Especially as we mention `--root-dir` specifically, the user will be able to check out what this means via `man lychee` or `lychee --help` and get further help.

thomas-zahner:
@katrinafyi Thank you for the PR and the extensive discussion on #2195! :rocket:

PR #2200, opened 2026-05-19
Credit NLnet

thomas-zahner:


mre:
Thanks for adding that. Renders nicely, too. 😊

mre:
As a follow-up, it would be cool to support the dark theme as well. Right now the "nl" is a bit hard to read on my screen. But, of course, that would just be a potential improvement in the future.

PR #2203, opened 2026-05-20
fix: count transport-layer failures in per-host statistics

### LINKED ISSUE #2202: Per-host statistics report 100% success rate when requests fail before receiving an HTTP response
## Summary

The per-host statistics table (introduced in #1929, shipped in v0.23.0 and unchanged through v0.24.2 / current `master`) under-counts requests for hosts whose links fail at the transport layer (DNS lookup, TCP connect, TLS handshake, "Connection failed"). Those failures never reach `HostStats::record_response`, so they're invisible to `total_requests`, the per-bucket error counters, and the derived `success_rate`. The hosts whose links genuinely failed get reported as 100% successful.

## Reproduction

A lychee run on our docs site failed with the following errors:

```
### Errors in build/site/.../configuration.html
* [ERROR] <https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/> (at 1040:127) | Connection failed. Check network connectivity and firewall settings

### Errors in build/site/.../introduction.html
* [ERROR] <https://kubernetes.io/docs/concepts/extend-kubernetes/operator/> (at 397:17) | Connection failed. Check network connectivity and firewall settings
```

The same report's `Per-host Statistics` section reports:

```
| Host          | Requests | Success Rate | Median Time | Cache Hit Rate |
| kubernetes.io | 74       | 100.0%       | 109ms       | 41.2%          |
```

74 requests, 100% success, even though two other requests to the same host failed. The two failures are absent from `Requests` (the real number should be 76) and absent from any error bucket, so the success-rate denominator never sees them.

## Root cause

`lychee-lib/src/ratelimit/host/host.rs::perform_request` (lines 175-197 in v0.24.2):

```rust
async fn perform_request(...) -> Result<CacheableResponse> {
    let start_time = Instant::now();
    let response = match self.client.execute(request).await {
        Ok(response) => response,
        Err(e) => {
            return Err(ErrorKind::NetworkRequest(e));  // <-- early return
        }
    };
    self.update_stats(response.status(), start_time.elapsed());
    ...
}
```

## Suggested fix

Add a method on `HostStats` to record outcomes that didn't produce a status code, and call it from the `Err` arm of `perform_request`. For example:

```rust
// lychee-lib/src/ratelimit/host/stats.rs
impl HostStats {
    /// Record a request that failed before receiving an HTTP response
    /// (DNS, TCP, TLS, connection reset, etc.)
    pub fn record_network_error(&mut self, request_time: Duration) {
        self.total_requests += 1;
        // Could also introduce a dedicated `network_errors` counter so the
        // failure mode shows up in the per-host serialization.
        self.request_times.push(request_time);
    }
}
```

```rust
// lychee-lib/src/ratelimit/host/host.rs::perform_request
Err(e) => {
    self.stats
        .lock()
        .unwrap()
        .record_network_error(start_time.elapsed());
    return Err(ErrorKind::NetworkRequest(e));
}
```

With that, our `kubernetes.io` row would correctly report `74 / 76 ≈ 97.4%` instead of `100%`, and the per-host table would once again be trustworthy for spotting suspicious hosts. A separate `network_errors` counter (exposed in the markdown formatter) would make the failure mode explicit. For example:

Before (current):

```
| Host          | Requests | Success Rate | Median Time | Cache Hit Rate |
| kubernetes.io | 74       | 100.0%       | 109ms       | 41.2%          |
```

After (with `network_errors` column):

```
| Host          | Requests | Success Rate | Network Errors | Median Time | Cache Hit Rate |
| kubernetes.io | 76       | 97.4%        | 2              | 109ms       | 41.2%          |
```

afalhambra-hivemq:
Fixes #2202.

## Problem

The per-host statistics table reports an inaccurate success rate for hosts whose links fail at the transport layer (DNS lookup, TCP connect, TLS handshake, connection reset, etc.). Those failures returned early from `Host::perform_request` and never reached `update_stats`, so they were invisible to `total_requests` and to every error bucket. In a real run the top-of-report `ResponseStats` flagged two `kubernetes.io` errors while the per-host table reported `kubernetes.io | 74 requests | 100.0% success` for the same host.

Full root-cause analysis with code references in #2202.

## Change

Add a `network_errors` counter on `HostStats` and a corresponding `record_network_error` method, then call it from the `Err` arm of `Host::perform_request`. The failed request now contributes to `total_requests` (so the success-rate denominator is correct) and to a dedicated `network_errors` bucket (so the failure mode is visible in the output).

Surfaces touched:

- `HostStats` field + recording method + updated `error_rate`.
- `Serialize` impl: adds `network_errors` to the JSON output.
- Markdown formatter: new `Network Errors` column on the per-host table.
- Detailed text formatter: new line shown when `network_errors > 0` (mirrors the existing pattern for `client_errors` / `server_errors`).
- Compact formatter: intentionally unchanged. Its `% success` column already reflects the corrected math, and adding a column to its rigid one-line layout would be intrusive. Happy to add it if you'd prefer consistency across all three formatters.

Before (current):

```
| Host          | Requests | Success Rate | Median Time | Cache Hit Rate |
| kubernetes.io | 74       | 100.0%       | 109ms       | 41.2%          |
```

After:

```
| Host          | Requests | Success Rate | Network Errors | Median Time | Cache Hit Rate |
| kubernetes.io | 76       | 97.4%        | 2              | 109ms       | 41.2%          |
```

## Tests

- New unit test `test_record_network_error_affects_totals_and_success_rate` in `lychee-lib/src/ratelimit/host/stats.rs` asserts that a network error increments `total_requests` and `network_errors`, leaves the HTTP-status buckets untouched, and lowers the success rate accordingly.
- Existing JSON snapshot in `lychee-bin/src/formatters/stats/json.rs` updated to include the new field.

## Verification

- `cargo test --workspace --all-targets`: 534 passed, 0 failed.
- `make lint` (workspace clippy): clean.
- `cargo clippy --workspace --no-deps -- -D warnings`: clean.

### DIFF
--- a/lychee-bin/src/formatters/host_stats/markdown.rs
+++ b/lychee-bin/src/formatters/host_stats/markdown.rs
@@ -32,6 +32,8 @@ struct HostStatsTableEntry {
     requests: u64,
     #[tabled(rename = "Success Rate")]
     success_rate: String,
+    #[tabled(rename = "Network Errors")]
+    network_errors: u64,
     #[tabled(rename = "Median Time")]
     median_time: String,
     #[tabled(rename = "Cache Hit Rate")]
@@ -52,6 +54,7 @@ fn host_stats_table(host_stats: &HostStatsMap) -> String {
                 host: hostname.clone(),
                 requests: stats.total_requests,
                 success_rate: format!("{:.1}%", stats.success_rate() * 100.0),
+                network_errors: stats.network_errors,
                 median_time,
                 cache_hit_rate: format!("{:.1}%", stats.cache_hit_rate() * 100.0),
             }
--- a/lychee-bin/src/formatters/host_stats/detailed.rs
+++ b/lychee-bin/src/formatters/host_stats/detailed.rs
@@ -38,6 +38,9 @@ impl Display for DetailedHostStats {
             if stats.server_errors > 0 {
                 writeln!(f, "  Server errors (5xx): {}", stats.server_errors)?;
             }
+            if stats.network_errors > 0 {
+                writeln!(f, "  Network errors: {}", stats.network_errors)?;
+            }
 
             if let Some(median_time) = stats.median_request_time() {
                 writeln!(
--- a/lychee-bin/src/formatters/stats/json.rs
+++ b/lychee-bin/src/formatters/stats/json.rs
@@ -113,6 +113,7 @@ mod tests {
       "rate_limited": 1,
       "client_errors": 0,
       "server_errors": 1,
+      "network_errors": 0,
       "median_request_time_ms": null,
       "cache_hits": 1,
       "cache_misses": 4,
--- a/lychee-lib/src/ratelimit/host/host.rs
+++ b/lychee-lib/src/ratelimit/host/host.rs
@@ -182,6 +182,11 @@ impl Host {
         let response = match self.client.execute(request).await {
             Ok(response) => response,
             Err(e) => {
+                // Record the network error in the per-host totals.
+                self.stats
+                    .lock()
+                    .unwrap()
+                    .record_network_error(start_time.elapsed());
                 // Wrap network/HTTP errors to preserve the original error
                 return Err(ErrorKind::NetworkRequest(e));
             }
--- a/lychee-lib/src/ratelimit/host/stats.rs
+++ b/lychee-lib/src/ratelimit/host/stats.rs
@@ -38,6 +38,8 @@ pub struct HostStats {
     pub server_errors: u64,
     /// Number of client error responses (4xx, excluding 429)
     pub client_errors: u64,
+    /// Number of network errors (DNS, TCP, TLS, connection reset, no HTTP response received)
+    pub network_errors: u64,
     /// Timestamp of the last successful request
     pub last_success: Option<Instant>,
     /// Timestamp of the last rate limit response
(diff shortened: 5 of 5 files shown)


thomas-zahner (approved):
@afalhambra-hivemq Awesome 🚀

@mre any additional thoughts?

afalhambra-hivemq:
> @afalhambra-hivemq Awesome 🚀
> 
> @mre any additional thoughts?

Thanks @thomas-zahner!

@mre - do you mind having a look at it? Thanks as well

lychee-bin/src/formatters/host_stats/markdown.rs
```diff
@@ -32,6 +32,8 @@ struct HostStatsTableEntry {
     requests: u64,
     #[tabled(rename = "Success Rate")]
     success_rate: String,
+    #[tabled(rename = "Network Errors")]
```
Souradip121, line 35: You mentioned intentionally leaving the compact formatter unchanged to avoid a rigid layout change which makes perfect sense. Just a suggestion, instead of a full new column, just append (2 net err) inline on the success rate when network_errors > 0:

kubernetes.io │ 76 reqs │ 97.4% success (2 net err) │ 109ms median │ 41.2% cached

That way compact users see why the number dropped without restructuring the layout. 

(Just a suggestion, you are happy to ignore this or let me know what do you think)

Souradip121 (commented):
@mre, @afalhambra-hivemq  what do you think of this??

lychee-bin/src/formatters/host_stats/markdown.rs
mre: "2 net err" is a bit too much jargon for my taste. As a user, I wouldn't know what that meant. 😅

mre: It's probably fine to skip it for the compact layout.

mre:
Looks good to me. Thanks for the contribution @afalhambra-hivemq. ⭐

PR #2204, opened 2026-05-21
Make auto-merge more conservative

thomas-zahner:
Only do auto-merges when the pull-request contains exactly one commit author. This prevents pull-requests which are manually amended from being merged automatically.

This prevents me and others from redoing the mistake done in https://github.com/lycheeverse/lychee/pull/2199#issuecomment-4507063631

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -9,10 +9,18 @@ jobs:
   auto-merge:
     runs-on: ubuntu-latest
     if: github.event.pull_request.user.login == 'dependabot[bot]' && github.repository == 'lycheeverse/lychee'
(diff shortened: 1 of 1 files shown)


katrinafyi (commented):
I might be mistaken, but I believe each `run` step uses its own shell so an early exit 0 won't stop the next step from running?

thomas-zahner:
@katrinafyi Oh, thank you, I think you are totally right. It would have to exit with a non-zero code, which is what I did first. But then I realised that the action should probably not fail. Using an `if` should do.

Btw, renovate bot would apparently handle such cases. So we could think about replacing dependabot with renovate bot. But I thought to keep the changes minimal for now.

PR #2208, opened 2026-05-25
Remove deprecated flags from codebase.

mre:
Eventually we have to get rid of the long-deprecated flags in lychee. And I believe now is as good of a time as ever. We don't have to wait for 1.0 to hit (I believe). 

This affects the following flags:
- `--exclude-file`, which got replaced by `.lycheeignore`
- `--base`, superceded by `--base-url`.

Wdyt about merging this in and announcing the removal with the next release?

If we decide to do that, I can prepare a PR to update the lychee docs and the lychee action.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,9 +377,6 @@ Options:
           The provided base URL value must either be a URL (with scheme) or an absolute path.
           Note that certain URL schemes cannot be used as a base, e.g., `data` and `mailto`.
 
-      --base <BASE>
-          Deprecated; use `--base-url` instead
-
       --basic-auth <BASIC_AUTH>
           Basic authentication support. E.g. `http://example.com username:password`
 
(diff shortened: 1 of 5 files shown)


katrinafyi (approved):
I'm happy with this. Indeed, we have made other breaking changes without such a long notice period as these options have had.

thomas-zahner (approved):
I totally agree with this change and see no issues.

Not related to flags, but noticed while searching for deprecations; maybe we could now remove this deprecated usage. It's probably okay to drop support for the old format after almost 2 years. What are your thoughts? Could be done in a separate PR.

https://github.com/lycheeverse/lychee/blob/ee86f0d3ec0bb0512cb19d1ca7008a9e73b21e54/lychee-lib/src/types/cookies.rs#L39

mre:
Thanks for review. I will create a follow-up PR for the deprecated code.

PR #2209, opened 2026-05-28
fix: hide "cannot send response to queue: SendError" message

### LINKED ISSUE #2189: lychee can panic with invalid input URL
Since 8bbeeb57f3055a97b27495610919ff1169b963f9 when passing multiple URLs to lychee, where one URL is "invalid" (i.e. returns a unsuccessful status code) lychee panics when it tries to use the queue once more. Is there maybe an easy way to exit lychee early instead of trying to reuse the queue? Maybe this behaviour is okay, the panic just doesn't look very nice.

# Example

```
cargo run 'https://example.com/' 'https://example.com/a'

thread 'tokio-rt-worker' (67049) panicked at lychee-bin/src/commands/check.rs:236:18:
cannot send response to queue: SendError { .. }
Error: Cannot read input content from URL: status code 404 Not Found. To check links in error pages, download and check locally instead.
```

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/2189

The SendError message was just a distracting panic that was printed alongside a genuine error in another input URL.

It should be okay to just silence the warning. If sending has failed, then the channel has dropped which means we are heading towards a main thread exit which would print any necessary messages.

### DIFF
--- a/lychee-bin/src/commands/check.rs
+++ b/lychee-bin/src/commands/check.rs
@@ -230,10 +230,7 @@ async fn request_channel_task(
             )
             .await;
 
-            send_resp
-                .send((guard, response))
-                .await
-                .expect("cannot send response to queue");
+            send_resp.send((guard, response)).await.unwrap_or_default();
         },
     )
     .await;


thomas-zahner (approved):
Thanks! That makes sense. I ran `cargo fmt` on your commit to fix the formatting.

Romasato:
Still seeing the issue where it does not abort now, but fails to generate report in the end.

mre:
Hey @Romasato, can you send us the lychee version you're using and perhaps a way to reproduce the issue? E.g. the commands you run (including inputs) so we can troubleshoot the issue? Ideally, can you create a new issue for that?

Issue #2108, opened 2026-03-27
Problems with fullwidth delimiters like fullwidth comma (U+FF0C)


adamlui:
For this file https://github.com/adamlui/js-utils/blob/ed75797d3a18eaea4dd52bd7acfcfd0b049983c1/docs/zh-cn/SECURITY.md

 After install lychee to https://github.com/adamlui/js-utils/blob/ed75797d3a18eaea4dd52bd7acfcfd0b049983c1/.github/workflows/check-links-on-push-pr.yml the 1st run is https://github.com/adamlui/js-utils/actions/runs/23633975539/job/68839177430

<img width="880" height="222" alt="Image" src="https://github.com/user-attachments/assets/59544c9c-1c5e-405d-ae2a-165b520fce28" />

(I purposely didn't wrap email in `<>` 

<img width="831" height="224" alt="Image" src="https://github.com/user-attachments/assets/1ffb7a31-161b-4d44-ab5d-fe1aa129c0c5" />

but should still work as GH hyperlinked it to mailto:security@tidelift.com

<img width="274" height="194" alt="Image" src="https://github.com/user-attachments/assets/64af5747-e025-4847-aac1-b30a9d1d4500" />

)

adamlui:
OK I got every error to go away: https://github.com/adamlui/js-utils/actions/runs/23635457611/job/68843651214

.. but it still feels wrong like it is now excluding the address instead of validating it:

<img width="592" height="558" alt="Image" src="https://github.com/user-attachments/assets/71b875ae-f076-4a6c-9ec5-bfde0a996c23" />

Also the 'could not find generate-pw' went awya w/ https://github.com/adamlui/js-utils/commit/d37fa72955e14bcce66cb2ebaf14ab0ecf8c065f looks due to rel link was hard to handle so replaced w/ abs (which GH is terrible at resolving anyway like from repo home MD preview it works different from docs/ view so I didn't mind but others might like using rel links so should still validate those)

katrinafyi:
I think there has been some confusion around the files. I looked into it and there are very similar files at different parts of the repository: generate-ip/README.md and generate-ip/docs/README.md. The relative link `../generate-pw` will work from the first file but not the second file. I think you noticed this because you said that it works different from [the docs/ view in Github](https://github.com/adamlui/js-utils/tree/ed75797d3a18eaea4dd52bd7acfcfd0b049983c1/generate-ip/docs#-generate-pw-) and, indeed, the link does not work there.

[...]

PR #2211, opened 2026-05-28
feat: plaintext link extraction based on GFM

katrinafyi:
Just putting this here to make the work more discoverable. This aims to fix https://github.com/lycheeverse/lychee/issues/2108 (and possibly other plaintext link delimiting issues).

Using GFM gives us a more rigorous foundation for plaintext link extraction. For example, it gives nice heuristics for trailing punctuation. It lets a `.` be part of a link if it is within a link (e.g., a file extension), but it trims it from the end of the link.

However, GFM is still not perfect. The [spec](https://github.github.com/gfm/#extended-email-autolink) is silent on the interaction with Unicode, but the GFM implementation on Github.com does implement some non-trivial handling. This Unicode handling still has some issues, as noted in https://github.com/lycheeverse/lychee/issues/2108#issuecomment-4229490943. Additionally, GFM is in C and its Unicode handling is mixed amongst its general tokenising so it's not easy to tease out.

We may just have to choose our own GFM-like heuristics based on some Unicode character classes. This is where I got stuck, as I don't have a good understanding of how users expect plaintext links to behave - especially in languages without spaces. There are competing forces where we want links to exclude adjacent prose text, but domains and URLs are increasingly becoming more internationalised so we must allow non-ASCII characters in links. Maybe we can just use a conservative heuristic for plaintext links, but allow _anything_ between `<...>` links?

### DIFF
--- a/lychee-lib/src/extract/plaintext.rs
+++ b/lychee-lib/src/extract/plaintext.rs
@@ -1,11 +1,8 @@
 use std::{borrow::Cow, ops::Range, sync::LazyLock};
 
-use regex::{CaptureMatches, Captures, Match, Regex};
+use regex::{Captures, Regex};
 
-use crate::{
-    types::uri::raw::{RawUri, SpanProvider},
-    utils::url,
-};
+use crate::types::uri::raw::{RawUri, SpanProvider};
 
 static GFM_AUTOLINKS_REGEX: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(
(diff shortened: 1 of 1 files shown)


mre:
I wonder what to do with this draft PR now that https://github.com/lycheeverse/lychee/issues/2108 returns a 404. 🤔

katrinafyi:
What happened to https://github.com/lycheeverse/lychee/issues/2108? 👀

I'm not fussed on whether this gets merged. Maybe we can just file an issue to upstream

mre:
Yeah, no idea what happened there. I didn't even know that issues could be replaced by a 404. Very strange.

PR #2215, opened 2026-05-30
Remove deprecated legacy cookie format support

### LINKED ISSUE #2213: Remove deprecated legacy cookie format support
Follow-up from PR #2208 (review comment by @thomas-zahner).

`CookieJar::load` in `lychee-lib/src/types/cookies.rs` still falls back to the legacy cookie store format via the deprecated `ReqwestCookieStore::load_json`. The new format has been the default for almost 2 years, so it's probably safe to drop support for the old format.

- Remove the legacy-format fallback branch and the `#[allow(deprecated)]` usage of `ReqwestCookieStore::load_json`.
- Simplify `CookieJar::load` to only support the current `cookie_store::serde::json` format.
- Update the doc comment that mentions the "legacy format".

This is a breaking change for anyone still using a cookie file in the old format, so it should be announced in the release notes.

mre:
Follow-up to #2208 (review comment by @thomas-zahner) and closes #2213.

`CookieJar::load` previously fell back to the legacy cookie store format via the deprecated `ReqwestCookieStore::load_json`. The new `cookie_store::serde::json` format has been the default for almost 2 years, so this drops support for the old format and the associated `#[allow(deprecated)]`.

### DIFF
--- a/lychee-lib/src/types/cookies.rs
+++ b/lychee-lib/src/types/cookies.rs
@@ -23,39 +23,26 @@ impl CookieJar {
     ///
     /// This function will return an error if
     /// - the file cannot be opened (except for `NotFound`) or
-    /// - if the file is not valid JSON in either new or legacy format
+    /// - if the file is not valid JSON
     pub fn load(path: PathBuf) -> Result<Self> {
-        match std::fs::File::open(&path).map(std::io::BufReader::new) {
-            Ok(mut reader) => {
+        let store = match std::fs::read_to_string(&path) {
+            // A missing or empty jar (e.g. a freshly `touch`ed file) simply
+            // starts out with an empty cookie store.
+            Err(e) if e.kind() == IoErrorKind::NotFound => ReqwestCookieStore::default(),
+            Ok(contents) if contents.trim().is_empty() => ReqwestCookieStore::default(),
+            Ok(contents) => {
                 info!("Loading cookies from {}", path.display());
-
-                // Try loading with new format first, fall back to legacy format
-                #[allow(clippy::single_match_else)]
-                let store = match cookie_store::serde::json::load(&mut reader) {
-                    Ok(store) => store,
-                    Err(_) => {
-                        // Reopen file for legacy format attempt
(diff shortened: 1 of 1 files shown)


mre:
I noticed that the removed legacy fallback used to silently handle empty files.
I've updated the code to make sure that will continue to be the case. So a `touch cookies.json` workflow (which the caching docs recommend for Docker) doesn't break.

lychee-lib/src/types/cookies.rs
```diff
@@ -23,39 +23,26 @@ impl CookieJar {
     ///
     /// This function will return an error if
     /// - the file cannot be opened (except for `NotFound`) or
-    /// - if the file is not valid JSON in either new or legacy format
+    /// - if the file is not valid JSON
     pub fn load(path: PathBuf) -> Result<Self> {
-        match std::fs::File::open(&path).map(std::io::BufReader::new) {
-            Ok(mut reader) => {
+        let store = match std::fs::read_to_string(&path) {
```
thomas-zahner, line 28: Does it make sense to read to a string to then use it as `contents.as_bytes`? Can't we just use `fs::read`?

Edit: I've missed `contents.trim().is_empty()`. Probably it's best to read to a string.

thomas-zahner (approved):
Awesome 🚀

lychee-lib/src/types/cookies.rs
mre: Yeah, the as_bytes was my first attempt. 😅 

We could also do:

```rust
Ok(contents) => {
    info!("Loading cookies from {}", path.display());
    let store: cookie_store::CookieStore = serde_json::from_str(&contents)
        .map_err(|e| ErrorKind::Cookies(format!("Failed to load cookies: {e}")))?;
    store
}
```

But then, we'd add `serde_json` as a direct dependency here.
Also, `as_bytes()` is just a view into the buffer, so at least we don't allocate twice. It's probably fine. 😉

thomas-zahner: Totally agree :+1:

PR #2218, opened 2026-05-30
Support HTTP request method fallback

### LINKED ISSUE #1327: Support HTTP request method fallback
Some misconfigured webservers don't accept HEAD requests, but using HEAD is more efficient and appropriate in general. As rough but effective work-around, allow the configuration key`method` to be a list `["head", "get"]` instead of a string, where Lychee's checking algorithm falls back to the subsequent HTTP method in the list if the former gives and error response. Regrettably, in practice the response codes returned upon HEAD requests by these misconfigured webservers vary (404, 403, etc.), so don't bother to implement a fallback condition evaluator.

mre:
Previously, lychee only supported a single request method for all servers: HEAD.

This caused trouble, because some servers do not accept HEAD requests. However, using HEAD is a sane default, as it is more efficient/lightweight than GET requests.

Allow the configuration key method to be a list ["head", "get"] instead of a string, where Lychee's checking algorithm falls back to the subsequent HTTP method in the list if the former gives an error response.

This will always retry the request until the list of request methods gets exhausted. The only exception are timeouts, in which case the iteration stops and the status gets returned immediately.

Fixes https://github.com/lycheeverse/lychee/issues/1327
@sanmai-NL fyi


lychee-lib/src/types/methods.rs
mre, line 1: Three reasons why this is cool:
1. We can enforce that the list of methods is non-empty.
2. We can support `.methods(Method::HEAD)` as well as `.methods(reqwest::Method::GET` and multiple methods.
3. We move some of the method parsing logic from bin into one place in lib.

examples/builder/builder.rs
```diff
@@ -26,7 +26,7 @@ async fn main() -> Result<()> {
         .user_agent("custom useragent")
         .allow_insecure(true)
         .custom_headers(headers)
-        .method(Method::HEAD)
+        .methods(Method::HEAD)
```
mre: Arguably, we could keep `method` here. Then we'd avoid the breaking change and it reads equally nicely. Open for feedback.

lychee-lib/src/ratelimit/host/host.rs
```diff
@@ -28,8 +28,16 @@ use crate::{
...
 
-/// Per-host cache for storing request results
-type HostCache = DashMap<Uri, CacheableResponse>;
+/// Identifies a request for caching and de-duplication purposes.
+///
+/// The [`Method`] is part of the key because the same URL can yield different
+/// results per method (e.g. a server may reject `HEAD` but accept `GET`).
+/// Keying by method ensures method fallback re-requests instead of reusing a
+/// previous method's cached response.
+type RequestKey = (Method, Uri);
```
mre, line 37: If someone has an idea for a better name, I'm all ears. But at the end, we use it for a cache, so maybe `Request*Key*` is fine?

lychee-bin/src/config/mod.rs
```diff
@@ -534,12 +534,17 @@ pub(crate) struct Config {
     #[arg(short, long)]
     retry_wait_time: Option<u64>,
 
-    /// Request method
+    /// Request method(s)
+    ///
+    /// Accepts a single method or a comma-separated list. When multiple methods
+    /// are given, lychee tries each one in order and returns the first success.
+    /// This is useful for servers that reject `HEAD` requests: `--method head,get`
+    /// falls back to `GET` when `HEAD` fails.
```
katrinafyi, line 542: I believe setting a non GET method will have the side-effect of disabling fragment checks. In particular, a HEAD,GET that succeeds on HEAD won't fall back to GET to get the HTML body for fragments.

Is this worth mentioning in this help text? I'm okay with this caveat existing - it's basically what already happens - it's just more noticeable when fallback methods apply.

examples/builder/builder.rs
katrinafyi: I truly don't mind either way, but I lean a tiny bit towards "method" because the toml and CLI flag keep the non-plural name as well.

katrinafyi (commented):
Happy in principle, though I haven't dived into actually running it. I'm also okay with the Methods abstraction.

lychee-bin/tests/cli.rs
```diff
@@ -2904,6 +2875,62 @@ The config file should contain every possible key for documentation purposes."
...
+            .arg("head,get")
+            .arg("-")
+            .write_stdin(server.uri())
+            .assert()
+            .success();
+    }
+
+    #[tokio::test]
+    async fn test_method_no_fallback_by_default() {
+        // The same server: HEAD is rejected (405), GET would succeed.
```
thomas-zahner: The same server as where? In the future maybe we'll add some tests in between these two CLI tests, leaving us with no clue what this refers to.

I think it would make a lot of sense to just move `cargo_bin_cmd!().arg("--method").arg("head")` into the above test, leaving us with a single server and two runs.

lychee-lib/src/checker/website.rs
```diff
@@ -266,10 +270,45 @@ impl WebsiteChecker {
     /// - The request failed.
     /// - The response status code is not accepted.
     async fn check_website_inner(&self, uri: &Uri, default_chain: &RequestChain) -> Status {
-        let request = self.host_pool.build_request(self.method.clone(), uri);
+        let mut last_status = None;
+
+        // Try each configured method in order and return the first success.
+        //
+        // Misconfigured servers reject requests with wildly varying status
```
thomas-zahner: To me it's not totally clear what `Misconfigured servers` refers to. I assume you mean violating the standard that HEAD should result in an identical status code as GET. But at this point we're not talking about HEAD & GET yet. It could be any method and users could use `POST, PUT, PATCH, GET`. So I feel like this prior assumption should be clarified or maybe this sentence could even be fully removed.

lychee-lib/src/types/methods.rs
thomas-zahner: This is awesome 🚀

thomas-zahner (approved):
Nice, I didn't spot any big issues. Just two small details.

examples/builder/builder.rs
mre: True. I've changed it back to `method`.

lychee-bin/src/config/mod.rs
mre: Keen eye! Yes, I've added a help text. Also added a user hint because this is easy to do now. (@thomas-zahner fyi)

lychee-bin/tests/cli.rs
mre: Yeah, you're right. Done.

lychee-lib/src/checker/website.rs
mre: Sure, I've updated the comment to make it less ambiguous. Thanks.

PR #2219, opened 2026-05-30
Make `lychee::latest` track the latest stable release, add `nightly` tag

### LINKED ISSUE #1981: Docker tag for latest stable release
I would like to be able to run the latest stable release. I thought that using `lycheeverse/lychee:latest-alpine` would achieve this but it seems like there is no difference between the `latest` and `master` Docker tags. Is this intentional? If so, would it be possible to add a `stable` tag that points to the latest release version?

mre:
The `latest` and `master` Docker tags were identical: `latest` was published on
every push to `master` (`type=raw,value=latest,enable={{is_default_branch}}`) and
the master build bakes in the `nightly` binary. So `lychee:latest` was effectively
a nightly build, not the latest stable release, which is surprising and violates
the usual Docker convention that `latest` means "latest stable release."

> [!NOTE]
>
> **The meaning of `lychee:latest` changes from "latest `master` build" to "latest
> stable release". Users who relied on `latest` for bleeding-edge should switch
> to `nightly`.**

This PR changes the tagging policy as follows:

- `latest` is published **only on stable release tags** (i.e. `lychee-v*`) and always points to
  the **most recent stable release**.
- `nightly` is a **new tag for the bleeding-edge build** from `master`. (The existing
  **`master`** tag is kept as an alias, so nothing breaks for users who
  reference it. We could consider dropping the `master` alias in the future, but
  it doesn't hurt to keep it around for a while until users have updated.)
- Release-tagged images now bake in the **exact** released binary
  (`LYCHEE_VERSION=<tag>`) instead of "whatever release is latest at build time",
  making `lychee:X.Y.Z` reproducible and removing a publish-time race. :)

All tags continue to ship an `-alpine` variant (e.g. `latest-alpine`).

So, in summary:

| Tag             | Description                                  |
| --------------- | -------------------------------------------- |
| `latest`        | Most recent stable release (recommended)     |
| `X.Y.Z` / `X.Y` | A specific release, e.g. `0.20.0` / `0.20`   |
| `nightly`       | Bleeding-edge build from the `master` branch |
| `master`        | Alias of `nightly` (removal t.b.d)           |
| `sha-<sha>`     | Build for a specific commit                  |

The README now contains a compact version of this table so that people can find
the right image for their use-case. And since the Docker workflow pushes
`README.md` as the DockerHub description, it propagates to DockerHub
automatically.

Fixes #1981

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -40,7 +40,7 @@ jobs:
       # are prefixed (`lychee-v1.2.3`) and the version is only recovered via the
       # `match=` regex, so `auto`'s detection is unreliable here. Gating `latest`
       # on the release-tag ref ourselves is the documented, predictable approach:
-      # https://github.com/docker/metadata-action#conditionally-tag-with-latest-for-a-specific-branch
+      # https://github.com/docker/metadata-action#latest-tag
       # Note: this gate assumes all `lychee-v*` tags are stable. If pre-release
       # tags are ever introduced, also exclude them here (e.g. `!contains(github.ref, '-')`).
       - name: Docker meta (debian)
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -26,10 +26,30 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
 
+      # Docker tagging policy
+      # ----------------------
+      # - `latest`           -> most recent stable release (release tag push only)
+      # - `X.Y.Z` / `X.Y`    -> a specific release
+      # - `nightly`/`master` -> bleeding-edge build from the default branch
+      # - `sha-<sha>`        -> a specific commit
+      # Each also gets an `-alpine` variant via the second metadata step.
+      #
+      # We intentionally set `flavor: latest=false` and add `latest` explicitly
+      # instead of relying on the default `latest=auto`. `auto` decides whether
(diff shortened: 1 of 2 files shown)


mre:
I tested the Docker image creation with a throwaway workflow and everything seems to have worked as expected. If I didn't overlook anything, this should be good to go. Fingers crossed because historically I didn't have much luck with Docker changes.

But overall, this aligns lychee's Docker image behavior with user's expectations. `latest` is the latest released version, `master` is `nightly`.

mre:
@haines fyi

thomas-zahner (approved):
Thank you for the investigation & detailed docs! I couldn't spot any issues.

As a side-note I started investigating this issue too, but got sidetracked into simplifying the Docker images by not depending on the GitHub releases. I've started this work in https://github.com/thomas-zahner/lychee/tree/reproducible-docker-images. But this is only indirectly related to this PR, so lets merge this PR first.

mre:
> I've started this work in https://github.com/thomas-zahner/lychee/tree/reproducible-docker-images. But this is only indirectly related to this PR, so lets merge this PR first.

Ah, great. Yes, looking forward to your independent work on the Dockerfile changes.

As discussed, I will go ahead and merge this, which will resolve the issue https://github.com/lycheeverse/lychee/issues/1981. Thanks for the review. 👍

Issue #2117, opened 2026-03-30
Handle DataDome protection


adamlui:
https://github.com/elonsucks/why.elonsucks.org/actions/runs/23771818083/job/69264816349

<img width="1083" height="125" alt="Image" src="https://github.com/user-attachments/assets/f8eeb5f3-68e5-4608-9692-a2126fd3df2c" />

- https://www.marketwatch.com/story/teslas-stock-falls-again-and-this-analyst-says-why-its-an-elon-musk-problem-a7ac0f01
- https://www.reuters.com/legal/exclusive-tesla-faces-us-criminal-probe-over-self-driving-claims-sources-2022-10-26/

thomas-zahner:
> if we include a wildcard accept with a lower priority, then a conforming web server shouldn't send us 406

Ah yes, I didn't consider this! Then it sounds quite feasible.

> I really want to avoid an "arms race" with the bot detection software, and I worry that adding a quirk for a specific protection service would lead us down that path

I see your point. Adding the language header with an included wildcard by default would be a change I'm okay with it. However, as @mre's snippet shows, they also seem to be requiring a browser-like user agent and accept header. Both of these are fields where lychee would be dishonest, so as you mention we shouldn't send them by default.

@katrinafyi So do you mean with "arms race" that you prefer to not handle the `x-datadome: protected` response header? Thinking about it I agree that this might be a really bad idea.

But with a list of websites, based on user reports, to handle the affected websites as a quirk where lychee intentionally sends dishonest headers, would you be okay with this? If so, then maybe we also don't have to start sending language headers by default.

thomas-zahner:
Hmm, yes such a guide or document would certainly be great. This also reminds me of "user hints" (https://github.com/lycheeverse/lychee/pull/2021). Maybe instead of implementing any quirks we could scan for `x-datadome: protected` response headers and then report the DataDome protection issue in a UX friendly manner and link to the user guide on how to work around the issue.

> but I'm a bit afraid that we might break some existing use-cases

I do see your point. It's quite excessive to send additional headers with every single request just in order to satisfy a few non-conforming bot detection mechanisms out there.

PR #2220, opened 2026-05-31
Add built-in host config defaults

### LINKED ISSUE #1960: Host config defaults
Collect useful real-world host-specific configurations and configure them by default. This is a follow-up feature of #1929. This also addresses the idea of [this comment](https://github.com/lycheeverse/lychee/pull/1929#issuecomment-3660091727).

This feature is comparable to [quirks](https://github.com/lycheeverse/lychee/blob/8d51c6003ded4ee5151755b5be28704aa30c1404/lychee-lib/src/quirks/mod.rs#L35) but specifically for HostConfigs. I've started in 6b4b61f9c58a38408730e018f419a256cf319af1, which can be used as a starting point.

One difficulty to consider is the merging/precedence of the different options. I guess our host specific "quirk" defaults should take precedence over user specified global defaults, but not over user-specified host defaults.

mre:
Adds built-in host-specific config defaults (like quirks, but for `HostConfig`s). Seeds two hosts that reject lychee's default user-agent:

| Host | User-Agent |
|------|-----------|
| `www.gnu.org` | `Mozilla/5.0` |
| `developers.redhat.com` | `curl/8.4.0` |

Adding more is a one-liner in `HostConfigs::builtin_defaults()`.

**Precedence:** user host config > built-in defaults > user global defaults. Built-ins are merged underneath user host config, and `effective_*` already falls back to the global config, so this works for free.

**Also:** `HostConfig::merge` now overrides headers by key instead of appending (no more duplicate `User-Agent`s). The existing `test_merge_hosts` only merged disjoint keys, so it's unaffected.

**Note:** `HostKey` matches exact hostnames, so `www.gnu.org` doesn't cover `gnu.org`. No opt-out flag, matching quirks.

Closes #1960.

### DIFF
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -367,14 +367,15 @@ impl ClientBuilder {
             .build()
             .map_err(ErrorKind::BuildRequestClient)?;
 
-        let client_map = self.build_host_clients(&redirect_history)?;
+        // User-specified host configuration takes precedence over the built-in
+        // defaults, which in turn take precedence over the global defaults.
+        // This must happen before building the host clients, since host headers
+        // are baked into the per-host reqwest client.
+        let hosts = self.hosts.clone().merge(HostConfigs::builtin_defaults());
 
-        let host_pool = HostPool::new(
-            self.rate_limit_config,
-            self.hosts,
-            reqwest_client,
-            client_map,
-        );
+        let client_map = self.build_host_clients(&hosts, &redirect_history)?;
+
+        let host_pool = HostPool::new(self.rate_limit_config, hosts, reqwest_client, client_map);
 
         let github_client = match self.github_token.as_ref().map(ExposeSecret::expose_secret) {
             Some(token) if !token.is_empty() => Some(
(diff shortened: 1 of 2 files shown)


mre:
A note of caution from my side: I'm somewhat skeptical about this approach in general.

Hardcoding host-specific user-agents is a real maintenance burden. The `curl/8.4.0` string, for example, will go stale and we'd be on the hook to bump these over time. The list also only ever grows, and every entry is a small judgement call about another site's firewall behaviour that can change without notice.

I mostly wanted to get this off my todo list and see it written down. I'm not convinced we should merge it as-is, and I'm happy to close it if others feel the same. Curious what people think.

thomas-zahner:
I actually do agree with your opinion on **not** merging this PR. I've already prepared a version mentioned in the related issue. (https://github.com/lycheeverse/lychee/commit/6b4b61f9c58a38408730e018f419a256cf319af1) I've almost created this PR multiple times. But each time I didn't because I was no longer convinced about doing this.

This is very much related to [this comment regarding the arms race](https://github.com/lycheeverse/lychee/issues/2117#issuecomment-4170786533) and hardcoding headers & values. Lets close this PR and issue as "won't do". As we established on https://github.com/lycheeverse/lychee/issues/2117 I would suspect that @katrinafyi would also prefer not to hardcode these values and not merge this PR. @katrinafyi please let me know if you disagree.

mre:
Yes, that makes the most sense. I agree. 👍

katrinafyi:
> As we established on #2117 I would suspect that @katrinafyi would also prefer not to hardcode these values and not merge this PR. 

Indeed, that's my opinion here as well and for the same reason. Thanks @thomas-zahner for raising

PR #2221, opened 2026-05-31
fix(ci): stop concurrent release uploads racing on assets

mre:
Release events fire `created`/`prereleased`/`published` together, spawning 3 concurrent runs that race `gh release upload` and drop assets (e.g. nightly aarch64-musl, which broke the arm64 Docker build). Trigger on `published` only, add a concurrency guard, and use `--clobber`.


mre:
This fixed the build. It was unrelated to the Docker changes in #2219. The root cause was a race on GitHub releases on push.

PR #2228, opened 2026-06-03
fix: align docker image version to Cargo.toml

cbachhuber:
To prevent that these go out of sync in future, we could use something like [this pre-commit hook](https://github.com/hofbi/dev-tools/blob/b58fd55d9fd6619f52ec584a942ac5245d7cffc2/README.md#sync-tool-versions). If you want me to, I'll propose that change in this repo.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -24,7 +24,7 @@
 - id: lychee-docker
   name: lychee-docker
   description: "Run 'lychee' docker image for fast, async, stream-based link checking (auto-installs)"
-  entry: lycheeverse/lychee:0.23.0
+  entry: lycheeverse/lychee:0.24.2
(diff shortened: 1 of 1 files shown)


mre:
Thanks for the fix and an automatic update would be great. I envision some kind of workflow, which bumps the version automatically.

cbachhuber:
So you create the tag, and the workflow rolls that out to the corresponding places in the codebase? I think https://github.com/marketplace/actions/write-tag-to-version-file might work for that.

Just setting expectations: I personally won't have time to implement this any time soon :/

mre:
I played around with it a bit. It's harder than I thought. Could only come up with super ugly, big solutions so far. Still investigating.

mre:
Think I found another angle in #2231.

PR #2229, opened 2026-06-03
Pass channel sender along to prevent deadlocks

soundofspace:
Hi, I just read both blog posts about "Five Years of Trying to Add Recursion to lychee". Thanks for those they were an interesting read.

The one thing I don't agree with is "Channels are the wrong tool for cyclic pipelines". They have served me really well, but the points you made about how they work in rust are really valid. So they could be the right tool if they are used in a slightly different way.

The easiest solution to make everything work with channels is not to store a single recursive/global sender, but to actually pass it along with inputs/requests/responses. By doing this the feedback loop's lifecycle is now tied to the actual work. This prevents deadlocks and effectively implements what is being doing here with a waitgroup and a guard. The guard here basically becomes the sender and works almost essentially the same: passing it along and relying on drop to signal work is finished


soundofspace:
This is PR is not ready for a full merge and could use some more cleanup / polishing and nicer UX. But I want to already request a review just to talk about this design in general.

It also seems tests are flacky in github ci? Doing the exact same tests locally work without any issue.

Also this only address this specific issue, it does not attempt to solve or handle any of the other difficulties related to recursing

katrinafyi (approved):
I'm happy with this in principle, thanks for the advice!

Re the pending structs, it would be nice if they could be combined into one generic struct somehow, but this is hard since the sender needs to take a (pending) request. I guess this is one thing that would've been easier with WaitGuard, since it doesn't have the type parameter. Do you have thoughts?

I feel like passing the sender around is also somewhat less semantic. It's harder to know at a glance why it's being propagated so much and it gives the owner more power (anyone could send more requests at any time).

Re using Option to reduce the size of the struct, I don't think it matters too much at all.

mre:
Thanks, @soundofspace. Change looks good to me. Do you have any thoughts on @katrinafyi's comment above?

lychee-bin/src/commands/check.rs
```diff
@@ -24,6 +23,21 @@ use crate::{ExitCode, cache::Cache};
 
 use super::CommandParams;
 
+struct PendingRequest {
+    // TODO this could option to be even light if recursion is not used
+    /// Ride along sender used to enqueue recursively discovered children.
+    ///
+    /// By passing this along with the request/response we can rely on native
+    /// rust ownership to prevent deadlocks and automatic cleanup
```
mre: I find this comment hard to read. Not exactly sure what it means. Can it be rephrased for simplicity?

soundofspace:
Hey, sorry I lost track of this!

I updated code to incorporate changes from latest main branch. As can be seen: passing a waitguard or a typed RequestQueue (could be the sender directly, but feels like this deserves a newtype) is almost exactly the same. At this point it's mostly a trade off between these two:

## lifetime
 A single channel that tracks liveness and handles recursion
```rust
ReceiverStream::new(recursive_channel_recv)
```
vs 
a channel that lives forever but is closed by an external future, which is also backed by another sender/receiver pair internally
```rust
ReceiverStream::new(recursive_channel_recv).take_until(waiter.wait())
```

## splitting responsibilities

A channel that is responsible for liveness and for recursion
- recursive_channel_recv / RequestQueue which does both

vs 
two independant parts 

- WaitGuard that tracks liveness
- recursive_channel_recv for recursion

--- 

I personally would argue that a single channel that does both here, makes more sense, since that maps quite nicely to how people expect channels to work in rust (and waitgroup is basically another sender/receiver pair). But I can also see why someone would want to split this. So I'm leaving this PR here mostly to showcase that this is possible and how it would work.

lychee-bin/src/commands/check.rs
```diff
@@ -179,6 +178,28 @@ pub(crate) async fn check(
...
+#[derive(Clone)]
+struct RequestQueue(mpsc::Sender<(RequestQueue, Request)>);
+
+impl RequestQueue {
+    /// Enqueues a recursively discovered request
+    async fn enqueue(
+        self,
+        request: Request,
+    ) -> Result<(), mpsc::error::SendError<(RequestQueue, Request)>> {
+        let sender = self.0.clone();
```
katrinafyi: is this clone necessary given self is already moved?

lychee-bin/src/commands/check.rs
```diff
@@ -61,81 +60,81 @@ pub(crate) async fn check(
 
     /* Input streams and channels (both initial and recursive) */
 
-    let (waiter, wait_guard) = WaitGroup::new();
+    let (recursive_channel_send, recursive_channel_recv) = mpsc::channel(max_concurrency);
+    let queue = RequestQueue(recursive_channel_send);
 
     // Split initial requests into: valid requests and request errors. Note that
-    // this stream closure *owns* a wait guard, so we must drop the closure after
+    // this stream closure *owns* the queue handle, so we must drop the closure after
```
katrinafyi: not specific to your pr, but an extra word would help clarity here

```suggestion
    // this stream closure *owns* the initial queue handle, so we must drop the closure after
```

lychee-bin/src/commands/check.rs
```diff
@@ -61,81 +60,81 @@ pub(crate) async fn check(
 
     /* Input streams and channels (both initial and recursive) */
 
-    let (waiter, wait_guard) = WaitGroup::new();
+    let (recursive_channel_send, recursive_channel_recv) = mpsc::channel(max_concurrency);
+    let queue = RequestQueue(recursive_channel_send);
```
katrinafyi: I would shadow this variable to make very clear that the unwrapped sender isn't used anymore (even though it's enforced by moving).

 you can leave all the closure arguments as `queue` though, otherwise they get too long.

```suggestion
    let recursive_channel_send = RequestQueue(recursive_channel_send);
```

katrinafyi (approved):
Happy with this! I now agree that this is a better way to combine the responsibilities.

Thanks again for the insight. At the time, using senders in channels for recursion didn't occur to me, but it is quite obvious in hindsight :)

I have a few minor changes suggested which I can easily push to the branch if you don't mind, just let me know!

katrinafyi:
Updating with main should fix these CI issues.

lychee-bin/src/commands/check.rs
soundofspace: Yes since we also put this exact same queue on the channel again so it can reused for further recursive work. But I slightly changed what we clone to make this clearer now, also switched it to &self, since it has no impact on the clone and will be easier to use when enqueueing multiple requests.

lychee-bin/src/commands/check.rs
soundofspace: applied

lychee-bin/src/commands/check.rs
soundofspace: applied

soundofspace:
> I have a few minor changes suggested which I can easily push to the branch if you don't mind, just let me know!

I added all the feedback you left, since I wanted to do a small refactor around the queue clone. If you have any other changes you want to push, definitely go for it, no need for an extra ping pong here. And thanks for looking at this!

lychee-bin/src/commands/check.rs
katrinafyi: Ohh very true, thanks!

katrinafyi:
Awesome! Looks good to me now, I'll let @mre do the final look and merge.

mre:
Hey, sorry for dropping the ball on this one. I have one more question before merging: is this PR intended strictly as a liveness simplification before recursion work lands, or as the chosen foundation for recursive fan-out? I think it's the latter, in which case I would require a test where one response creates more children than the channel capacity before relying on this architecture. Could you add one?

mre:
To elaborate a bit on what I mean, I believe bounded recursive fan-out can still deadlock.

Even with the new design, the data flow remains a bounded cyclic pipeline:

- First we [create a channel with capacity `max_concurrency`](https://github.com/soundofspace/lychee/blob/091850c7d29c8c1f3e93adf7d5f81bddfca11296/lychee-bin/src/commands/check.rs#L63).
- Then we serially processes discovered children with `for_each` [here](https://github.com/soundofspace/lychee/blob/091850c7d29c8c1f3e93adf7d5f81bddfca11296/lychee-bin/src/commands/check.rs#L130).
- Then [this](https://github.com/soundofspace/lychee/blob/091850c7d29c8c1f3e93adf7d5f81bddfca11296/lychee-bin/src/commands/check.rs#L133) awaits `queue.enqueue(req)`.
- The same `all_done` future is the only code polling `recursive_channel_recv`.

If one response eventually produces more children than the channel’s available capacity, `enqueue().await` can block while `for_each` is still processing that response's child vector. Because the receiver is upstream in the same future, it cannot be polled to free capacity. A page with more than `max_concurrency` discovered links is quite a realistic scenario.

I'd like to note that this structure predates the PR, and `recursive_uris` is currently hardcoded to `vec![]`, so it is not a new regression. However, the PR should not be interpreted as demonstrating that channels now prevent all recursion-related deadlocks. Before we can enable recursion, we should add a test for that case and ensure that it passes. I'm wondering if we should add a (disabled) test for that case before merging this PR, although it will likely fail or require some refactoring to pass.

And there is a chance that I'm missing something and that it's all fine, sorry.

katrinafyi:
@mre I think your worry is correct. I think the recursive channel might just have to be unbounded, I don't see another way to avoid that deadlock. A test would definitely be good.

mre:
So I'm a bit confused and not sure how to make progress here. @soundofspace, did you find the time to look into the comments? In the meantime, there's a new PR at https://github.com/lycheeverse/lychee/pull/2283, which looks like it aims to address the issue, but I have not looked into that yet. Any thoughts on how to continue?

katrinafyi:
I can make some committs which fix the bounded recursion issue later this week. It should be very small :)

PR #2231, opened 2026-06-04
ci: check lychee-docker pre-commit version matches Cargo.toml

mre:
Follow-up to #2228. CI now fails if the docker tag in `.pre-commit-hooks.yaml` drifts from the `Cargo.toml` version, so the release PR catches it instead of someone noticing later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,13 @@ jobs:
         with:
           components: clippy, rustfmt
       - uses: Swatinem/rust-cache@v2
+      - name: Check lychee-docker pre-commit version matches Cargo.toml
+        run: |
+          version=$(cargo metadata --format-version=1 --no-deps | jq -r '.packages[] | select(.name == "lychee") | .version')
+          if ! grep -q "lycheeverse/lychee:$version" .pre-commit-hooks.yaml; then
+            echo "::error::Bump the lycheeverse/lychee docker tag in .pre-commit-hooks.yaml to $version"
+            exit 1
+          fi
       - name: Run cargo fmt (check if all code is rustfmt-ed)
         run: cargo fmt --all --check
       - name: Run cargo clippy (deny warnings)


.github/workflows/ci.yml
```diff
@@ -53,6 +53,11 @@ jobs:
         with:
           components: clippy, rustfmt
       - uses: Swatinem/rust-cache@v2
+      - name: Check lychee-docker pre-commit version matches Cargo.toml
+        run: |
+          version=$(grep -m1 '^version' Cargo.toml | sed -E 's/.*"(.*)".*/\1/')
+          grep -q "lycheeverse/lychee:$version" .pre-commit-hooks.yaml \
+            || { echo "::error::Bump the lycheeverse/lychee docker tag in .pre-commit-hooks.yaml to $version"; exit 1; }
```
mre: How this works: 
On release, `release-plz` bumps `Cargo.toml`, the release PR's CI turns red, and we have to manually fix the one line in `.pre-commit-hooks.yaml`. 

It's not automatic, but very simple, self-documenting, and avoids the silent drift.

wdyt?

mre:
@cbachhuber wdyt?

.github/workflows/ci.yml
cbachhuber: Nice, I like it! It's appropriately simple given your use case. And if you're annoyed by this some time in future, you can still automate more 👍

cbachhuber (approved):

katrinafyi (approved):
I like the idea.

Is there a `cargo` command we can use to get the version rather than grep the toml? Also, I think it could be more readable to use a bash if statement rather than `||`.

How does the interaction with release-plz work, in particular how does it interact with the automatic rebase of the PR? 

Are you able to push one commit fixing the pre commit version and release-plz will stack it's commits on top of that one? Or, does release-plz overwrite your commit and you'll need to bump pre-commit as the last commit before merging?

mre:
> Is there a cargo command we can use to get the version rather than grep the toml? Also, I think it could be more readable to use a bash if statement rather than ||.

Oh, yeah, I think there is one. 🤔 Can't remember from the top of my head. Let me check.

> How does the interaction with release-plz work, in particular how does it interact with the automatic rebase of the PR?

The check is very simple. It happens on every pull request. The `Release-plz` workflow only runs on pushes to `master` and on new releases, so it happens later.

> Are you able to push one commit fixing the pre commit version and release-plz will stack it's commits on top of that one? Or, does release-plz overwrite your commit and you'll need to bump pre-commit as the last commit before merging?

As long as the build is green when we merge a PR, it's fine. Everything will work as before. The only difference is this sanity-check for the pre commit version. So if we bump the version in Cargo.toml, but forget to bump the pre commit version, it will be a CI error.

release-plz will always base its work on top of latest `master`, which means there are no problems unless we broke the build before (by merging a faulty PR) in which case the lint step on the release-plz PR will also fail, but we can fix that in a separate PR. After that, release-plz will update its own PR and everything will be green again. 😄

mre:
We could do

```sh
cargo metadata --format-version=1 --no-deps | jq -r '.packages[0].version'
```

wdyt?

thomas-zahner:
@mre So this yields to failing CI when encountering a version mismatch. Alternatively (or additionally?) we could create an action which automatically bumps the specified version. This is what we do with the table of contents for the README as well as (I think) two version references in our docs.

mre:
> we could create an action which automatically bumps the specified version

yes, that would be better. 🤔 We can close this PR if we can somehow bump the version automatically. 👍

mre:
Moved it over to cargo-metadata. Not any shorter, but arguably cleaner.

@thomas-zahner: I'm going to go ahead and merge this. We can always replace it with a more sophisticated mechanism like an action in the future. But for now, this should hedge us against forgetting to bump the pre-commit version. 😉

PR #2233, opened 2026-06-09
fix(markdown): show raw URI for request item errors

### LINKED ISSUE #2143: `<error:>` leaks into markdown format output for unparseable URLs.

YforC:
﻿## Why

For malformed URLs that fail while creating a request item, the markdown formatter was rendering the internal placeholder URI `<error:>`. That leaks implementation detail into user-facing markdown output.

Fixes #2143.

## What Changed

When formatting `RequestError::CreateRequestItem`, markdown output now uses the original raw URI text instead of the internal placeholder URI. A regression test covers the malformed URL case and asserts `<error:>` is not emitted.

## Verification

- `cargo fmt`
- `git diff --check`

I could not run the Rust test locally on this Windows machine because the toolchain is missing MSVC `link.exe`.

### DIFF
--- a/lychee-bin/src/formatters/stats/markdown.rs
+++ b/lychee-bin/src/formatters/stats/markdown.rs
@@ -6,7 +6,7 @@ use std::{
 use super::StatsFormatter;
 use anyhow::Result;
 use http::StatusCode;
-use lychee_lib::{InputSource, ResponseBody, Status};
+use lychee_lib::{InputSource, RequestError, ResponseBody, Status};
 use std::fmt::Write;
 use tabled::{
     Table, Tabled,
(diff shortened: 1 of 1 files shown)


mre:
Thanks for the PR. The linter is currently failing; can you take a look?

But on a more general note, I'm not sure if the PR will be merged in its current form.
See https://github.com/lycheeverse/lychee/issues/2143#issuecomment-4561103663 for discussion.

Although it looks relatively straightforward, this not an easy task. We are planning to solve the issue on a more fundamental level.

mre:
Closing because this is stale now. Also see discussion at https://github.com/lycheeverse/lychee/issues/2143#issuecomment-4587497270.

PR #2236, opened 2026-06-20
Append fallback extensions when the pre-existing extension has a space

### LINKED ISSUE #2235: fallback-extensions doesn't apply when file name has a dot
_Originally posted by @larrytalley in [#1503](https://github.com/lycheeverse/lychee/issues/1503#issuecomment-4745421581)_

> I am seeing incorrect results with some links that contain spaces and other punctuation. Here is an example:
> 
> `[x1902-1906-78109-koserefsky-kaltag.md]:
>  [ERROR] file:///Users/ltalley/Mailboats%20Project/N.%20C.%20Co.,%20which%20has%20the%20U.%20S.%20mail%20contract%201902 (at 26:3) | Wikilink file:///Users/ltalley/Mailboats%20Project/N.%20C.%20Co.,%20which%20has%20the%20U.%20S.%20mail%20contract%201902 not found at /Users/ltalley/Mailboats Project/N. C. Co., which has the U. S. mail contract 1902`
> 
> the file does exist:
> `% ls *which\ has\ the*
> N. C. Co., which has the U. S. mail contract 1902.md`
>  

 _Originally posted by @nobkd in [#1503](https://github.com/lycheeverse/lychee/issues/1503#issuecomment-4747458238)_

> I would have recommended [`--fallback-extensions md`](https://lychee.cli.rs/guides/cli/#--fallback-extensions) but I think that wouldn't work, as your link contains a dot, and lychee therefore likely thinks of the part after the last dot as the extension and would only replace that part..
> 
> Maybe it should be enforced, that there can't be any space in the file extension? Unsure about that.
> 
> <details>
> <summary>EDIT: Btw here a constructed example:</summary>
> 
> ```md
> # `test.md`
> 
> [hiii](<file. hi>)
> ```
> 
> ```md
> # `file. hi.md`
> ```
> 
> ```bash
> $ lychee test.md --root-dir . --fallback-extensions md
> 1/1 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                             Issues found in 1 input. Find details below.
> 
> [test.md]:
>  [ERROR] file:///path/to/file.%20hi (at 4:1) | File not found. Check if file exists and path is correct
> 
> 🔍 1 Total (in 1ms) 🔗 1 Unique ✅ 0 OK 🚫 1 Error
> ```
> 
> </details> 

 _Originally posted by @larrytalley in [#1503](https://github.com/lycheeverse/lychee/issues/1503#issuecomment-4755678697)_

> I reproduced the issue with the command:
> 
> lychee --include-wikilinks --base-url "/Users/ltalley/Mailboats Project" --fallback-extensions md --default-extension md x1902-1906-78104*.md
> 13/13 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                              Issues found in 1 input. Find details below.
> 
> [x1902-1906-78104-eagle-tanana.md]:
>  [ERROR] file:///Users/ltalley/Mailboats%20Project/N.%20C.%20Co.,%20which%20has%20the%20U.%20S.%20mail%20contract%201902 (at 26:3) | Wikilink file:///Users/ltalley/Mailboats%20Project/N.%20C.%20Co.,%20which%20has%20the%20U.%20S.%20mail%20contract%201902 not found at /Users/ltalley/Mailboats Project/N. C. Co., which has the U. S. mail contract 1902
> 
> 🔍 13 Total (in 7ms) 🔗 10 Unique ✅ 12 OK 🚫 1 Error
> 
> ltalley@MacBookPro14-2026 Mailboats Project % ls N.\ C.\ Co.,\ which\ has\ the\ U.\ S.\ mail\ contract\ 1902.md 
> N. C. Co., which has the U. S. mail contract 1902.md
> ltalley@MacBookPro14-2026 Mailboats Project %

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/2235 but I'm not totally sure whether this fix is right.

Maybe fallback-extensions should always just append the fallback extension rather than trying to replace a pre-existing extension? That would be a neater fix to this bug. But maybe some use cases would want a link like `x.html` to fall back to a `x.md` file (for example, when doing markdown to html generation). Not sure. In any case, I've added some test cases which show the behaviour.

### DIFF
--- a/lychee-lib/src/checker/file.rs
+++ b/lychee-lib/src/checker/file.rs
@@ -628,4 +628,37 @@ mod tests {
             Ok("filechecker/same_name.html")
         );
     }
+
+    #[tokio::test]
+    async fn test_fallback_extensions_with_embedded_dots() {
+        let checker = FileChecker::new(
+            &BaseInfo::none(),
+            vec!["md".to_string(), "gz".to_string()],
+            None,
+            FragmentCheckerOptions {
+                check_anchor_fragments: true,
+                check_text_fragments: false,
+            },
+            false,
+        )
+        .unwrap();
+
+        // fallback extensions *should* be applied when a file name
+        // has dots but also has a space after the last dot. we assume
+        // file extensions cannot have spaces in them
+        assert_resolves!(
+            &checker,
+            "fallback-extensions/file. hi",
+            Ok("fallback-extensions/file. hi.md")
+        );
+
+        // fallback extensions should *not* be applied when there is
+        // already a file extension, to avoid doubling up and getting the
+        // wrong file.
+        assert_resolves!(
+            &checker,
+            "fallback-extensions/a.tar",
+            Err(InvalidFilePath(_))
+        );
+    }
 }
--- a/lychee-lib/src/checker/file.rs
+++ b/lychee-lib/src/checker/file.rs
@@ -1,7 +1,6 @@
 use http::StatusCode;
 use log::warn;
 use std::borrow::Cow;
-use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
 
 use crate::checker::wikilink::resolver::WikilinkResolver;
@@ -177,12 +176,14 @@ impl FileChecker {
 
         let mut path_buf = path.to_path_buf();
 
-        // If existing "extension" has spaces, we assume it is not a real extension
-        // and we want to append fallbacks _after_ the full original filename.
-        if let Some(ext) = path.extension().map(OsStr::as_encoded_bytes)
-            && ext.iter().any(u8::is_ascii_whitespace)
+        // If existing "extension" has spaces, we assume it is not a real extension.
+        // In this case, we want to append fallbacks _after_ the full original filename.
+        if let Some(ext) = path.extension()
+            && ext.as_encoded_bytes().iter().any(u8::is_ascii_whitespace)
         {
-            path_buf.add_extension("temporary-ext");
+            let mut ext = ext.to_os_string();
+            ext.push(".extra-extension");
+            path_buf.set_extension(ext);
         }
 
         // Try fallback extensions, replacing any current extension in the path.
--- a/lychee-lib/src/checker/file.rs
+++ b/lychee-lib/src/checker/file.rs
@@ -182,8 +182,10 @@ impl FileChecker {
             && ext.as_encoded_bytes().iter().any(u8::is_ascii_whitespace)
             && let Some(first_fallback) = &self.fallback_extensions[..].first()
         {
-            path_buf.push(".");
-            path_buf.push(first_fallback);
+            let mut ext = ext.to_os_string();
+            ext.push(".");
+            ext.push(first_fallback);
+            path_buf.set_extension(ext);
         }
 
         // Try fallback extensions, replacing any current extension in the path.
--- a/lychee-lib/src/checker/file.rs
+++ b/lychee-lib/src/checker/file.rs
@@ -178,14 +178,14 @@ impl FileChecker {
         let mut path_buf = path.to_path_buf();
 
         // If existing "extension" has spaces, we assume it is not a real extension
-        // and we want to append fallbacks _after_ the (full) original filename.
+        // and we want to append fallbacks _after_ the full original filename.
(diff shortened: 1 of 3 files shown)


lychee-lib/src/checker/file.rs
```diff
@@ -174,8 +174,19 @@ impl FileChecker {
             return Ok(path.to_path_buf());
         }
 
-        // Try fallback extensions
         let mut path_buf = path.to_path_buf();
+
+        // If existing "extension" has spaces, we assume it is not a real extension.
+        // In this case, we want to append fallbacks _after_ the full original filename.
+        if let Some(ext) = path.extension()
+            && ext.as_encoded_bytes().iter().any(u8::is_ascii_whitespace)
```
mre, line 182: I'm sure it makes sense but why not [is_whitespace](https://doc.rust-lang.org/std/primitive.char.html#method.is_whitespace)?

mre (approved):
Thanks. I think the fix looks fine and it would unblock a few users, so I'm happy to merge it.

To your point about appending extensions, I agree that it might break a bunch of use-cases and in general, it would be less intuitive than the current behavior. E.g. as a user, I'd expect the extensions to be exactly the ones I specified in the option, because otherwise I'd have to look at the docs or the source code to find out what's going on. 😉

lychee-lib/src/checker/file.rs
katrinafyi: Oh, because the extension is an OsStr which is not guaranteed to be utf8. So we only get a `&[u8]` and we can only reliably detect ASCII. Lmk if you want a comment :)

mre: Ah, of course. Sorry.

lychee-lib/src/checker/file.rs
```diff
@@ -174,8 +174,19 @@ impl FileChecker {
...
+        {
+            let mut ext = ext.to_os_string();
+            ext.push(".extra-extension");
```
mre: That's perhaps the only little wrinkle. I'm guessing you didn't want to refactor the code too much to fix that issue? It's okay to merge, but can we add a comment about that so we know that we should refactor that in the future?

katrinafyi: I've changed it to avoid the magic string in https://github.com/lycheeverse/lychee/pull/2236/changes/bf010683fb998d148938f7fca89e37a27505723d..caa4e985d434facd15c2c7579754bf02a05c2c0b, let me know if that's better.

And yeah, I wanted to reuse the existing code as much as possible so I just append a new extension first to avoid overwriting the space-containing extension.

mre: Yes, that's nice. Thanks.

mre:
This seems fine to me. Thanks for the update!

One additional test worth having would be to test that the second fallback still works.
For example:

```rust
assert_resolves!(
    &checker,
    "fallback-extensions/file. hi",
    Ok("fallback-extensions/file. hi.gz")
);
```

where only `file. hi.gz` exists and the fallback list is `["md", "gz"]`.

That proves the first fallback is only used to reshape the path, and the later loop still correctly tries all fallback extensions. Right now, your change covers `file. hi.md`; adding the `.gz` variant would make the tricky `set_extension` behavior a bit clearer. wdyt?

Either way, that's a minor thing, and we could merge it either way, so I've approved it.

mre (approved):

lychee-lib/src/checker/file.rs
```diff
@@ -174,8 +174,21 @@ impl FileChecker {
...
         }
 
-        // Try fallback extensions
         let mut path_buf = path.to_path_buf();
+
+        // If existing "extension" has spaces, we assume it is not a real extension.
+        // In this case, we want to append fallbacks _after_ the full original filename.
+        if let Some(ext) = path.extension()
+            && ext.as_encoded_bytes().iter().any(u8::is_ascii_whitespace)
+            && let Some(first_fallback) = &self.fallback_extensions[..].first()
```
mre: This is slightly confusing. 
It can just be

```rust
&& let Some(first_fallback) = self.fallback_extensions.first()
```

mre:
Perfect. That looks very fine now. Thanks, @katrinafyi, for adding the additional test.

PR #2237, opened 2026-06-20
fix: expand tilde in local file links

### LINKED ISSUE #1503: Weird/incorrect behaviour when checking local links
I'm using lychee 0.15.1 on Arch Linux.

Here is an example markdown document:

```md
# Incorrect or strange behaviour
## Links not being checked
- [ ] [Test1](~/Downloads/not checked because of space)
- [ ] [Test3](/home/nonsense/not-checked2-absolute-path)

## Links checking not working because of tilde
- [Test4](~/incorrect-check)
<!-- ✗ [ERR] file:///home/me/Documents/~/incorrect-check | Failed: Cannot find file -->

# Links checked correctly
- [Test5](./correct-check)
<!-- ✗ [ERR] file:///home/me/Documents/correct-check | Failed: Cannot find file -->
```

```sh
$ lychee test.md 
  2/2 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                             Issues found in 1 input. Find details below.

[test.md]:
✗ [ERR] file:///home/fenuks/Dokumenty/notatki/correct-check | Failed: Cannot find file
✗ [ERR] file:///home/fenuks/Dokumenty/notatki/~/incorrect-check | Failed: Cannot find file

🔍 2 Total (in 0s) ✅ 0 OK 🚫 2 Errors
```

It seems that lychee skips links with absolute path or spaces, and it doesn't handle treats links with `~` as absolute links.

xub5210-droid:
## Summary
- expand `~/...` links against the user home directory when resolving local file links
- keep `~/...` unchanged for non-file bases such as HTTP pages
- add regression coverage for local file and HTTP bases

Fixes #1503.

## Tests
- cargo fmt --check
- cargo test -p lychee-lib types::base_info::tests -- --nocapture
- cargo test -p lychee-lib collector::tests::test_file_path_with_base -- --nocapture
- cargo clippy -p lychee-lib --all-targets -- -D warnings

### DIFF
--- a/lychee-lib/src/types/base_info.rs
+++ b/lychee-lib/src/types/base_info.rs
@@ -289,7 +289,6 @@ impl BaseInfo {
     /// Returned errors include [`ErrorKind::RootRelativeLinkWithoutRoot`]
     /// and [`ParseError::RelativeUrlWithoutBase`] (within [`ErrorKind::ParseUrl`]).
     #[expect(clippy::unnested_or_patterns, reason = "more readable here")]
-    #[expect(clippy::match_same_arms, reason = "we need to comment one of the arms")]
     pub fn resolve_relative_link(&self, rel: &RelativeUri<'_>) -> Result<Url, ErrorKind> {
         match (self, &rel) {
             (Self::None, RelativeUri::Root(_)) | (Self::NoRoot(_), RelativeUri::Root(_)) => {
@@ -300,7 +299,9 @@ impl BaseInfo {
 
             (Self::None, _) => Err(ParseError::RelativeUrlWithoutBase),
 
-            (Self::NoRoot(base), RelativeUri::Local(text)) => base.join(text),
+            (Self::NoRoot(base), RelativeUri::Local(text)) => {
+                base.join(&expand_tilde_for_file_base(base, text))
+            }
 
             // `(Self::NoRoot, RelativeUri::Scheme)` happens when a link like `///a` occurs
             // within a local file without root-dir. note the triple slash because file
@@ -318,7 +319,10 @@ impl BaseInfo {
             }
 
             (Self::Full { origin, path }, rel) => {
-                origin.join(path).and_then(|x| x.join(rel.link_text()))
+                let text = rel.link_text();
+                origin
+                    .join(path)
+                    .and_then(|x| x.join(&expand_tilde_for_file_base(origin, text)))
             }
         }
         .map_err(|e| ErrorKind::ParseUrl(e, rel.link_text().to_string()))
@@ -361,6 +365,14 @@ impl BaseInfo {
         } else {
             self.resolve_relative_link(&rel)
         }
+    }
+}
+
+fn expand_tilde_for_file_base<'a>(base: &Url, text: &'a str) -> Cow<'a, str> {
+    if base.scheme() == "file" && (text == "~" || text.starts_with("~/")) {
+        shellexpand::tilde(text)
+    } else {
+        Cow::Borrowed(text)
     }
 }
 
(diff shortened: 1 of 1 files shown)


lychee-lib/src/types/base_info.rs
```diff
@@ -482,6 +494,32 @@ mod tests {
...
+        let base = BaseInfo::try_from("file:///home/me/docs/page.md").unwrap();
+
+        let url = base
+            .parse_url_text("~/Downloads/not checked because of space")
+            .unwrap();
+
+        assert_eq!(
+            url,
+            Url::from_file_path(PathBuf::from(
+                shellexpand::tilde("~/Downloads/not checked because of space").as_ref()
```
katrinafyi, line 508: What is "not checked because of space" for?

lychee-lib/src/types/base_info.rs
```diff
@@ -300,7 +299,9 @@ impl BaseInfo {
 
             (Self::None, _) => Err(ParseError::RelativeUrlWithoutBase),
 
-            (Self::NoRoot(base), RelativeUri::Local(text)) => base.join(text),
+            (Self::NoRoot(base), RelativeUri::Local(text)) => {
+                base.join(&expand_tilde_for_file_base(base, text))
```
katrinafyi, line 303: rather than injecting your function into all the applicable cases at the end, if I were to do this, I'd probaby do the tilde expansion earlier before this function.

katrinafyi (changes requested):
Was there consensus on whether this is something that should be done? https://github.com/lycheeverse/lychee/issues/1503#issuecomment-2578983151 was unsure.

I've added some quick code comments, but the main question is whether the issues raised in https://github.com/lycheeverse/lychee/issues/1503#issuecomment-2578983151 are resolved.

mre:
> Was there consensus on whether this is something that should be done? [#1503 (comment)](https://github.com/lycheeverse/lychee/issues/1503#issuecomment-2578983151) was unsure.
> 
> I've added some quick code comments, but the main question is whether the issues raised in [#1503 (comment)](https://github.com/lycheeverse/lychee/issues/1503#issuecomment-2578983151) are resolved.

Yeah, so I took another look and the only exception where `~` does get expanded on Windows is inside PowerShell. Or at least I think that's what's going on according to [this page](http://xahlee.info/powershell/powershell_path_expansion.html).

I also tested it in an online editor (because I don't have access to a Windows machine). Here is my [test script](https://compilebytes.com/share/97a-KKVvaL).

With that, I'd say the correct way is to detect if we run inside PowerShell on Windows, but it's probably fine if we just leave it out and add a comment which mentions that Windows is complicated and that we deliberately leave that case unsupported for now.

katrinafyi:
Hmm personally, I don't think we should be expanding tildes in URLs. A URL is not used in the shell so I don't think comparing to a Unix or Windows shell is the right thing to do. 

Rather, we should look at whether any browsers or markdown editors would expand tildes in this position. A browser won't expand a tilde - it's a valid URL character and, indeed, many university URLs use it like https://cs.au.dk/~amoeller/spa/spa.pdf (which could be a relative link like `~amoeller/spa/spa.pdf`). This leaves the question of markdown editors, and I'm less familiar with markdown editors.

We should also consider whether expanding tilde makes sense in HTML files. I think it doesn't.

So, I would err on the side of not expanding tildes unless someone makes a compelling case. And I don't know if @xub5210-droid is making a case or just picking up the issue.

Also, if someone desperately needs this feature, it can be done... with remaps :)

Ranveersingh1113:
Thanks for the ping, @mre. I went through the diff and the thread, and tried to put
the design question on firmer footing by checking how `~/…` links actually resolve.

I resolved `~/Downloads/file.md` against a `file://` base using both Python's RFC-3986
urljoin and Node's WHATWG URL parser (the browser's parser), and rendered the same link
through python-markdown and markdown-it. In every case `~` is kept literal:
  - file base  -> file:///home/me/docs/~/Downloads/file.md
  - http base  -> "https://example.com/docs/~/profile"
  - markdown   -> "<a href="~/Downloads/file.md">"

So no browser, URL parser, or Markdown renderer expands `~`. To me that turns this into
a correctness point rather than a preference: lychee should validate a link the way the
consumer will resolve it, and consumers leave `~` literal. Expanding it inside lychee
would make a `~/…` link pass here while the published page still 404s — a false negative.

So I lean with @katrinafyi: I don't think we should expand tildes, and `remap` is the
right escape hatch for the rare person who really wants it. (FWIW the `~amoeller/…` worry
isn't triggered by this PR specifically — it only matches `~` and `~/…`, not `~user/…`,
and there's a test confirming HTTP bases are untouched.)

One more note: this PR only covers the tilde half of #1503; the "links with spaces"
behaviour in that issue isn't addressed here.

Happy to help!!

mre:
OK, so the verdict is to close this? @katrinafyi, @xub5210-droid is that fine with you?

mre:
I'll go ahead and close this, sorry. Nothing wrong with the code per-se, but it's a risky change and might be semantically incorrect. Unconditional tilde-expansion can have unexpected side effects for web-URLs, as we've uncovered. (E.g. for university URLs like https://stanford.edu/~sachour)

Even though the code fixes some other edge-cases, it's tricky to get the behavior right for all cases.

PR #2239, opened 2026-06-22
fix(quirks): check GitHub #readme fragments via REST API

DiazSk:
## Summary

When lychee encounters a GitHub URL of the form
`github.com/{owner}/{repo}[/tree/{branch}]#readme`, it currently fetches the full
GitHub HTML page to locate the `#readme` anchor. This is fragile and rate-limit-prone
(HTTP 429/403 under unauthenticated load). The `#readme` fragment names a resource
(the repository README file), not a DOM anchor — so the correct fix is to call the
GitHub REST API rather than scrape HTML.

Closes #2128

## Changes

Adds a new `Quirk` in `lychee-lib/src/quirks/mod.rs` that:

1. Detects `github.com/{owner}/{repo}[/tree/{branch}]#readme` URLs via a static
   `LazyLock<Regex>` (pattern mirrors the existing `GITHUB_BLOB_*` quirks).
2. Rewrites the request to `https://api.github.com/repos/{owner}/{repo}/readme`
   (with optional `?ref={branch}`) and sets `Accept: application/vnd.github.v3+json`.
3. A `200 OK` from the API is treated as success — fragment resolution is skipped
   because the rewritten URL carries no fragment.

Also adds:
- `GitHubTokenHandler` in `checker/website.rs` — injects `Authorization: Bearer`
  for `api.github.com` requests after Quirks has rewritten the URL. Holds
  `Option<String>` so the chain is always statically defined.
- `RemapChainHandler` in `checker/website.rs` — applies `--remap` rules after
  Quirks rewrites the URL, which enables integration-test interception via wiremock.

No new dependencies. No `octocrab` usage. Pure-reqwest layer only.

## Test Plan

```bash
cargo test -p lychee-lib -- quirks   # unit tests
cargo test -p lychee-bin -- cli      # integration tests (wiremock)
```

New unit test `test_github_readme_quirk` uses `#[rstest]` + `#[case]` (same
pattern as the existing `test_github_blob_line_fragment_regex`) with 4 cases:
base repo, `/tree/{branch}` variant, blob URL (no match), non-readme fragment
(no match).

Two new integration tests assert that `Authorization: Bearer` and
`Accept: application/vnd.github.v3+json` are present on the final request to
the mocked API endpoint.

## Acceptance Criteria

- [x] New unit tests added covering match/no-match/branch cases
- [x] All existing quirk tests still pass
- [x] Integration tests verify `Authorization: Bearer` and `Accept` headers end-to-end
- [x] Follows project test conventions (`rstest` + `#[case]`, `MockRequest` wrapper)
- [x] No new dependencies added
- [x] No breaking changes to the public API


katrinafyi:
Under what circumstances would you expect this link check to succeed or fail? FOr example, should it fail if a repository doesn't have a readme? Can this be tested?

> Should the branch be forwarded via `?ref={branch}`

Yes.

DiazSk:
Thanks for the review, @katrinafyi 

**Success/fail semantics:**

The GitHub REST API `GET /repos/{owner}/{repo}/readme` returns:
- `200 OK` — the repo has a README → link succeeds  
- `404 Not Found` — the repo has no README (or doesn't exist) → link fails  
- `403 Forbidden` — private repo without a valid token → link fails  

So yes, if `github.com/owner/repo#readme` points to a repo with no README, the check correctly fails which matches user intent (they linked to something that doesn't render). lychee's existing `Status::new()` already maps non-200 responses to `ErrorKind::RejectedStatusCode` without any extra handling from us.

**Testability:**

The unit tests here only cover URL rewriting (since the actual HTTP call happens downstream of the quirk). For the HTTP behavior, I can add an integration test using `wiremock` (already used in `checker/website.rs`) that mocks the API returning 404 and asserts lychee reports the link as broken happy to add that in this PR if you'd like it.

**Branch fix:**

Thanks for catching that I just pushed a fix that forwards the branch via `?ref={branch}` when present. Updated `test_github_readme_tree_branch_request` to assert the correct expected URL (`?ref=main`). The no-branch case still rewrites to the bare `/readme` endpoint as before.

katrinafyi:
Yes add mock integration test in lychee-bin/tests/cli.rs The mock test should also check that a `Authorization: Bearer` header is present.

lychee-lib/src/quirks/mod.rs
```diff
@@ -112,6 +116,24 @@ impl Default for Quirks {
...
+                name: "check GitHub README existence via API",
+                pattern: &GITHUB_README_PATTERN,
+                rewrite: |mut request, captures| {
+                    let owner = captures.name("owner").unwrap().as_str();
+                    let repo = captures.name("repo").unwrap().as_str();
+                    let api_url = if let Some(branch) = captures.name("branch") {
+                        format!("https://api.github.com/repos/{owner}/{repo}/readme?ref={}", branch.as_str())
+                    } else {
+                        format!("https://api.github.com/repos/{owner}/{repo}/readme")
+                    };
```
katrinafyi: use captures.expand to do the mandatory fields, then parse to url, then use https://docs.rs/url/2.5.4/url/struct.Url.html#method.query_pairs_mut to optionally add the ref while handling encoding.

lychee-lib/src/quirks/mod.rs
```diff
@@ -315,4 +338,55 @@ mod tests {
...
+        let url = Url::parse("https://github.com/lycheeverse/lychee#readme").unwrap();
+        let request = Request::new(Method::GET, url);
+        let modified = Quirks::default().apply(request);
+        let expected_url =
+            Url::parse("https://api.github.com/repos/lycheeverse/lychee/readme").unwrap();
+        assert_eq!(
+            MockRequest(modified),
+            MockRequest::new(Method::GET, expected_url)
+        );
+    }
```
katrinafyi: i'd fold all of these tests together and just test the input url and output url using rstest. if the url changes, the pattern must have matched so u don't need to test is_match separately. you can also use Option if you want.

katrinafyi:
don't inject GitHub token into quirks. rather, just add a new entry to the Chain list in website.rs which attaches the token to all api.github.com requests.

DiazSk:
The wayback_suggestion_real failure is a pre-existing flaky test that hits web.archive.org over the network it timed out in CI but is unrelated to this PR. Could you re-run the failed job?

lychee-lib/src/checker/website.rs
```diff
@@ -218,11 +243,16 @@ impl WebsiteChecker {
...
-            Box::new(credentials),
-            Box::new(self.clone()),
-        ]);
+        let mut chain_items: Vec<Box<dyn Handler<Request, Status> + Send>> =
+            vec![Box::new(Quirks::default())];
+        if let Some(token) = &self.github_token {
+            chain_items.push(Box::new(GitHubTokenHandler {
+                token: token.clone(),
+            }));
+        }
```
katrinafyi: I'd just make GitHubTokenHandler hold an option to keep the Chain statically-defined.

lychee-lib/src/checker/website.rs
```diff
@@ -515,11 +546,62 @@ mod tests {
...
+        let request = Request::new(Method::GET, url);
+
+        let ChainResult::Next(req) = handler.handle(request).await else {
+            panic!("expected ChainResult::Next");
+        };
+        assert!(
+            req.headers().get("Authorization").is_none(),
+            "Authorization header should not be set for non-api.github.com hosts"
+        );
+    }
```
katrinafyi: i think the GitHubTokenHandler is very simple and these tests are too mundane to be useful.

lychee-bin/tests/cli.rs
```diff
@@ -4638,6 +4638,51 @@ exclude_path = ["exclude_package.txt"]
...
+
+        let url = format!("{}/repos/no-readme-repo/readme", mock_server.uri());
+
+        cargo_bin_cmd!()
+            .arg("--no-progress")
+            .arg("-")
+            .write_stdin(url)
+            .assert()
+            .failure();
+    }
```
katrinafyi: CLI tests should be integration tests, but these do not exercise the fragment quirk logic at all, nor do they check for authorization header. before https://github.com/lycheeverse/lychee/pull/2239/changes/cfcdb4fd79e3a31c4e3685aab0dd4c395b48a8e0, you passed in a github URL and remapped api.github.com to wiremock. why did you get rid of that?

lychee-lib/src/quirks/mod.rs
```diff
@@ -24,6 +24,12 @@ static GITHUB_BLOB_LINE_FRAGMENT_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^https://github\.com/(?<user>.*?)/(?<repo>.*?)/blob/(?<path>.*?)#L\d+(?:-L?\d+)?$")
         .unwrap()
 });
+static GITHUB_README_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(
+        r"^https://github\.com/(?<owner>[^/]+)/(?<repo>[^/]+)(?:/tree/(?<branch>[^/]+))?#readme$",
```
katrinafyi, line 29: it's not uncommon to have a branch with / in its name, e.g. https://github.com/lycheeverse/lychee/tree/feat/per-host-rate-limiting/lychee-lib. thoughts?

katrinafyi (commented):
Some questions and changes....

lychee-lib/src/quirks/mod.rs
DiazSk: Done. The rewrite closure now uses captures.expand("https://api.github.com/repos/$owner/$repo/readme", &mut api_url) for the mandatory fields, then url.query_pairs_mut().append_pair("ref", branch.as_str()) for the optional branch parameter. Using query_pairs_mut means / in branch names is safely percent-encoded as %2F automatically.

lychee-lib/src/quirks/mod.rs
DiazSk: Done. All four README tests are now a single test_github_readme_quirk rstest that maps input URL → expected output URL. Dropped the separate is_match test since a URL change already implies the pattern matched. Added a fourth case covering branches with / in the name (see thread 6).

lychee-lib/src/checker/website.rs
DiazSk: Done. The field is now token: Option<String> and the chain is always statically defined as [Quirks, GitHubTokenHandler, RemapChainHandler, credentials, WebsiteChecker] no more conditional push.

lychee-lib/src/checker/website.rs
DiazSk: Removed both tests. Agreed the logic is trivial enough to not justify direct unit coverage.

lychee-bin/tests/cli.rs
DiazSk: You're right that those tests were wrong. I removed them in cfcdb4f because I'd discovered that --remap fires on the original URI in Client::check(), before the chain runs so --remap "api.github.com → mock" with a github.com#readme input never intercepts anything, because Quirks hasn't rewritten the URL yet when remap runs.

To fix this properly rather than just removing the tests, I added a RemapChainHandler to the default chain (between GitHubTokenHandler and credentials). It applies the configured remap rules inside the chain, after Quirks and after token injection. The pipeline is now:

1. Quirks: github.com/repo#readme → api.github.com/repos/owner/repo/readme
2. GitHubTokenHandler: host is api.github.com → injects Authorization: Bearer
3. RemapChainHandler: pattern api.github.com matches → rewrites to mock_server
4. WebsiteChecker: sends request to mock with Bearer header already set 

The original tests are restored and now correctly verify the full end-to-end path.

lychee-lib/src/quirks/mod.rs
DiazSk: Thanks for catching that I changed the branch capture group from [^/]+ to [^#]+, which captures everything up to the #readme fragment and allows slashes. Also added a test case for the exact example you gave (feat/per-host-rate-limiting/lychee-lib), and the expected output is ?ref=feat%2Fper-host-rate-limiting%2Flychee-lib query_pairs_mut handles the encoding correctly.

katrinafyi: I asked for your thoughts, not a guess at a quick fix. And the quick fix is not right because `/lychee-lib` is a path and shuoldn't be part of the branch name.

What are your thoughts?

lychee-bin/tests/cli.rs
katrinafyi: This is a tricky case because it doesn't fit nicely into the existing pipeline, but applying the remaps again is a bad idea for many reasons.

Because it's a tricky case, I think it's okay to do something slightly tricky to test it. Inside GitHubTokenHandler, you could add a `#[cfg(test)]` condition that if it's in testing and a certain environment variable is set, it will additionally change the host to be that value. Then, you could pass the mock server URL through this special environment variable.

But idk, @mre @thomas-zahner what're your thoughts?

katrinafyi (changes requested):
B y the way, do you know that the purpose of conventional commits is to make it easier to extract a semi-structured changelog from the commit log? There's not much point in using CC in commits of a WIP PR that will get squashed.

lychee-lib/src/quirks/mod.rs
DiazSk: You're right @katrinafyi, [^#]+ is incorrect, and I should have thought about this thoroughly before going to a hasty fix. The main issue is that GitHub's /tree/{ref}/{path} URL is unclear when branch names include slashes: tree/a/b/c#readme might imply branch=a/b, dir=c or branch=a/b/c, dir=''. To resolve this correctly, we must call the GitHub branches API to determine which branches exist; we cannot do it based solely on the URL.

For the scope of this feature (validating #readme fragments), I'd go with [^/]+ and keep the limitation apparent. With [^/]+, URLs like tree/feat/per-host-rate-limiting/lychee-lib#readme do not fit the pattern and fall through to a standard HTTP check, which is safe and honest. This quirk is most useful for root-level README links (owner/repo#readme or tree/main#readme), when branch names with slashes are ignored.

To handle the whole situation, we'd need to either (a) try several branch/path splits over the GitHub API, or (b) use GET /repos/{owner}/{repo}/readme/{dir}?ref={branch}, which we can't do without knowing the split. I'm happy to create the disambiguation if you believe it's worth the complexity, but my feeling is that [^/]+ with a disclosed constraint is the best choice here.

DiazSk:
> B y the way, do you know that the purpose of conventional commits is to make it easier to extract a semi-structured changelog from the commit log? There's not much point in using CC in commits of a WIP PR that will get squashed.

A fair point, and you are correct. I was using the format instinctively without considering why it existed. I'll use simple descriptive messages for the remaining commits.

lychee-lib/src/quirks/mod.rs
katrinafyi: > I'd go with [^/]+ and keep the limitation apparent. 

Yeah

lychee-lib/src/checker/website.rs
```diff
@@ -13,10 +14,52 @@ use crate::{
...
+            let uri = Uri {
+                url: request.url().clone(),
+            };
+            if let Ok(Some(remap)) = remaps.remap(&uri) {
+                *request.url_mut() = remap.new.url;
+            }
+        }
+        ChainResult::Next(request)
+    }
+}
```
mre, line 61: I'm a bit confused why we need those handlers? Maybe we can add some more documentation around that. Especially the RemapChainHandler is puzzling me. I thought we always execute remaps as part of the request chain?

mre:
Thanks a lot @katrinafyi for shepherding this PR. 

This seems fine to me, although I'm a bit puzzled by the RequestChain changes. Not sure why that is needed (?).

I would not mind merging this, but I want to get more than one approval first to see that we're all aligned.
@katrinafyi, @thomas-zahner wdyt?

lychee-lib/src/checker/website.rs
DiazSk: GitHubTokenHandler: the Quirk rewrite is a plain function pointer (fn(Request, Captures) -> Request) and cannot capture state, so token injection can't happen inside the Quirk itself. The handler fires after Quirks has rewritten the URL to api.github.com, checks the host, and injects Authorization: Bearer. It is a no-op when no token is configured.

RemapChainHandler: remaps currently execute on the original URI in Client::check(), before the chain runs. With input github.com/repo#readme and --remap api.github.com localhost:PORT, the remap evaluates against the original URL (no match) and the Quirks rewrite to api.github.com happens too late for the pre-chain remap to catch it. The handler re-applies remaps post-Quirks so the integration tests can use wiremock correctly: the mock receives the post-Quirks API request with the Bearer header already set. Without it, those tests silently call the real GitHub API instead of the mock.

If consolidating all remap execution into the chain and removing the pre-chain step is the preferred direction, I can restructure that way, but it felt out of scope for this PR.

mre: Yeah, I like that better. Not sure what the implications of that might be, but maybe remaps could indeed be a normal part of the chain. It's one straightforward pass and we have less individual handlers doing odd things like that, which will later bite us.

mre: Quick update, and sorry for the confusion: I take it back. After discussing this internally, we probably don't want to have a single handler for the entire remap chain. The reason is that the chain only affects websites right now and doesn't handle files. So moving it into the chain would break existing functionality. 

It's probably okay for now to accept your code as-is, with the minor caveat that we have to do a make-over in the future and resolve some of the underlying design issues, but that's out-of-scope for this change.

So, all good from my end for now and no further work is needed at this point. Please ignore my last comment above.

mre (approved):

mre:
Holding off on the merge to get one more review first.

katrinafyi (commented):
It's probably okay as is. I have these outstanding comments in mind:

- `#readme` fragment checking applies regardless of --include-fragments
- A failed `#readme` fragment check will appear as a 404 rather than a fragment-specific error.
- Applying the remaps twice could cause unexpected things. For instance, the remap history will not be recorded by second remap pass.

To be honest, these problems might be inevitable with the approach I suggested (sorry for leading you down the wrong path). It was a lofty goal to re-use the quirks system for Github, but I think the process of this PR has shown that Github checking does need more specialised and centralized handling.

DiazSk:
> It's probably okay as is. I have these outstanding comments in mind:
> 
> * `#readme` fragment checking applies regardless of --include-fragments
> * A failed `#readme` fragment check will appear as a 404 rather than a fragment-specific error.
> * Applying the remaps twice could cause unexpected things. For instance, the remap history will not be recorded by second remap pass.
> 
> To be honest, these problems might be inevitable with the approach I suggested (sorry for leading you down the wrong path). It was a lofty goal to re-use the quirks system for Github, but I think the process of this PR has shown that Github checking does need more specialised and centralized handling.

Thanks for summarizing these. Agreed on all three points:

- The --include-fragments bypass is a real gap; #readme resolves via an API call rather than fragment validation, so the flag has no meaningful effect here.
- The 404 vs fragment error distinction would need a wrapper to translate the API response into a proper fragment error type, which feels out of scope.
- The double-remap is a side effect of RemapChainHandler re-applying rules post-quirks. As @mre noted, the chain architecture will need a proper overhaul to handle this cleanly across files and websites.

These are good future issues to track. If you're comfortable with the current state given those caveats, I'll leave it as is.

mre:
Okay, I think this is mergeable, but quirks probably were the wrong place to add this.

As a rule of thumb, we should use quirks only when all of this is true:

- transformation is purely syntactic,
- no credentials,
- no network call needed to decide,
- no provider-specific error semantics,
- normal website checking remains correct after rewrite.

Or at least, that's how I think of quirks right now.

It could be that in the future are more than just syntactic replacements, but that's a dangerous alley to follow. It also didn't help how quirks have to interface with remaps for this PR.

In the context of this PR, we make a network call, so that's a bit of a no-go. We should clean this up soon.
On top of that, the regex-based approach becomes unwieldy at this stage. It's about as far as I'd push the regex engine before moving to something more "state-based."

Since GitHub is such an important part of our pipeline, maybe it's time to add a GitHub specific checker now.

I will think a little longer before merging this. The reason is that a lot of the code will go away in a future version of this and the issue is not pressing enough to get it merged quickly. Hope that's okay @DiazSk.

DiazSk:
That makes complete sense. Looking at those five criteria, this PR violates three of them: it makes a network call, uses credentials, and produces provider-specific error semantics (a 404 instead of a proper fragment error). You're right that quirks was the wrong layer.

Happy to wait while you think it through. If it would be more useful, I could take a first pass at a dedicated GitHub checker instead and close this PR in favor of that. Just let me know which direction you'd prefer.

mre:
Working on it in https://github.com/lycheeverse/lychee/pull/2254 if you'd like to take a look and perhaps help with the review. 😊 I'd be very thankful for any thoughts.

PR #2243, opened 2026-06-24
feat: list ignored (IGNORED) links in the run summary

### LINKED ISSUE #997: Make `IGNORED` URLs failures more visible in the summary
# Issue summary

I experienced a slight usability issue on a recent change from lychee 0.10 to 0.11.

From my limited understanding of running lychee (without digging into the code), `IGNORED` URLs were just skipped in lychee 0.10, but now seem to be *failures* in lychee 0.11.
 
If that is correct, my suggestion is that `IGNORED` URLs are listed in the summary explicitly, to perhaps make it more apparent that `IGNORED` URLs are the cause of failure.

# Example minimal repository

This [minimal repository](https://github.com/StevenMaude/lychee-issue-991-demo) uses lychee-action to check one valid and one invalid URL (with `hhttps://` protocol). There are two workflow runs with the lychee summaries shown below.

In that repository, the problem was obvious: there are only two URLs to check.

In the actual working repository, the problem wasn't obvious: there are more than two thousand URLs to check.

This took me a while to figure out. That included a diversion into trying to figure out if the problem was with lychee-action's `entrypoint.sh`, because I didn't see any errors in the lychee summary. It was only when I realised that the summary "Total" count didn't match the sum of the other counts that I noticed that there was a broken URL.

#### lychee 0.11 (:x: exit code 2)

https://github.com/StevenMaude/lychee-issue-991-demo/commit/3700ec69b48c0b204bc9ef526cf3920b71d6472e

```
✔ [200] https://github.com/
? [IGNORED] hhttps://github.com/ | Unsupported: Error creating request client
```

| Status        | Count |
|---------------|-------|
| 🔍 Total      | 2     |
| ✅ Successful | 1     |
| ⏳ Timeouts   | 0     |
| 🔀 Redirected | 0     |
| 👻 Excluded   | 0     |
| ❓ Unknown    | 0     |
| 🚫 Errors     | 0     |

#### lychee 0.10 (:heavy_check_mark: passing)

https://github.com/StevenMaude/lychee-issue-991-demo/commit/b6f23df5c3831502ca5cf0df424b00cb52635fc8

```
✔ [200] https://github.com/
? [IGNORED] hhttps://github.com/ | Unsupported: Error creating request client
```

| Status        | Count |
|---------------|-------|
| 🔍 Total      | 1     |
| ✅ Successful | 1     |
| ⏳ Timeouts   | 0     |
| 🔀 Redirected | 0     |
| 👻 Excluded   | 0     |
| ❓ Unknown    | 0     |
| 🚫 Errors     | 0     |

samiuls25:
## Summary

Closes #997.

When lychee can't turn a URL into a request (e.g. a malformed scheme like `hhttps://` or an unsupported scheme like `slack://`), it records the response as `Unsupported` and displays it as `IGNORED`. Until now, these links were counted in the summary but never listed, so the only hint that something was ignored was that `Total` didn't equal the sum of the visible categories (easy to miss in a repo with thousands of links).

This PR lists ignored links per input, mirroring how Errors, Timeouts, Redirects and Suggestions are already broken out.

## What changed

The root cause was in the data layer, not just formatting. `ResponseStats` counted unsupported responses but never stored them (so they fell through `add_response_status`'s final `else { return; }`), thus the formatters had no data to list. So here's what I changed:

- **`response.rs`** - adds an `unsupported_map` to `ResponseStats`, populated alongside the existing error/timeout/excluded maps (via `Status::is_unsupported()`).
- **`markdown.rs`** - adds an `## Ignored per input` section.
- **`compact.rs`** - lists ignored links per input, including inputs whose only finding is ignored links.
- **`detailed.rs`** - adds an `Ignored in <source>` section.
- **`json.rs`** - the new field is serialized, so JSON output now includes `unsupported_map` (snapshot updated).

### Example (markdown)

Before, the report said `⛔ Unsupported | 2` but listed none of them. After:

    ## Ignored per input

    ### Ignored in README.md

    * [IGNORED] <hhttps://example.com> (at 6:1) | Unsupported: ...

## Also included: a small related fix

While in `detailed.rs` I noticed the `⛔ Unsupported` summary row printed `stats.errors` instead of `stats.unsupported`, so it showed the error count under the Unsupported label. Fixed as a **separate commit** (`fix(stats): use the unsupported count in the detailed summary`) since it's on-topic for #997. Happy to split it into its own PR if you'd prefer or exclude it.

## A design choice worth flagging

In the compact formatter I included ignored links in the `Issues found in N inputs` count so the header matches the number of detail blocks shown. Ignored links don't affect the exit code (`is_success()` is unchanged). If we'd rather keep "Issues" strictly for errors/timeouts, I can drop them from that count while still listing them.

## Testing

- New unit tests: ignored responses are counted and stored (`response.rs`), plus per-formatter render tests in `markdown.rs` / `compact.rs` / `detailed.rs`, including the "only ignored links" case.
- All existing stats tests still pass; full `cargo test --bin lychee` is green and `cargo clippy` is clean.

### DIFF
--- a/lychee-bin/src/formatters/stats/detailed.rs
+++ b/lychee-bin/src/formatters/stats/detailed.rs
@@ -53,7 +53,7 @@ impl Display for DetailedResponseStats {
         write_stat(f, "👻 Excluded", stats.excludes, true)?;
         write_stat(f, "❓ Unknown", stats.unknown, true)?;
         write_stat(f, "🚫 Errors", stats.errors, true)?;
-        write_stat(f, "⛔ Unsupported", stats.errors, false)?;
+        write_stat(f, "⛔ Unsupported", stats.unsupported, false)?;
 
         let response_formatter = get_response_formatter(&self.mode);
 
@@ -155,7 +155,7 @@ mod tests {
 👻 Excluded.........0
 ❓ Unknown..........0
 🚫 Errors...........1
-⛔ Unsupported......1
+⛔ Unsupported......0
 
 Errors in https://example.com/
 [404] https://github.com/mre/idiomatic-rust-doesnt-exist-man (at 1:1) | 404 Not Found
@@ -231,7 +231,7 @@ Host: example.com
 👻 Excluded.........0
 ❓ Unknown..........0
 🚫 Errors...........0
-⛔ Unsupported......0
+⛔ Unsupported......1
 
 Ignored in https://example.com/
 [IGNORED] https://example.com/ignored | Unsupported: URL is missing a hostname"
--- a/lychee-bin/src/formatters/stats/detailed.rs
+++ b/lychee-bin/src/formatters/stats/detailed.rs
@@ -72,6 +72,17 @@ impl Display for DetailedResponseStats {
             write_stats(f, "Redirects", source, stats.redirect_map.get(source))?;
         }
 
+        // Ignored (unsupported) links get their own section so they are not
+        // mislabeled as errors, and so inputs whose only finding is an ignored
+        // link are still listed.
+        for (source, responses) in super::sort_stats_iter(stats.unsupported_map.iter()) {
+            write!(f, "\n\nIgnored in {source}")?;
+
+            for response in responses {
+                write!(f, "\n{}", response_formatter.format_response(response))?;
+            }
+        }
+
         Ok(())
     }
 }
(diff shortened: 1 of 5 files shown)


lychee-bin/src/formatters/stats/detailed.rs
```diff
@@ -72,6 +72,17 @@ impl Display for DetailedResponseStats {
...
+        // Ignored (unsupported) links get their own section so they are not
+        // mislabeled as errors, and so inputs whose only finding is an ignored
+        // link are still listed.
+        for (source, responses) in super::sort_stats_iter(stats.unsupported_map.iter()) {
+            write!(f, "\n\nIgnored in {source}")?;
+
+            for response in responses {
+                write!(f, "\n{}", response_formatter.format_response(response))?;
+            }
+        }
```
mre, line 84: This raises the question whether we need to align our wording in the code. Sometimes we deal with "unsupported" links and sometimes with "ignored" links. These are not the same, but they are related.

It's not a blocker for that PR, just an observation.

mre (approved):

mre:
This seems all fine to me. Thanks. 😊

PR #2245, opened 2026-06-26
fix: default-extension not applied to local files, and tweak help

### LINKED ISSUE #2244: `--default-extension` is not extracting links without URL-scheme
PR that introduced the feature: #1842

I found this, while looking for `--fallback-extensions` for #2235 / <https://github.com/lycheeverse/lychee/issues/1503#issuecomment-4745421581>, but wanted to make sure it's not just my user error. Basically means, I'm not depending on this.

---

`--default-extension` is not extracting links without URL-scheme.
**TLDR:** Maybe the flag is never used properly (on files?) / the `file_type_hint` gets lost?

```md
$ lychee --version
lychee 0.24.2
```

### Test case:

```md
# `my-extensionless-file`

[link 1 with scheme](https://lychee.cli.rs/)
[link 2 with scheme](mailto:44443899+nobkd@users.noreply.github.com)
[schemeless link 1](file.txt)
[schemeless link 2](//otherfile.txt)
```

Also create a file with the same content but with an extension, as `my-extension-file.md`.

```md
$ lychee my-extensionless-file --root-dir . --default-extension md --dump --include-mail
https://lychee.cli.rs/
mailto:44443899+nobkd@users.noreply.github.com
```

vs with the extension:

```md
$ lychee my-extension-file.md --root-dir . --default-extension md --dump --include-mail
https://lychee.cli.rs/
mailto:44443899+nobkd@users.noreply.github.com
file:///path/to/file.txt
file://otherfile.txt/
```

Also without `--default-extension md` gives the same result as with it:

```md
$ lychee my-extensionless-file --root-dir . --dump --include-mail
https://lychee.cli.rs/
mailto:44443899+nobkd@users.noreply.github.com
```

### Additional:

It feels like the file is just treated like unknown file format.
Soooo maybe the file is not really processed as markdown even with the flag?

The test [`test_default_extension_option`](https://github.com/lycheeverse/lychee/blob/467197f13e30c16befe54b8f83c17bf91952a741/lychee-bin/tests/cli.rs#L3502-L3517) also just checks the links for the remote link.

---

Slightly **off-topic**, but some remarks:

I at first thought, [`--default-extension`](https://lychee.cli.rs/guides/cli/#--default-extension) was like `--fallback-extensions`. Maybe the description could be worded a bit better to clarify it's not. Maybe it's just a me problem though. 🤷

For example: Maybe don't call it "applied"?

> This is the default file extension that is applied to files without an extension.
> - [Lychee Docs](https://lychee.cli.rs/guides/cli/#--default-extension)

Something like the following could be added:

> `--default-extension` is a flag to interpret (treat) extension-less input files as the provided format, for link extraction, it has no influence on finding files by extension-less link.

or 

> This is only used to handle file content, and does not append the extension to locate files. For that, see `--fallback-extensions`

But this were just two ideas, not sure if they're any good.

This of course only makes sense, when I understood the flag correctly.

katrinafyi:
This implementation is not the best because any plaintext file is changed to the default extension, but a file could genuinely be a plaintext file. To do this properly would need something like a TryFrom for FileType, but this conflicts with the From which is heavily used. It's probably not a big deal right now because the HTML and markdown extractors also perform plaintext extraction.

Also tries to improve the CLI help text based on suggestions in issue.

Fixes https://github.com/lycheeverse/lychee/issues/2244

### DIFF
--- a/benches/src/extract.rs
+++ b/benches/src/extract.rs
@@ -42,10 +42,10 @@ fn benchmark(c: &mut Criterion) {
 
     runtime.block_on(async {
         inputs = vec![
-            Input::path_content("../fixtures/bench/elvis.html", None)
+            Input::path_content("../fixtures/bench/elvis.html", None, None)
                 .await
                 .unwrap(),
-            Input::path_content("../fixtures/bench/arch.html", None)
+            Input::path_content("../fixtures/bench/arch.html", None, None)
                 .await
                 .unwrap(),
         ];
--- a/lychee-lib/src/types/input/input.rs
+++ b/lychee-lib/src/types/input/input.rs
@@ -13,6 +13,7 @@ use crate::{ErrorKind, LycheeResult};
 use async_stream::try_stream;
 use futures::stream::{Stream, StreamExt};
 use log::debug;
+use std::ffi::OsStr;
 use std::io::IsTerminal;
 use std::path::{Path, PathBuf};
 use tokio::io::{AsyncReadExt, stdin};
@@ -257,17 +258,11 @@ impl Input {
         let path = path.into();
         let content = Self::get_content(&path, preprocessor).await?;
 
-        // This implementation is not the best. Any plaintext file is changed
-        // to the default extension, but a file could genuinely be plaintext.
-        // To do this properly would need something like a TryFrom for FileType,
-        // but this conflicts with the From which is heavily used.
-        //
-        // It's not a big deal right now because the HTML and markdown extractors
-        // also perform plaintext extraction.
-        let file_type = match (FileType::from(&path), file_type_hint) {
-            (FileType::Plaintext, Some(default)) => default,
-            (file_type, _) => file_type,
-        };
+        let ext = path.extension().and_then(OsStr::to_str);
+        let file_type = ext
+            .and_then(FileType::from_extension)
+            .or(file_type_hint)
+            .unwrap_or_default();
 
         Ok(InputContent {
             file_type,
--- a/lychee-lib/src/types/input/input.rs
+++ b/lychee-lib/src/types/input/input.rs
@@ -165,7 +165,7 @@ impl Input {
                     Ok(source) => {
                         let content_result = match source {
                             ResolvedInputSource::FsPath(path) => {
-                                Self::path_content(&path, preprocessor.as_ref()).await
+                                Self::path_content(&path, preprocessor.as_ref(), self.file_type_hint).await
                             },
                             ResolvedInputSource::RemoteUrl(url) => {
                                 resolver.url_contents(*url).await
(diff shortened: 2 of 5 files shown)


benches/src/extract.rs
```diff
@@ -42,10 +42,10 @@ fn benchmark(c: &mut Criterion) {
 
     runtime.block_on(async {
         inputs = vec![
-            Input::path_content("../fixtures/bench/elvis.html", None)
+            Input::path_content("../fixtures/bench/elvis.html", None, None)
```
mre, line 45: Tangent: In this case, the additional `None` is fine, but we have a few places where we use a lot of `None`s and booleans now and it gets confusing. We'll need to think about refactoring that soon.

lychee-lib/src/types/input/input.rs
```diff
@@ -252,12 +252,18 @@ impl Input {
...
         preprocessor: Option<&Preprocessor>,
+        file_type_hint: Option<FileType>,
     ) -> LycheeResult<InputContent> {
         let path = path.into();
         let content = Self::get_content(&path, preprocessor).await?;
 
+        let file_type = match (FileType::from(&path), file_type_hint) {
+            (FileType::Plaintext, Some(default)) => default,
+            (file_type, _) => file_type,
+        };
```
mre: I really liked your PR description:

> This implementation is not the best because any plaintext file is changed to the default extension, but a file could genuinely be a plaintext file. To do this properly would need something like a TryFrom for FileType, but this conflicts with the From which is heavily used. It's probably not a big deal right now because the HTML and markdown extractors also perform plaintext extraction.

Do you think we could add a short version of that as a comment here?

mre (commented):
lgtm. one minor comment

lychee-lib/src/types/input/input.rs
katrinafyi: I just realised we can use `FileType::from_extension` to avoid the Plaintext as default behaviour, so we can handle it all directly! I think it's a lot better now and avoids this issue. https://github.com/lycheeverse/lychee/pull/2245/changes/61f7825fc962ac7d2e81d3bd5ff3d9d619874572

katrinafyi:
In the latest changes, I've also changed the text a bit. It was surprisingly hard to write an unambiguous leading sentence for the option description. This is what I came up with, let me know if it's okay. also @nobkd wdyt :)

      --default-extension <EXTENSION>
          Parse input sources without a recognised file type by using the given
          file extension.

          This affects how links are extracted from certain input sources. This
          is useful for stdin inputs, files without extensions, and files with
          unknown extensions.

          Examples:
            --default-extension md
            --default-extension html

          To automatically append file extensions when locating files, see
          `--fallback-extensions`.

nobkd:
That sounds quite nice! Thanks for taking this on :)

lychee-lib/src/types/input/input.rs
mre: Oh wow, yeah, you're right!

mre (approved):
Looks great! Thanks. 😊

PR #2250, opened 2026-06-30
Recognize block-style heading attributes

### LINKED ISSUE #2249: Recognize `{#id}` heading anchors in `--include-fragments`
## Description

With `--include-fragments`, lychee recognizes a heading's explicit id when it is written as a **trailing inline** attribute (`### Heading {#id}`), but not when it is written as a **preceding block** attribute:

```markdown
{#id}
### Heading
```

I stumble upon this subtelty with MyST's `attrs_block` extension, which allow and recognize this form.

I check this behavior with lychee 0.24.2.

## Reproduction

Create a new `test.md` file that contain:

```markdown
# Attribute-list anchor test

- [trailing inline form](#inline-id)
- [preceding block form](#block-id)

### Heading one {#inline-id}

Content.

{#block-id}
### Heading two

Content.
```

Run:

```console
$ lychee --include-fragments --offline --no-progress test.md
[ERROR] file://.../test.md#block-id (at 4:3) | Cannot find fragment

🔍 2 Total  🔗 2 Unique  ✅ 1 OK  🚫 1 Error
```

`#inline-id` resolves; `#block-id` does not.

A workaround is to use a raw `<a id="..."></a>` immediately before the heading, but that's quite ugly when Lychee already understand the `{#inline-id}` notation.

sjh9714:
## Summary

Fixes #2249.

`--include-fragments` already recognizes heading IDs written inline as `### Heading {#id}`. This adds the same fragment recognition for MyST-style preceding block attributes such as:

```markdown
{#block-id}
### Heading
```

The change is scoped to Markdown fragment extraction. A block attribute is only applied to the immediately following heading, and the normal generated heading fragment is still emitted.

## Changes

- Track a pending `{#id}` paragraph before a heading during Markdown fragment extraction.
- Add regression coverage for the block attribute form while keeping generated heading IDs intact.

## Testing

- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p lychee-lib test_extract_fragments_from_preceding_heading_attributes -- --nocapture` (failed before the fix, passed after)
- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p lychee-lib extract::markdown -- --nocapture`
- `PATH="$HOME/.cargo/bin:$PATH" cargo fmt --all -- --check`
- `PATH="$HOME/.cargo/bin:$PATH" cargo clippy -p lychee-lib --all-targets --all-features -- -D warnings`
- `PATH="$HOME/.cargo/bin:$PATH" cargo test -p lychee-lib`
- `git diff --check`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/lychee-lib/src/extract/markdown.rs
+++ b/lychee-lib/src/extract/markdown.rs
@@ -486,6 +486,35 @@ or inline like `https://bar.org` for instance.
     }
 
     #[test]
+    fn test_extract_fragments_from_preceding_heading_attributes_ignores_non_heading_blocks() {
+        let markdown = r"
+{#not-a-heading-id}
+
+This paragraph consumes the pending block attribute.
+
+### Heading two
+";
+
+        let actual = extract_markdown_fragments(markdown);
+
+        assert!(!actual.contains("not-a-heading-id"));
+        assert!(actual.contains("heading-two"));
+    }
+
+    #[test]
+    fn test_extract_fragments_from_preceding_heading_attributes_ignores_missing_id() {
+        let markdown = r"
+{.class key=value #}
+### Heading two
+";
+
+        let actual = extract_markdown_fragments(markdown);
+
+        assert!(!actual.contains(""));
+        assert!(actual.contains("heading-two"));
+    }
+
+    #[test]
     fn test_skip_verbatim() {
         let expected = vec![
             RawUri {
--- a/lychee-lib/src/extract/markdown.rs
+++ b/lychee-lib/src/extract/markdown.rs
@@ -279,8 +279,18 @@ fn raw_uri(dest_url: &CowStr<'_>, span: RawUriSpan) -> Vec<RawUri> {
 /// [heading attribute]: https://github.com/pulldown-cmark/pulldown-cmark/blob/94e7011a22a235e3abc5c5d6f3615192a6b1e42b/pulldown-cmark/specs/heading_attrs.txt
 pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
     let mut in_heading = false;
+    let mut in_paragraph = false;
     let mut heading_text = String::new();
+    // Explicit ID parsed from pulldown-cmark's inline heading attribute support,
+    // e.g. `## Heading {#custom-id}`.
     let mut heading_id: Option<CowStr<'_>> = None;
+    // Explicit ID parsed from a standalone attribute paragraph immediately
+    // before a heading, e.g. `{#custom-id}\n## Heading`.
+    let mut heading_block_id: Option<HeadingId> = None;
+    let mut paragraph_text: Option<String> = None;
+    // Block attributes are emitted as the paragraph before the heading they
+    // belong to, so keep one pending ID until the next block starts.
+    let mut pending_heading_block_id: Option<HeadingId> = None;
     let mut id_generator = GithubHeadingIdGenerator::new();
 
     let mut out = HashSet::new();
@@ -288,12 +298,25 @@ pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
     for event in Parser::new_ext(input, md_extensions()) {
         match event {
             Event::Start(Tag::Heading { id, .. }) => {
+                heading_block_id = pending_heading_block_id.take();
                 heading_id = id;
                 in_heading = true;
+            }
+            Event::Start(Tag::Paragraph) => {
+                pending_heading_block_id = None;
+                paragraph_text = Some(String::new());
+                in_paragraph = true;
+            }
+            Event::Start(_) if !in_heading && !in_paragraph => {
+                pending_heading_block_id = None;
             }
             Event::End(TagEnd::Heading(_)) => {
                 if let Some(frag) = heading_id.take() {
                     out.insert(frag.to_string());
+                }
+
+                if let Some(frag) = heading_block_id.take() {
+                    out.insert(frag.into_string());
                 }
 
                 if !heading_text.is_empty() {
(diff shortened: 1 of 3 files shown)


lychee-lib/src/extract/markdown.rs
```diff
@@ -392,6 +428,21 @@ or inline like `https://bar.org` for instance.
         assert_eq!(actual, expected);
     }
 
+    #[test]
+    fn test_extract_fragments_from_preceding_heading_attributes() {
```
mre, line 474: We should add a few more negative tests and one cli test with a fixture.

lychee-lib/src/extract/markdown.rs
```diff
@@ -316,6 +342,16 @@ pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
     out
 }
 
+fn extract_block_heading_id(text: &str) -> Option<String> {
```
mre: I think we should introduce a type here. Like `HeadingId` and then have a `TryFrom<&str> for HeadingId`. Or maybe a method on `HeadingId`, which returns an `Option<Self>`

lychee-lib/src/extract/markdown.rs
```diff
@@ -275,23 +275,40 @@ fn raw_uri(dest_url: &CowStr<'_>, span: RawUriSpan) -> Vec<RawUri> {
 /// It means a single heading such as `## Frag 1 {#frag-2}` would generate two fragments.
 pub(crate) fn extract_markdown_fragments(input: &str) -> HashSet<String> {
     let mut in_heading = false;
+    let mut in_paragraph = false;
```
mre, line 282: This list of fields starts to get unwieldy. Can we add a comment for each field to describe what it does?

E.g. describing what is the difference between `heading_id` and `heading_block_id`. And I'm not even sure if we need both.

mre (commented):
Thx. Took a quick look.

lychee-lib/src/extract/markdown.rs
```diff
@@ -275,23 +275,40 @@ fn raw_uri(dest_url: &CowStr<'_>, span: RawUriSpan) -> Vec<RawUri> {
...
     let mut heading_text = String::new();
     let mut heading_id: Option<CowStr<'_>> = None;
+    let mut heading_block_id: Option<String> = None;
```
katrinafyi: a

katrinafyi (commented):
Hi :) I have a drive-by comment.

It would be nice to avoid the need to copy all of the paragraph text into a String, since it could be quite large. AFAICT, copying would only be needed if the parser returned the heading attribute across different Text events, which I think it would only do if it was split up by another tag (like bold, italic, etc). That shouldn't be the case here.

Separately, is there a spec for this markdown extension? I would also be curious about the handling of these edge cases and whether they act as expected according to the spec (if there is one). That said, some of these are quite mean and it's probably fine to overapproximate the fragment identifiers so maybe not a big deal.
````md
{ #spaces-around }
# a

a {#text-before}
# a

{#text-after} a
# a

a
{#line-before}
# a

{#line-between}

# a

```
{#inside-code-block}
```
# a

`{#inside-code-inline}`
# a

*{#inside-bold}*
# a

> {#inside-quote}
> # a

{ #multiple-attributes } { #multiple-attributes2 }
# a

{} #between-empty-sets {}
# a

{{ #double-braced }}
# a

{{#double-braced-no-space}}
# a

{
#across-lines
}
# a
````

sjh9714:
Thanks for the review. I pushed a follow-up that:

* Adds the `HeadingId` wrapper.
* Documents the `heading-id` vs. `block-heading-id` state.
* Avoids collecting non-attribute paragraphs.
* Adds negative unit coverage and a CLI fixture case.

Checks run locally:

* `cargo test -p lychee-lib extract::markdown`
* `cargo test -p lychee --features check_example_domains test_fragments --test cli`
* `cargo fmt --check`

mre:
@sjh9714, going forward, please make sure all comments are cleaned up before posting. We prefer human commentary here, even if imperfect. It's more productive to hear your own thoughts so we can help you best.

lychee-lib/src/extract/markdown.rs
sjh9714: Addressed in the current diff. I added negative unit coverage for non-heading block attributes and missing IDs, plus a CLI fixture case in `fixtures/fragments/file1.md` / `lychee-bin/tests/cli.rs`.

lychee-lib/src/extract/markdown.rs
sjh9714: Addressed in the current diff. The parser state now has comments for the inline `heading_id`, standalone block `heading_block_id`, collected `paragraph_text`, and pending block ID state so the distinction is visible near the fields.

sjh9714:
Got it, thanks for pointing this out. I’ll take more care with my comments going forward and make sure they’re in my own words before posting.

mre:
@katrinafyi, the paragraph String buffer is now gone. :) Instead, we inspect borrowed source slices only, so we only allocate the extracted ID.

I also checked your examples against MyST’s [attrs_block implementation](https://github.com/executablebooks/mdit-py-plugins/blob/519953da0a394a92534bc27580f4e7415462df44/mdit_py_plugins/attrs/index.py#L184) in mdit-py-plugins and added unit and integration coverage. We now match its handling of multiple groups on one line: `{#first} {#second}` uses `first` and ignores the trailing group. We still reject nested braces inside the first group.

I also added tests for quoted attribute values, so things like `title="100%"` work without treating `#text` inside a value as an ID or a quoted `}` as the end of the group. Thanks for the thorough examples!

Overall, I'm still not 100% happy with that code. We made attribute parsing much harder to understand, but that's mainly because attribute parsing is hard. In the future, I'd like us to refactor that code. It's okay for now, though, I'd say.

mre:
The failing test is unrelated. See https://github.com/lycheeverse/lychee/pull/2294

mre:
@sjh9714, thanks for the initial work on this.

PR #2251, opened 2026-07-01
Feat issue 1998 domain summary

### LINKED ISSUE #1998: Domain-grouped summary of checked links
## Description

It would be helpful to have an option to display a summary of all checked links grouped by domain at the end of a lychee run. This would provide a quick overview of which external services/sites are being referenced and how many links point to each.

## Use Case

When checking documentation or large websites with many external links, understanding the distribution of links across domains helps with:

- **Identifying domain-specific issues**: If many links to a single domain fail, it may indicate the domain is down or blocking requests, rather than individual broken links
- **Auditing external dependencies**: Quickly see which external services your documentation relies on most heavily
- **Debugging exclusion patterns**: Verify that exclusion rules are working as expected by seeing which domains remain in the check
- **Maintenance planning**: Prioritize fixing links to domains with the most references

## Suggested Output Format

```
Domain Summary (25 domains, 342 links checked):

  docs.example.com         89 links  [✓ 89]
  github.com               67 links  [✓ 65, ✗ 2]
  api.example.org          45 links  [✓ 44, ? 1]
  stackoverflow.com        38 links  [✓ 38]
  ...
  (showing top 10 of 25 domains, use --domain-summary=all for full list)
```

## Possible Options

- `--domain-summary` or `-D`: Enable domain summary (could accept values like `none`, `top:N`, `all`)
- `--domain-summary-min`: Only show domains with at least N links
- Integration with existing `--format` options (JSON, markdown, etc.)

## Considerations

For runs with many domains, displaying all of them could be overwhelming. Possible mitigations:
- Default to showing top N domains by link count
- Provide a threshold option to filter low-count domains
- In JSON output, always include full data regardless of display limits

beihaogu:
Adds `(N domains, M links checked)` to the `📊 Per-host Statistics` header in all three `--host-stats` formatters (compact, detailed, markdown), where:

- `N` = number of unique hosts in `HostStatsMap`
- `M` = sum of `total_requests` across all hosts

**Before:**

```
📊 Per-host Statistics
```

**After:**

```
📊 Per-host Statistics (23 domains, 113 links checked)
```

### Notes

- Reused `--host-stats` rather than adding a new flag, matching the direction set in the issue discussion.
- `total_requests` is summed for "links checked" — the run-level `Total` counter includes duplicates across input files, which would over-represent per-domain coverage. Summing `total_requests` keeps the header consistent with the rows beneath it.
- No singular/plural handling (`1 domain` vs `N domains`); kept identical to the format `@mre` posted. Happy to add if preferred.
- **Drive-by:** `host_stats/compact.rs` previously called `HostStatsMap::sorted()` three times per render. Reduced to one call to match `detailed.rs` / `markdown.rs`.
- Snapshot tests in `formatters/stats/{compact,detailed}.rs` updated to include the new header substring.

`cargo test`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt --check` all pass.

Closes #1998

---


mre:
Hey @beihaogu, thanks for the PR.

I looked through your code and had a few remarks.
Instead of adding comments, I realized that it was easier to create a new PR, which fixes the same issue [here](https://github.com/lycheeverse/lychee/pull/2252).

My hope is that this helps you learn more about the codebase and how I'd structure the solution. My goal is to keep the changes self-contained inside the host stats map, which avoids the redundancy in the individual formatters. 

If you don't mind, can you review my PR and let me know if you have any questions?

mre:
I went ahead and merged the other one, https://github.com/lycheeverse/lychee/pull/2252. But thanks, @beihaogu, for looking into it and doing the first version of the fix. 👍

PR #2252, opened 2026-07-03
Add domains and links to the Per-host stats header

mre:
This adds the status line discussed in https://github.com/lycheeverse/lychee/issues/1998#issuecomment-3798779520 to the host stats output.

(Admittedly, the entire code around stats formatting could be much nicer. It is very imperative at the moment. But I decided to keep the changes small and focused. A bigger refactor could be done in the future.)

Related to: #1998

PR #2253, opened 2026-07-05
Add short per-domain statistics

mre:
This is the second part of #1998, where we show per-domain short statistics like so if `--host-stats` is enabled:

```
📊 Per-host Statistics (23 domains, 116 links checked)
  github.com                     67 links  [✓ 67]
  lychee.cli.rs                  17 links  [✓ 17]
  img.shields.io                  4 links  [✓ 4]
  docs.rs                         4 links  [✓ 4]
  ngi.eu                          4 links  [? 4]
  developer.mozilla.org           2 links  [✓ 2]
  nlnet.nl                        2 links  [✓ 2]
  chocolatey.org                  1 links  [✓ 1]
  opensource.org                  1 links  [✓ 1]
  pre-commit.com                  1 links  [✓ 1]
  doc.rust-lang.org               1 links  [✓ 1]
  raw.githubusercontent.com       1 links  [✓ 1]
  gitlab.torproject.org           1 links  [✓ 1]
  nixos.org                       1 links  [✓ 1]
  hub.docker.com                  1 links  [✓ 1]
  www.apache.org                  1 links  [✓ 1]
  docs.github.com                 1 links  [✓ 1]
  rust-lang.org                   1 links  [✓ 1]
  hello-rust.github.io            1 links  [✓ 1]
  crates.io                       1 links  [✓ 1]
  scoop.sh                        1 links  [✓ 1]
  brew.sh                         1 links  [✓ 1]
  www.macports.org                1 links  [✓ 1]
```

It's a follow-up of #2252. Once both PRs are merged, #1998 is considered done.

### DIFF
--- a/lychee-bin/src/formatters/host_stats/compact.rs
+++ b/lychee-bin/src/formatters/host_stats/compact.rs
@@ -1,6 +1,6 @@
 use std::fmt::{self, Display};
 
-use super::{status_summary, write_host_heading};
+use super::{host_heading, status_summary};
 use crate::formatters::color::{NORMAL, color};
 use lychee_lib::ratelimit::HostStatsMap;
 
@@ -14,7 +14,7 @@ impl Display for CompactHostStats {
             return Ok(());
         };
 
-        write_host_heading(f, "\n📊 ", host_stats)?;
+        writeln!(f, "{}", host_heading("\n📊 ", host_stats))?;
 
         let sorted_hosts = host_stats.sorted();
         let hostname_width = sorted_hosts
@@ -25,12 +25,14 @@ impl Display for CompactHostStats {
             .max(10);
 
         for (hostname, stats) in sorted_hosts {
+            let status_summary = status_summary(&stats);
+            let cache_summary = stats.cache_summary();
+
             color!(
                 f,
                 NORMAL,
-                "  {hostname:<width$}  {:>6} reqs  {}",
+                "  {hostname:<width$}  {:>6} reqs  {cache_summary:>12}    {status_summary}",
                 stats.total_requests,
-                status_summary(&stats),
                 width = hostname_width,
             )?;
             writeln!(f)?;
--- a/lychee-bin/src/formatters/host_stats/compact.rs
+++ b/lychee-bin/src/formatters/host_stats/compact.rs
@@ -1,7 +1,7 @@
 use std::fmt::{self, Display};
 
-use super::write_host_heading;
-use crate::formatters::color::{DIM, NORMAL, color};
+use super::{status_summary, write_host_heading};
+use crate::formatters::color::{NORMAL, color};
 use lychee_lib::ratelimit::HostStatsMap;
 
 pub(crate) struct CompactHostStats {
@@ -16,37 +16,22 @@ impl Display for CompactHostStats {
 
         write_host_heading(f, "\n📊 ", host_stats)?;
 
-        let separator = "─".repeat(60);
-        color!(f, DIM, "{}", separator)?;
-        writeln!(f)?;
-
         let sorted_hosts = host_stats.sorted();
-
-        // Calculate optimal hostname width based on longest hostname
-        let max_hostname_len = sorted_hosts
+        let hostname_width = sorted_hosts
             .iter()
             .map(|(hostname, _)| hostname.len())
             .max()
-            .unwrap_or(0);
-        let hostname_width = (max_hostname_len + 2).max(10); // At least 10 chars with padding
+            .unwrap_or(0)
+            .max(10);
 
         for (hostname, stats) in sorted_hosts {
-            let median_time = stats
-                .median_request_time()
-                .map_or_else(|| "N/A".to_string(), |d| format!("{:.0}ms", d.as_millis()));
-
-            let cache_hit_rate = stats.cache_hit_rate() * 100.0;
-
             color!(
                 f,
                 NORMAL,
-                "{:<width$} │ {:>6} reqs │ {:>6.1}% success │ {:>8} median │ {:>6.1}% cached",
-                hostname,
+                "  {hostname:<width$}  {:>6} reqs  {}",
                 stats.total_requests,
-                stats.success_rate() * 100.0,
-                median_time,
-                cache_hit_rate,
-                width = hostname_width
+                status_summary(&stats),
+                width = hostname_width,
             )?;
             writeln!(f)?;
         }
--- a/lychee-bin/src/formatters/host_stats/detailed.rs
+++ b/lychee-bin/src/formatters/host_stats/detailed.rs
@@ -1,6 +1,6 @@
 use std::fmt::{self, Display};
 
-use super::write_host_heading;
+use super::host_heading;
 use lychee_lib::ratelimit::HostStatsMap;
 
 pub(crate) struct DetailedHostStats {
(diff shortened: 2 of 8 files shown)


lychee-bin/src/formatters/host_stats/compact.rs
```diff
@@ -14,39 +14,24 @@ impl Display for CompactHostStats {
             return Ok(());
         };
 
-        write_header(f, "📊 ", host_stats)?;
-
-        let separator = "─".repeat(60);
-        color!(f, DIM, "{}", separator)?;
-        writeln!(f)?;
+        write_header(f, &section_header("📊 ", host_stats))?;
```
mre: I'm still not 100% convinced that this "function-style" is the best way going forward, but it was too tempting not to clean up that part while I was at it.

In the long run, I'd like to look into using a trait for the formatter, cleaning up some of the boilerplate in a better way.

lychee-bin/src/formatters/host_stats/mod.rs
```diff
@@ -10,20 +10,71 @@ pub(crate) use compact::CompactHostStats;
...
+/// `[✓ 65, ✗ 2, ? 1]`
+/// where
+/// - `✓` indicates successful requests,
+/// - `✗` indicates failed requests, and
+/// - `?` indicates unknown requests.
+fn status_summary(stats: &HostStats) -> String {
+    let failed_requests = stats.rate_limited + stats.client_errors + stats.server_errors;
+    let unknown_requests = stats
+        .total_requests
+        .saturating_sub(stats.successful_requests + failed_requests);
```
katrinafyi, line 33: *micro*-nit / commentary: is saturating ineffective since unchecked `+` is used all around it?

i guess it still avoids the subtraction overflowing to super big. but all the unchecked `+` could overflow to an unexpectedly small number. but it's all u64 so it's basically impossible anyway.

lychee-bin/src/formatters/host_stats/compact.rs
```diff
@@ -14,39 +14,24 @@ impl Display for CompactHostStats {
...
-
-        // Calculate optimal hostname width based on longest hostname
-        let max_hostname_len = sorted_hosts
+        let hostname_width = sorted_hosts
             .iter()
             .map(|(hostname, _)| hostname.len())
             .max()
-            .unwrap_or(0);
-        let hostname_width = (max_hostname_len + 2).max(10); // At least 10 chars with padding
+            .unwrap_or(0)
```
katrinafyi, line 24:
```suggestion
            .unwrap_or_default()
```
.. or maybe not

katrinafyi (approved):
Looks reasonable and should be an improvement in readability/usability.

Of course, the % cached would be hidden. Is it something worth keeping? We have had past feature requests around adding a "unique count" to the totals line, so it seems like something users are generally interested in. But I'm not sure how best to represent it in the symbols.

I also think that cleaning up the formatters would be nice, but it can happen in the long run. Not sure of the best approach right now.

lychee-bin/src/formatters/host_stats/mod.rs
mre: yeah, true. We could change all of the calculations to be saturating. So I could change it to saturating_sub and saturating_add everywhere. 
But as you said, overflowing a u64 is unlikely and it can look noisy.
The reason why I used saturating_sub for this specific case is that it's easier to underflow.

I think the proper solution would be to not do those calculations here at all and instead move them into a proper type. I could refactor it? Might be a bit overkill, idk.

lychee-bin/src/formatters/host_stats/compact.rs
mre: Hm yeah, not sure. ...or_default() is risky sometimes if the type gets changed and now the default is an unexpected value. But in this case it's certainly fine. I just prefer to be explicit if possible. (E.g. if we change the type to be a richer type and not just a number, this will fail loudly, which is good.)

lychee-bin/src/formatters/host_stats/mod.rs
katrinafyi: ya happy to leave it as is. don't need to go overboard

mre:
Update!

I've added back the cache statistics:

image

@katrinafyi, do you have time for one more round of review before I merge this? I've moved some stuff around and integrated your feedback.

lychee-bin/src/formatters/host_stats/detailed.rs
```diff
@@ -13,8 +13,9 @@ impl Display for DetailedHostStats {
             return Ok(());
         };
 
-        write_host_heading(f, "\n📊 ", host_stats)?;
-        writeln!(f, "---------------------")?;
+        let heading = host_heading("📊 ", host_stats);
+        writeln!(f, "\n{heading}")?;
+        writeln!(f, "{}", "-".repeat(heading.chars().count()))?;
```
katrinafyi, line 18: no change needed but chars(), while better than byte length, could still be inaccurate because rust chars are codepoints rather than graphemes. so, for example, ZWJs and emoji variation selectors are separate chars despite taking no width.

but it's definitely close enough and better than manual width. a very minor thing :) no changes needed

lychee-bin/src/formatters/host_stats/compact.rs
```diff
@@ -14,39 +14,26 @@ impl Display for CompactHostStats {
...
-            let cache_hit_rate = stats.cache_hit_rate() * 100.0;
+            let status_summary = status_summary(&stats);
+            let cache_summary = stats.cache_summary();
 
             color!(
                 f,
                 NORMAL,
-                "{:<width$} │ {:>6} reqs │ {:>6.1}% success │ {:>8} median │ {:>6.1}% cached",
-                hostname,
+                "  {hostname:<width$}  {:>6} reqs  {cache_summary:>12}    {status_summary}",
```
katrinafyi, line 34: nit: can we put cache at the end of the line? users are probably more interested in the ticks and crosses and it's good to have those right next to the request count. if it's at the end of the line, I think the cache % need not be aligned.

katrinafyi (approved):
Very happy with this now, thanks for the changes! Tiny nitpick comments but definitely ready for merge.

lychee-bin/src/formatters/host_stats/detailed.rs
mre: Fair. Idk now how to improve that with code from the standard library alone. Is there a way? Crates for graphemes exist, but not much in the stdlib.

lychee-bin/src/formatters/host_stats/compact.rs
mre: Let me look into that. I had it at the end in a prior version, but it looked a bit out-of-place, to be honest. :/

lychee-bin/src/formatters/host_stats/detailed.rs
katrinafyi: I don't think so in stdlib. And yeah, no need to pull in a crate just for this.

lychee-bin/src/formatters/host_stats/compact.rs
katrinafyi: Ah, that's very possible. In that case, I'm happy for you to make a choice on what looks best. I don't have strong feelings about it

mre (commented):
I'll go ahead and merge this. The formatting can be changed later, in case we find that we'd like to move the cached requests back to the end. Thanks for the feedback, @katrinafyi. Super helpful as always. 🌟

PR #2254, opened 2026-07-05
Move all GitHub checks into dedicated module

mre:
This PR moves GitHub-specific link handling out of the generic website checker and request quirks.

What changed:

- GitHub URL parsing now lives in `checker::github::url`.
- GitHub API calls now go through `checker::github::api` and the existing `HostPool`.
- GitHub request rewrites now live in `checker::github::request` instead of generic `Quirks`.
- `WebsiteChecker` delegates GitHub behavior before and after the normal website path.
- The old `types::uri::github::GithubUri` parser and `Octocrab` dependency are gone.

This is still a draft. I mainly want to check whether this module split feels right before polishing the edge cases further.


katrinafyi:
Just some quick thoughts.. Yeah, I think this is the right approach. Thanks for getting started!

I think we should centralise around the GithubUrl enum. The other files like api.rs are probably less important in a first pass.

There is a architecture decision of how to hook it into the website checker. I don't love the before+after functions. Maybe a GithubChecker should wrap a WebsiteChecker, so then the github checker naturally has control over whether to delegate to the website checker.

I also see that github/request.rs does some rewriting and logging. I'm increasingly thinking that we need to abstract the remaps/redirect history feature into a generic "url transformation history". It would be something like a `Vec<(Url, TransformKind)>` and could record remaps, redirects, github, maybe quirks, and it would increase transparency around these various URL transformations.

But these are just very rough thoughts and it doesn't have to happen all at once (or at all). I'll let you keep working on it and exploring the solution space :)

lychee-lib/src/checker/github/api.rs
```diff
@@ -0,0 +1,267 @@
...
+                private: None,
+            },
+        }
+    }
+
+    pub(crate) async fn check_repo_readme(
+        &self,
+        owner: &str,
+        repo: &str,
+        ref_: Option<&str>,
```
thomas-zahner, line 81: Is the trailing underscore intentional? Seems to repeat itself quite a lot in this PR.

lychee-lib/src/checker/github/api.rs
```diff
@@ -0,0 +1,267 @@
...
+use url::Url;
+
+use crate::{Result, Status, Uri, ratelimit::HostPool};
+
+const GITHUB_API_BASE_URL: &str = "https://api.github.com";
+const GITHUB_API_ACCEPT: &str = "application/vnd.github+json";
+
+#[derive(Debug)]
+pub(crate) struct RepoStatus {
+    pub(crate) status: Status,
```
thomas-zahner, line 15: Some doc comments would help here

lychee-lib/src/checker/github/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+                .check_repo_readme(&owner, &repo, ref_.as_deref())
+                .await,
+        )
+    }
+
+    /// Fall back to the GitHub API after a normal website request failed.
+    ///
+    /// This keeps private repository checks working without leaving GitHub API
+    /// details in the generic website checker.
+    pub(crate) async fn check_after_website_failure(&self, status: Status, uri: &Uri) -> Status {
```
thomas-zahner, line 66: As @katrinafyi mentions the before/after distinction feels a bit funky. But I'm not sure what the best approach is. Making the GitHub checker wrap a Websitechecker might indeed make sense.

thomas-zahner (commented):
I really like this draft already. I also really like how this enables us to simply drop the dependency on octocrab

PR #2258, opened 2026-07-09
feat(config): support per-host accept status codes

### LINKED ISSUE #2223: Per-host `accept` Option
It would be nice to be allowed to override the `accept` config in the per host settings. I am running into a situation where an external site I don't control (but link to a bunch) is blocking crawlers. I am getting 403 from it, which is fine and probably as long as it's not a 404 I don't care.

I fixed it by adding 403, to my `accept`, but I don't want to accidentally allow 403s on my own website.

costajohnt:
Closes #2223.

Adds a per-host `accept` override so a host can define its own set of accepted HTTP status codes, alongside the existing per-host `concurrency` / `request_interval` / `headers`:

```toml
[hosts."example.com"]
accept = [200, 429]
```

## Semantics

Per-host `accept` **replaces** the global `accept` for that host (it does not union with it), matching how the other per-host options override their globals. This follows the direction discussed in the issue (a host's accepted codes are exactly what's listed). Hosts without an override keep using the global `accept`, so existing configs are unaffected.

## Implementation

- `HostConfig` gains `accept: Option<StatusCodeSelector>` plus an `effective_accept(global)` resolver (mirrors `effective_concurrency` / `effective_request_interval`), with `merge` and `Default` handling.
- `HostPool::effective_accept` resolves the accepted set for a given URL/URI, falling back to the global set when the host has no override.
- The website checker and the cache-hit path both resolve per-host before turning a status code into a `Status`. The cache path resolves from the post-remap cache-key host, so cross-host cache hits are judged by the correct host's codes.
- Added a `Serialize` impl for `StatusCodeSelector` (round-trips through its existing `FromStr`), needed now that `HostConfig` carries one.

## Tests

- `effective_accept` replace semantics + fallback to global, and `merge` precedence.
- `StatusCodeSelector` serialize/deserialize round-trip (ranges and single codes).
- Live-check path (wiremock): a per-host `accept: 429` makes a 429 response succeed, and a host without the override still fails on 429.
- Cache: a cached status is judged by the cache-key host's override, verified with a genuine request-host to cache-key-host remap.

`cargo test`, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt` are all clean.

## Note

One pre-existing behavior to flag: `Status::from_cache_status` returns a cached `Ok` as success without re-checking it against the accept set, so a *more restrictive* per-host override won't re-downgrade codes already cached as OK from a previous run (same as tightening the global `accept` between runs today). The common case here (accepting *more* codes, e.g. 429) goes through the cached-`Error` branch, which does re-check. Happy to address the cached-`Ok` reinterpretation separately if you'd like.

### DIFF
--- a/lychee-lib/src/ratelimit/pool.rs
+++ b/lychee-lib/src/ratelimit/pool.rs
@@ -1,7 +1,7 @@
 use dashmap::DashMap;
-use http::Method;
+use http::{Method, StatusCode};
 use reqwest::{Client, Request};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 use crate::ratelimit::{
@@ -92,6 +92,26 @@ impl HostPool {
             .request(method, uri.url.clone())
             .build()
             .map_err(ErrorKind::BuildRequestClient)
+    }
+
+    /// Resolve the effective set of accepted status codes for the host of
+    /// `key` (a URL or URI).
+    ///
+    /// If the host has its own `accept` override it fully **replaces** the
+    /// `global` set; otherwise `global` is returned unchanged. Falls back to
+    /// `global` when `key` has no valid host.
+    #[must_use]
+    pub fn effective_accept<K>(&self, key: K, global: &HashSet<StatusCode>) -> HashSet<StatusCode>
+    where
+        HostKey: TryFrom<K>,
+    {
+        match HostKey::try_from(key) {
+            Ok(host_key) => self
+                .host_configs
+                .get(&host_key)
+                .map_or_else(|| global.clone(), |config| config.effective_accept(global)),
+            Err(_) => global.clone(),
+        }
     }
 
     /// Get an existing host or create a new one for the given hostname
--- a/lychee-bin/src/cache.rs
+++ b/lychee-bin/src/cache.rs
@@ -271,6 +271,68 @@ mod tests {
         }
     }
 
+    /// The per-host `accept` override on a cache hit must be resolved from the
+    /// *cache-key* host (post-remap), not the original request host. Here the
+    /// request host (`request-host.com`) has no override and remaps to a
+    /// different cache-key host (`accepted-host.com`) which accepts 429. The
+    /// cached 429 is judged by the cache-key host's override, so it succeeds.
+    /// If resolution keyed off the request URL instead, this would fail.
+    #[tokio::test]
+    async fn test_cache_hit_resolves_accept_from_cache_key_host() {
+        use lychee_lib::remap::Remaps;
+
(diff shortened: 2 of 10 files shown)


costajohnt:
Checking in on this one, it's been about a month. No rush, just want to make sure it didn't fall through the cracks. CI is green and I'm happy to rebase onto current master or split the change up if that makes review easier.

mre:
Sorry, yeah, I did want to take a look, but haven't had the bandwidth yet. Overall it looks good and I don't think we'll need a lot of changes and it's great that you added all the tests.

If someone has the time for a review, I'd appreciate it. 🙃

lychee-lib/src/ratelimit/pool.rs
```diff
@@ -94,6 +94,26 @@ impl HostPool {
...
+        global: &HashSet<StatusCode>,
+    ) -> HashSet<StatusCode>
+    where
+        HostKey: TryFrom<K>,
+    {
+        HostKey::try_from(key)
+            .ok()
+            .and_then(|host_key| self.host_configs.get(&host_key))
+            .and_then(|config| config.accept.as_ref())
+            .map_or_else(|| global.clone(), |selector| selector.clone().into())
```
mre, line 114: I removed the double-fallback to `global` here to make this part a little cleaner.

lychee-lib/src/types/status.rs
```diff
@@ -99,28 +99,17 @@ impl Status {
...
-            CacheStatus::Error(code) => {
-                if let Some(code) = code
-                    && accepted.contains(&code)
-                {
-                    return Self::Cached(CacheStatus::Ok(code));
+            CacheStatus::Ok(code) | CacheStatus::Error(Some(code)) => {
+                if accepted.contains(&code) {
+                    Self::Cached(CacheStatus::Ok(code))
+                } else {
+                    Self::Cached(CacheStatus::Error(Some(code)))
```
mre, line 111: This now respects the current `accept` config for cached status codes as well. :) I think that's what most users expect.

E.g. if a status code 200 is cached, but no longer accepted in a consecutive run, it will be interpreted as an error.

mre:
I made a couple of small tweaks to make the code more ergonomic. The one significant change is how cached status codes are handled.

Previously, a cached success was always treated as successful, even if the current `accept` setting no longer allowed that status code. Now both cached successes and errors are checked against the current accepted codes, including per-host overrides. For example, a cached 200 is rejected if the host is configured to accept only 429, just as a fresh response would be. I added regression tests and updated the existing CLI test to reflect this.

I'll go ahead and merge this now. Thanks for your pull request, @costajohnt, and apologies for the long delay.

PR #2265, opened 2026-07-26
tests: Use live github API to test markdown heading fragments

katrinafyi:
Removes the committed special-casing.**html** and fetches it during the test run, using the Github API to convert a committed special-casing.md to HTML.

Uses a GITHUB_TOKEN environment variable (if it is set), but still works without a github token. We just ignore 429 errors if they occur.

Alternatively, we could skip the test if there is no Github token. Since Github actions sets GITHUB_TOKEN, this will still run the test in CI which should be enough.

The markdown file generation is now ported to Rust and we have a test case to check that the Github generated fragments match our algorithm. Doing the update of the special-casing.md file is still manual, but there is now another test case to check it lines up and this should make it easier. 

Follow-up to https://github.com/lycheeverse/lychee/pull/2153#discussion_r3566566048


mre:
Looks fine to me.

PR #2266, opened 2026-07-26
test: add missing `cfg(test)` and fix new 1.97 warnings

katrinafyi:
I'm guessing that this was accidentally missed. Or, is there a reason that this `mod test` should not be `cfg(test)`?

After this, https://github.com/lycheeverse/lychee/pull/2262 should be unblocked.


thomas-zahner (approved):
Thanks! I can't remember a good reason for omitting the test attribute, I must have simply forgotten to add it.

PR #2271, opened 2026-07-31
fix: retry HTTP/2 protocol failures over HTTP/1.1

### LINKED ISSUE #2264: HTTP/2 stream error on a live site (ampcode.com); no way to force HTTP/1.1
## Summary

`lychee` consistently fails on `https://ampcode.com/` with an HTTP/2 stream error, while the site is demonstrably alive. There is currently no flag to force HTTP/1.1, so the only workaround is excluding the domain.

## Reproduction

```
$ lychee --no-progress https://ampcode.com/
    0: error sending request for url (https://ampcode.com/)
    1: client error (SendRequest)
    2: http2 error
    3: stream error detected: unspecific protocol error detected
```

Reproduced in two environments:
- lychee 0.24.2 via Homebrew on macOS (Darwin 25.5.0)
- lycheeverse/lychee-action@v2 on ubuntu-latest GitHub runners (multiple runs over several hours, so not transient)

## The site is fine

`curl` succeeds from the same machines over both protocols:

```
$ curl -s -o /dev/null -w '%{http_code}\n' https://ampcode.com/            # negotiates h2
200
$ curl -s -o /dev/null -w '%{http_code}\n' --http1.1 https://ampcode.com/
200
```

ALPN negotiation itself succeeds for curl (server accepts h2, cert from Google Trust Services). The failure is at the h2 stream level after connection setup.

## Ruled out

- Not transient: consistent across ~6 runs, two networks, local + CI.
- Not user-agent sniffing at the HTTP layer: `-u "Mozilla/5.0 ..."` (Chrome UA) fails identically.
- Not the request path: the bare domain root fails.

Since curl's h2 works and reqwest's h2 gets a PROTOCOL_ERROR-class stream error, this looks like either an h2-fingerprint-based bot mitigation on the server/CDN side killing non-browser h2 clients, or an interop issue between the h2 crate and this server. Either way lychee reports a live link as broken with no recourse.

## Suggested fixes

1. Expose reqwest's `http1_only()` as a CLI option (e.g. `--http1`), so affected sites can be checked over HTTP/1.1 instead of being excluded entirely.
2. Better: automatic downgrade retry - when an h2 request fails with a stream/protocol error (not an HTTP status), retry once over an HTTP/1.1-only client before reporting the link broken. Browsers do effectively this, which is why such sites look fine to humans.

Happy to test a build against this domain if that helps.

w3lld1:
## Summary

I added a single HTTP/1.1 fallback attempt when an HTTP/2 stream or protocol error prevents a request from completing.

The fallback client preserves the configured headers, cookies, timeouts, TLS settings, redirects, and per-host overrides. Other network failures continue to use the existing retry behavior without protocol downgrade.

## Validation

- `cargo fmt --all --check`
- `cargo test -j 2 -p lychee-lib ratelimit::host::host::tests --locked` (3 passed)
- `cargo clippy -j 2 -p lychee-lib --lib --tests --locked -- -D warnings`
- `RUST_LOG=lychee_lib=debug cargo run -j 2 -p lychee --locked -- --no-progress https://ampcode.com/` (HTTP/2 failure triggered the HTTP/1.1 fallback; 151 links OK, 0 errors)

The full workspace suite is deferred to CI because it is disproportionate for this constrained environment.

Fixes #2264


ashkulz:
@mre can you review this? It's hitting us on a different site: https://quickbooks.intuit.com :sweat_smile:

katrinafyi:
Quick look, apologies if I'm misunderstanding anything.

- I don't love having to carry two clients around, it's easy to get out of sync and this is a very pervasive change. Could you instead modify a failed http2 request using https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.version ?
- Is there a better way to detect the HTTP2 error than matching on error strings..?
- Thoughts on retry logic? Should the HTTP1 request restart the retry counter for other transient network errors? It would be nice if it could count as one retry, but I don't think that's possible in the framework.

mre:
@w3lld1, did you find the time to look into @katrinafyi's suggestions? They make sense to me. No worries if not, but if this turns out to be a blocker for people, we might alternatively continue the work on the branch to get this merged. 😉

w3lld1:
Thanks — addressed all three points in `98ed252`:

- the failed request is now retried by setting its HTTP version to HTTP/1.1, so there is only one client per host;
- HTTP/2 failures are identified by the typed `h2::Error` in the source chain rather than error-message matching;
- the downgrade uses the existing retry loop, consumes one retry, and subsequent transient failures continue from the same retry budget over HTTP/1.1.

Fresh validation on the pushed commit:
- focused status/error tests: 5 passed;
- website checker tests: 3 passed;
- `cargo fmt --all -- --check`;
- `git diff --check`.

lychee-lib/src/ratelimit/host/host.rs
```diff
@@ -216,6 +208,13 @@ impl Host {
         Ok(response)
     }
 
+    fn record_network_error(&self, request_time: Duration) {
```
katrinafyi: inline this function and keep the original comments.

lychee-lib/src/types/status.rs
```diff
@@ -8,6 +8,17 @@ use http::StatusCode;
 use serde::ser::SerializeStruct;
 use serde::{Serialize, Serializer};
 
+fn is_http2_error(error: &(dyn Error + 'static)) -> bool {
```
katrinafyi: inline this function

lychee-lib/src/client.rs
```diff
@@ -439,11 +439,10 @@ impl ClientBuilder {
             .map(|(host, config)| {
                 let mut headers = self.default_headers()?;
                 headers.extend(config.headers.clone());
-                let client = self
+                let builder = self
```
katrinafyi: undo this change??

lychee-lib/src/types/status.rs
```diff
@@ -96,6 +107,14 @@ impl Status {
         }
     }
 
+    /// Returns `true` when the status wraps an HTTP/2 transport error.
+    pub(crate) fn is_http2_error(&self) -> bool {
+        let Status::Error(ErrorKind::NetworkRequest(error)) = self else {
+            return false;
+        };
+        is_http2_error(error)
+    }
```
katrinafyi, line 106: this logic should go in retry.rs which already has `get_source_error_type` and other similar error introspection.

katrinafyi (changes requested):
I have some code structure comments.

But much more crucially, this does not fix the issue for either of the reported URLs.
```

[a.md]:
 [ERROR] https://ampcode.com/ (at 3:1) | HTTP/2 protocol error. Server may not support HTTP/2 properly
 [ERROR] https://quickbooks.intuit.com/ (at 1:1) | HTTP/2 protocol error. Server may not support HTTP/2 properly

🔍 2 Total (in 411ms) 🔗 2 Unique ✅ 0 OK 🚫 2 Errors
```

I'm not thrilled about this. How much are you using LLMs to make this PR and how much are you verifying it?

w3lld1:
You're right — the previous version did not verify the actual retry control flow closely enough. The typed `h2::Error` detection was correct, but `Status::should_retry()` returned false first, so `retry_request()` exited before it could switch the request to HTTP/1.1.

I fixed that in `3884be0` by classifying typed HTTP/2 transport errors as retryable in `retry.rs`. The next retry now reaches the existing version switch and consumes the same retry budget. I also moved the error-chain logic into `retry.rs`, restored the original inline network-error accounting/comments, and reverted the unrelated client formatting change.

Validation on the pushed commit:
- `cargo test -p lychee-lib retry::tests --jobs 2` (3 passed);
- `cargo fmt --all -- --check`;
- `cargo clippy -p lychee-lib --lib --tests --jobs 2 -- -D warnings`;
- `git diff --check`.

I have not treated the two external-site checks as passing locally; the focused validation above covers the control-flow bug and typed classification. A rerun against the reported sites is still needed to confirm their current server behavior.

PR #2273, opened 2026-08-05
Feature: state why requests were excluded

### LINKED ISSUE #2142: Excluded links don't say which rule triggered the exclusion
Excluded links don't say which rule triggered the exclusion. This makes it hard to debug with multiple `--exclude` patterns.

Here's an example:

```
lychee --no-progress -v \
  --exclude 'wikipedia\.org' \
  --exclude 'licensebuttons\.net' \
  --exclude 'example\.com' \
  fixtures/TEST.md
[EXCLUDED] https://en.wikipedia.org/wiki/Static_program_analysis (at 7:1) | This is due to your 'exclude' values
[EXCLUDED] https://licensebuttons.net/p/zero/1.0/88x31.png (at 16:2) | This is due to your 'exclude' values
[EXCLUDED] http://example.com/ (at 20:1) | This is due to your 'exclude' values
[EXCLUDED] https://example.com/ (at 21:1) | This is due to your 'exclude' values
[EXCLUDED] mailto:test@example.com (at 23:1) | This is due to your 'exclude' values
[EXCLUDED] mailto:test2@example.com (at 24:8) | This is due to your 'exclude' values
🔍 12 Total (in 370ms) 🔗 12 Unique ✅ 6 OK 🚫 0 Errors 👻 6 Excluded
```

What it could look like after the fix

```
[EXCLUDED] https://en.wikipedia.org/wiki/Static_program_analysis (at 7:1) | Excluded by pattern: `wikipedia\.org`
[EXCLUDED] https://licensebuttons.net/p/zero/1.0/88x31.png (at 16:2) | Excluded by pattern: `licensebuttons\.net`
[EXCLUDED] http://example.com/ (at 20:1) | Excluded by pattern: `example\.com`
[EXCLUDED] https://example.com/ (at 21:1) | Excluded by pattern: `example\.com`
[EXCLUDED] mailto:test@example.com (at 23:1) | Excluded: mail checking is disabled (use --include-mail to enable)
[EXCLUDED] mailto:test2@example.com (at 24:8) | Excluded: mail checking is disabled (use --include-mail to enable)
```

I would like to get some input on this. Is it worth adding this information? Maybe only in verbose mode? Maybe in the final report output?

The only tedious part is that changing `Status::Excluded` from `Excluded` to `Excluded(ExcludeReason)` forces us to update every pattern match on that variant in the codebase. But there aren't many.

beihaogu:
Closes https://github.com/lycheeverse/lychee/issues/2142.

```
Excluded links now name the rule that excluded them instead of all printing This is due to your 'exclude' values:

[EXCLUDED] https://en.wikipedia.org/wiki/Static_program_analysis (at 7:1) | Excluded by pattern: `wikipedia\.org`
[EXCLUDED] https://licensebuttons.net/p/zero/1.0/88x31.png (at 16:2) | Excluded by pattern: `licensebuttons\.net`
[EXCLUDED] http://example.com/ (at 20:1) | Excluded by pattern: `example\.com`
[EXCLUDED] https://example.com/ (at 21:1) | Excluded by pattern: `example\.com`
[EXCLUDED] mailto:test@example.com (at 23:1) | Excluded: mail checking is disabled (use --include-mail to enable)
[EXCLUDED] mailto:test2@example.com (at 24:8) | Excluded: mail checking is disabled (use --include-mail to enable)
```

Status::Excluded carries an ExcludeReason now, so all the formatters (plain, color, emoji, task, JSON, JUnit) get this for free. Built-in exclusions name the flag that controls them where there is one, e.g. Excluded: `192.168.1.5` is a private IP address (use --exclude-private=false to check it).

One deviation from what we agreed in the issue: I'd suggested keeping the generic message by default and adding "check verbose mode for details". That turns out not to work as non-verbose runs don't print excluded lines at all, so the hint has nowhere to appear, and the only non-verbose surfaces are --format json/junit, where sending a machine-readable consumer back to -v seems worse. So the reason is just always in the status. Happy to gate it if you'd rather.

Breaking for library users: Status::Excluded becomes Status::Excluded(ExcludeReason), and the details string in JSON / message in JUnit change.

(FalsePositive still just says "known false positive" without naming which built-in rule matched. FALSE_POSITIVE_SET is a RegexSet too so the same trick works, but it means changing the signature of the public is_false_positive — left it out, can add if you want it here.)

### DIFF
--- a/lychee-bin/src/cache.rs
+++ b/lychee-bin/src/cache.rs
@@ -149,8 +149,8 @@ fn cache_hit(
     cache_key: &Uri,
     value: &CacheValue,
 ) -> Response {
-    let status = if client.is_excluded(cache_key) {
-        Status::Excluded
+    let status = if let Some(reason) = client.exclusion_reason(cache_key) {
+        Status::Excluded(reason)
     } else {
         // Can't impl `Status::from(v.value().status)` here because the
         // `accepted` status codes might have changed from the previous run
--- a/lychee-lib/src/chain/mod.rs
+++ b/lychee-lib/src/chain/mod.rs
@@ -11,7 +11,7 @@
 //! the handler to the chain.
 //!
 //! [pattern]: https://github.com/lpxxn/rust-design-pattern/blob/master/behavioral/chain_of_responsibility.rs
-use crate::Status;
+use crate::{ExcludeReason, Status};
 use async_trait::async_trait;
 use core::fmt::Debug;
 use std::sync::Arc;
@@ -224,7 +224,7 @@ impl<'a> ClientRequestChains<'a> {
 
         // Consider the request to be excluded if no chain element has converted
         // it to a `ChainResult::Done`
-        Status::Excluded
+        Status::Excluded(ExcludeReason::RequestChain)
     }
 }
 
--- a/lychee-lib/src/client.rs
+++ b/lychee-lib/src/client.rs
@@ -13,7 +13,7 @@
     clippy::default_trait_access,
     clippy::used_underscore_binding
 )]
-use crate::{BasicAuthExtractor, remap::Remap};
+use crate::{BasicAuthExtractor, ExcludeReason, remap::Remap};
 use std::{collections::HashSet, sync::Arc, time::Duration};
 
 use http::{
@@ -566,12 +566,14 @@ impl Client {
         let start = std::time::Instant::now(); // Measure check time
         let remap = self.remap(&mut uri)?.inspect(|r| debug!("Remapping {r}"));
 
-        let (status, redirects) = match uri.scheme() {
-            _ if self.is_excluded(&uri) => (Status::Excluded, None),
-            _ if uri.is_tel() => (Status::Excluded, None), // We don't check tel: URIs
-            _ if uri.is_file() => (self.check_file(&uri).await, None),
-            _ if uri.is_mail() => (self.check_mail(&uri).await, None),
-            _ => self.check_website(&uri).await,
+        let (status, redirects) = if let Some(reason) = self.exclusion_reason(&uri) {
+            (Status::Excluded(reason), None)
+        } else {
+            match uri.scheme() {
+                _ if uri.is_file() => (self.check_file(&uri).await, None),
+                _ if uri.is_mail() => (self.check_mail(&uri).await, None),
+                _ => self.check_website(&uri).await,
+            }
         };
 
         Ok(Response::new(
@@ -614,6 +616,12 @@ impl Client {
     #[must_use]
     pub fn is_excluded(&self, uri: &Uri) -> bool {
         self.filter.is_excluded(uri)
+    }
+
+    /// Returns why the given `uri` should be ignored from checking.
+    #[must_use]
+    pub fn exclusion_reason(&self, uri: &Uri) -> Option<ExcludeReason> {
+        self.filter.exclusion_reason(uri)
     }
 
     /// Checks the given URI of a website.
@@ -678,7 +686,7 @@ mod tests {
 
     use super::ClientBuilder;
     use crate::{
-        BasicAuthExtractor, ErrorKind, Redirect, Redirects, Request, Status, Uri,
+        BasicAuthExtractor, ErrorKind, ExcludeReason, Redirect, Redirects, Request, Status, Uri,
         chain::{ChainResult, Handler, RequestChain},
         remap::{Remap, Remaps},
     };
@@ -1077,7 +1085,7 @@ mod tests {
         #[async_trait]
         impl Handler<Request, Status> for ExampleHandler {
             async fn handle(&mut self, _: Request) -> ChainResult<Request, Status> {
-                ChainResult::Done(Status::Excluded)
+                ChainResult::Done(Status::Excluded(ExcludeReason::RequestChain))
             }
         }
 
@@ -1091,6 +1099,6 @@ mod tests {
 
         let result = client.check("http://example.com");
         let res = result.await.unwrap();
-        assert_eq!(res.status(), &Status::Excluded);
+        assert_eq!(res.status(), &Status::Excluded(ExcludeReason::RequestChain));
     }
 }
--- a/lychee-lib/src/filter/mod.rs
+++ b/lychee-lib/src/filter/mod.rs
@@ -15,7 +15,7 @@ pub type Excludes = regex_filter::RegexFilter;
 /// You can exclude paths and files based on regex patterns.
 pub type PathExcludes = regex_filter::RegexFilter;
 
-use crate::Uri;
+use crate::{ExcludeReason, Uri};
 
 /// These domains are explicitly defined by RFC 2606, section 3 Reserved Example
 /// Second Level Domain Names for describing example cases and should not be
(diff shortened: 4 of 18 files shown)


lychee-bin/src/cache.rs
mre, line 161: Did clippy suggest that? I find it harder to read than a simple match.

```rust
let status = match client.exclusion_reason(cache_key) {
    Some(reason) => Status::Excluded(reason),
    None => {
        // Can't impl `Status::from(v.value().status)` here because the
        // `accepted` status codes might have changed from the previous run
        // and they may have an impact on the interpretation of the status
        // code.
        client.host_pool().record_persistent_cache_hit(cache_key);
        Status::from_cache_status(value.status, accept)
    }
}
```

lychee-lib/src/chain/mod.rs
```diff
@@ -224,11 +224,10 @@ impl<'a> ClientRequestChains<'a> {
 
         // Consider the request to be excluded if no chain element has converted
         // it to a `ChainResult::Done`
-        Status::Excluded
+        Status::Excluded(ExcludeReason::RequestChain)
```
mre, line 227: I wonder if we should call this exclude reason differently. At the end, this happens if a request cannot be handled by lychee, so maybe `ExcludeReason::UnhandledRequest`?

lychee-lib/src/filter/mod.rs
```diff
@@ -192,20 +192,51 @@ impl Filter {
         matches!(self.includes, Some(ref includes) if includes.is_match(input))
     }
 
+    /// The reason the IP address of `uri` is excluded, if it is
+    ///
+    /// Mirrors [`Self::is_ip_excluded`], but reports which of the address
+    /// ranges matched so the corresponding flag can be named.
+    fn ip_reason(&self, uri: &Uri) -> Option<ExcludeReason> {
```
mre, line 199: How about we align the method names a bit?

```suggestion
    fn ip_excluded_reason(&self, uri: &Uri) -> Option<ExcludeReason> {
```

lychee-lib/src/filter/mod.rs
```diff
@@ -192,20 +192,51 @@ impl Filter {
...
+        // Only reached for URIs with an IP host, so `host_ip` is always `Some`
+        let address = || uri.host_ip().map(|ip| ip.to_string()).unwrap_or_default();
```
mre, line 201: I think the closure is premature optimization and the string allocation is fine even if we end up not using it below. So I'd prefer an early return here:

```rust
let address = uri.host_ip()?.to_string();
```

I find this easier to read.

Then you can also just use `address` instead of `address()` below.

lychee-lib/src/filter/mod.rs
```diff
@@ -215,58 +246,81 @@ impl Filter {
...
-            || self.is_mail_excluded(uri)
-            || uri.is_tel()
-            || is_example_domain(uri)
-            || is_unsupported_domain(uri)
-        {
-            return true;
+    pub fn exclusion_reason(&self, uri: &Uri) -> Option<ExcludeReason> {
+        if self.is_scheme_excluded(uri) {
+            return Some(ExcludeReason::Scheme(uri.scheme().to_owned()));
+        }
```
mre, line 253: Can you add an empty line between each if block?

lychee-lib/src/filter/mod.rs
```diff
@@ -215,58 +246,81 @@ impl Filter {
...
+        if self.is_host_excluded(uri) {
+            return Some(ExcludeReason::Host(
+                uri.domain().unwrap_or_default().to_owned(),
```
mre, line 256: I can see that you use `unwrap_or_default()` to fulfill the `Option` return value, but I'm not a huge fan of this pattern in general. A safer way might be if `is_host_excluded` already returned the host:

```rust
if let Some(host) = self.is_host_excluded(uri) {
    return Some(ExcludeReason::Host(host));
}
```

lychee-lib/src/filter/mod.rs
```diff
@@ -215,58 +246,81 @@ impl Filter {
...
+        }
+        if let Some(reason) = self.ip_reason(uri) {
+            return Some(reason);
+        }
+        if self.is_mail_excluded(uri) {
+            return Some(ExcludeReason::Mail);
+        }
+        if uri.is_tel() {
+            return Some(ExcludeReason::Telephone);
         }
```
mre, line 267: Nice.

lychee-lib/src/filter/mod.rs
```diff
@@ -215,58 +246,81 @@ impl Filter {
...
 
         let input = uri.as_str();
 
+        let builtin_domain_reason = if is_example_domain(uri) {
+            Some(ExcludeReason::ExampleDomain)
+        } else if is_unsupported_domain(uri) {
+            Some(ExcludeReason::UnsupportedDomain)
```
mre, line 274: It would be nice if we could report the unsupported domain as well. Same pattern as above:

```rust
} else if let Some(domain) = is_unsupported_domain(uri) {
    Some(ExcludeReason::UnsupportedDomain(domain))
}
```

lychee-lib/src/filter/mod.rs
```diff
@@ -215,58 +246,81 @@ impl Filter {
...
+            None
+        };
+        if let Some(builtin) = builtin_domain_reason {
+            // Built-in domain rules exclude the URI either way. Still prefer
+            // reporting a matching user-provided pattern, which makes the
+            // configured rule visible without changing include precedence.
+            if self.is_includes_match(input) {
+                return Some(builtin);
+            }
+            return Some(self.pattern_reason(input).unwrap_or(builtin));
```
mre, line 285: I believe you could also use a let-chain to get rid of the slightly awkward `self.pattern_reason(input).unwrap_or(builtin)`:

```rust
if let Some(builtin) = builtin_domain_reason {
    // If not an includes match, try to prefer the user-provided pattern reason
    if !self.is_includes_match(input) 
        && let Some(pattern) = self.pattern_reason(input) 
    {
        return Some(pattern);
    }

    return Some(builtin);
}
```

lychee-lib/src/filter/mod.rs
```diff
@@ -215,58 +246,81 @@ impl Filter {
...
+        }
+
         if self.is_includes_empty() {
             if self.is_excludes_empty() {
-                // Both excludes and includes rules are empty:
-                // *Presumably included* unless it's a false positive
-                return is_false_positive(input);
+                return is_false_positive(input).then_some(ExcludeReason::FalsePositive);
```
mre, line 290: Can you add back the comment here? I found it to be helpful.

```rust
// Both excludes and includes rules are empty;
// presumably included unless it's a false positive.
```

lychee-lib/src/filter/mod.rs
```diff
@@ -215,58 +246,81 @@ impl Filter {
...
-            return false;
+            return None;
         }
 
-        // Exclude well-known false-positives
-        // Performed after checking includes to allow user-overwrites
-        if is_false_positive(input)
-                // Previous checks imply input is not explicitly included.
-                // If exclude rules are empty, then *presumably excluded*
-                || self.is_excludes_empty()
```
mre, line 251: Same here: I believe we should bring back those comments. They are useful to humans for understanding the flow.

lychee-lib/src/filter/mod.rs
```diff
@@ -418,6 +472,167 @@ mod tests {
...
+            Some(ExcludeReason::UnsupportedDomain)
+        );
+    }
+
+    /// A user-provided pattern is reported in preference to a built-in domain
+    /// rule, because that is the rule the user can act on. Includes still take
+    /// precedence over the user's excludes.
+    ///
+    /// Note: this exercises the same branch as reserved example domains, which
+    /// cannot be tested here because `EXAMPLE_DOMAINS` is empty under `cfg(test)`.
```
mre, line 613: Shall we remove this part of the comment?

mre: Actually, I'm not even sure if we need the test at all. Exclusion order might not be something we'd want to explicitly guarantee.

lychee-lib/src/filter/regex_filter.rs
```diff
@@ -36,6 +36,17 @@ impl RegexFilter {
...
+    /// Returns the first regular expression that matches the input.
+    #[must_use]
+    pub fn matching_pattern(&self, input: &str) -> Option<&str> {
+        self.regex
+            .matches(input)
+            .iter()
+            .next()
+            .and_then(|index| self.regex.patterns().get(index))
+            .map(String::as_str)
+    }
```
mre, line 48: Here's another attempt:

```rust
/// Returns the first regular expression pattern that matches the input.
#[must_use]
pub fn matching_pattern(&self, input: &str) -> Option<&str> {
    self.regex
        .matches(input)
        .into_iter()
        .next()
        .map(|index| &self.regex.patterns()[index])
}
```

I would guess that this works because `regex::RegexSet` returns an iterator from `.matches(input).into_iter()`, so you can avoid calling `.iter()` explicitly. Also, `self.regex.patterns()[index]` directly returns a `&str`, which makes `.get()` and `.map(String::as_str)` redundant. We can do that because `index` is guaranteed to be a valid index returned by `RegexSet::matches`, so self.regex.patterns()[index] is completely safe.

lychee-lib/src/types/status.rs
```diff
@@ -16,6 +16,82 @@ const ICON_ERROR: &str = "✗";
...
+    /// The IP address is in a private range.
+    PrivateIp(String),
+    /// The IP address is in the link-local range.
+    LinkLocalIp(String),
+    /// Telephone links are not checked.
+    Telephone,
+    /// Reserved example domains are not checked.
+    ExampleDomain,
+    /// The domain is not supported.
+    UnsupportedDomain,
```
mre, line 42: As mentioned above, we should capture the unsupported domain for reporting here:

```rust
UnsupportedDomain(String),
```

lychee-lib/src/types/status.rs
```diff
@@ -16,6 +16,82 @@ const ICON_ERROR: &str = "✗";
...
+    /// The resource is a known false positive.
+    FalsePositive,
+    /// No user-provided include pattern matched the resource.
+    NotIncluded,
+    /// Mail checking support is not enabled in this build.
+    MailFeatureDisabled,
+    /// A request chain excluded the resource.
+    RequestChain,
```
mre, line 50: As mentioned above, not a huge fan of this variant. It does not express the intent behind the exclusion. Alternative name: `UnhandledRequest`.

lychee-lib/src/types/status.rs
```diff
@@ -300,9 +376,81 @@ impl From<ErrorKind> for Status {
...
 
+    /// The rendered reason is what users see for every excluded link, so pin
+    /// down the wording of every variant.
+    ///
+    /// This also covers the variants that cannot be reached from the filter
+    /// unit tests: `ExampleDomain` (the example domain set is empty under
+    /// `cfg(test)`) and `MailFeatureDisabled` (only produced when the
+    /// `email-check` feature is off).
+    #[test]
+    fn test_exclude_reason_messages() {
```
mre, line 390: Please use `rstest` for this test; see other tests for reference.

lychee-lib/src/client.rs
```diff
@@ -566,12 +566,14 @@ impl Client {
...
-            _ if uri.is_mail() => (self.check_mail(&uri).await, None),
-            _ => self.check_website(&uri).await,
+        let (status, redirects) = if let Some(reason) = self.exclusion_reason(&uri) {
+            (Status::Excluded(reason), None)
+        } else {
+            match uri.scheme() {
+                _ if uri.is_file() => (self.check_file(&uri).await, None),
+                _ if uri.is_mail() => (self.check_mail(&uri).await, None),
+                _ => self.check_website(&uri).await,
+            }
```
mre, line 576: I think you can now use match-guards in if-let! 😃 

```rust
let (status, redirects) = match uri.scheme() {
    _ if let Some(reason) = self.exclusion_reason(&uri) => (Status::Excluded(reason), None),
    _ if uri.is_file() => (self.check_file(&uri).await, None),
    _ if uri.is_mail() => (self.check_mail(&uri).await, None),
    _ => self.check_website(&uri).await,
};
```

mre (commented):
Sorry for the late review. A few comments.

Issue #2275, opened 2026-08-07
Wikilink resolution skips the target as written


uncoolclub:
With `--include-wikilinks`, a wikilink whose target includes a file extension is not found, even when the file is present in the wikilink index.

## Reproduction

```console
$ find . -type f
./note.md
./sub/image.png
./sub/Note.md
```

```console
$ cat note.md
[[image.png]]
[[Note]]
```

```console
$ lychee --offline --no-progress --include-wikilinks \
    --fallback-extensions md \
    --base-url "$PWD" \
    note.md

[ERROR] file:///tmp/w/image.png (at 1:3) | Wikilink file:///tmp/w/image.png not found at /tmp/w/image.png
🔍 2 Total 🔗 2 Unique ✅ 1 OK 🚫 1 Error
```

`[[Note]]` resolves to `sub/Note.md`, confirming that the index lookup itself works.

`[[image.png]]` fails even though `sub/image.png` is present in the index.

## Cause

`WikilinkResolver::resolve` in `lychee-lib/src/checker/wikilink/resolver.rs` only looks up paths produced from the configured fallback extensions:

```rust
for ext in &self.fallback_extensions {
    let mut candidate = path.to_path_buf();
    candidate.set_extension(ext);

    if let Some(resolved) = self.checker.contains_path(&candidate) {
        return Ok(resolved);
    }
}
```

With `--fallback-extensions md`, `image.png` is converted to `image.md`. The original `image.png` path is never checked.

When no fallback extensions are configured, the loop does not run at all, so no index lookup is attempted.

## Expected behavior

The target as written should be checked before applying fallback extensions:

1. `image.png`
2. `image.md`

This preserves explicit filenames while still allowing extensionless wikilinks such as `[[Note]]` to resolve through the configured fallback extensions.

## Suggested fix

Look up the original path before iterating over fallback candidates.

I noticed this while working on #2274 and left it out of that PR because it is a separate issue.

## Version

`lychee 0.24.2`, reproduced on an unmodified checkout of `master` at c59caafb.

PR #2276, opened 2026-08-07
fix: append fallback extensions past a pre-existing extension

### LINKED ISSUE #2274: --fallback-extensions resolves dotted filenames against the wrong file
`--fallback-extensions` replaces the last segment of a dotted filename instead of appending to it.

As a result, `/a.b` never resolves to `a.b.html`. Worse, if an unrelated `a.html` exists, the link is reported as valid against the wrong file.

This is the no-space counterpart of #2235. That case was fixed by #2236 for filenames containing spaces, but dotted filenames still follow the replacement path.

## Reproduction

Given these files:

```console
$ ls
a.b.html  t.html
```

```console
$ cat t.html
<a href="/a.b">dot</a>
```

Run lychee with `html` as a fallback extension:

```console
$ lychee --offline --no-progress --fallback-extensions html --root-dir . t.html
[ERROR] file:///tmp/verify/a.b (at 1:10) | File not found. Check if file exists and path is correct
🔍 1 Total (in 4ms) 🔗 1 Unique ✅ 0 OK 🚫 1 Error
```

`a.b.html` exists, but lychee does not find it.

Now add an unrelated `a.html`:

```console
$ ls
a.b.html  a.html  t.html
```

Running the same command now succeeds:

```console
$ lychee --offline --no-progress --fallback-extensions html --root-dir . t.html
🔍 1 Total (in 0s) 🔗 1 Unique ✅ 1 OK 🚫 0 Errors
```

The link is reported as valid even though it resolves against `a.html`, not `a.b.html`.

The false positive is particularly concerning because it is much harder to notice than a missing-file error.

## Cause

In `apply_fallback_extensions` in `lychee-lib/src/checker/file.rs`:

```rust
// Try fallback extensions, replacing any current extension in the path.
for ext in &self.fallback_extensions {
    path_buf.set_extension(ext);
    if path_buf.is_file() {
        return Ok(path_buf);
    }
}
```

`PathBuf::set_extension` replaces the existing suffix. For `a.b`, this produces `a.html`, so `a.b.html` is never tried.

The whitespace-specific branch added in #2236 appends the fallback extension instead, so the two paths handle similar filenames differently.

## Expected behavior

For `/a.b` with `html` configured as a fallback extension, lychee should try `a.b.html`.

It is not always possible to know whether `.b` is intended as an extension or simply part of the filename. Trying both candidates, with the appended form first, could preserve the existing replacement behavior while avoiding this false positive:

1. `a.b.html`
2. `a.html`

I would like to open a PR for this if the approach makes sense to you.

## Context

I found this while checking a documentation site whose pretty URLs contain version numbers, for example `/use/migrate-to-10.0.0`, served from `migrate-to-10.0.0.html`. These links currently have to be excluded from checking.

## Version

Tested with:

* lychee `0.24.2`
* nightly from `2026-08-04`

I also tested the whitespace case from #2236 against both builds: nightly passes, `0.24.2` fails. The dotted filename case fails on both.

uncoolclub:
Closes #2274.

`apply_fallback_extensions` currently uses `PathBuf::set_extension`, which replaces the existing extension instead of appending a fallback extension.

For `/a.b` with `--fallback-extensions html`, lychee only tries:

```text
a.html
```

It never tries:

```text
a.b.html
```

If an unrelated `a.html` exists, the link can therefore be reported as valid against the wrong file.

## Why append first

The option is documented as appending extensions. From `lychee-bin/src/config/mod.rs:244`, also repeated in the README:

> To automatically **append** file extensions when locating files, see `--fallback-extensions`.

Appending also preserves the path written by the author. Replacing can discard part of the filename and resolve to a different document.

For example:

```text
a.tar -> a.tar.gz
```

is preferable to:

```text
a.tar -> a.gz
```

#2236 already introduced append behavior for filenames whose extension contains whitespace. This change applies the same rule consistently to all paths.

Replacement remains available as a fallback, so existing cases such as:

```text
b.non-existing -> b.gz
```

continue to work when the appended candidate does not exist.

## Changes

A new `fallback_candidates` helper in `checker/mod.rs` generates candidates in this order for each configured extension:

1. Append the extension
2. Replace the existing extension

Configured extension precedence is unchanged. For example:

```text
--fallback-extensions html,htm
```

still prefers `html` over `htm`.

The helper is now shared by:

* `FileChecker::apply_fallback_extensions`
* `WikilinkResolver::resolve`

This removes the duplicated replacement logic and gives both call sites the same fallback behavior.

The whitespace-specific branch in `FileChecker` is no longer needed for appending. Keeping it would duplicate the extension after general append handling, for example:

```text
file. hi.md.md
```

Its other half is kept. A whitespace "extension" is still never replaced, since it is taken to be part of the file name:

```text
file. hi -> file.md
```

is not produced. #2236 added that rule to avoid exactly the wrong-file resolution this PR is about, and dropping it would reintroduce it.

Candidates are constructed with `file_name()` and `with_file_name()` rather than by pushing onto the full path. This preserves the existing trailing-slash normalization:

```text
/a/b/ -> /a/b.html
```

rather than:

```text
/a/b/.html
```

`PathBuf::add_extension` would express this more directly, but it requires Rust 1.91 while the current MSRV is 1.88.

When a path has no extension, appending and replacing produce the same result, so the duplicate candidate is skipped. This remains the common path for pretty URLs.

The `--fallback-extensions` help text and the matching README block now state the candidate order, so it can be read without going to the source.

## Tests

The existing embedded-dot test now expects:

```text
a.tar -> a.tar.gz
```

New fixtures cover the false-positive case reported in #2274:

```text
c.d -> c.d.md
```

An unrelated `c.md` is also present to verify that lychee does not resolve the link against the wrong file.

Replacement behavior is still covered by:

```text
b.non-existing -> b.gz
```

Two tests guard the wrong-file cases. `e. hi` must not resolve to the existing `e.md`, and on the wikilink side `Page.v1` must reach `Page.v1.md` rather than the existing `Page.md`. Reverting either half of the rule makes them fail.

A new test pins the candidate order. With `--fallback-extensions html,htm` the order is:

```text
a.b.html
a.html
a.b.htm
a.htm
```

so extension precedence outranks the preference for appending. The fixtures make the two orderings distinguishable: `d.html` and `d.e.htm` both exist, and a link to `d.e` must resolve to `d.html`. Reordering the candidates to try every appended form first makes that assertion fail.

Validation results:

* `lychee-lib`: 505 tests passed
* `cargo clippy -- -D warnings`: passed
* `cargo fmt --check`: passed
* `cargo +1.88.0 check`: passed

Two CLI failures, `test_fragments` and `test_fragments_regression`, also reproduce on an unmodified checkout and are unrelated to this change.

## Behavior changes to review

Appending broadens which files can resolve. In particular:

```text
a.tar -> a.tar.gz
```

now succeeds. This is intentional and matches the documented append behavior.

One separate issue remains unchanged: `WikilinkResolver::resolve` does not try the bare path before fallback extensions. As a result:

```text
[[image.png]]
```

with `--fallback-extensions md` can look for `image.md` and miss an existing `image.png`.

Filed separately as #2275, and left out of scope here.

### DIFF
--- a/lychee-lib/src/checker/mod.rs
+++ b/lychee-lib/src/checker/mod.rs
@@ -2,7 +2,45 @@
 //!
 //! This module contains all checkers, which are responsible for checking the status of a URL.
 
+use std::path::{Path, PathBuf};
+
 pub(crate) mod file;
 pub(crate) mod mail;
 pub(crate) mod website;
 pub(crate) mod wikilink;
+
+/// Returns fallback candidates in configured extension order.
+///
+/// For each extension, the appended form is tried before the replaced form.
+/// Only the appended form is produced for a path without an extension, or with
+/// one containing whitespace, which is taken to be part of the file name.
+pub(crate) fn fallback_candidates<'a>(
+    path: &'a Path,
+    extensions: &'a [String],
+) -> impl Iterator<Item = PathBuf> + 'a {
+    let can_replace = path.extension().is_some_and(|extension| {
+        !extension
+            .as_encoded_bytes()
+            .iter()
+            .any(u8::is_ascii_whitespace)
+    });
+
+    extensions.iter().flat_map(move |extension| {
+        // Operate on the file name so trailing slashes are normalized like
+        // `set_extension`: `/a/b/` becomes `/a/b.html`, not `/a/b/.html`.
+        let appended = path.file_name().map(|file_name| {
+            let mut file_name = file_name.to_os_string();
+            file_name.push(".");
+            file_name.push(extension);
+            path.with_file_name(file_name)
+        });
+
+        let replaced = can_replace.then(|| {
+            let mut candidate = path.to_path_buf();
+            candidate.set_extension(extension);
+            candidate
+        });
+
+        appended.into_iter().chain(replaced)
+    })
+}
--- a/lychee-lib/src/checker/mod.rs
+++ b/lychee-lib/src/checker/mod.rs
@@ -2,7 +2,88 @@
 //!
 //! This module contains all checkers, which are responsible for checking the status of a URL.
 
+use std::path::{Path, PathBuf};
+
 pub(crate) mod file;
 pub(crate) mod mail;
 pub(crate) mod website;
 pub(crate) mod wikilink;
+
+/// Returns the candidates to try for `path`, in the order they should be tried.
+///
+/// The path itself comes first, then every extension appended to the full file
+/// name, then every extension replacing an existing one. Appended candidates
+/// are exhausted first because they preserve the name as written.
+///
+/// Replacement is skipped when the existing extension contains ASCII
+/// whitespace. Replacing it could discard part of the intended file name and
+/// resolve to an unrelated file.
+pub(crate) fn fallback_candidates<'a>(
+    path: &'a Path,
+    extensions: &'a [String],
+) -> impl Iterator<Item = PathBuf> + 'a {
+    let has_replaceable_extension = path.extension().is_some_and(|extension| {
+        !extension
+            .as_encoded_bytes()
+            .iter()
+            .any(u8::is_ascii_whitespace)
+    });
+
+    let original = std::iter::once(path.to_path_buf());
+
+    // Build from the file name rather than the whole path so that trailing
+    // slashes normalize like `set_extension`: `/a/b/` gives `/a/b.html`.
+    let appended = extensions.iter().filter_map(move |extension| {
+        path.file_name().map(|file_name| {
+            let mut file_name = file_name.to_os_string();
+            file_name.push(".");
+            file_name.push(extension);
+            path.with_file_name(file_name)
+        })
+    });
+
+    let replaceable = if has_replaceable_extension {
+        extensions
+    } else {
+        &[]
+    };
+
+    let replaced = replaceable.iter().map(move |extension| {
+        let mut candidate = path.to_path_buf();
+        candidate.set_extension(extension);
+        candidate
+    });
+
+    original.chain(appended).chain(replaced)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn candidates(path: &str, extensions: &[&str]) -> Vec<String> {
+        let extensions: Vec<String> = extensions.iter().map(ToString::to_string).collect();
+        fallback_candidates(Path::new(path), &extensions)
+            .map(|candidate| candidate.to_string_lossy().into_owned())
+            .collect()
+    }
+
+    #[test]
+    fn appending_is_exhausted_before_replacing() {
+        assert_eq!(
+            candidates("a.tar", &["gz", "md"]),
+            ["a.tar", "a.tar.gz", "a.tar.md", "a.gz", "a.md"]
+        );
+    }
+
+    #[test]
+    fn a_path_without_an_extension_has_nothing_to_replace() {
+        assert_eq!(candidates("a", &["gz", "md"]), ["a", "a.gz", "a.md"]);
+    }
+
+    #[test]
+    fn an_extension_with_whitespace_is_not_replaced() {
+        // Replacing here would drop " hi" and could resolve to an unrelated `e.md`.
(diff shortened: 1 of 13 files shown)


lychee-lib/src/checker/mod.rs
```diff
@@ -2,7 +2,45 @@
...
+            file_name.push(".");
+            file_name.push(extension);
+            path.with_file_name(file_name)
+        });
+
+        let replaced = can_replace.then(|| {
+            let mut candidate = path.to_path_buf();
+            candidate.set_extension(extension);
+            candidate
+        });
```
mre: I would refactor this part into its own function as well. This way, you can write a few small tests to make sure it works as expected. (The code looks fine to me.)

mre: Overall, I really like the structure of `fallback_candidates` and it looks very idiomatic to me. Good work!

mre (approved):
Good work! Fixes the problem and makes the code easier to read overall. Thanks! 😊

katrinafyi:
Quick comments. 
- I think that if a link ends in `/`, then we should not apply any fallback extensions.
- Since the test files are rather trivial and only their filenames matter, it might be cleaner if the test cases create the test files at runtime into a temp directory.
- This is subjective, but it might make more sense to try appending all of the fallbacks, then try replacing all of the fallbacks. I think that appending is generally safer and less likely to point to a wrong file.
- I don't think there's much need to keep the special case excluding "extensions" with space. It will get handled properly by the appending logic.
- I think that the fix to https://github.com/lycheeverse/lychee/issues/2275 will be tiny - just yield the original path as the first item in fallback_candidates. If this is the case, it may as well be done in this PR :)
- I'm not sure what you're talking about with "Two CLI failures, test_fragments and test_fragments_regression, also reproduce on an unmodified checkout and are unrelated to this change." I don't see the failures in main, but nor do I see them in your CI checks...

uncoolclub:
Thanks! All six points were helpful. 🙇‍♀️
I pushed a revision.

### Updated

* **Candidate order:** now tries the original path first, then all appends, then all replacements. I agree this better avoids resolving to a file the URL did not actually name.
* **#2275:** folded in. Trying the original path first also exposed a `WikilinkIndex` issue where a directory like `Page.v1/` could shadow `Page.v1.md`; lookup now only considers files.
* **Tests:** moved the new cases to runtime temp directories. The existing fixtures remain because `test_fallback_extensions` still scans that directory.
* **CLI failures:** you were right, and I removed this from the PR description. The failures were caused by my local `GITHUB_TOKEN`; without it, the full CLI suite passes (173/173), consistent with CI.
* **`fallback_candidates` tests (@mre):** added direct unit tests for the candidate sequence, so the ordering contract is pinned without filesystem fixtures. Thanks for the review!

Two points still need a decision:

### Whitespace guard

I tried removing it, but that reintroduces false matches. With fallback `md`, if `e. hi.md` does not exist but an unrelated `e.md` does:

```text
e. hi      miss
e. hi.md   miss   (append)
e.md       HIT    (replace drops " hi")
```

Trying appends first does not prevent the replacement once all appended candidates miss, so I kept the guard and added `test_extension_with_whitespace_is_not_replaced`.

If there is another way to prevent `e. hi` from resolving to `e.md`, I'm happy to use that instead.

### Trailing slash

I agree that a link ending in `/` should not get fallback extensions.

I looked into implementing that, but the distinction is lost before `fallback_candidates` is called: `request.rs` strips the empty trailing segment from `file:` URLs, and `apply_index_files` loses it again by joining the default `.` index. As a result, `same_name` and `same_name/` both reach this logic as `same_name/.`.

Preserving that distinction would require changing the earlier normalization rather than `fallback_candidates` itself, and `test_fallback_extensions_on_directories` currently asserts the existing behavior.

I left that change out of this revision for now because it expands the scope beyond fallback candidate generation. Would you prefer that I include the normalization change in this PR, or handle it separately?

uncoolclub:
@katrinafyi Whenever you get a chance, would you mind taking another look? 🙇‍♀️

mre:
This looks fine to me. I've rebased on top of `master` and all tests are still passing. Merging. Good job and thanks a lot for your contribution, @uncoolclub. 🪩

Issue #2278, opened 2026-08-10
Document anchor fragment check compatibilty


hoijui:
..., at least to some degree of detail, within the `--help` output.

I personally undertook adding this functionality to an other rust based link checker, and I have a slight idea what that really means. One needs to get the target document/content, and regardless of its content, parse it (as a modern browser would), and get a list of the anchors therein. For full support, one would basically have to run "half" a browser-engine. I am sure you only support a sub-set of what a browser might, so it would be important for users to have more or less of an idea what is actually supported and to what degree.
To me, this is by far the most interesting feature of any link-checker, as it is vital for checking integrity of documentation, and it is supported by very few checkers (I had not found a single one when I last checked).
I am willing to document it (make a PR) if you give me a rough description or idea where I could check.

katrinafyi:
Just checking, is your idea about the feature of https://lychee.cli.rs/guides/cli/#--include-fragments or something different?

hoijui:
yes, that!
Hmm... maybe I understand the feature wrong.. I understand it so, that the target URL/document/content is checked, whether the anchor in the URL exists in it.

mre:
Thanks for looking into that.

On top of what Kait posted, we have some more documentation on anchor support [here](https://lychee.cli.rs/recipes/anchors/). We could extend it.

If you want to give it a try, you could start with the test fixtures [here](https://github.com/lycheeverse/lychee/tree/master/fixtures/fragments) and [here](https://github.com/lycheeverse/lychee/tree/master/fixtures/fragments-fallback-extensions). If you check the directory history, you'll find the relevant PRs that added the functionality and extract the details from there.

hoijui:
I made a first draft here:
https://github.com/lycheeverse/lychee/pull/2281

I am sure you'll want it changed. I am also unsure if having a link there makes sense at all.

katrinafyi:
I took a look!

I'm also interested in what you added to the other link checker. Was it a similar feature or did it also run some Javascript like a browser?

It might interest you to see that we also have this issue tracking the failed fragment checks in Github.com URLs which use Javascript:
- https://github.com/lycheeverse/lychee/issues/1729

hoijui:
[...]

PR #2281, opened 2026-08-10
Extend --include-fragments docu #2278

hoijui:
A first draft, including the info I would want to see

### DIFF
--- a/lychee-bin/src/config/mod.rs
+++ b/lychee-bin/src/config/mod.rs
@@ -519,16 +519,10 @@ pub(crate) struct Config {
     ///
     /// If provided without a value, defaults to `anchor-only`.
     ///
-    /// This check will not work on all URLs.
-    /// The rough requirements are, that the target content be Markdown or HTML,
-    /// and that it follows certain, common conventions.
-    /// If for example JavaScript needs to be executed
-    /// to get to the actual HTML content that might contain the anchor,
-    /// the check will not see the anchor (e.g. report a miss),
-    /// as this scenario would require running this checker to run a browser engine,
-    /// which would make it orders of magnitude heavier on computing resources.
-    /// More info about this check can be found here:
-    /// <https://lychee.cli.rs/recipes/anchors/>
+    /// Fragment checking does not work for every URL. In general, the target
+    /// content needs to be Markdown or HTML and follow common conventions.
+    /// Anchors that only appear after JavaScript execution may be missed, since
+    /// lychee does not run a browser engine.
     #[arg(
         long,
         default_missing_value = "anchor-only",
--- a/lychee-bin/src/config/mod.rs
+++ b/lychee-bin/src/config/mod.rs
@@ -518,6 +518,17 @@ pub(crate) struct Config {
     /// or `full` to check both.
     ///
     /// If provided without a value, defaults to `anchor-only`.
+    ///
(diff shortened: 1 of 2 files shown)


lychee-bin/src/config/mod.rs
```diff
@@ -518,6 +518,17 @@ pub(crate) struct Config {
...
+    /// This check will not work on all URLs.
+    /// The rough requirements are, that the target content be Markdown or HTML,
+    /// and that it follows certain, common conventions.
+    /// If for example JavaScript needs to be executed
+    /// to get to the actual HTML content that might contain the anchor,
+    /// the check will not see the anchor (e.g. report a miss),
+    /// as this scenario would require running this checker to run a browser engine,
+    /// which would make it orders of magnitude heavier on computing resources.
+    /// More info about this check can be found here:
+    /// <https://lychee.cli.rs/recipes/anchors/>
```
katrinafyi: Makes sense. I shortened it a bit because we don't have to justify ourself in the help text. I think this limitation is very reasonable and unsurprising.

I also don't feel a strong need to list markdown + HTML.  `id` attributes are fundamentally a HTML thing so other file types don't usually make sense. We support Markdown because Markdown is often converted to HTML. It's also in the linked page, but if you feel strongly about this it can be added back.

```suggestion
    /// If enabled, this check will only succeed if the anchors exist in the
    /// static content of the target page (e.g., it will not run JavaScript).
    /// See also: <https://lychee.cli.rs/recipes/anchors/>
```

katrinafyi (commented):
If you have it checked out locally, you can run `make readme` to generate the new README text and fix the failing test. If not let me know and I can do it.

lychee-bin/src/config/mod.rs
mre: I would not mind adding this, but I'd shorten it a bit. Here's an attempt; what do you think?

```suggestion
    /// Fragment checking does not work for every URL. In general, the target
    /// content needs to be Markdown or HTML and follow common conventions.
    /// Anchors that only appear after JavaScript execution may be missed, since
    /// lychee does not run a browser engine.
```

And in general, I would restrain from adding a link. I know that we use them in other places, but the link could break in the future (which would be quite ironic, given that we're building a link checker). Open for discussion.

Also, from your issue description, I assumed that you wanted the documentation to be more specific about the concrete "flavors" of fragments we resolve. For example, without any further information, we assume that fragments follow GitHub flavored Markdown syntax in Markdown documents. Documenting this could be follow-up work, but we should tackle that in a separate PR on the [lychee docs repo](https://github.com/lycheeverse/lycheeverse.github.io).

hoijui: Thank you two! :-)
puh.. I slightly lean towards @mre 's suggestion, but am fine with both.
Also that with the Markdown flavor is indeed a good info to have as well.

In the link checker I worked on, I do it pretty much like you guys - just bare content parsing, without a browser engine.

katrinafyi: I'm also happy with @ mre's :)

katrinafyi (approved):
Thanks!

mre (approved):

mre:
Thanks for the contribution @hoijui.

hoijui:
Thank you for the tool and the process; was a pleasure!

PR #2282, opened 2026-08-16
Add a Debian install section

barnumbirr:
[pkg.haus](https://pkg.haus) is a signed APT archive carrying lychee for Debian stable, testing and unstable (amd64/arm64), built from source at upstream release tags. This adds a Debian section next to the other package managers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,14 @@ snap install lychee
 apk add lychee
 ```
 
+### Debian
+
+Available in the [pkg.haus](https://pkg.haus) archive. Follow the [setup instructions](https://pkg.haus) to configure the repository, then install with:
+
+```sh
+sudo apt install lychee
+```
+
 ### macOS
 
 Via [Homebrew](https://brew.sh):


katrinafyi:
It's not clear to me whether this "pkg.haus" is a well-established and trustworthy project. It looks brand new.

I wouldn't want to add this at the moment, but what do others think?

barnumbirr:
Totally fair concern, and you're right: pkg.haus itself is brand new (it went live this week).

The packaging behind it isn't, though. I've been building and releasing Debian packages under my personal account for years, for example:

- [alacritty-debian](https://github.com/barnumbirr/alacritty-debian) (since 2020, 41 releases)
- [zola-debian](https://github.com/pkghaus/zola-debian) (since 2021, 26 releases, recently moved into the pkghaus org)
- [i3-gaps-debian](https://github.com/barnumbirr/i3-gaps-debian) (since 2020)

alacritty-debian and i3-gaps-debian are archived as both are now officially packaged in Debian.

pkg.haus consolidates that work into a proper signed APT archive instead of loose `.deb` files on release pages. Everything is public and auditable: packages are built from upstream source at release tags in GitHub Actions ([pkghaus/action-debian-build](https://github.com/pkghaus/action-debian-build)), and the archive itself is generated by CI ([pkghaus/apt](https://github.com/pkghaus/apt)), so you can trace any published package back to the exact build run.

That said, no hard feelings if you'd rather wait until the project has more of a track record. Happy to answer any questions either way.

thomas-zahner:
@barnumbirr Thank you for including lychee in your APT archive.

I agree with @katrinafyi that we should only provide install instructions for popular & well established platforms. In my opinion the install section is already rather bloated and it might make sense to trim it down a bit in the future. If Debian/Ubuntu included lychee officially in their package distribution that's something I'd document, but I prefer to not list unofficial archives.

mre:
Thanks for the work on the APT archive, Martin. I think we all agree to close this PR, but we'd still be thankful for any ongoing maintenance of the package in your archive.

PR #2285, opened 2026-08-18
feat: add --cache-location to configure the request cache path

### LINKED ISSUE #1700: support relocating `.lycheecache`
https://lychee.cli.rs/recipes/caching/

`--cache-location some/path/lychee/cache`
`--cache-location /other/path/lychee/cache`

also:

https://github.com/lycheeverse/lychee/blob/master/lychee.example.toml

Other tools:

https://eslint.org/docs/latest/use/command-line-interface#--cache-location
https://prettier.io/docs/cli#--cache-location

ChrisJr404:
This adds a `--cache-location` option (and matching `cache_location` config key) so the on-disk request cache can be written somewhere other than the hard-coded `.lycheecache` in the current directory.

Closes #1700

The cache path was previously a compile-time constant, so every project that runs with `--cache` accumulates a `.lycheecache` file at the repo root, which then has to be added to `.gitignore`, and there was no way to redirect it into a build/scratch directory or an XDG-style cache dir. This came up a few times on the issue, including a CI setup where writing into the working tree is awkward. With this change you can point the cache wherever you want:

```
lychee --cache --cache-location target/lychee/cache .
```

The default is unchanged: when the option/key is omitted the cache still lives at `.lycheecache`, so existing setups keep working exactly as before.

A few notes on the implementation:

- Following the repo convention of avoiding clap's own defaults, the field is a private `Option<PathBuf>` with the default annotated in the doc comment (`[default: .lycheecache]`) and resolved in a `cache_location()` getter. It's wired into `check_default_values!`, so the existing `test_default_values` test verifies the documented default matches the actual one.
- The three call sites in `main.rs` (cache freshness check, load, store) now use `cfg.cache_location()` instead of the constant.
- Added a CLI test (`test_cache_location_custom_path`) that runs with a custom path and asserts the cache is written there and that no `.lycheecache` is created.
- Documented the key with a commented example in `lychee.example.toml`.

`cargo test` (config unit tests + the cache CLI tests) and `cargo clippy` pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,11 @@ Options:
           comma-separated list of excluded status codes. This example will not cache results
           with a status code of 429, 500 and 501.
 
+      --cache-location <PATH>
+          Filesystem path used to store the on-disk request cache
+
+          [default: .lycheecache]
+
       --cookie-jar <COOKIE_JAR>
           Read and write cookies using the given file. Cookies will be stored in the
           cookie jar and sent with requests. New cookies will be stored in the cookie jar
--- a/lychee-bin/src/config/mod.rs
+++ b/lychee-bin/src/config/mod.rs
@@ -275,6 +275,12 @@ pub(crate) struct Config {
     #[arg(long, verbatim_doc_comment)]
     cache_exclude_status: Option<StatusCodeSelector>,
 
+    /// Filesystem path used to store the on-disk request cache
+    ///
+    /// [default: .lycheecache]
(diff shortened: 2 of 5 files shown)


mre:
Thanks, looks good.

Note: this is not a blocker for future XDG support. Later, we can resolve an XDG path instead of `.lycheecache`, while keeping the precedence: `--cache-location` > TOML config key `cache_location` > default location (XDG).

PR #2288, opened 2026-08-21
fix: extract issues due to not switching HTML parser states

### LINKED ISSUE #2287: `<` followed by a letter inside `<script>` silently drops the links after it
Hi, thank you for lychee. We added it to our docs CI a little while ago and it has been a pleasure to set up.

While checking a static export I noticed that most of a page's links were never reported. It comes down to a `<` immediately followed by a letter inside a `<script>`.

`page.html`:

```html
<!doctype html><html><body>
<a href="/before">b</a>
<script>for(var b=0;b<a.length;b+=2){}</script>
<a href="/after">a</a>
</body></html>
```

```
lychee --dump --extensions html --root-dir . -- page.html
```

Only `/before` comes out. `/after` is gone, the exit code is 0 and nothing is logged. Links come back at the next `</script>`, so when none follows, the rest of the document is skipped.

`1 < a` with a space and `1<2` are fine. `<!--` does the same thing, and `<textarea>` behaves the same way.

The same shape of input in other elements goes the other way. Links that are not in the document get reported:

```html
<title>x<a href="/inside-title">y</title>
<style>a{}<a href="/inside-style">z</style>
```

Both `/inside-title` and `/inside-style` come out.

This isn't only a synthetic case for us. Next.js writes `$RT=performance.now();for(var b=0;b<a.length;b+=2)` into every page it streams, which is the snippet above. On one 2.9 MB page of our exported site the skipped range is 966,912 bytes long and covers 986 `href`/`src` attributes. The file has 273 unique GitHub links and lychee reports 1.

I found #1905 while searching, but its fix is already in the nightly I tested and the snippet above still reproduces there.

Tested on lychee v0.24.2 and the nightly build from 2026-08-18, macOS (arm64).

katrinafyi:
Fixes https://github.com/lycheeverse/lychee/issues/2287

Simply uses `emitter.naively_switch_states(true)`, as suggested by https://github.com/untitaker/html5gum/blob/0.8.4/examples/tokenize_with_state_switches.rs. This is functionally the same as the old code calling `naive_next_state`:
https://github.com/lycheeverse/lychee/blob/0a96dc2fdc99faf5e1cd6ccc429b27da32f25985/lychee-lib/src/extract/html/html5gum.rs#L287-L294

I did a small refactor to factor out this common setup which is used for both link and fragment extraction.

### DIFF
--- a/lychee-lib/src/extract/html/html5gum.rs
+++ b/lychee-lib/src/extract/html/html5gum.rs
@@ -363,7 +363,7 @@ fn run_extractor<S: SpanProvider>(
     emitter.naively_switch_states(true); // https://github.com/untitaker/html5gum/blob/0.8.4/examples/tokenize_with_state_switches.rs
 
     let mut tokenizer = Tokenizer::new_with_emitter(buf, emitter);
-    assert!(tokenizer.next().is_none());
+    debug_assert!(tokenizer.next().is_none());
     extractor
 }
 
--- a/lychee-lib/src/extract/html/html5gum.rs
+++ b/lychee-lib/src/extract/html/html5gum.rs
@@ -363,7 +363,7 @@ fn run_extractor<S: SpanProvider>(
     emitter.naively_switch_states(true); // https://github.com/untitaker/html5gum/blob/0.8.4/examples/tokenize_with_state_switches.rs
 
     let mut tokenizer = Tokenizer::new_with_emitter(buf, emitter);
-    debug_assert!(tokenizer.next().is_none());
+    //debug_assert!(tokenizer.next().is_none());
     extractor
 }
 
--- a/lychee-lib/src/extract/html/html5gum.rs
+++ b/lychee-lib/src/extract/html/html5gum.rs
@@ -830,6 +830,7 @@ mod tests {
     }
 
     /// Should ignore `<a` that occurs within a `<script>`.
+    /// This prevents a regression of [#2287](https://github.com/lycheeverse/lychee/issues/2287).
     #[test]
     fn test_html_in_script() {
         let input = r#"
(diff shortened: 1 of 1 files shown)


lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -804,4 +812,52 @@ mod tests {
...
+                a
+                href="https://example.com/1"
+            >First</a>
+        </body>
+        "#;
+        let uris = extract_html(input, false);
+        assert_eq!(uris, vec![]);
+    }
+
+    /// Should ignore `<a` that occurs within a `<script>`.
```
thomas-zahner, line 835: Maybe we could mention the original issue:

```suggestion
    /// Should ignore `<a` that occurs within a `<script>`.
    /// This prevents a regression of [#2287](https://github.com/lycheeverse/lychee/issues/2287).
```

lychee-lib/src/extract/html/html5gum.rs
```diff
@@ -352,6 +352,21 @@ impl<S: SpanProvider> Callback<(), usize> for &mut LinkExtractor<S> {
...
+    include_verbatim: bool,
+    span_provider: S,
+) -> LinkExtractor<S> {
+    let mut extractor = LinkExtractor::new(span_provider, include_verbatim);
+
+    let mut emitter = CallbackEmitter::new(&mut extractor);
+    emitter.naively_switch_states(true); // https://github.com/untitaker/html5gum/blob/0.8.4/examples/tokenize_with_state_switches.rs
+
+    let mut tokenizer = Tokenizer::new_with_emitter(buf, emitter);
+    assert!(tokenizer.next().is_none());
```
thomas-zahner: I think that it might make sense to convert this to a debug assertion. If this condition was triggered we should probably not panic in release builds.

```suggestion
    debug_assert!(tokenizer.next().is_none());
```

thomas-zahner (approved):
Looks good to me :) thank you for the investigation

lychee-lib/src/extract/html/html5gum.rs
katrinafyi: Sorry it took so long! I had a guess that `tokenizer.next()` must always be called and this turned out to be right (without it, nothing gets extracted). I've now split up the next() call from the assertion. https://github.com/lycheeverse/lychee/pull/2288/commits/61a11c44d771e8409539df369c5cbe0b1426bb43

thomas-zahner: No worries! Huh interesting, that's not something I would have expected

PR #2293, opened 2026-09-03
fix: unset `GIT_` vars to fix pre-commit in worktree

### LINKED ISSUE #2292: pre-commit hook fails from a linked git worktree — hook-exported GIT_DIR points the wrapper's git calls at the user's repository
## Summary

`scripts/lychee_pre_commit.sh` runs `git` against the pre-commit cache clone of lychee to work out which version to install.
It runs as a git hook, and git exports `GIT_DIR` to hooks.
From a linked `git worktree` that variable is an absolute path to the user's repository, so those `git` calls read the user's repository instead of the lychee checkout, find no `lychee-*v*` tag, and the hook exits 100:

```
lychee...................................................................Failed
- hook id: lychee
- exit code: 100

lychee pre-commit requires 'rev' to be a versioned release tag, such as 'lychee-v0.XX.0'. please update the 'rev' field in your .pre-commit-config.yaml
```

The `rev` is a correct versioned release tag.
The same repository, same config, same `rev` passes from the main checkout and fails from a linked worktree.

There is a second effect covered in step 4: on a repository with no tags of its own, the hook runs `git fetch --tags` inside the user's repository.

## Step 1 — what a hook actually receives

A `language: script` hook that echoes its environment, run in a repository with one linked worktree:

| stage    | checkout        | GIT_DIR                                | GIT_WORK_TREE |
|----------|-----------------|----------------------------------------|---------------|
| pre-push | main checkout   | unset                                  | unset         |
| pre-push | linked worktree | `/…/main/.git/worktrees/wt` (absolute) | unset         |

That asymmetry is the whole bug, and it is why this has gone unnoticed: git exports `GIT_DIR` when the git dir is not the default `.git` beside the working directory, which is exactly the linked-worktree case (and the submodule case).

githooks(5) states the situation directly:

> Environment variables, such as GIT_DIR, GIT_WORK_TREE, etc., are exported so that Git commands run by the hook can correctly locate the repository. If your hook needs to invoke Git commands in a foreign repository or in a different working tree of the same repository, then it should clear these environment variables so they do not interfere with Git operations at the foreign location.

The wrapper is that case: it invokes git in a foreign repository, the pre-commit cache clone.

## Step 2 — what those variables do to the wrapper's git calls

The wrapper starts by locating its own checkout (lines 11 to 13), then reads the version from a tag (lines 23 to 38):

```bash
pushd "$(dirname "$0")" >/dev/null
LYCHEE_DIR="$(git rev-parse --show-toplevel)"
popd >/dev/null
...
if [[ -z "$(git tag)" ]]; then
    git fetch --tags &>/dev/null
fi
tag="$(git describe --tags --exact-match --match 'lychee-*v*' 2>/dev/null || true)"
```

Run from the cache clone's `scripts/` directory, those two reads behave like this:

```
(no GIT_DIR)           git rev-parse --show-toplevel -> ~/.cache/pre-commit/repoXXXX           correct
GIT_DIR=<user gitdir>  git rev-parse --show-toplevel -> ~/.cache/pre-commit/repoXXXX/scripts   = $PWD
GIT_DIR=<user gitdir>  git describe --tags --exact-match --match 'lychee-*v*'
                       -> fatal: No names found, cannot describe anything.
```

Two separate misfires:

- `--show-toplevel` returns the work tree root, and with `GIT_DIR` set and `GIT_WORK_TREE` unset git treats the current directory as that root, so `LYCHEE_DIR` becomes the `scripts/` subdirectory rather than the checkout.
- `git describe` resolves refs through `GIT_DIR`, so it describes the user's HEAD. No `lychee-*v*` tag is there, `tag` stays empty, the regex fails, and the script exits 100 with the message above.

The second one is what the user sees.

## Step 3 — end-to-end reproduction

```bash
mkdir -p /tmp/lychee-wt && cd /tmp/lychee-wt
git init -q main && git init -q --bare remote.git
cd main && git remote add origin ../remote.git
cat > .pre-commit-config.yaml <<'YAML'
repos:
    rev: lychee-v0.24.2
    hooks:
      - id: lychee
        stages: [pre-push]
        types: [markdown]
        args: [--no-progress]
YAML
echo '# doc' > doc.md
git add -A && git commit -qm init --no-verify
pre-commit install --install-hooks -t pre-push

git push origin master                       # 1. from the main checkout
git worktree add -q ../wt -b topic
cd ../wt && git commit -qm x --allow-empty --no-verify
git push origin topic                        # 2. from the linked worktree
```

Observed:

```
=== 1. MAIN CHECKOUT push ===
lychee...................................................................Passed
=== 2. LINKED WORKTREE push ===
lychee...................................................................Failed
- hook id: lychee
- exit code: 100
lychee pre-commit requires 'rev' to be a versioned release tag, ...
```

## Step 4 — the hook fetches into the user's repository

The version block reaches for tags before it describes:

```bash
if [[ -z "$(git tag)" ]]; then
    git fetch --tags &>/dev/null
fi
```

Under the leak both commands run against the user's repository, so a repository with no tags of its own takes that branch on every push: a network fetch that writes refs into the user's repo, silenced by `&>/dev/null`.

Verified rather than inferred, continuing the scratch repo from step 3 — a tag created only in the remote, never fetched by the worktree:

```bash
git -C ../remote.git tag leaked-proof-tag master
cd ../wt
git tag                    # (empty)
git push origin topic      # hook runs, fails as before
git tag                    # leaked-proof-tag
```

```
tags before: []
tags after:  [leaked-proof-tag ]
```

A link checker performed a network fetch and wrote refs into the repository it was checking.

## Scope

- Affects linked `git worktree` checkouts, and submodules for the same reason (git exports `GIT_DIR` and `GIT_INDEX_FILE` there too).
- Does not affect `pre-commit run` invoked directly: no git hook, no exported `GIT_DIR`, so the wrapper resolves correctly.
- Independent of the pre-commit stage; measured here on `pre-push`.
- Not fixable in pre-commit: the same class has been filed there three times against other hooks and closed each time — pre-commit/pre-commit#1972, pre-commit/pre-commit#2295, pre-commit/pre-commit#3609.

katrinafyi:
Closes https://github.com/lycheeverse/lychee/issues/2292

### DIFF
--- a/scripts/lychee_pre_commit.sh
+++ b/scripts/lychee_pre_commit.sh
@@ -7,6 +7,9 @@ if [[ "$first_arg" == LYCHEE_VERSION=* ]]; then
     exec env "$first_arg" "$0" "$@"
 fi
 
+# https://git-scm.com/docs/githooks#:~:text=it%20should%20clear%20these%20environment%20variables
+unset $(git rev-parse --local-env-vars)
+
 # Find and navigate to the pre-commit cache folder where lychee
 # is checked out. something like: ~/.cache/pre-commit/repo7r00atq6/
 pushd "$(dirname "$0")" >/dev/null


mre:
Hum, any idea why the tests don't pass?

katrinafyi:
Not sure. Seems to be a 429 but does also seem to be consistent. We could try to remove the problematic 10Gb URL.

mre:
Either way, I guess the failure is unrelated to the branch's code, so I'll go ahead and merge it. 😄 Thanks for the fix.

PR #2294, opened 2026-09-06
test: replace flaky large file URLs with local fixture

mre:
Replace the large-file CLI test's proof.ovh.net dependency with a local streaming HTTP fixture. This preserves the lazy-download and fragment coverage without relying on a rate-limited third-party host.

See the discussion over at #2293.

### DIFF
--- a/lychee-bin/tests/cli.rs
+++ b/lychee-bin/tests/cli.rs
@@ -17,7 +17,7 @@ mod cli {
         collections::{HashMap, HashSet},
         error::Error,
         fs::{self, File},
-        io::{BufRead, Write},
+        io::{BufRead, Read, Write},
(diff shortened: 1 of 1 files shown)


mre:
I decided to remove the `/fragment` endpoint again. Fragment behavior is already covered by dedicated tests, so this now keeps only the small raw TCP fixture needed to model an incomplete 10 GB response. `wiremock` buffers response bodies and delays before sending headers, which cannot represent this case. The listener handles two connections because the test invokes lychee twice against the same URL: once with `--include-fragments` and no fragment, and once with a fragment but without `--include-fragments`.

thomas-zahner:
That's a great alternative, thanks for the fix!

PR #2304, opened 2026-09-20
Fix duplicate JUnit test names for repeated links

### LINKED ISSUE #2302: GitLab's Junit parser treats same test names as a signle test, resulting in not showing all failed links
When I integrated lychee into my CI/CD pipeline on GitLab, the resulting xml can have multiple testcases with the same name. GitLab parser only treats the first find as a valid test, ignoring other tests with the same name. This is a problem when the same link appears in multiple files, because only the first find would be flagged as failed.

Would it be possible to add a feature to make the testcases' names unique?

mre:
Thanks @Nextross for reporting this! I’ve added the source file and available line/column information to JUnit test names so GitLab can distinguish the same URL in different files. I also removed the outcome prefixes so a link keeps the same test name when it goes from failing to passing. Fixes #2302.


mre:
@thomas-zahner, you know more about it than me. Can you review the PR when you find the time?

lychee-bin/src/formatters/stats/junit.rs
```diff
@@ -127,6 +133,118 @@ mod tests {
...
+                    )
+                })
+                .collect(),
+            ..Default::default()
+        };
+        let cases = junit_testcases(stats);
+        assert_eq!(cases.len(), 2);
+        let names: HashSet<_> = cases.iter().map(|case| case.name.to_string()).collect();
+
+        assert_eq!(
```
thomas-zahner, line 180: I feel like it shouldn't be necessary to pin down the specific names in these tests. But then I'm also okay with doing it. I would think that testing the property `same_url_in_different_files_has_distinct_names` doesn't mean we need to examine the exact contents. I.e. when collecting it as set, we can just assert the length of the set.

lychee-bin/src/formatters/stats/junit.rs
```diff
@@ -127,6 +133,118 @@ mod tests {
...
+                ]
+                .into_iter()
+                .map(|span| response(Some(span), Status::Ok(StatusCode::OK)))
+                .collect(),
+            )]),
+            ..Default::default()
+        };
+        let cases = junit_testcases(stats);
+        assert_eq!(cases.len(), 4);
+        let names: HashSet<_> = cases.iter().map(|case| case.name.to_string()).collect();
```
thomas-zahner, line 208: Same here, this line and below isn't strictly necessary. I we want, we can create a single, probably one-line, test to test the structure, and simplify these "property tests" to really only test the specific property.

thomas-zahner (approved):
Looks good to me :)

